PR #19, opened 2025-02-03
Plumb Vote Certificate into BankingStage

carllin:
#### Problem
Vote Certificate is missing from block creation in BankingStage

#### Summary of Changes
For the new consensus mechanism, every block X that descends from block Y needs a "certificate", essentially a vector of 67% of VoteTransaction included in block X for every block in the range [Y, X]

Otherwise block X is considered invalid

Now by the time the leader makes a block X, we can assume he has all the certificates sitting around, so he just needs to make sure BankingStage will include those Vec<VoteTransaction> in the block.

In this PR, starting from ReplayStage which holds these certificates:

1. In maybe_start_leader() when we set the block in PoH via poh_recorder.lock().unwrap().set_bank(&tpu_bank);, also pass along the necessary certificates
PohRecorder saves these certificates in BankStarthttps://github.com/anza-xyz/agave/blob/68536a665b97f80eb25c97360a9de1abdcf723dc/poh/src/poh_recorder.rs#L66-L69

2. PohRecorder saves these certificates in BankStarthttps://github.com/anza-xyz/agave/blob/68536a665b97f80eb25c97360a9de1abdcf723dc/poh/src/poh_recorder.rs#L66-L69
in the `WorkingBank`

3.  BankingStage main loop finds BankStart when it makes the block: [https://github.com/anza-xyz/agave/blob/68536a665b97f80eb25c97360a9de1abdcf723dc/core/src/banking_stage.rs#L688
consumer](https://github.com/anza-xyz/alpenglow/blob/8463a4323fdb604bee0208a9bc0e63f035656ef1/core/src/banking_stage.rs#L771-L789) For banking id  thread = 1 the tpu thread adds the certificate to the UnprocessedTransactionStorage::VoteStorage for votes.

5. Then when we drain from the vote storage we also drain from the vote_certificate to include in the block

Fixes #

### DIFF
--- a/core/src/banking_stage/unprocessed_transaction_storage.rs
+++ b/core/src/banking_stage/unprocessed_transaction_storage.rs
@@ -23,7 +23,9 @@ use {
     solana_runtime::bank::Bank,
     solana_runtime_transaction::runtime_transaction::RuntimeTransaction,
     solana_sdk::{
-        clock::FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET, hash::Hash, saturating_add_assign,
+        clock::{Slot, FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET},
+        hash::Hash,
+        saturating_add_assign,
         transaction::SanitizedTransaction,
     },
     solana_svm::transaction_error_metrics::TransactionErrorMetrics,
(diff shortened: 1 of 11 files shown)


core/src/banking_stage/unprocessed_transaction_storage.rs
```diff
@@ -52,8 +54,33 @@ pub struct ThreadLocalUnprocessedPackets {
     thread_type: ThreadType,
 }
 
+#[derive(Debug)]
+struct UnprocessedVoteCertificate {
+    vote_certificate: Vec<Arc<ImmutableDeserializedPacket>>,
```
apfitzge, line 59: just FYI you probably don't want to base things off of this type since it will somewhat soon be deleted

apfitzge: `ImmutableDeserializedPacket` I mean

carllin: Noted!

poh/src/poh_recorder.rs
```diff
@@ -43,6 +44,8 @@ use {
     thiserror::Error,
 };
 
+pub type VoteCertificate = Vec<PacketBatch>;
```
wen-coding, line 47: Will you use this VoteCertificate for Notarization/Skip/Finalization?

carllin: nah this is outdated, new stuff is all in the new vote_certificate module

PR #22, opened 2025-02-05
Add vote certificate module

carllin:
#### Summary of Changes
1. Add ability to create vote certificates and ingest them into a bank
2. Block BankingStage/Poh from processing/recording anything until the vote certificate is recorded first

Fixes #

### DIFF
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -66,6 +66,8 @@ pub type WorkingBankEntry = (Arc<Bank>, (Entry, u64));
 pub struct BankStart {
     pub working_bank: Arc<Bank>,
     pub bank_creation_time: Arc<Instant>,
+    #[cfg(feature = "alpenglow")]
+    pub contains_valid_certificate: Arc<AtomicBool>,
 }
 
 impl BankStart {
@@ -265,6 +267,8 @@ pub struct WorkingBank {
     pub min_tick_height: u64,
     pub max_tick_height: u64,
     pub transaction_index: Option<usize>,
+    #[cfg(feature = "alpenglow")]
+    pub contains_valid_certificate: Arc<AtomicBool>,
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -401,6 +405,8 @@ impl PohRecorder {
         self.working_bank.as_ref().map(|w| BankStart {
             working_bank: w.bank.clone(),
             bank_creation_time: w.start.clone(),
+            #[cfg(feature = "alpenglow")]
+            contains_valid_certificate: w.contains_valid_certificate.clone(),
         })
     }
 
@@ -688,6 +694,8 @@ impl PohRecorder {
             bank,
             start: Arc::new(Instant::now()),
             transaction_index: track_transaction_indexes.then_some(0),
+            #[cfg(feature = "alpenglow")]
+            contains_valid_certificate: Arc::new(AtomicBool::new(false)),
         };
         trace!("new working bank");
         assert_eq!(working_bank.bank.ticks_per_slot(), self.ticks_per_slot());
@@ -971,6 +979,15 @@ impl PohRecorder {
                 .ok_or(PohRecorderError::MaxHeightReached)?;
             if bank_slot != working_bank.bank.slot() {
                 return Err(PohRecorderError::MaxHeightReached);
+            }
+
+            // TODO: allow certificate through
+            #[cfg(feature = "alpenglow")]
+            if !working_bank
+                .contains_valid_certificate
+                .load(Ordering::Relaxed)
+            {
+                panic!("We should not be recording transactions before the certificate is written");
             }
 
             let (mut poh_lock, poh_lock_us) = measure_us!(self.poh.lock().unwrap());
--- a/core/src/banking_stage/decision_maker.rs
+++ b/core/src/banking_stage/decision_maker.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "alpenglow")]
+use std::sync::atomic::Ordering;
+
 use {
     solana_poh::poh_recorder::{BankStart, PohRecorder},
     solana_sdk::{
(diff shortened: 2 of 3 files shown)


core/src/vote_certificate.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    pub fn is_valid(&mut self, total_stake: u64) -> bool {
+        (self.stake as f64 / total_stake as f64) > SUPERMAJORITY
+    }
+
+    pub fn commit(&self, new_bank: &Arc<Bank>) -> bool {
+        assert_eq!(self.parent_bank.slot(), new_bank.parent_slot());
+        assert_eq!(self.parent_bank.hash(), new_bank.parent_hash());
+        let summary = self.consumer.process_transactions(
```
carllin, line 130: @apfitzge this is where I added in your recommendation to prcesss the votes directly first. 

Workflow will look like

1. poh_recorder::set_bank() in replay so that the recorder has a place to record
2. then call commit() here on the vote certificate

core/src/vote_certificate.rs
```diff
@@ -0,0 +1,159 @@
...
+    ) -> VoteCertificate {
+        let mut vote_certificate = VoteCertificate::new(poh_recorder, parent_bank, self.slot);
+        for (pubkey, vote) in self.certificate.into_iter() {
+            let _res = vote_certificate.add_vote(&pubkey, vote, get_stake(&pubkey));
+        }
+        vote_certificate
+    }
+}
+
+pub struct VoteCertificate {
```
wen-coding, line 66: We already have too many things under core/src. Would this only be used by banking_stage? Would consensus code need it too?

core/src/vote_certificate.rs
```diff
@@ -0,0 +1,159 @@
...
+    // Either:
+    // 1. The frozen bank for `slot` if the certificate is
+    // notarization votes
+    // 2. The last notarized bank < slot if the certificate is skip votes
+    parent_bank: Arc<Bank>,
+    // Must be either all skip votes, or all notarization votes.
+    // We keep separate certificates for each type
+    certificate: HashMap<Pubkey, RuntimeTransaction<SanitizedTransaction>>,
```
wen-coding, line 74: So we only keep one transaction per pubkey? Could we have more than one certificates for a pubkey?

poh/src/poh_recorder.rs
```diff
@@ -66,6 +66,8 @@ pub type WorkingBankEntry = (Arc<Bank>, (Entry, u64));
 pub struct BankStart {
     pub working_bank: Arc<Bank>,
     pub bank_creation_time: Arc<Instant>,
+    #[cfg(feature = "alpenglow")]
+    pub contains_valid_certificate: Arc<AtomicBool>,
```
wen-coding, line 70: Would love to move things out of poh. But I can live with current design if that's too much work.

core/src/vote_certificate.rs
carllin: The primary consumer/tracker of these for each slot would be replay_stage which is starting the leader banks, so it seems most sense to keep it here

core/src/vote_certificate.rs
carllin: I think leader could potentially also include votes for the same pubkeys for earlier slots, but since it's not critical to the certificate for this particular slot, I think we can track those elsewhere

poh/src/poh_recorder.rs
carllin: BankingStage already checks `BankStart` every time it wants to ingest transactions, so it seems natural to use this struct to host the flag to block transaction processing

core/src/banking_stage/consumer.rs
```diff
@@ -108,6 +109,13 @@ impl Consumer {
         }
     }
 
+    pub fn create_consumer(poh_recorder: &RwLock<PohRecorder>) -> Consumer {
+        let (replay_vote_sender, _replay_vote_receiver) = unbounded();
+        let committer = Committer::new(None, replay_vote_sender, Arc::default());
+        let transaction_recorder: TransactionRecorder = poh_recorder.read().unwrap().new_recorder();
```
apfitzge, line 115: just take this directly instead of the `RwLock<PohRecorder>`?

is this fn test only? the vote sender will always error if it sends, is that okay?

core/src/vote_certificate.rs
apfitzge: how many vote certificates will we have? is runtime a concern here?

will this delay the sending of bank to banking stage?

carllin: one vote certificate per slot

> will this delay the sending of bank to banking stage?

Yeah, banking stage will not be able to write any transactions until these votes are written out

core/src/banking_stage/consumer.rs
carllin: > is this fn test only? 

No its used by replay to process/record certificates into the bank/poh before allowing banking stage to write

Created here: https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/vote_certificate.rs#L84

Used here: https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/vote_certificate.rs#L130-L133

> the vote sender will always error if it sends, is that okay

Yeah because we're only using this to record a vote certificate from replay into a bank, we already have the votes in replay, so nobody needs to hear about these votes again

core/src/vote_certificate.rs
apfitzge: 🤔 alternatively, just reserve CUs and grab locks for the transactions, then send the bank out to BS.
Then it doesn't need to wait on the execution/commit for **all** transactions, just conflicting ones?

carllin: Thought about this, but it seemed like it required more refactoring of the consumer logic. Maybe I should stop being lazy and just do the surgery though :)

Would this be the rough steps:

1. Getting the pre-results https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/banking_stage/consumer.rs#L450-L473
2. Grabbng the CUs https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/banking_stage/consumer.rs#L488-L495
3. Grabbing the locks https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/banking_stage/consumer.rs#L500-L506
4. Handing the bank off to PohRecorder
5. Calling execute https://github.com/anza-xyz/alpenglow/blob/a21182ebadd2f679aab2eb1541421f8e171eda1c/core/src/banking_stage/consumer.rs#L511-L512

apfitzge: Yeah those seem like the right steps.
I wonder if the best way to do this is to just add a closure argument to consumer's processing fn.
For regular workers closure will just do nothing.
For your vote worker, closure would set the bank in poh recorder.

That way we don't need to have some additional code path.

carllin: yeah that would work well 

This is perfect because needed a way to detect in the scenario that some of these vote transactions fail, need to be able to map them to stake to figure out if at least 2/3 landed, otherwise report an error because the block will be invalid

PR #36, opened 2025-02-07
Add basic skeleton of certificate tracking

carllin:
#### Problem
Determines if we can start a leader block

#### Summary of Changes
1. Add new alpenglow consensus module
2. Add SkipPool for tracking supermajority skips
3. Add certificates for tracking notarizations
4. Add vote storage for persisting validator votes across reboot

Fixes #
Entrypoint is `maybe_start_leader()`

### DIFF
--- a/core/src/alpenglow_consensus/vote_history.rs
+++ b/core/src/alpenglow_consensus/vote_history.rs
@@ -91,7 +91,7 @@ impl VoteHistory {
     }
 
     pub fn is_slot_skipped(&self, slot: Slot) -> bool {
-        !self.skip_votes.iter().any(|vote| {
+        self.skip_votes.iter().any(|vote| {
             vote.skip_range()
                 .expect("must be a skip vote")
                 .contains(&slot)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,275 @@
...
+        Ok(())
+    }
+
+    fn get_notarization_certificate(&self, slot: Slot) -> Option<Vec<VersionedTransaction>> {
+        self.certificates
+            .get(&(slot - 1, CertificateType::Notarize))
+            .map(|certificate| certificate.get_certificate())
+    }
+
+    fn commit_certificate(
```
carllin: @wen-coding assuming no failures here 🤞 

TODO: Please fix

wen-coding: Please always put TODO in the code. TODO in comments will get forgotten.

core/Cargo.toml
```diff
@@ -119,7 +119,7 @@ solana-cost-model = { workspace = true, features = ["dev-context-only-utils"] }
 solana-ledger = { workspace = true, features = ["dev-context-only-utils"] }
 solana-logger = { workspace = true }
 solana-net-utils = { workspace = true, features = ["dev-context-only-utils"] }
-solana-poh = { workspace = true, features = ["dev-context-only-utils"] }
+solana-poh = { workspace = true, features = [ "dev-context-only-utils"] }
```
wen-coding: why do we want a whitespace here?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+    AddSkipPoolFailed(#[from] skip_pool::AddVoteError),
+
+    #[error("Add vote to vote certificate failed: {0}")]
+    AddVoteCertificateFailed(#[from] vote_certificate::AddVoteError),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Vote {
+    Notarize(Slot),
+    Skip(RangeInclusive<Slot>),
```
wen-coding, line 34: I'm guessing that you want Skip(slot_range) instead of Skip(slot) because it's easier to pack one skip cert for all slots skipped. But in the simplest version I think the algorithm skips one block at a time (line 13), this makes processing Vote easier because you don't need to special case Skip. And we can pack multiple skips in the same block if we want to prove we can skip slot X to slot Y?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    pub fn maybe_start_leader(
+        &self,
+        my_pubkey: &Pubkey,
+        my_leader_slot: Slot,
+        bank_forks: &RwLock<BankForks>,
+        poh_recorder: &RwLock<PohRecorder>,
+        total_stake: u64,
+    ) -> Option<Arc<Bank>> {
+        let parent_bank = bank_forks.read().unwrap().get(self.highest_notarized_slot);
```
wen-coding: Hmm, I think the actual trigger is that you either have finalize or skip my_leader_slot-1?

We probably need a data structure recording final decision for each slot? (skip or finalized hash X)

Also currently the paper proposed sending all 4 blocks belonging to one leader back to back, but we can implement that later.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+            // and could immediately start building your slot off an earlier parent.
+            if !parent_bank.is_frozen() {
+                return None;
+            }
+        } else {
+            // We haven't started replaying the parent yet
+            return None;
+        }
+
+        let parent_bank = parent_bank.unwrap();
```
wen-coding: Between line 195 and building your block `self.highest_notarized_slot` might change, so I really think we should check the status of `my_leader_slot-1` instead. (Of course, special case bank 0.)

core/src/alpenglow_consensus/vote_history_storage.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+pub struct EtcdTlsConfig {
+    pub domain_name: String,
+    pub ca_certificate: Vec<u8>,
+    pub identity_certificate: Vec<u8>,
+    pub identity_private_key: Vec<u8>,
+}
+
+impl EtcdVoteHistoryStorage {
```
wen-coding, line 185: I would prefer different storage for vote history to be in separate files, like what we are doing for TowerStorage. Separating it into cores/src/consensus/alpenglow_vote_history_storage/???

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum CertificateType {
+    Notarize,
+    Skip,
+    Finalize,
+}
+
+pub struct CertificatePool {
+    // Notarization and finalization vote certificates
+    certificates: BTreeMap<CertificateId, VoteCertificate>,
```
wen-coding, line 82: I think in the current paper all decisions are made per slot, is there anywhere we need to make decision based on <slot, certificateType> pair? If we need that we should probably just have one certificate pool (move skip pool into here).

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+    // Pool of latest skip votes per validator
+    skip_pool: SkipPool,
+    // Highest slot with a notarized certificate
+    highest_notarized_slot: Slot,
```
wen-coding, line 86: Do you need RwLock for this variable?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,276 @@
...
+                bank.slot(),
+                e
+            );
+            return false;
+        }
+
+        let runtime_transactions = runtime_transactions.unwrap();
+        let summary = consumer.process_transactions(bank, &Instant::now(), &runtime_transactions);
+
+        if summary.reached_max_poh_height {
```
wen-coding: I guess in the future this will be a timer? Do we still need to bind with PoH here?

core/src/alpenglow_consensus/certificate_pool.rs
carllin: this will most likely be synchronous with replay stage, and replay stage is the one inserting votes, so there won't be new certificates. 

If a skip certificate already exists for `my_leader_slot-1` then the leader was too slow and its acceptable to just start your block

core/src/alpenglow_consensus/certificate_pool.rs
carllin: Not for now, all synchronous with replay

core/src/alpenglow_consensus/certificate_pool.rs
carllin: For now yes, PohRecorder is necessary to  record transactions and send to broadcast

core/src/alpenglow_consensus/certificate_pool.rs
carllin: skip pool is a different structure because skips are ranges. Ranges aren't most efficiently tracked per slot, so makes sense to keep in them in different structures than per-slot based certs like Notarization and Finalization

core/src/alpenglow_consensus/certificate_pool.rs
carllin: >  I think the actual trigger is that you either have finalize or skip my_leader_slot-1

Not sufficient because you could be missing skip certificates/notarizations for slots earlier than my_leader_slot - 1

Here we just brain-dead check if there is a notarization and then a consecutive row of skip certificates up to `my_leader_slot-1`

core/src/alpenglow_consensus/certificate_pool.rs
carllin: Skips could potentially take up too much space in a long partition, I think it's easier and faster to just compress them

wen-coding: These are certificates. No matter how big your long-standing partition is, the algorithm is the same as before. If you don't have a notar/skip certificate for slot X, your local timer will not even start for slot X+1.
- If you are just starting up, you should catch up first with all the messages in Gossip so that you are in sync with your partition
- If you are in sync with everyone else in your partition, and you are working on a certificate for slot X, then others could only be in slot X+1. They shouldn't be in slot X+2, because they shouldn't have a cert for slot X+1, otherwise you'll see all the nota/skip votes in Gossip and you would have that cert as well.

In any case, I don't think inside the network partition everyone will be too far off such that we need to consider sending and processing skip slots as a range.

Also the skips are even smaller than Notar/Final, it only contains one slot, no hash. How much do we save by compressing them? I worry more about the code complexity introduced.

carllin: It's possible to have mutliple skips in a row which means multiple entrires in gossip. It's cleaner to only have a single entry in gossip so you don't have to worry about losing information in gossip every time you publish a new skip.

wen-coding: If before the partition everyone agreed on slot X, then either:
1. your partition has >2/3 stake, for all blocks belonging to leaders in your partition, your partition is just finalizing them. For blocks belonging to leaders not in your partition, you are getting skip certs in sequence for all of them.
or
2. your partition has <2/3 stake, then you don't have certs for anything because you don't have enough stake to confirm either notar or skip, your partition will basically be stuck on slot X+1, no new blocks for your partition

If all partitions has <2/3 stake, then everyone is stuck on slot X+1, there are no skips in a row.

The majority partition doesn't care about old skips at all, once the majority partition proceeds to slot Y+1, you can discard all the old skips belonging to slot Y.

If you are a minority partition recovering after the partition, apparently you don't care about processing the skips or notars in Gossip any more, because you don't have any of the blocks confirmed by the majority partition. You should just repair all the blocks you are missing. When you get the new block confirmed by the majority partition, you automatically get all the notar/skip certs in the block, you don't care about the stale info in Gossip either.

In either case I don't think it helps to pay special attention to old skips in Gossip.

Anything I'm missing?

carllin: > your partition has >2/3 stake, for all blocks belonging to leaders in your partition, your partition is just finalizing them. For blocks belonging to leaders not in your partition, you are getting skip certs in sequence for all of them.

This is not necessarily true because partitions don't have to be perfect splits, message delays can cause skips as well, for instance if the leader in the partition's messages are all delayed by > delta, so all validators timeout. This is also why you can get a skip certificate  and a notarization for the same slot for instance.

If you repeat this on every slot, you can get multiple skips in a row

> The majority partition doesn't care about old skips at all, once the majority partition proceeds to slot Y+1, you can discard all the old skips belonging to slot Y.

You can't completely discard them because you need to commit those skips into the next leader block. 

You're right though in gossip you probably don't need to store that many skips. I'm not sure if this changes in any future/optimistic fast path.

As for handling the ranges, I don't think the code is that much more complex, I think it's actually easier to reason about if your skip votes are compact/and more flexible to work with. 

Compression is also worth something, what if there are thousands of skips in a row for instance

Not opposed to somebody else rewriting that portion though if they feel strongly about this/doesn't affect future fast path

wen-coding: That's a good point. Also I think this does depend on the optimistic fast path we may introduce in SLC.

Maybe let's talk about data structure design a bit in SLC so everyone can be on the same page?

I'm fine with checking in like this but maybe change it later if we decide differently in SLC.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,715 @@
...
+    },
+    thiserror::Error,
+};
+
+pub type CertificateId = (Slot, CertificateType);
+
+#[derive(Debug, Error, PartialEq)]
+pub enum AddVoteError {
+    #[error("Vote already exists for this pubkey")]
+    AddSkipPoolFailed(#[from] skip_pool::AddVoteError),
```
wen-coding: Your comment implies this is a duplication error while your Error name seems to be more general?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,715 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum NewHighestCertificate {
+    Notarize(Slot),
+    Skip(Slot),
+    Finalize(Slot),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Vote {
+    Notarize(Slot),
```
wen-coding, line 33: Hmm, I thought we need the Hash in Notarize as well?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,715 @@
...
+
+pub type CertificateId = (Slot, CertificateType);
+
+#[derive(Debug, Error, PartialEq)]
+pub enum AddVoteError {
+    #[error("Vote already exists for this pubkey")]
+    AddSkipPoolFailed(#[from] skip_pool::AddVoteError),
+
+    #[error("Add vote to vote certificate failed: {0}")]
+    AddVoteCertificateFailed(#[from] vote_certificate::AddVoteError),
```
wen-coding: nit: AddVoteCertificateFailed sounds like you are saving a complete vote certificate, while you are really trying to AddToCertificatePool here?

If you like the AddToCertificatePool name, then it makes sense to rename the previous one AddToSkipPoolFailed.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+            highest_notarized_slot: 0,
+            highest_finalized_slot: 0,
+        }
+    }
+
+    pub fn add_vote(
+        &mut self,
+        vote: Vote,
+        transaction: VersionedTransaction,
+        validator_key: &Pubkey,
```
wen-coding: Probably best to make it clear which key this is, I'm guessing this is the vote account pubkey?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+        validator_stake: u64,
+        total_stake: u64,
+    ) -> Result<Option<NewHighestCertificate>, AddVoteError> {
```
wen-coding, line 114: Who is the caller of this method? Does it care about NewHighestCertificate every time? If the caller is adding several votes mixed together, maybe it only cares about what new cert there is at the very end?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+                    )));
+                }
+            }
+        }
+        Ok(None)
+    }
+
+    fn get_notarization_certificate(&self, slot: Slot) -> Option<Vec<VersionedTransaction>> {
+        self.certificates
+            .get(&(slot - 1, CertificateType::Notarize))
```
wen-coding: Why do we need to use slot-1 here?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,716 @@
...
+        bank: &Arc<Bank>,
+        poh_recorder: &RwLock<PohRecorder>,
+        certificate: Vec<VersionedTransaction>,
+    ) -> bool {
+        let consumer = Consumer::create_consumer(poh_recorder);
+        let runtime_transactions: Result<Vec<RuntimeTransaction<SanitizedTransaction>>, _> =
+            certificate
+                .into_iter()
+                .map(|versioned_tx| {
+                    // Short circuits on first error
```
wen-coding: Why do we need short circuits on first error?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,723 @@
...
+                ("slot", bank.slot(), i64)
+            );
+            // TODO: check if 2/3 of the stake landed, otherwise return false
+            return false;
+        }
+
+        true
+    }
+
+    pub fn highest_certificate_slot(&self) -> Slot {
```
wen-coding, line 202: highest_certificate_slot seems vague, you really mean highest slot with notarize or skip cert, you don't really care about finalization cert here.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,723 @@
...
+                        self.highest_finalized_slot = self.highest_finalized_slot.max(vote_slot);
+                        if old_highest_finalized_slot != self.highest_finalized_slot {
+                            return Ok(Some(NewHighestCertificate::Finalize(
+                                self.highest_finalized_slot,
+                            )));
+                        }
+                    }
+                }
+            }
+            Vote::Skip(skip_range) => {
```
wen-coding, line 149: Err, if we have  a range in Skip, then it's questionable what total_stake we use here, because this skip range could cross Epoch boundary.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,723 @@
...
+
+        true
+    }
+
+    pub fn highest_certificate_slot(&self) -> Slot {
+        self.highest_notarized_slot
+            .max(*self.skip_pool.max_skip_certificate_range().end())
+    }
+
+    pub fn maybe_start_leader(
```
wen-coding: Do you think we should maybe move this logic elsewhere? Seems too much function put into CertificatePool

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    fn pubkey(&self) -> Pubkey {
+        self.0
+    }
+}
+
+/// Dynamic Segment Tree that works with any type implementing `HasStake`
+struct DynamicSegmentTree<T: Ord + Clone + HasStake> {
+    tree: BTreeMap<usize, Vec<T>>, // Single Vec<T> per slot (first occurrence = add, second = remove)
```
wen-coding: Is the key type actually Slot?

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+    #[error("Newer skip vote {0:?} than {1:?} already exists for this pubkey")]
+    TooOld(RangeInclusive<u64>, RangeInclusive<u64>),
+
+    #[error("Overlapping skip vote old {0:?} and new {1:?}")]
+    Overlapping(RangeInclusive<u64>, RangeInclusive<u64>),
+}
+
+/// A trait for objects that provide a stake value.
+pub trait HasStake {
+    fn stake_value(&self) -> i64;
```
wen-coding: I think Stake is u64?

core/src/alpenglow_consensus/certificate_pool.rs
carllin: yeah, everything other than the check should probably should be split out into a function in replay_stage

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+            }
+            if let Some(events) = self.tree.get_mut(&end) {
+                events.retain(|v| v.pubkey() != new_value.pubkey());
+            }
+        }
+    }
+
+    /// Queries the first slot range where the accumulated stake exceeds `threshold`
+    /// Returns the range and contributing Pubkeys
+    fn query(&self, threshold: f64) -> Option<(RangeInclusive<Slot>, Vec<T>)> {
```
wen-coding: nit: maybe make the argument name `threshold` more obvious we are using a stake not percentage here.

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+        if add {
+            self.tree.entry(start).or_default().push(new_value.clone());
+            self.tree.entry(end).or_default().push(new_value);
+        } else {
+            if let Some(events) = self.tree.get_mut(&start) {
+                events.retain(|v| v.pubkey() != new_value.pubkey());
+            }
+            if let Some(events) = self.tree.get_mut(&end) {
+                events.retain(|v| v.pubkey() != new_value.pubkey());
+            }
```
wen-coding: It looks like add and remove doesn't share too much code, would it be simpler to just split this into two functions: add() and remove()?

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+                    skip_range,
+                ));
+            }
+
+            self.segment_tree.update(
+                *prev_skip_vote.skip_range.start() as usize,
+                *prev_skip_vote.skip_range.end() as usize,
+                (*pubkey, -(stake as i64)), // stake doesn't actually matter here
+                false,
+            );
```
wen-coding, line 170: So when there is a new range the old range is always removed?
Could it be one validator vote skip for slot 100~101, decided to Notarize slot 102, then wants to skip slot 103?
It now sends [103, 103] because it doesn't want to skip 102, but now its skip for [100, 101] will be removed and not counted when we need skip certs?

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        // Add new skip range
+        self.segment_tree.update(
+            *skip_range.start() as usize,
+            *skip_range.end() as usize,
+            (*pubkey, stake as i64), // Add stake
```
wen-coding: Across epoch boundaries the stake might change for the same pubkey.

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.skips.insert(
+            *pubkey,
+            SkipVote {
+                skip_range: skip_range.clone(),
+                skip_transaction,
+            },
+        );
+
+        // Find the largest range where the cumulative stake exceeds 2/3 of total stake
+        let threshold = (2.0 * total_stake as f64) / 3.0; // Calculate required stake threshold
```
wen-coding, line 190: Same as above, across epoch boundary the total_stake might change.

Also, put constants into a const?

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -0,0 +1,343 @@
...
+impl SkipPool {
+    /// Initializes the `SkipPool`
+    pub fn new() -> Self {
+        Self {
+            skips: HashMap::new(),
+            segment_tree: DynamicSegmentTree::new(),
+            max_skip_certificate_range: 0..=0, // Default range
+        }
+    }
+
```
wen-coding, line 136: We probably should have a purge() operation, when some slot X is finalized, everything older than X can be discarded?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,73 @@
...
+#[derive(Debug, Error, PartialEq)]
+pub enum AddVoteError {
+    #[error("Vote already exists for this pubkey")]
+    DuplicateVote,
+
+    #[error("Transaction failed: {0}")]
+    TransactionFailed(#[from] TransactionError),
+}
+
+pub const SUPERMAJORITY: f64 = 2f64 / 3f64;
```
wen-coding: nit: should we share this const with skip_pool which also needs this?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub struct VoteCertificate {
+    // Must be either all notarization or finalization votes.
+    // We keep separate certificates for each type
+    certificate: HashMap<Pubkey, VersionedTransaction>,
```
wen-coding, line 24: Should we use BTreeMap here so in the values would be sorted by keys?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,73 @@
...
+    // Total stake of all the slots in the certificate
+    stake: u64,
```
wen-coding: nit: use Stake type?

core/src/alpenglow_consensus/vote_history_storage.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    fn store(&self, _saved_vote_history: &SavedVoteHistoryVersions) -> Result<()> {
+        Ok(())
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct FileVoteHistoryStorage {
+    pub vote_history_path: PathBuf,
+}
```
wen-coding, line 123: Hmm, should we save certificates as well?

core/Cargo.toml
carllin: fixed

core/src/alpenglow_consensus/certificate_pool.rs
carllin: This is not the official vote object, just a placeholder until the vote program gets pulled in

core/src/alpenglow_consensus/certificate_pool.rs
carllin: Changed!

core/src/alpenglow_consensus/certificate_pool.rs
carllin: changed

core/src/alpenglow_consensus/certificate_pool.rs
carllin: yeah this was primarily convenient for testing

The caller is replaystage which will be adding votes kind of like we add votes to tower right now. 

The current function for tower::add_vote returns a new root, I was thinking this would do something similar for new finalized certificates for parity

core/src/alpenglow_consensus/certificate_pool.rs
carllin: This is a bug, good catch!

core/src/alpenglow_consensus/certificate_pool.rs
carllin: im assuming here there will be future logic to ensure that transactions in a certificate cannot fail. Otherwise you run into danger that the certificate is incomplete.

Updated comment

core/src/alpenglow_consensus/certificate_pool.rs
carllin: This could technically also be the finalized certificate because finalized is stronger than notarized, I will update

core/src/alpenglow_consensus/certificate_pool.rs
carllin: Yeah, left an outstanding issue to think about this: https://github.com/anza-xyz/alpenglow/issues/38

I think naively you could have two separate pools for two different epochs, and when you have a vote that spans two epochs, you break up the range into the two parts, one per epoch.

More complicated solution is splitting votes across epoch in the same tree, not sure it's worth it

core/src/alpenglow_consensus/skip_pool.rs
carllin: yup, changed!

core/src/alpenglow_consensus/skip_pool.rs
carllin: that's right, changed!

core/src/alpenglow_consensus/skip_pool.rs
carllin: renamed to `threshold_stake`

core/src/alpenglow_consensus/skip_pool.rs
carllin: yeah, that makes sense

core/src/alpenglow_consensus/skip_pool.rs
carllin: After notarizing slot 102, in order to skip slot 103, this validator either had to:

1. See a notarization certificate for slot 102, in which case the previous skips don't matter because in our simple fork choice model right now, we will always try to build off of 102 + any skips

2. See a skip certificate for slot 102, in which case the validator should update their notarization vote to a skip vote, so they should have `skip(100, 102)`

core/src/alpenglow_consensus/vote_history_storage.rs
carllin: yeah saving latest finalization/skip/notarization certificates is a good idea: https://github.com/anza-xyz/alpenglow/issues/39

core/src/alpenglow_consensus/vote_certificate.rs
carllin: fixed!

core/src/alpenglow_consensus/vote_certificate.rs
carllin: I haven't needed an ordered map yet, primarily just need to query for the highest which is tracked in a different field

core/src/alpenglow_consensus/vote_certificate.rs
carllin: yeah I can move it to the top level module

core/src/alpenglow_consensus/certificate_pool.rs
carllin: ah yeah, forgot to clean this up, good catch

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: I assume get_certificate() below will be used for saving the actual certificates in the block you created, just thinking whether there would be any benefit to sort them by key. Not sure there is, for debugging maybe?

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// No vote since restart
+    #[default]
+    Uninitialized,
+    /// Non voting validator
+    NonVoting,
+    /// Hot spare validator
+    HotSpare,
+    /// Successfully generated vote tx with blockhash
+    Blockhash(Slot, Hash),
+}
```
wen-coding, line 21: I'm suddenly confused, why do we need to track this?

And do we need to save VoteHistory if we save all the (completed or incomplete) certs?

core/src/banking_stage/consumer.rs
```diff
@@ -108,6 +109,13 @@ impl Consumer {
         }
     }
 
+    pub fn create_consumer(poh_recorder: &RwLock<PohRecorder>) -> Consumer {
+        let (replay_vote_sender, _replay_vote_receiver) = unbounded();
+        let committer = Committer::new(None, replay_vote_sender, Arc::default());
+        let transaction_recorder: TransactionRecorder = poh_recorder.read().unwrap().new_recorder();
+        Consumer::new(committer, transaction_recorder, QosService::new(0), None)
```
wen-coding: Would 0 always work here? Do we need to reserve an unique ID for our cert service?

core/src/alpenglow_consensus/skip_pool.rs
carllin: Don't think it's as necessary here because every validator has at most one vote, might be worth to clean outdated validators at every epoch boundary though if their stake drops to 0: https://github.com/anza-xyz/alpenglow/issues/40

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,737 @@
...
+        Ok(None)
+    }
+
+    fn get_notarization_certificate(&self, slot: Slot) -> Option<Vec<VersionedTransaction>> {
+        self.certificates
+            .get(&(slot, CertificateType::Notarize))
+            .map(|certificate| certificate.get_certificate())
+    }
+
+    fn commit_certificate(
```
wen-coding: nit: maybe add a comment here saying the certificate is always committed first?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -0,0 +1,737 @@
...
+        bank: &Arc<Bank>,
+        poh_recorder: &RwLock<PohRecorder>,
+        certificate: Vec<VersionedTransaction>,
+    ) -> bool {
+        let consumer = Consumer::create_consumer(poh_recorder);
+        let runtime_transactions: Result<Vec<RuntimeTransaction<SanitizedTransaction>>, _> =
+            certificate
+                .into_iter()
+                .map(|versioned_tx| {
```
wen-coding: For the votes already in certificate, would the banking stage pack them again?

core/src/replay_stage.rs
```diff
@@ -2084,53 +2090,78 @@ impl ReplayStage {
...
-            poh_slot,
+            my_leader_slot,
             parent_slot
         );
 
-        if let Some(next_leader) = leader_schedule_cache.slot_leader_at(poh_slot, Some(&parent)) {
+        if let Some(next_leader) =
+            leader_schedule_cache.slot_leader_at(my_leader_slot, Some(&parent))
+        {
             if !has_new_vote_been_rooted {
```
wen-coding, line 2183: has_new_vote_been_rooted still tracks the old Vote format? Do we still need this vote for Alpenglow? If you are the leader for slot X and you have skip/notar cert for slot X-1 you'd start building your block right?

core/src/replay_stage.rs
```diff
@@ -2084,53 +2090,78 @@ impl ReplayStage {
...
-            warn!("{} already have bank in forks at {}?", my_pubkey, poh_slot);
+        if bank_forks.read().unwrap().get(my_leader_slot).is_some() {
+            warn!(
+                "{} already have bank in forks at {}?",
+                my_pubkey, my_leader_slot
+            );
             return false;
         }
         trace!(
             "{} poh_slot {} parent_slot {}",
```
wen-coding: nit: poh_slot -> my_leader_slot

core/src/replay_stage.rs
```diff
@@ -2084,53 +2090,78 @@ impl ReplayStage {
...
 
-        trace!("{} reached_leader_slot", my_pubkey);
+            trace!("{} reached_leader_slot", my_pubkey);
+        }
+
+        #[cfg(feature = "alpenglow")]
+        {
+            parent_slot = cert_pool.highest_unskipped_certificate_slot();
+            // maybe my next slot
+            my_leader_slot = cert_pool.highest_certificate_slot() + 1;
```
wen-coding: I guess we haven't implemented the "send all 4 slots belonging to the same leader together" optimization yet? (Algorithm 2 line 6: just send out following blocks without waiting for cert)

core/src/replay_stage.rs
```diff
@@ -2140,86 +2171,110 @@ impl ReplayStage {
...
 
-            update_bank_forks_and_poh_recorder_for_new_tpu_bank(
+                update_bank_forks_and_poh_recorder_for_new_tpu_bank(
+                    bank_forks,
+                    poh_recorder,
+                    tpu_bank,
+                    track_transaction_indexes,
+                );
+            }
+            #[cfg(feature = "alpenglow")]
```
wen-coding: Err, it becomes harder to figure which code block 

> #[cfg(feature = "alpenglow")]

covers in a large function. Can we split code covered by `#[cfg(feature = "alpenglow")]` and those by `#[cfg(feature = "alpenglow")]` into different functions?

core/src/replay_stage.rs
carllin: nope, no optimizations

core/src/alpenglow_consensus/certificate_pool.rs
carllin: don't think so, should fail the duplicate transaction check

core/src/alpenglow_consensus/certificate_pool.rs
carllin: split out the functionality into banking_stage

core/src/replay_stage.rs
carllin: think the purpose of this flag is to avoid creating a duplicate block for a slot you already created: https://github.com/solana-labs/solana/pull/15607

core/src/alpenglow_consensus/certificate_pool.rs
carllin: Don't think that should be a function comment,  this can be used to commit any certificate

Added it to where the function is called

core/src/alpenglow_consensus/vote_history.rs
carllin: Yeah, you don't want to send a Finalization if you sent a Skip, it's slashable

core/src/banking_stage/consumer.rs
carllin: Hmm yeah this is a good catch, generally this is only used for metrics and is set to `(thread_index as u32).saturating_add(NUM_VOTE_PROCESSING_THREADS)`

Maybe something like INT_MAX might work

core/src/replay_stage.rs
carllin: yeah was torn on this, felt like it might be easier to catch newly introduced functionality from the main repo if we had everything in the same function, especially if it's sharing a lot of functionality.

But you're right, it would be a lot more readable

PR #43, opened 2025-02-18
Add alpenglow poh

carllin:
#### Problem
Poh needlessly hashes/complicated for alpenglow

#### Summary of Changes
1.  On Alpenglow signal, PohService shuts down its current tick producer and migrates to a simple Alpenglow version that only records and ticks once to signal the end of the slot.
2. Poh  sets bank tick_count to `max_tick_height - 1`  right before it ticks to mimick bank being complete

Tested the above, this is sufficient for a single leader to continually produce Alpenglow blocks on a single node network because replay won't verify those blocks

TODO: 
Change replay verification on migration to stop counting number of ticks/verifying hashes. Only thing necessary is entry count and reading the last tick as a signal for the end of the block

Fixes #

### DIFF
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -10,8 +10,6 @@
 //! For Entries:
 //! * recorded entry must be >= WorkingBank::min_tick_height && entry must be < WorkingBank::max_tick_height
 //!
-#[cfg(feature = "dev-context-only-utils")]
-use solana_ledger::genesis_utils::{create_genesis_config, GenesisConfigInfo};
 use {
     crate::{leader_bank_notifier::LeaderBankNotifier, poh_service::PohService},
     crossbeam_channel::{
@@ -36,11 +34,16 @@ use {
         num::Saturating,
         sync::{
             atomic::{AtomicBool, Ordering},
-            Arc, Mutex, RwLock,
+            Arc, Mutex,
         },
         time::{Duration, Instant},
     },
     thiserror::Error,
+};
+#[cfg(feature = "dev-context-only-utils")]
+use {
+    solana_ledger::genesis_utils::{create_genesis_config, GenesisConfigInfo},
+    std::sync::RwLock,
 };
 
 pub const GRACE_TICKS_FACTOR: u64 = 2;
@@ -1087,6 +1090,7 @@ impl PohRecorder {
         poh_config: &PohConfig,
         poh_timing_point_sender: Option<PohTimingSender>,
         is_exited: Arc<AtomicBool>,
+        is_alpenglow_enabled: bool,
     ) -> (Self, Receiver<WorkingBankEntry>, Receiver<Record>) {
         let tick_number = 0;
         let poh = Arc::new(Mutex::new(Poh::new_with_slot_info(
@@ -1137,8 +1141,8 @@ impl PohRecorder {
                 delay_leader_block_for_pending_fork,
                 last_reported_slot_for_pending_fork: Arc::default(),
                 is_exited,
-                is_alpenglow_enabled: false,
-                use_alpenglow_tick_produer: false,
+                is_alpenglow_enabled,
+                use_alpenglow_tick_produer: is_alpenglow_enabled,
             },
             working_bank_receiver,
             record_receiver,
@@ -1149,6 +1153,7 @@ impl PohRecorder {
     /// * bank - the LastId's queue is updated on `tick` and `record` events
     /// * sender - the Entry channel that outputs to the ledger
     #[allow(clippy::too_many_arguments)]
+    #[cfg(feature = "dev-context-only-utils")]
     pub fn new(
         tick_height: u64,
         last_entry_hash: Hash,
@@ -1161,6 +1166,7 @@ impl PohRecorder {
         is_exited: Arc<AtomicBool>,
     ) -> (Self, Receiver<WorkingBankEntry>, Receiver<Record>) {
         let delay_leader_block_for_pending_fork = false;
+        let is_alpenglow_enabled = false;
         Self::new_with_clear_signal(
             tick_height,
             last_entry_hash,
@@ -1174,6 +1180,7 @@ impl PohRecorder {
             poh_config,
             None,
             is_exited,
+            is_alpenglow_enabled,
         )
     }
 
@@ -1205,6 +1212,7 @@ impl PohRecorder {
     }
 }
 
+#[cfg(feature = "dev-context-only-utils")]
 fn do_create_test_recorder(
     bank: Arc<Bank>,
     blockstore: Arc<Blockstore>,
@@ -1254,6 +1262,7 @@ fn do_create_test_recorder(
     (exit, poh_recorder, poh_service, entry_receiver)
 }
 
+#[cfg(feature = "dev-context-only-utils")]
 pub fn create_test_recorder(
     bank: Arc<Bank>,
     blockstore: Arc<Blockstore>,
@@ -1268,6 +1277,7 @@ pub fn create_test_recorder(
     do_create_test_recorder(bank, blockstore, poh_config, leader_schedule_cache, false)
 }
 
+#[cfg(feature = "dev-context-only-utils")]
 pub fn create_test_recorder_with_index_tracking(
     bank: Arc<Bank>,
     blockstore: Arc<Blockstore>,
@@ -1917,6 +1927,7 @@ mod tests {
                 &PohConfig::default(),
                 None,
                 Arc::new(AtomicBool::default()),
+                false,
             );
         poh_recorder.set_bank_for_test(bank);
         poh_recorder.clear_bank();
--- a/poh/src/poh_service.rs
+++ b/poh/src/poh_service.rs
@@ -111,65 +111,73 @@ impl PohService {
         record_receiver: Receiver<Record>,
     ) -> Self {
         let poh_config = poh_config.clone();
+        let is_alpenglow_enabled = poh_recorder.read().unwrap().is_alpenglow_enabled;
         let tick_producer = Builder::new()
             .name("solPohTickProd".to_string())
             .spawn(move || {
-                if poh_config.hashes_per_tick.is_none() {
-                    if poh_config.target_tick_count.is_none() {
-                        Self::low_power_tick_producer(
-                            poh_recorder.clone(),
-                            &poh_config,
-                            &poh_exit,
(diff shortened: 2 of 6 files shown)


poh/src/poh_recorder.rs
```diff
@@ -318,6 +318,8 @@ pub struct PohRecorder {
     delay_leader_block_for_pending_fork: bool,
     last_reported_slot_for_pending_fork: Arc<Mutex<Slot>>,
     pub is_exited: Arc<AtomicBool>,
+    pub is_aspenglow_enabled: bool,
```
bw-solana: is this the luxury resort version of alpenglow?

runtime/src/bank.rs
```diff
@@ -6034,6 +6034,10 @@ impl Bank {
         self.tick_height.load(Relaxed)
     }
 
+    pub fn set_tick_height(&self, tick_height: u64) {
+        self.tick_height.store(tick_height, Relaxed)
```
bw-solana, line 6036: Does this need to have more stringent ordering?

poh/src/poh_service.rs
```diff
@@ -160,6 +189,65 @@ impl PohService {
...
+                        // By this point the leader should have committed their certificates,
+                        // so it's safe to start the timer
+                        start: Instant::now(),
+                    });
+                    info!("current ns per slot: {}", leader_bank.ns_per_slot);
+                    current_slot_time = Duration::from_nanos(leader_bank.ns_per_slot as u64);
+                }
+            }
+
+            if let Some(CurrentLeaderBank { bank, start }) = &current_leader_bank {
```
bw-solana:
```
let Some() else {
    continue;
}
```

would be nice here to reduce nesting

poh/src/poh_service.rs
```diff
@@ -160,6 +189,65 @@ impl PohService {
...
+                Self::read_record_receiver_and_process(
+                    &poh_recorder,
+                    &record_receiver,
+                    current_slot_time,
```
bw-solana: should this be `remaining_slot_time`?

poh/src/poh_service.rs
```diff
@@ -131,17 +138,39 @@ impl PohService {
...
+                        record_receiver.clone(),
                         Self::target_ns_per_tick(
                             ticks_per_slot,
                             poh_config.target_tick_duration.as_nanos() as u64,
                         ),
                     );
                 }
+
+                // Migrate to alpenglow PoH
+                if !poh_exit.load(Ordering::Relaxed)
```
bw-solana: I'm assuming this will eventually need to be feature gated so that block producer and replay are in sync?

carllin: the `poh_recorder.read().unwrap().is_aspenglow_enabled` flag is only set by ReplayStage if the alpenglow feature is activated, so the entire migration here will be feature gated

poh/src/poh_recorder.rs
carllin: heard the ass-penglow resort is kind of ass

poh/src/poh_service.rs
carllin: Ah yup, good catch

runtime/src/bank.rs
carllin: I don't think it's less strict than the current state, but that doesn't mean it's necessarily correct xD

I think this upholdsthe same ordering as the current `register_tick()` function which does the current tick incrementation https://github.com/anza-xyz/alpenglow/blob/fad37613b33b2148636719cb3d06e3cab6fb0c45/runtime/src/bank.rs#L3072 because the usage will be:

1. pohservice  set_tick_height
2. pohservice -> poh_recorder -> register_tick (today relaxed store)
3. replay_stage checking if bank.is_complete() with ticks to freeze (today relaxed read)

Guessing it's ok today because we have to acquire locks before reading any of the shared state in bank like the blockhash_queue(). We could switch to acquire-release if we modify bank `is_complete()` and `register_tick()`

poh/src/poh_recorder.rs
```diff
@@ -318,6 +318,8 @@ pub struct PohRecorder {
     delay_leader_block_for_pending_fork: bool,
     last_reported_slot_for_pending_fork: Arc<Mutex<Slot>>,
     pub is_exited: Arc<AtomicBool>,
+    pub is_alpenglow_enabled: bool,
+    pub is_poh_service_migrated: bool,
```
wen-coding: nit: not sure what "is_poh_service_migrated" means here, and how it's different from is_alpenglow_enabled above. How about names like "use_alpenglow_tick_produer"?

poh/src/poh_recorder.rs
```diff
@@ -632,11 +634,28 @@ impl PohRecorder {
             ))
     }
 
+    pub fn migrate_poh_to_alpenglow(&mut self) {
```
wen-coding: nit: Are we doing anything other than setting PoH to low power mode here? This name is vague.

poh/src/poh_recorder.rs
```diff
@@ -879,6 +898,33 @@ impl PohRecorder {
...
 
+    pub fn tick_alpenglow(&mut self, slot_max_tick_height: u64) {
+        let (poh_entry, tick_lock_contention_us) = measure_us!({
+            let mut poh_l = self.poh.lock().unwrap();
+            poh_l.tick()
+        });
+        self.tick_lock_contention_us += tick_lock_contention_us;
+
+        if let Some(poh_entry) = poh_entry {
+            self.tick_height = slot_max_tick_height;
```
wen-coding, line 912: I'm confused, so the plan is we will only have one tick forever in all Alpenglow blocks? Do we need to change tick verification in replay at the same time?

poh/src/poh_service.rs
```diff
@@ -131,17 +138,39 @@ impl PohService {
...
+                    // TODO: Can essentailly replace this with no ticks
+                    // once we properly remove poh/entry verification in replay
+                    {
+                        let mut w_poh_recorder = poh_recorder.write().unwrap();
+                        w_poh_recorder.migrate_poh_to_alpenglow();
+                        w_poh_recorder.is_poh_service_migrated = true;
+                    }
+                    Self::alpenglow_tick_producer(poh_recorder, &poh_exit, record_receiver);
+                }
+                //}
```
wen-coding: nit: remove this line?

poh/src/poh_service.rs
```diff
@@ -160,6 +189,65 @@ impl PohService {
...
+        record_receiver: Receiver<Record>,
+    ) {
+        info!("starting alpenglow tick producer");
+        let mut current_leader_bank = None;
+        let mut current_slot_time = Duration::from_millis(DEFAULT_MS_PER_SLOT);
+        let leader_bank_notifier = poh_recorder.read().unwrap().new_leader_bank_notifier();
+        while !poh_exit.load(Ordering::Relaxed) {
+            // Wait for a new leader bank to be set in PohRecorder
+            let leader_bank = leader_bank_notifier
+                .get_or_wait_for_in_progress(Duration::from_millis(50))
```
wen-coding, line 211: Since you can't do much if you don't have a leader bank, is 50ms a good time interval to wait here?

poh/src/poh_service.rs
```diff
@@ -110,17 +117,17 @@ impl PohService {
                 if poh_config.hashes_per_tick.is_none() {
```
wen-coding: If Alpenglow is already enabled, should we skip this if statement?

poh/src/poh_service.rs
```diff
@@ -160,6 +189,65 @@ impl PohService {
...
+                if Some(leader_bank.slot())
+                    != current_leader_bank.as_ref().map(
+                        |current_leader_bank: &CurrentLeaderBank| current_leader_bank.bank.slot(),
+                    )
+                {
+                    current_leader_bank = Some(CurrentLeaderBank {
+                        bank: leader_bank.clone(),
+                        // By this point the leader should have committed their certificates,
+                        // so it's safe to start the timer
+                        start: Instant::now(),
```
wen-coding, line 225: If we are packing a bank then the validator got restarted in the middle, what will happen? I'm guessing we will just repack the whole thing, so start will be correct again?

Is it possible that we already sent out some shreds then the validator is restarted  before the leader bank finishes? Would the start be wrong here?

poh/src/poh_service.rs
```diff
@@ -160,6 +189,65 @@ impl PohService {
...
+                    });
+                    info!("current ns per slot: {}", leader_bank.ns_per_slot);
```
wen-coding, line 227: Does this value change? Do we need to log every time we start a leader bank?

carllin: We can adjust Poh difficulty today, so I think it would be nice to be able to adjust slot times if we ever want to shorten them, which was on the roadmap

poh/src/poh_service.rs
carllin: Just like today we always check blockstore to see if we have existing shreds before we create a leader bank

poh/src/poh_service.rs
carllin: I think other components like banking_stage used 50ms, so I used that. It just needs to be reasonable enough where we can occasionally check the exit condition for poh_service

poh/src/poh_service.rs
carllin: removed

poh/src/poh_recorder.rs
carllin: yeah just one tick in all alpenglow blocks to signal the end of a slot

Yeah, need to remove tick verification in replay when alpenglow is enabled, that can be done in another PR

poh/src/poh_recorder.rs
carllin: replaced!

carllin: replaced!

poh/src/poh_recorder.rs
carllin: it's just migrating to the alpenglow poh tick producer, renamed to `migrate_to_alpenglow_poh`

poh/src/poh_service.rs
carllin: replaced!

poh/src/poh_service.rs
carllin: yup, thats a good point, should skip the migration if we have a frozen bank on startup greater than the first alpenglow slot, added!

PR #44, opened 2025-02-18
Add alpenglow push vote path

carllin:
#### Problem
Alpenglow voting isn't handled in gossip or sent to tpu's

#### Summary of Changes
Add a wider fanout to tpu's and also handle alpenglow push vote into gossip

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -4392,6 +4392,9 @@ pub(crate) mod tests {
     use {
         super::*,
         crate::{
+            alpenglow_consensus::vote_history_storage::{
+                NullVoteHistoryStorage, VoteHistoryStorage,
+            },
             consensus::{
                 progress_map::{ValidatorStakeInfo, RETRANSMIT_BASE_DELAY_MS},
                 tower_storage::{FileTowerStorage, NullTowerStorage},
@@ -7672,6 +7675,7 @@ pub(crate) mod tests {
             ..
         } = replay_blockstore_components(None, 10, None::<GenerateVotes>);
         let tower_storage = NullTowerStorage::default();
+        let vote_history_storage: NullVoteHistoryStorage = NullVoteHistoryStorage::default();
 
         let VoteSimulator {
             mut validator_keypairs,
@@ -7762,6 +7766,7 @@ pub(crate) mod tests {
             &cluster_info,
             &poh_recorder,
             &tower_storage,
+            &vote_history_storage,
             vote_info,
             Arc::new(connection_cache),
         );
@@ -7867,6 +7872,7 @@ pub(crate) mod tests {
             &cluster_info,
             &poh_recorder,
             &tower_storage,
+            &vote_history_storage,
             vote_info,
             Arc::new(connection_cache),
         );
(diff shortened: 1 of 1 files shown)


core/src/voting_service.rs
```diff
@@ -122,14 +132,38 @@ impl VotingService {
...
+                error!("Unable to save vote history to storage: {:?}", err);
+                std::process::exit(1);
+            }
+            measure.stop();
+            trace!("{measure}");
+        }
+
         // Attempt to send our vote transaction to the leaders for the next few slots
         const UPCOMING_LEADER_FANOUT_SLOTS: u64 = FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET;
+        const UPCOMING_LEADER_ALPENGLOW_FANOUT_SLOTS: u64 = 20;
```
wen-coding, line 150: Now we do want to push all-to-all, do we still need to change the fanout here?

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
         }
     }
 
+    /// Evict the oldest vote if gossip is full
+    fn find_alpenglow_vote_index_to_evict(&self) -> u8 {
+        let self_pubkey = self.id();
+        let mut num_crds_votes = 0;
+        let vote_index = {
+            let gossip_crds =
+                self.time_gossip_read_lock("gossip_read_push_vote", &self.stats.push_vote_read);
```
wen-coding: Add our own stats to separate from old votes stats?

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
...
+            (0..MAX_LOCKOUT_HISTORY as u8)
+                .filter_map(|ix| {
+                    let vote = CrdsValueLabel::Vote(ix, self_pubkey);
```
wen-coding, line 817: I still think we should give each type of votes its own queue, so that skips won't kick finalize out and vice versa.

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
...
+                    let vote: &CrdsData = gossip_crds.get(&vote)?;
+                    num_crds_votes += 1;
+                    match &vote {
+                        CrdsData::Vote(_, vote) => Some((vote.wallclock, ix)),
```
wen-coding, line 821: We won't have timestamp in finalize and skip.

core/src/voting_service.rs
```diff
@@ -122,14 +132,38 @@ impl VotingService {
             trace!("{measure}");
         }
 
+        if let VoteOp::PushAlpenglowVote {
+            saved_vote_history, ..
+        } = &vote_op
+        {
+            let mut measure = Measure::start("tower storage save");
```
wen-coding: nit: tower-> vote history

gossip/src/cluster_info.rs
carllin: think we can eventually switch to having constant indexes for each vote type, don't think its very liekely given it takes 32 votes to overwrite right now

gossip/src/cluster_info.rs
carllin: all crds gossip data is tagged with a wallclock

PR #45, opened 2025-02-18
remove consensus logic from replay

carllin:
#### Problem
Replay lacks alpenglow awareness

#### Summary of Changes
1. Call the Poh migration path from replay, but doesn't handle lockouts/epoch boundaries safely yet
2.  Handles starting a leader in replay
3. Handles pushing votes from replay

With this, we can now spin up a single alpenglow leader in a one node cluster and watch it make blocks

This builds on  https://github.com/anza-xyz/alpenglow/pull/44 and https://github.com/anza-xyz/alpenglow/pull/43, only the last commit is relevant
Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -2781,8 +2781,7 @@ impl ReplayStage {
         // TODO: Handle that finalization certificate can be received before we
         // have even frozen/replayed the bank
         if let Some(new_root) = new_alpenglow_root {
-            cert_pool.purge(new_root);
-            Self::check_and_handle_new_root(
+            Self::alpenglow_handle_new_root(
                 vote_bank,
                 new_root,
                 bank_forks,
@@ -2801,6 +2800,7 @@ impl ReplayStage {
                 vote_signatures,
                 epoch_slots_frozen_slots,
                 drop_bank_sender,
+                cert_pool,
             )?;
         }
         // TODO: update the commitmment cache
@@ -4724,6 +4724,51 @@ impl ReplayStage {
             }
         }
         duplicate_confirmed_forks
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    fn alpenglow_handle_new_root(
+        vote_bank: &Bank,
+        new_root: Slot,
+        bank_forks: &RwLock<BankForks>,
+        progress: &mut ProgressMap,
+        blockstore: &Blockstore,
+        leader_schedule_cache: &Arc<LeaderScheduleCache>,
+        accounts_background_request_sender: &AbsRequestSender,
+        rpc_subscriptions: &Arc<RpcSubscriptions>,
+        block_commitment_cache: &Arc<RwLock<BlockCommitmentCache>>,
+        heaviest_subtree_fork_choice: &mut HeaviestSubtreeForkChoice,
+        bank_notification_sender: &Option<BankNotificationSenderConfig>,
+        duplicate_slots_tracker: &mut DuplicateSlotsTracker,
+        duplicate_confirmed_slots: &mut DuplicateConfirmedSlots,
+        unfrozen_gossip_verified_vote_hashes: &mut UnfrozenGossipVerifiedVoteHashes,
+        has_new_vote_been_rooted: &mut bool,
+        voted_signatures: &mut Vec<Signature>,
+        epoch_slots_frozen_slots: &mut EpochSlotsFrozenSlots,
+        drop_bank_sender: &Sender<Vec<BankWithScheduler>>,
+        cert_pool: &mut CertificatePool,
+    ) -> Result<(), SetRootError> {
+        cert_pool.purge(new_root);
+        Self::check_and_handle_new_root(
+            vote_bank,
+            new_root,
+            bank_forks,
+            progress,
+            blockstore,
+            leader_schedule_cache,
+            accounts_background_request_sender,
+            rpc_subscriptions,
+            block_commitment_cache,
+            heaviest_subtree_fork_choice,
+            bank_notification_sender,
+            duplicate_slots_tracker,
+            duplicate_confirmed_slots,
+            unfrozen_gossip_verified_vote_hashes,
+            has_new_vote_been_rooted,
+            voted_signatures,
+            epoch_slots_frozen_slots,
+            drop_bank_sender,
+        )
     }
 
     #[allow(clippy::too_many_arguments)]


core/src/voting_service.rs
```diff
@@ -122,14 +132,38 @@ impl VotingService {
             trace!("{measure}");
         }
 
+        if let VoteOp::PushAlpenglowVote {
+            saved_vote_history, ..
+        } = &vote_op
+        {
+            let mut measure = Measure::start("tower storage save");
```
wen-coding: nit: not tower storage here

core/src/voting_service.rs
```diff
@@ -122,14 +132,38 @@ impl VotingService {
...
+                error!("Unable to save vote history to storage: {:?}", err);
+                std::process::exit(1);
+            }
+            measure.stop();
+            trace!("{measure}");
+        }
+
         // Attempt to send our vote transaction to the leaders for the next few slots
         const UPCOMING_LEADER_FANOUT_SLOTS: u64 = FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET;
+        const UPCOMING_LEADER_ALPENGLOW_FANOUT_SLOTS: u64 = 20;
```
wen-coding: Why 20? We should probably put reasoning here.

core/src/voting_service.rs
```diff
@@ -122,14 +132,38 @@ impl VotingService {
...
         #[cfg(test)]
         static_assertions::const_assert_eq!(UPCOMING_LEADER_FANOUT_SLOTS, 2);
+
+        let leader_fanout = {
+            match &vote_op {
+                // Alpenglow relies on leaders to propagate votes otherwise we have to rely
+                // on gossip, especially true for skip votes
```
wen-coding: I thought we would send all-to-all as well? Otherwise we can't guarantee 3 delta... We can discuss this in SLC

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
         }
     }
 
+    /// Evict the oldest vote if gossip is full
+    fn find_alpenglow_vote_index_to_evict(&self) -> u8 {
+        let self_pubkey = self.id();
+        let mut num_crds_votes = 0;
+        let vote_index = {
+            let gossip_crds =
+                self.time_gossip_read_lock("gossip_read_push_vote", &self.stats.push_vote_read);
```
wen-coding: nit: push_alpenglow_vote

And maybe give alpenglow vote its own stats?

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
...
+            (0..MAX_LOCKOUT_HISTORY as u8)
+                .filter_map(|ix| {
+                    let vote = CrdsValueLabel::Vote(ix, self_pubkey);
```
wen-coding: And our own label as well, don't share the same queue with the old votes.

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
...
     }
 
+    /// Evict the oldest vote if gossip is full
+    fn find_alpenglow_vote_index_to_evict(&self) -> u8 {
+        let self_pubkey = self.id();
+        let mut num_crds_votes = 0;
+        let vote_index = {
+            let gossip_crds =
+                self.time_gossip_read_lock("gossip_read_push_vote", &self.stats.push_vote_read);
+            (0..MAX_LOCKOUT_HISTORY as u8)
```
wen-coding: Is 32 enough for Alpenglow? You are sharing the same vote buffer for Notar/Skip/Final, and in the async execution world Final might be further away from Notar.

gossip/src/cluster_info.rs
```diff
@@ -803,6 +803,34 @@ impl ClusterInfo {
...
+                .filter_map(|ix| {
+                    let vote = CrdsValueLabel::Vote(ix, self_pubkey);
+                    let vote: &CrdsData = gossip_crds.get(&vote)?;
+                    num_crds_votes += 1;
+                    match &vote {
+                        CrdsData::Vote(_, vote) => Some((vote.wallclock, ix)),
```
wen-coding: Wait, there was a should_evict_vote check in the original code, we don't need it any more?

Also we need to think carefully whether we should evict the oldest vote. Could all the finalize votes be kicked out because we have too many skip/notar on the newer votes?

I'm inclined to say "please give each type of vote its own queue" now...

core/src/banking_stage.rs
```diff
@@ -812,6 +812,11 @@ pub fn alpenglow_update_bank_forks_and_poh_recorder_for_new_tpu_bank(
 ) -> bool {
     let tpu_bank = bank_forks.write().unwrap().insert(tpu_bank);
     let parent_slot = tpu_bank.parent_slot();
+    info!(
```
wen-coding: Do we need the info log here? Can we just log at maybe_start_leader?

core/src/replay_stage.rs
```diff
@@ -668,6 +680,34 @@ impl ReplayStage {
...
+
+            // Alpenglow specific objects
+            // TODO: Properly initialize VoteHistory from storage
+            let mut vote_history = VoteHistory::new(my_pubkey, bank_forks.read().unwrap().root());
+            let mut first_alpenglow_slot = bank_forks
+                .read()
+                .unwrap()
+                .root_bank()
+                .feature_set
+                .activated_slot(&solana_feature_set::alpenglow::id());
```
wen-coding: When validators enter the next Epoch at different speed, would it happen that some validators don't have enough votes for root to enter the new Epoch, so they were forever stuck in the old Epoch? (Because there are no new TowerBFT votes to confirm the fork they are on maybe?)

Should we enable Alpenglow at Epoch N+2 to make sure everyone's on the same page?

Also, we are checking this outside the loop? Could we enter a new Epoch inside the loop?

core/src/replay_stage.rs
```diff
@@ -759,402 +799,559 @@ impl ReplayStage {
...
-                    &ancestor_hashes_replay_update_sender,
-                    &mut purge_repair_slot_counter,
-                );
-                process_popular_pruned_forks_time.stop();
-
-                // Check to remove any duplicated slots from fork choice
-                let mut process_duplicate_slots_time = Measure::start("process_duplicate_slots");
-                if !tpu_has_bank {
-                    Self::process_duplicate_slots(
+                if !is_alpenglow_migration_complete {
```
wen-coding, line 812: Is it possible to move all non_alpenglow operations into a separate function for readability?

core/src/replay_stage.rs
```diff
@@ -759,402 +799,559 @@ impl ReplayStage {
...
+                        process_ancestor_hashes_duplicate_slots_time.as_us(),
+                        process_duplicate_confirmed_slots_time.as_us(),
+                        process_unfrozen_gossip_verified_vote_hashes_time.as_us(),
+                        process_popular_pruned_forks_time.as_us(),
+                        process_duplicate_slots_time.as_us(),
+                        dump_then_repair_correct_slots_time.as_us(),
+                        retransmit_not_propagated_time.as_us(),
+                    );
+                } else {
+                    // Alpenglow specific logic
```
wen-coding: similarly, move all Alpenglow specific logic into separate function for better readability?

core/src/replay_stage.rs
```diff
@@ -759,402 +799,559 @@ impl ReplayStage {
...
+                    let highest_frozen_bank = bank_forks.read().unwrap().highest_frozen_bank();
+                    assert!(highest_frozen_bank.is_frozen());
+                    if poh_recorder.read().unwrap().start_slot() != highest_frozen_bank.slot() {
+                        // Important to keep Poh somewhat accurate for
+                        // parts of the system relying on PohRecorder::would_be_leader()
+
+                        // Note that like today, if this occurs during your leader slot
+                        // it will cause you to dump your leader slot. This should be ok because
+                        // the fact that there is a greater/valid slot than your own must mean there
+                        // was a skip certificate for your slot, so it's ok to abandon your leader slot
```
wen-coding: Hmm, your comment implies:
if poh_recorder.read().unwrap().start_slot() **<** highest_frozen_bank.slot() {
but your actual comparison uses **!=**
Should we reset poh_recorder if start_slot is larger than highest_frozen_bank.slot()?

core/src/replay_stage.rs
```diff
@@ -759,402 +799,559 @@ impl ReplayStage {
...
-                let mut retransmit_not_propagated_time =
-                    Measure::start("retransmit_not_propagated_time");
-                Self::retransmit_latest_unpropagated_leader_slot(
-                    &poh_recorder,
-                    &retransmit_slots_sender,
-                    &mut progress,
-                );
-                retransmit_not_propagated_time.stop();
+                    // Try to finalize the highest notarized block
+                    let highest_notarized_slot = cert_pool.highest_notarized_slot();
```
wen-coding: Should we finalize the highest non-finalized block instead?

Could it be notarizations keep coming:
Slot 100 notarized -> Didn't get to finalize Slot 100 before Slot 101 is also notarized -> Didn't get to finalize Slot 101 before Slot 102 is also notarized ...

I know we will eventually get all of them finalized, but that might mean Slot 100 finalization takes more than 3 delta.

core/src/replay_stage.rs
```diff
@@ -759,402 +799,559 @@ impl ReplayStage {
...
+                    if vote_history.latest_finalize_vote.slot() != highest_notarized_slot
+                        && !vote_history.is_slot_skipped(highest_notarized_slot)
+                    {
+                        let maybe_vote_bank =
+                            bank_forks.read().unwrap().get(highest_notarized_slot);
+                        if let Some(vote_bank) = maybe_vote_bank {
+                            if vote_bank.is_frozen() {
```
wen-coding: If highest_notarized_slot is not frozen, I think we should start repair somewhere?

core/src/replay_stage.rs
```diff
@@ -2125,14 +2295,33 @@ impl ReplayStage {
                     }
                 }
             } else {
-                // We are now in full alpenglow mode
+                // We are in full alpenglow mode
+                info!(
+                    "maybe_start_leader hucs: {}, hcs: {}",
```
wen-coding: hucs and hcs are not easy to decode, maybe "unskipped" and "has_certificate"?

Why don't we just print highest notarized/finalized/skipped slot here?

core/src/replay_stage.rs
```diff
@@ -2589,50 +2884,147 @@ impl ReplayStage {
...
+    ) -> GenerateVoteTxResult {
+        if !bank.is_startup_verification_complete() {
+            info!("startup verification incomplete, so unable to vote");
+            return GenerateVoteTxResult::Failed;
+        }
+
+        if authorized_voter_keypairs.is_empty() {
+            return GenerateVoteTxResult::NonVoting;
+        }
+        if let Some(slot) = wait_to_vote_slot {
```
wen-coding, line 2946: Does this still work in Alpenglow? In PoH if you wait long enough some leader will eventually start building for a future slot X you are waiting on, but I believe now we need to send skip?

core/src/replay_stage.rs
carllin: repair will happen in the background like it does today based on votes in gossip, we just need to have the repair structure read the new votes https://github.com/anza-xyz/alpenglow/issues/56

core/src/replay_stage.rs
carllin: I think this is only used to manipulate tests in local cluster, should be fine

core/src/replay_stage.rs
carllin: migration will not be blocked on votes from the previous epoch, you just need to respect lockout and switching rules for all proposed alpenglow slots until at least one alpenglow slot gets a notarization

core/src/replay_stage.rs
carllin: It's impossible for start slot > highest_frozen_bank, we only ever start leader banks from parents that are

1. notarized
2. frozen

which necessarily means the highest frozen bank > start slot, added an assert and a comment

core/src/replay_stage.rs
carllin: yup, refactored

core/src/replay_stage.rs
carllin: refactored

core/src/replay_stage.rs
```diff
@@ -2126,14 +2204,37 @@ impl ReplayStage {
...
+                // TODO: need to keep the ticks around for parent slots in previous epoch
+                // because reset below will delete those ticks
+                info!(
+                    "initiating alpenglow migration from maybe_start_leader() for slot {}",
+                    maybe_my_leader_slot
+                );
+                Self::initiate_alpenglow_migration(poh_recorder, is_alpenglow_migration_complete);
+            }
+        }
+
```
wen-coding, line 2466: Can't comment on line 2254:
Suddenly confused:
```
if parent_slot < *first_alpenglow_slot.as_ref().unwrap_or(&u64::MAX) {
            assert!(parent_bank.is_frozen());
}
```
Why don't we need to make sure parent_bank is frozen if we are past first alpenglow slot?

carllin: Ah yeah, this is a bug from moving the start leader logic out of the certificate pool, where we used to check the alpenglow slot is frozen before starting the leader.

Because in alpenglow mode we set the parent_slot to `cert_pool.highest_not_skip_certificate_slot()`, we can have a certificate for a non frozen slot if our replay is slow. This assert exists in today's code because parent bank is always frozen

We need to return in alpenglow mode if  the parent bank isn't frozen. I need to refactor this alpegnlow-specific logic anyways so will do that

wen-coding: Sounds perfect. When you refactor, please try to split TowerBFT and Alpenglow logic as much as possible. It's really hard to read the code when the two are mixed together. I know we have to do it somehow for the migration, but hopefully we can at least split things into smaller functions.

carllin: split up the logic in `maybe_start_leader()`

PR #59, opened 2025-02-28
PohService needs to set `use_alpenglow_tick_producer` flag on startup 

carllin:
#### Problem
flag `use_alpenglow_tick_producer` is not enabled at startup

#### Summary of Changes

Fixes #

### DIFF
--- a/poh/src/poh_service.rs
+++ b/poh/src/poh_service.rs
@@ -159,17 +159,6 @@ impl PohService {
                     && poh_recorder.read().unwrap().is_alpenglow_enabled
                     {
                         info!("Migrating poh service to alpenglow tick producer");
-                        // Important this is called *before* any new alpenglow
-                        // leaders call `set_bank()`, otherwise, the old PoH
-                        // tick producer will still tick in that alpenglow bank
-                        //
-                        // TODO: Can essentailly replace this with no ticks
-                        // once we properly remove poh/entry verification in replay
-                        {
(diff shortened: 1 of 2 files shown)


poh/src/poh_service.rs
```diff
@@ -159,24 +159,25 @@ impl PohService {
                     && poh_recorder.read().unwrap().is_alpenglow_enabled
```
wen-coding, line 159: Do you still need this test? Maybe just say the only way it exits the previous loop is either:
1. poh_exit is true
2. alpenglow is enabled
So only exit the loop if 1 is true.
Then you can move the info loop to line 169~175?

carllin: The info log is only necessary if we started in not alpenglow mode, and then migrated to alpenglow. If we started in alpenglow mode, there's no need to print that we're migrating

wen-coding: Hmm ok, we can also do:
if poh_exit.load(Ordering::Relaxed) {
     return;
} else { // must exit because of Alpenglow
    info(...)
}
Your choice.

carllin: updated!

carllin: hmm actually I might like the original better, I want to explicitly check that alpenglow was enabled in case somebody else adds another path/error on which poh would exit

wen-coding: Ok, that's fair.

PR #63, opened 2025-03-02
integrate aspenglow vote program

AshwinSekar:
chido is working on getting buildkite to recognize the PAT

### DIFF
--- a/ci/docker-run.sh
+++ b/ci/docker-run.sh
@@ -95,7 +95,7 @@ if [[ -z "$SOLANA_DOCKER_RUN_NOSETUID" ]]; then
     --user "$(id -u):$(id -g)"
     --volume "/etc/passwd:/etc/passwd:ro"
     --volume "/etc/group:/etc/group:ro"
-    --volume "/var/lib/buildkite-agent/.ssh:/var/lib/buildkite-agent/.ssh:ro"
+    --volume "/var/lib/buildkite-agent:/var/lib/buildkite-agent"
   )
 fi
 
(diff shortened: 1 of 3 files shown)


.github/workflows/cargo.yml
```diff
@@ -1,4 +1,6 @@
 name: Cargo
+with:
+  token: ${{ secrets.AG_PAT }}
```
AshwinSekar: @yihau i've added the PAT as a secret to this repo, but not sure how we can put it into the buildkite agent so they can ssh for the dependency:
```
alpenglow-vote = { git = "ssh://git@github.com/solana-program/alpenglow-vote.git", rev = "8942fe9" }
```

yihau: I spent some time understanding the real issue. it's a long story. tldr, I will need some time to update the Buildkite agent part and will provide an update here.
btw, we can remove `.github/*` modification in this PR. we don't run Github Actions in this repo.

yihau: can use this patch as a workaround: https://github.com/anza-xyz/alpenglow/commit/1390cc6a2e01763e0909ace7fb79fbac9b4f1b6d

core/src/alpenglow_consensus/vote_history_storage.rs
```diff
@@ -168,159 +168,3 @@ impl VoteHistoryStorage for FileVoteHistoryStorage {
         Ok(())
     }
 }
-
-pub struct EtcdVoteHistoryStorage {
```
AshwinSekar, line 172: @carllin removed this guy while i was here since it's no longer supported and planned for removal in 3.0

carllin: amazing, kill it all

PR #66, opened 2025-03-03
Handle set new root on new frozen slots

carllin:
#### Problem
Slots can get a finalization certificate before we freeze them

#### Summary of Changes
On bank freeze, check if we have the finalization certificate

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -768,7 +768,7 @@ impl ReplayStage {
                     let r_bank_forks = bank_forks.read().unwrap();
                     (r_bank_forks.ancestors(), r_bank_forks.descendants())
                 };
-                let completed_slots = Self::replay_active_banks(
+                let new_frozen_slots = Self::replay_active_banks(
                     &blockstore,
                     &bank_forks,
                     &my_pubkey,
@@ -803,10 +803,10 @@ impl ReplayStage {
                     first_alpenglow_slot,
                     &mut is_alpenglow_migration_complete,
                 );
-                let did_complete_bank = !completed_slots.is_empty();
+                let did_complete_bank = !new_frozen_slots.is_empty();
                 replay_active_banks_time.stop();
 
-                let new_frozen_finalized_slot = completed_slots
+                let new_frozen_finalized_slot = new_frozen_slots
                     .into_iter()
                     .filter(|slot| cert_pool.is_finalized_slot(*slot))
                     .max();
@@ -3761,7 +3761,7 @@ impl ReplayStage {
         // TODO: See if processing of blockstore replay results and bank completion can be made thread safe.
         let mut tx_count = 0;
         let mut execute_timings = ExecuteTimings::default();
-        let mut completed_slots = vec![];
+        let mut new_frozen_slots = vec![];
         for replay_result in replay_result_vec {
             if replay_result.is_slot_dead {
                 continue;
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -805,8 +805,50 @@ impl ReplayStage {
...
+                    .max();
+                if let Some(new_frozen_root) = new_frozen_root {
+                    let (current_root_slot, new_root_bank) = {
+                        let r_bank_forks = bank_forks.read().unwrap();
+                        (
+                            r_bank_forks.root(),
+                            r_bank_forks.get(new_frozen_root).unwrap(),
+                        )
+                    };
+                    assert!(new_frozen_root > current_root_slot);
```
AshwinSekar: 👍 finalized have to descend from each other, replay cannot proceed out of order

core/src/replay_stage.rs
```diff
@@ -770,7 +770,7 @@ impl ReplayStage {
                     let r_bank_forks = bank_forks.read().unwrap();
                     (r_bank_forks.ancestors(), r_bank_forks.descendants())
                 };
-                let did_complete_bank = Self::replay_active_banks(
+                let completed_slots = Self::replay_active_banks(
```
wen-coding: nit: I think complete is pretty vague here, we really mean "slots_completed_replay"?

core/src/replay_stage.rs
```diff
@@ -805,8 +805,50 @@ impl ReplayStage {
                     first_alpenglow_slot,
                     &mut is_alpenglow_migration_complete,
                 );
+                let did_complete_bank = !completed_slots.is_empty();
                 replay_active_banks_time.stop();
 
+                let new_frozen_root = completed_slots
+                    .into_iter()
+                    .filter(|slot| cert_pool.is_finalized_slot(*slot))
+                    .max();
```
wen-coding, line 812: Is it possible that we have:
slot 100 -> 101
101 got finalized cert but somehow the finalization of 100 hasn't fully propagated so it doesn't have a finalization cert?

core/src/replay_stage.rs
```diff
@@ -805,8 +805,50 @@ impl ReplayStage {
...
+                if let Some(new_frozen_root) = new_frozen_root {
+                    let (current_root_slot, new_root_bank) = {
+                        let r_bank_forks = bank_forks.read().unwrap();
+                        (
+                            r_bank_forks.root(),
+                            r_bank_forks.get(new_frozen_root).unwrap(),
+                        )
+                    };
+                    assert!(new_frozen_root > current_root_slot);
+                    if let Err(e) = Self::alpenglow_handle_new_root(
```
wen-coding: Should we have two sources of new_alpenglow_root, one in handle_votable_alpenglow_bank and one in here? Should we just have one set_root step at the end of Alpenglow replay loop (so no matter whether your own vote or replayed votes nailed a new cert, we just have one check_and_handle_new_root in the loop?)

core/src/replay_stage.rs
carllin: yes

core/src/replay_stage.rs
carllin: yup that makes sense, there will be a third one as well once we add ingesting votes from gossip. I moved it out into the main replay loop, also was good to cut down on a lot of the arguments passed to the alpenglow vote functions 😃

core/src/replay_stage.rs
carllin: renamed to new_frozen_slots

PR #69, opened 2025-03-04
feat(voting_service): broadcast votes to all validators in an all-to-all fashion.

ksn6:
All-to-all broadcast votes to all other validators in `voting_service` and ingest those votes into replay.

### DIFF
--- a/core/src/staked_validators_cache.rs
+++ b/core/src/staked_validators_cache.rs
@@ -72,9 +72,12 @@ impl StakedValidatorsCache {
         cluster_info: &ClusterInfo,
         update_time: Instant,
     ) {
-        let bank_forks = self.bank_forks.read().unwrap();
+        let banks = {
+            let bank_forks = self.bank_forks.read().unwrap();
+            [bank_forks.root_bank(), bank_forks.working_bank()]
+        };
 
-        let epoch_staked_nodes = [bank_forks.root_bank(), bank_forks.working_bank()].iter().find_map(|bank| bank.epoch_staked_nodes(epoch)).unwrap_or_else(|| {
+        let epoch_staked_nodes = banks.iter().find_map(|bank| bank.epoch_staked_nodes(epoch)).unwrap_or_else(|| {
             error!("StakedValidatorsCache::get: unknown Bank::epoch_staked_nodes for epoch: {epoch}");
             Arc::<HashMap<Pubkey, u64>>::default()
         });
--- a/core/src/voting_service.rs
+++ b/core/src/voting_service.rs
@@ -47,16 +47,6 @@ pub enum VoteOp {
     },
 }
 
-impl VoteOp {
-    fn tx(&self) -> &Transaction {
-        match self {
-            VoteOp::PushVote { tx, .. } => tx,
-            VoteOp::PushAlpenglowVote { tx, .. } => tx,
-            VoteOp::RefreshVote { tx, .. } => tx,
-        }
-    }
-}
-
 #[derive(Debug, Error)]
 enum SendVoteError {
     #[error(transparent)]
@@ -132,7 +122,7 @@ impl VotingService {
     fn broadcast_tower_vote(
         cluster_info: &ClusterInfo,
         poh_recorder: &RwLock<PohRecorder>,
-        vote_op: &VoteOp,
+        tx: &Transaction,
         connection_cache: &Arc<ConnectionCache>,
     ) {
         // Attempt to send our vote transaction to the leaders for the next few
@@ -142,16 +132,8 @@ impl VotingService {
             FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET.saturating_add(1);
         #[cfg(test)]
         static_assertions::const_assert_eq!(UPCOMING_LEADER_FANOUT_SLOTS, 3);
-        const UPCOMING_LEADER_ALPENGLOW_FANOUT_SLOTS: u64 = 20;
 
-        let leader_fanout = {
-            match &vote_op {
-                // Alpenglow relies on leaders to propagate votes otherwise we have to rely
-                // on gossip, especially true for skip votes
-                VoteOp::PushAlpenglowVote { .. } => UPCOMING_LEADER_ALPENGLOW_FANOUT_SLOTS,
-                _ => UPCOMING_LEADER_FANOUT_SLOTS,
-            }
-        };
+        let leader_fanout = UPCOMING_LEADER_FANOUT_SLOTS;
 
         let upcoming_leader_sockets = upcoming_leader_tpu_vote_sockets(
             cluster_info,
@@ -164,21 +146,21 @@ impl VotingService {
             for tpu_vote_socket in upcoming_leader_sockets {
                 let _ = send_vote_transaction(
                     cluster_info,
-                    vote_op.tx(),
+                    tx,
                     Some(tpu_vote_socket),
                     connection_cache,
                 );
             }
         } else {
             // Send to our own tpu vote socket if we cannot find a leader to send to
-            let _ = send_vote_transaction(cluster_info, vote_op.tx(), None, connection_cache);
+            let _ = send_vote_transaction(cluster_info, tx, None, connection_cache);
         }
     }
 
     fn broadcast_alpenglow_vote(
         slot: Slot,
         cluster_info: &ClusterInfo,
-        vote_op: &VoteOp,
+        tx: &Transaction,
         connection_cache: Arc<ConnectionCache>,
         staked_validators_cache: &mut StakedValidatorsCache,
     ) {
@@ -186,12 +168,12 @@ impl VotingService {
             .get_staked_validators_by_slot(slot, cluster_info, Instant::now());
 
         if staked_validator_tpu_sockets.is_empty() {
-            let _ = send_vote_transaction(cluster_info, vote_op.tx(), None, &connection_cache);
+            let _ = send_vote_transaction(cluster_info, tx, None, &connection_cache);
         } else {
             for tpu_vote_socket in staked_validator_tpu_sockets {
                 let _ = send_vote_transaction(
                     cluster_info,
-                    vote_op.tx(),
+                    tx,
                     Some(*tpu_vote_socket),
                     &connection_cache,
                 );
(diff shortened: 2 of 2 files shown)


core/src/next_leader.rs
```diff
@@ -1,38 +1,49 @@
...
 
-/// Returns a list of tpu vote sockets for the leaders of the next N fanout
-/// slots. Leaders and sockets are deduped.
-pub(crate) fn upcoming_leader_tpu_vote_sockets(
+/// Returns a deduplicated list of TVU sockets for all TVU peers, with the next few leaders
+/// prioritized first.
+pub(crate) fn leader_prioritized_tvu_peer_sockets(
     cluster_info: &ClusterInfo,
     poh_recorder: &RwLock<PohRecorder>,
+    connection_cache: &Arc<ConnectionCache>,
```
carllin: why do we need to change passing just the protocol to passing in the entire connection cache object?

core/src/next_leader.rs
```diff
@@ -1,38 +1,49 @@
...
-    };
+    // Get the keys for the upcoming few leaders
+    let poh_recorder = poh_recorder.read().unwrap();
+    let leader_keys =
+        (0..fanout_slots).filter_map(|n_slots| poh_recorder.leader_after_n_slots(n_slots));
+
+    // Then, get the keys for all eligible validators; this will likely include keys from the
+    // upcoming few leaders as well.
+    let tvu_peer_keys = cluster_info
+        .all_tvu_peers()
```
carllin: `all_tvu_peers()` will also include unstaked nodes, I think we want to limit this to just the staked set

We also probably don't want to copy the entire tvu set every time, a cache would be nice.

I think you're probably looking to reuse the `ClusterNodesCache` object that Behzad's already built, which caches the set of tvu peers and distinguishes the staked nodes

core/src/next_leader.rs
```diff
@@ -1,38 +1,49 @@
...
+        sync::{Arc, RwLock},
+    },
 };
 
-/// Returns a list of tpu vote sockets for the leaders of the next N fanout
-/// slots. Leaders and sockets are deduped.
-pub(crate) fn upcoming_leader_tpu_vote_sockets(
+/// Returns a deduplicated list of TVU sockets for all TVU peers, with the next few leaders
+/// prioritized first.
+pub(crate) fn leader_prioritized_tvu_peer_sockets(
```
carllin: introducing a test for this module would be appreciated :)

core/src/next_leader.rs
ksn6: fixed

core/src/next_leader.rs
ksn6: fixed?

core/src/next_leader.rs
ksn6: I ended up getting rid of the function - does that count? :)

I do think we need to test `StakedValidatorsCache`, though. Once we've basically agreed on the structure, I'll test this.

core/src/voting_service.rs
```diff
@@ -83,6 +84,83 @@ fn send_vote_transaction(
...
+    staked_validator_tpu_sockets: Vec<SocketAddr>,
+
+    /// Bank forks
+    bank_forks: Arc<RwLock<BankForks>>,
+
+    /// Protocol
+    protocol: Protocol,
+}
+
+impl StakedValidatorsCache {
```
carllin: this could probably live in its own module for use by other components

core/src/next_leader.rs
carllin: yeah the cache is good, tests would be appreciated!

ksn6: done

core/src/voting_service.rs
ksn6: fixed

core/src/replay_stage.rs
```diff
@@ -8445,13 +8445,16 @@ pub(crate) mod tests {
             )
         };
 
+        let mut staked_validators_cache =
+            StakedValidatorsCache::new(bank_forks.clone(), connection_cache.protocol());
+
```
carllin: might be cleaner to just pass a bank_forks to `VotingService` here https://github.com/anza-xyz/alpenglow/blob/41115f9ac05f183e234daa3214196cc68b8d32dc/core/src/tvu.rs#L334-L342 and have it own the `StakedValidatorsCache`

core/src/staked_validators_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+        net::SocketAddr,
+        sync::{Arc, RwLock},
+    },
+};
+
+/// Maintain `SocketAddr`s associated with all staked validators for a particular epoch and
+/// protocol (e.g., UDP, QUIC).
+pub struct StakedValidatorsCache {
+    /// The epoch for which we have cached our stake validators list
+    cached_epoch: Epoch,
```
carllin: because when we go from one epoch to the next there might be slot slots we replay at either end, might be worth caching a few epochs, right now I think ClusterNodes caches CLUSTER_NODES_CACHE_NUM_EPOCH_CAP = 5

core/src/staked_validators_cache.rs
```diff
@@ -0,0 +1,316 @@
...
+            .epoch_schedule()
+            .get_epoch(slot);
+
+        if cur_epoch == self.cached_epoch {
+            false
+        } else {
+            self.cached_epoch = cur_epoch;
+            self.refresh(slot, cluster_info);
+
+            true
```
carllin: gossip entries can update much more often than once an epoch, i.e. some validators can crash, come back online, or switch to a different machine. 

Might be good to implement a TTL of about 5 seconds like `ClusterNodes` does,

core/src/replay_stage.rs
ksn6: Yep so I originally considered having `VotingService` own `StakedValidatorsCache`; we'd benefit by having `handle_vote` be a "member function"; i.e., `handle_vote` would have to take in `&mut self` where `self` refers to `VotingService.

Making `handle_vote` a member function is tricky, since `VotingService`'s `new(...)` function below sets up `thread_hdl`, which needs to have created a `VotingService` prior to creation due to invocation of `handle_vote`.

Happy to find a way to make this change if you feel strongly about this / have a fix I've missed!

image

core/src/staked_validators_cache.rs
ksn6: fixed

core/src/staked_validators_cache.rs
ksn6: fixed

core/src/voting_service.rs
```diff
@@ -176,12 +166,79 @@ impl VotingService {
...
+            let mut measure = Measure::start("tower storage save");
+            if let Err(err) = tower_storage.store(saved_tower) {
+                error!("Unable to save tower to storage: {:?}", err);
+                std::process::exit(1);
+            }
+            measure.stop();
+            trace!("{measure}");
+
+            Self::broadcast_tower_vote(cluster_info, poh_recorder, &vote_op, &connection_cache);
+        } else if let VoteOp::PushAlpenglowVote {
```
carllin: more idiomatic to use a match here instead of the if/else if

core/src/staked_validators_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+            build_epoch_stakes(num_nodes, num_zero_stake_nodes, num_vote_accounts);
+
+        let GenesisConfigInfo { genesis_config, .. } = create_genesis_config(genesis_lamports);
+
+        let (bank_forks, cluster_info) = StakedValidatorsCacheHarness::new(&genesis_config)
+            .with_vote_accounts(slot_num, keypair_map, vahm, protocol)
+            .bank_forks();
+
+        // Create our staked validators cache
+        let mut svc = StakedValidatorsCache::new(bank_forks, protocol, Duration::from_secs(5), 5);
```
carllin, line 386: should set a smaller ttl constant to be used in tests so that tests don't take 5 seconds

core/src/staked_validators_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+            .get_epoch(slot)
+    }
+
+    fn refresh_cache_entry(
+        &mut self,
+        epoch: Epoch,
+        cluster_info: &ClusterInfo,
+        update_time: Instant,
+    ) {
+        let bank_forks = self.bank_forks.read().unwrap();
```
carllin: I would release this lock as soon as possible because it's critical in many other areas of the code, and risks deadlock as you'r grabbing the `cluster_info` lock below when you call `lookup_contact_info()` below. 

Make a scoped block where you clone out the banks you need: https://github.com/anza-xyz/agave/blob/f4c4f6d79a8b2324ffee815d8504cc0b90390566/core/src/banking_stage/transaction_scheduler/receive_and_buffer.rs#L164-L169

core/src/staked_validators_cache.rs
ksn6: tests don't take 5 seconds; the `get_staked_validators_x` methods take in an `Instant` as an argument, allowing us to simulate accessing the caches at made up times without having to actually wait

EDIT: concretely, tests (in unoptimized mode) currently take ~0.15s to run, and run in time independent of the ttl constant.

core/src/voting_service.rs
ksn6: fixed

core/src/staked_validators_cache.rs
ksn6: nice catch - fixed

core/src/voting_service.rs
```diff
@@ -174,23 +146,89 @@ impl VotingService {
...
+        connection_cache: Arc<ConnectionCache>,
+        staked_validators_cache: &mut StakedValidatorsCache,
+    ) {
+        let (staked_validator_tpu_sockets, _) = staked_validators_cache
+            .get_staked_validators_by_slot(slot, cluster_info, Instant::now());
+
+        if staked_validator_tpu_sockets.is_empty() {
+            let _ = send_vote_transaction(cluster_info, tx, None, &connection_cache);
+        } else {
+            for tpu_vote_socket in staked_validator_tpu_sockets {
```
AshwinSekar, line 173: Should we send to the next leader first here? Also maybe an adjustable delay before sending batches of votes?

PR #70, opened 2025-03-05
Notify rpc of new root and supermajority roots

carllin:
#### Problem
Alpenglow doesn't need to rely on `AggregateCommitmentService` to notify rpc of new roots

#### Summary of Changes
1. All alpenglow roots are supermajority roots and OC 
2. No aggregation is necessary, directly notify rpc of new roots

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -1267,10 +1267,10 @@ impl ReplayStage {
                             .then_some(r_bank_forks.get(maybe_new_root).unwrap())
                     };
                     if let Some(new_root_bank) = maybe_new_root_bank {
-                        let new_root = new_root_bank.slot();
+                        let new_root_slot = new_root_bank.slot();
                         if let Err(e) = Self::alpenglow_handle_new_root(
                             &new_root_bank, // unnecessary here, just filling out a random bank
-                            new_root,
+                            new_root_slot,
                             &bank_forks,
                             &mut progress,
                             &blockstore,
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -1267,16 +1267,16 @@ impl ReplayStage {
                             .then_some(r_bank_forks.get(maybe_new_root).unwrap())
                     };
                     if let Some(new_root_bank) = maybe_new_root_bank {
+                        let new_root = new_root_bank.slot();
```
wen-coding: nit: new_root-> new_root_slot ?

core/src/replay_stage.rs
```diff
@@ -1295,6 +1295,12 @@ impl ReplayStage {
                             );
                             return;
                         }
+                        rpc_subscriptions.notify_subscribers(CommitmentSlots {
+                            slot: new_root,
+                            root: new_root,
+                            highest_confirmed_slot: new_root,
```
wen-coding: Is there any value advertising slots who got notarize cert but not finalization cert? I'm mostly worried about the case where we would do async execution later, so that finalization may be several slots away from notarization, but I guess for now this is fine.

core/src/replay_stage.rs
```diff
@@ -1295,6 +1295,12 @@ impl ReplayStage {
                             );
                             return;
                         }
+                        rpc_subscriptions.notify_subscribers(CommitmentSlots {
+                            slot: new_root,
```
wen-coding: slot should be the current slot right?

carllin: In today's commitment_service, it's equivalent to the slot we just voted on to make the new root:

```
let mut new_block_commitment = BlockCommitmentCache::new(
            block_commitment,
            aggregation_data.total_stake,
            CommitmentSlots {
                slot: aggregation_data.bank.slot(),
                root: aggregation_data.root,
                highest_confirmed_slot: aggregation_data.root,
                highest_super_majority_root,
            },
        );
 ```

which in this case, we voted on the the new root

core/src/replay_stage.rs
carllin: notarization doesn't guarantee finality, so it's not safe to advertise it under OC/root levels of finality

We can introduce another finality level in the future for this if desired

Issue #24, opened 2025-02-05
Add persistent storage for votes, equivalent to TowerStorage today


carllin:
#### Problem

VoteHistoryStorage has a dummy implementation in Tvu, properly initialize it from storage

PR #74, opened 2025-03-06
feat(validator): add persistent storage for votes.

ksn6:
Address https://github.com/anza-xyz/alpenglow/issues/24.


core/src/validator.rs
```diff
@@ -1,6 +1,10 @@
 //! The `validator` module hosts all the validator microservices.
 
+use crate::alpenglow_consensus::vote_history_storage::{
+    FileVoteHistoryStorage, VoteHistoryStorage,
```
wen-coding: please move this into the use on line 9.

ksn6: fixed

core/src/validator.rs
```diff
@@ -1490,6 +1493,7 @@ impl Validator {
             &poh_recorder,
             tower,
             config.tower_storage.clone(),
+            config.vote_history_storage.clone(),
```
carllin, line 1496: this tower is read/initialized here https://github.com/anza-xyz/alpenglow/blob/24c69eeecc30e3a9d94be0d03aea45fa3719b30f/core/src/validator.rs#L1459-L1471, which handles a lot of edge cases (see below)

In a follow up PR, we should pass initialize and pass a vote history from storage on startup, and write some tests to show that the vote history prevents people from voting in a way that would get them slashed.

1. notarize on anything less than the last notarization in the vote history
2. notarize on another version of the last slot in the vote history
3. finalize on anything they've already voted skip on 
4. Brainstorm some other cases like what if we start from a snapshot less than the last root in the vote history. This means bank_forks.root() < vote_history.latest_finalize slot
    
We only write finalize into VoteHistory for blocks we've replayed https://github.com/anza-xyz/alpenglow/blob/24c69eeecc30e3a9d94be0d03aea45fa3719b30f/core/src/replay_stage.rs#L2798 so can we assume the shreds are in the blockstore database? We can error out or just not vote until the root/repair is updated to at least that slot

PR #75, opened 2025-03-07
Fix skip pool bugs

carllin:
#### Problem
Skip pool doesn't properly handle disjoint skip votes for consecutive slots

For instance:

1. 66% has skip range (1, 10)
2. 1% has skip range (2,2)
3. 1% has skip range (3,3)

Should return skip (2, 3), but the current skip pool will just return (2, 2)

Skip pool also rejects vote (2, 2) and (2, 5) as overlapping votes, even though they're not overlapping

#### Summary of Changes
Fix both of the above

Fixes #

### DIFF
--- a/core/src/alpenglow_consensus/skip_pool.rs
+++ b/core/src/alpenglow_consensus/skip_pool.rs
@@ -20,6 +20,9 @@ pub enum AddVoteError {
 
     #[error("Overlapping skip vote old {0:?} and new {1:?}")]
     Overlapping(RangeInclusive<Slot>, RangeInclusive<Slot>),
+
+    #[error("Zero stake")]
+    ZeroStake,
 }
 
 /// A trait for objects that provide a stake value.
@@ -147,6 +150,9 @@ impl SkipPool {
         stake: Stake,
         total_stake: Stake,
     ) -> Result<(), AddVoteError> {
+        if stake == 0 {
+            return Err(AddVoteError::ZeroStake);
+        }
         // Remove previous skip vote if it exists
         if let Some(prev_skip_vote) = self.skips.get(pubkey) {
             if prev_skip_vote.skip_range == skip_range {
@@ -262,6 +268,27 @@ mod tests {
     }
 
     #[test]
+    fn test_add_vote_zero_stake() {
+        let mut pool = SkipPool::new();
+        let validator = Pubkey::new_unique();
+        let skip_range = 1..=1;
+        let skip_tx = dummy_transaction();
+        let stake = 0;
+        let total_stake = 100;
+
+        assert_eq!(
+            pool.add_vote(
+                &validator,
+                skip_range.clone(),
+                skip_tx.clone(),
+                stake,
+                total_stake,
+            ),
+            Err(AddVoteError::ZeroStake)
+        );
+    }
+
+    #[test]
     fn test_add_singleton_range() {
         let mut pool = SkipPool::new();
         let validator = Pubkey::new_unique();
@@ -365,7 +392,7 @@ mod tests {
             .unwrap();
         assert_eq!(pool.max_skip_certificate_range, RangeInclusive::new(1, 6));
     }
-    
+
     #[test]
     fn test_update_existing_vote() {
         let mut pool = SkipPool::new();


core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -73,10 +73,16 @@ impl<T: Ord + Clone + Debug + HasStake> DynamicSegmentTree<T> {
     fn query(&self, threshold_stake: f64) -> Option<(RangeInclusive<Slot>, Vec<T>)> {
         let mut accumulated = 0f64;
         let mut start = None;
+        let mut prev_slot = None;
```
wen-coding, line 79: nit: prev_slot is vague, do you mean prev_end_slot or prev_start_slot?

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -156,7 +160,15 @@ impl SkipPool {
...
-            if skip_range.start() <= prev_skip_vote.skip_range.end() {
+            if skip_range.start() < prev_skip_vote.skip_range.end() {
+                return Err(AddVoteError::Overlapping(
+                    prev_skip_vote.skip_range.clone(),
+                    skip_range,
+                ));
+            }
+            if skip_range.start() == prev_skip_vote.skip_range.end()
+                && prev_skip_vote.skip_range.start() != prev_skip_vote.skip_range.end()
+            {
```
wen-coding: Hmm, I think your original intent is: If someone voted skip(x, y), he shouldn't vote skip(x', y') where x' > x, so he's un-skipping slots x to x'.

Should we compare skip_range.start() with prev_skip_vote.skip_range.start() instead?

Also, this means when we skip, if we don't notarize any new slot, we will keep the original skip slot forever?

And does it mean if some malfunctioning validator keeps skipping everything, we will keep the old_skip_start in this tree forever? I thought in TowerBFT we can discard everything older than root?

Is this check worth the hassle? Maybe we should rely on slashing instead?

core/src/alpenglow_consensus/skip_pool.rs
carllin: each entry in the tree can contain multiple starts/ends, so it's not possible to tell if its either or both

core/src/alpenglow_consensus/skip_pool.rs
carllin: > Should we compare skip_range.start() with prev_skip_vote.skip_range.start() instead?

Yeah you're right, much simpler and clearer. Updated, thanks!

> Also, this means when we skip, if we don't notarize any new slot, we will keep the original skip slot forever?

If we don't skip any new slots, then this slot will never be deleted yes

> And does it mean if some malfunctioning validator keeps skipping everything, we will keep the old_skip_start in this tree forever? I thought in TowerBFT we can discard everything older than root

Yes for now, but there's an issue for cleanup: https://github.com/anza-xyz/alpenglow/issues/40

> Is this check worth the hassle? Maybe we should rely on slashing instead?

Yeah it's worth it because the other code relies on the supermajority skip range never shrinking

wen-coding: I think I'm mainly worried about the attack case. We only have ~2000 validators for now and each one can only send one range. If we allow multiple ranges per pubkey, then tree size will grow when some attacker sends us tons of non-overlapping ranges. The tree size will also grow larger when there are more staked validators (and we have no lower limit on # of stake?)
And we do query() so full traversal of the tree whenever we add skip vote, so theoretically this may slow us down.

carllin: We don't allow more than one skip range per validator key, the newest skip range always replaces the old one https://github.com/anza-xyz/alpenglow/blob/b6371226a1f9ae117fc6c736a403e9eef6e0ba44/core/src/alpenglow_consensus/skip_pool.rs#L178-L183

carllin: also added sanity check to ignore zero stake validators: https://github.com/anza-xyz/alpenglow/pull/75/commits/c95015e377ffc566bc9f1dc444576a1e916b5baa

PR #76, opened 2025-03-07
Add alpenglow vote state to VoteAccount

wen-coding:
VoteAccount now only has old VoteState, add Alpenglow VoteState to it.

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -105,6 +105,7 @@ impl VoteAccount {
         }
     }
 
+    // This should be really cheap as deserialize is a zero-copy pointer re-cast.
     pub fn alpenglow_vote_state(&self) -> Option<&AlpenglowVoteState> {
         match &self.0.vote_state {
             VoteAccountState::TowerBFT(_) => None,


core/src/replay_stage.rs
```diff
@@ -4288,115 +4286,117 @@ impl ReplayStage {
...
-                bank_vote_state.root_slot = Some(local_root);
-                bank_vote_state
-                    .votes
-                    .retain(|lockout| lockout.slot() > local_root);
-                info!(
-                    "Local root is larger than on chain root, \
-                     overwrote bank root {:?} and updated votes {:?}",
-                    bank_vote_state.root_slot, bank_vote_state.votes
-                );
+        if let Some(vote_state) = vote_account.vote_state() {
```
AshwinSekar: Can we use 
```
let Some(vote_state) = vote_account.vote_state() else {
  return;
}
```
To reduce the indentation

ci/do-audit.sh
```diff
@@ -59,6 +59,28 @@ cargo_audit_ignores=(
...
+  # Crate:     ring
+  # Version:   0.17.3
+  # Title:     Some AES functions may panic when overflow checking is enabled.
+  # Date:      2025-03-06
+  # ID:        RUSTSEC-2025-0009
+  # URL:       https://rustsec.org/advisories/RUSTSEC-2025-0009
+  # Solution:  Upgrade to >=0.17.12
+  # Dependency tree:
+  # ring 0.17.3
+  --ignore RUSTSEC-2025-0009
```
AshwinSekar: don't think you need this now that carl has rebased master

core/src/consensus.rs
```diff
@@ -406,92 +406,96 @@ impl Tower {
...
             trace!("{} {} with stake {}", vote_account_pubkey, key, voted_stake);
-            let mut vote_state = TowerVoteState::from(account.vote_state().clone());
-            for vote in &vote_state.votes {
-                lockout_intervals
-                    .entry(vote.last_locked_out_slot())
-                    .or_default()
-                    .push((vote.slot(), key));
-            }
+            // Vote lockout is only relevant in TowerBFT
+            if let Some(vote_state) = account.vote_state() {
```
AshwinSekar: Also here, `let Some = ... else continue` just makes it easier to review what changed

core/src/replay_stage.rs
wen-coding: Done

ci/do-audit.sh
wen-coding: Removed

core/src/consensus.rs
wen-coding: Done.

core/src/commitment_service.rs
```diff
@@ -202,8 +202,11 @@ impl AggregateCommitmentService {
             let vote_state = if pubkey == node_vote_pubkey {
                 // Override old vote_state in bank with latest one for my own vote pubkey
                 node_vote_state.clone()
+            } else if let Some(vote_state) = account.vote_state() {
+                TowerVoteState::from(vote_state.clone())
             } else {
-                TowerVoteState::from(account.vote_state().clone())
+                // No lockup in AlpenglowVoteAccounts
```
carllin, line 208: this shouldn't be possible if the `bank.vote_accounts()` function only returns the non-alpenglow vote accounts, can just assert failure here

core/src/consensus.rs
```diff
@@ -406,7 +406,11 @@ impl Tower {
                 continue;
             }
             trace!("{} {} with stake {}", vote_account_pubkey, key, voted_stake);
-            let mut vote_state = TowerVoteState::from(account.vote_state().clone());
+            // Vote lockout is only relevant in TowerBFT
+            let Some(vote_state) = account.vote_state() else {
```
carllin, line 410: can just unwrap here, otherwise we have mixed up alpenglow and non alpenglow accounts

core/src/consensus.rs
```diff
@@ -1618,9 +1621,16 @@ impl Tower {
         bank: &Bank,
     ) {
         if let Some(vote_account) = bank.get_vote_account(vote_account_pubkey) {
-            self.vote_state = TowerVoteState::from(vote_account.vote_state().clone());
-            self.initialize_root(root);
-            self.initialize_lockouts(|v| v.slot() > root);
+            if let Some(vote_state) = vote_account.vote_state() {
```
carllin, line 1624: can just unwrap

core/src/vote_simulator.rs
```diff
@@ -67,6 +67,7 @@ impl VoteSimulator {
         }
     }
 
+    // TODO(wen): Add a method for Alpenglow.
```
carllin: We can probably just delete this vote simulator after the alpenglow migration, it's probably only useful for testing the handoff transition from tower/lockouts to alpenglow.

ledger-tool/src/main.rs
```diff
@@ -1335,6 +1318,18 @@ fn main() {
...
                         .help("The bootstrap validator's identity, vote and stake pubkeys"),
                 )
+                .arg(
+                    Arg::with_name("bootstrap_validator_alpenglow_count")
+                        .short("alpenglow")
+                        .long("bootstrap-validator-alpenglow-count")
+                        .value_name("ALPENGLOW_COUNT")
+                        .requires("bootstrap_validator")
+                        .takes_value(true)
+                        .help("How many bootstrap validators to creat as Alpenglow account. \
```
carllin: nit: creat -> create

core/src/replay_stage.rs
```diff
@@ -4330,73 +4331,73 @@ impl ReplayStage {
...
-        for slot in frozen_banks.iter().map(|b| b.slot()) {
-            if !progress
-                .get_fork_stats(slot)
-                .expect("All frozen banks must exist in fork stats")
-                .computed
-            {
-                continue;
+            // adopt the bank vote state
+            tower.vote_state = bank_vote_state;
+
```
carllin: nit: getting rid of this newline might minimize the diff

vote/src/vote_account.rs
```diff
@@ -83,13 +99,133 @@ impl VoteAccount {
...
+    }
+
+    pub fn last_timestamp(&self) -> BlockTimestamp {
+        match &self.0.vote_state {
+            VoteAccountState::TowerBFT(vote_state) => vote_state.last_timestamp.clone(),
+            VoteAccountState::Alpenglow(vote_state) => vote_state.latest_timestamp_legacy_format(),
+        }
+    }
+
+    pub fn get_root_slot(&self) -> Option<Slot> {
```
carllin: nit: get_root_slot -> root_slot or root to be consistent with the other root getter functions in the codebase

rpc/src/rpc.rs
```diff
@@ -1181,35 +1181,18 @@ impl JsonRpcRequestProcessor {
...
-                    epoch_credits
-                        .iter()
-                        .skip(epoch_credits.len() - MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY)
-                        .cloned()
-                        .collect()
-                } else {
-                    epoch_credits.clone()
-                };
+                let mut epoch_credits = account.epoch_credits();
+                epoch_credits.truncate(MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY);
```
carllin: this isn't functionally equivalent to the current

```
epoch_credits
                        .iter()
                        .skip(epoch_credits.len() - MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY)
                        .cloned()
                        .collect()
```

because the current one takes the last `MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY` items while this truncate takes the first `MAX_RPC_VOTE_ACCOUNT_INFO_EPOCH_CREDITS_HISTORY` items

carllin: In general let's leave all functional code changes other than replacing `vote_state` with `vote_account.vote_state()` to another PR

vote/src/vote_account.rs
```diff
@@ -321,50 +457,79 @@ impl From<VoteAccount> for AccountSharedData {
 impl TryFrom<AccountSharedData> for VoteAccount {
     type Error = Error;
     fn try_from(account: AccountSharedData) -> Result<Self, Self::Error> {
-        if !solana_sdk_ids::vote::check_id(account.owner()) {
+        // TODO(wen): this goes into slona_sdk_ids later
```
carllin: nit slona_sdk_ids -> solona_sdk_ids

vote/src/vote_account.rs
```diff
@@ -321,50 +457,79 @@ impl From<VoteAccount> for AccountSharedData {
...
+        let is_alpenglow = if alpenglow_vote::check_id(account.owner()) {
```
carllin, line 478: I've been using the scep program: https://github.com/anza-xyz/alpenglow/blob/41115f9ac05f183e234daa3214196cc68b8d32dc/core/src/replay_stage.rs#L691

might be good to use the same one so we can search and replace later

vote/src/vote_account.rs
```diff
@@ -321,50 +457,79 @@ impl From<VoteAccount> for AccountSharedData {
...
+            true
+        } else if solana_sdk_ids::vote::check_id(account.owner()) {
+            false
+        } else {
             return Err(Error::InvalidOwner(*account.owner()));
-        }
+        };
 
         // Allocate as Arc<MaybeUninit<VoteAccountInner>> so we can initialize in place.
```
carllin, line 486: nit: comment needs an update we're no longer using `MaybeUninit`

vote/src/vote_account.rs
```diff
@@ -321,50 +457,79 @@ impl From<VoteAccount> for AccountSharedData {
...
-        let mut inner = Arc::new(MaybeUninit::<VoteAccountInner>::uninit());
-        let inner_ptr = Arc::get_mut(&mut inner)
-            .expect("we're the only ref")
-            .as_mut_ptr();
+        let mut inner = Arc::new(VoteAccountInner {
```
carllin: I don't think it's safe here to change from `MaybeUninit` which gurantees the drop() function of the types aren't called if initialization fails.

This is critical for safety in case of errors as described by the comment here: https://github.com/anza-xyz/alpenglow/blame/575564bae49dafb96fb955386809d08ebaea8201/vote/src/vote_account.rs#L516

As per that comment, If `VoteAccountInner` is no longer uninit, we can't guarantee the drop() function won't run

Also perf implications described by Alessandro here: https://github.com/anza-xyz/alpenglow/commit/d53152835046f544310f0ef53fa6719bda8387e6

core/src/commitment_service.rs
wen-coding: Hmm we talked about two possibilities:
1. bank.vote_accounts() returns all VoteAccount, TowerBFT or Alpenglow
2.bank.vote_accounts() returns only TowerBFT, and bank.alpenglow_vote_accounts() returns only Alpenglow accounts
If we do 2, code complexity is the same or a little bit higher, because in some place (e.g. stake calculation) where you originally wanted to just loop through bank.vote_accounts(), you now loop through bank.alpenglow_vote_accounts() as well.
This PR is putting Alpenglow VoteState into VoteAccount just to do 1. Are you saying we should do 2 now? Or are you saying we should add bank.towerbft_vote_accounts() which only does filtering?

core/src/vote_simulator.rs
wen-coding: Changed comment.

rpc/src/rpc.rs
wen-coding: Reverted.

vote/src/vote_account.rs
wen-coding: Done.

vote/src/vote_account.rs

vote/src/vote_account.rs
wen-coding: Actually, the test is now failing because alpenglow_vote_state::id is embedded in shared data when I do AlpenglowVoteState::new_for_tests(), so I'll keep this for now.

vote/src/vote_account.rs
wen-coding: That's a good point, I think I still need to set enum to TowerBFT, removed initialization.

wen-coding: Turns out I need a fake initialization to make the enum correctly marked as TowerBFT, but still using the MaybeUninit so no drop is happening.

core/src/commitment_service.rs
carllin: I think we do 2 because based on the migration plan, the stakes cache object

```
pub struct Stakes<T: Clone> {
    /// vote accounts
    vote_accounts: VoteAccounts,

    /// stake_delegations
    stake_delegations: ImHashMap<Pubkey, T>,

    /// unused
    unused: u64,

    /// current epoch, used to calculate current stake
    epoch: Epoch,

    /// history of staking levels
    stake_history: StakeHistory,
}
```

can be modified to track an extra `alpenglow_vote_accounts` which we return

wen-coding: Err I really don't like looping through both vote_accounts and alpenglow_vote_accounts everywhere in the code, we need to keep two loops until the migration is completely over. But I can maybe implement that so we can compare the code, I think option 2 is going to be ugly.

carllin: Would we need to loop both? User calls either alpenglow_vote_accounts() or vote_accounts() depending on which one they need

wen-coding: Let me look at all places I touched:
commitment_service:
  TowerBFT only
consensus:
  lockout: TowerBFT only
  last_voted_slot: need to check both accounts because we are using one pubkey
  record_bank_vote: TowerBFT only
 replay: mostly one or the other
 vote_simulator: TowerBFT only
 ledger_tool: fork graph is mostly TowerBFT related, create snapshot probably needs to support both
 blockstore_processor: TowerBFT only
 rpc: support both
 bank (get timestamp): need to query both
 epoch_stakes and epoch_rewards: need to support both
 
 So not too bad, about 5 places. The stakes one is going to be the ugliest, it passes vote_state everywhere.
 
 Thoughts?

vote/src/vote_account.rs
```diff
@@ -337,22 +478,38 @@ impl TryFrom<AccountSharedData> for VoteAccount {
...
-            // Try to deserialize in place
-            if let Err(e) = VoteState::deserialize_into_uninit(account.data(), vote_state) {
+            if is_alpenglow {
+                addr_of_mut!((*inner_ptr).vote_state).write(VoteAccountState::Alpenglow(
+                    *AlpenglowVoteState::deserialize(account.data())?,
+                ));
+            } else {
+                // SAFELY initialize `vote_state` in place, we just need some value here to make sure
+                // enum is properly initialized with TowerBFT variant.
+                addr_of_mut!((*inner_ptr).vote_state).write(VoteAccountState::TowerBFT(
```
wen-coding, line 508: @alessandrod two questions:
1. Does this look correct? I had to write a dummy value in here, otherwise the enum isn't properly initialized and local tests core dump at line 495 below.
2. Do we need this MaybeUninit and unserialize in place for our new VoteState? I heard we needed it for the old one because it used bincode, in our case it's zero copy so it's really cheap, I think we no longer need it?

core/src/consensus.rs
wen-coding: We might have mixed up alpenglow and non alpenglow accounts for a while during the migration

vote/src/vote_account.rs
AshwinSekar: for some context, the new `VoteState` is zero copy:
```    
pub fn deserialize(vote_account_data: &[u8]) -> Result<&VoteState, ProgramError> {
    spl_pod::bytemuck::pod_from_bytes::<VoteState>(vote_account_data)
}
```
`bytemuck` adds some slight overhead as it checks alignments 
https://github.com/Lokathor/bytemuck/blob/d1d918b2879c62fa946a9534ab8ee0760106b0fa/src/internal.rs#L162

When cu optimizing the new vote program I plan to remove bytemuck and handle the cast similar to the `p-token` program which should remove this overhead.
https://github.com/febo/p-token/blob/main/interface/src/state/mod.rs#L53

Even with bytemuck this should be much faster than serde/bincode which is what I believe `vote_account.rs` was created to solve.

PR #78, opened 2025-03-10
skip_pool: add skip_certified check, multi cert support & fix bugs

AshwinSekar:
#### Problem
The replay voting loop needs the ability to query if a certain slot is `skip_certified`

Certain situations can cause multiple skip range certificates to be available in the skip pool (see `test_multi_cert` for an example). In these scenarios we need the ability to query all skip certs to see if the voting loop slot is `skip_certified`.

Current skip pool impl also has incorrect behavior around consecutive slots and contributors (see `test_consecutive_slots` and `test_contributor_removed`)

#### Summary of Changes
Add functions for `skip_certified`, and a general multi cert scan fn. Does not remove the `query` implementation, will potentially remove when integrating the voting loop / maybe start leader.

### DIFF
--- a/core/src/alpenglow_consensus/skip_pool.rs
+++ b/core/src/alpenglow_consensus/skip_pool.rs
@@ -134,6 +134,8 @@ impl<T: Ord + Clone + Debug + HasStake> DynamicSegmentTree<T> {
             }
         }
 
+        debug_assert_eq!(accumulated, 0f64);
+        debug_assert!(current_cert.is_none());
         certs
     }
 }


core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -233,6 +348,53 @@ impl SkipPool {
...
+    }
+
+    /// Is `slot` contained in any skip certificates
+    pub fn skip_certified(&mut self, slot: Slot, total_stake: Stake) -> bool {
+        if self
+            .certificate_ranges
+            .iter()
+            .any(|range| range.contains(&slot))
+        {
+            // If we are already have a certificate no reason to rescan (potentially costly)
```
AshwinSekar, line 290: the assumption here is if a slot is already part of a skip certificate then it's fine to not update, as once part of a skip certificate it cannot be removed. Also the `up_to_date` is to avoid scanning the range every time a new vote is added, however if you feel it's clearer we can mimic the behavior in `query`

generally I think we need to have protections around ingesting all-to-all  / gossip when catching up as it could overwrite votes from replay that are necessary to skip certify "old" slots leading to catch up stalling.
i don't know the best way to deal with this yet, will explore.

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -57,20 +103,81 @@ impl<T: Ord + Clone + Debug + HasStake> DynamicSegmentTree<T> {
...
     }
 
+    fn scan_certificates(
+        &self,
+        threshold_stake: f64,
+    ) -> impl Iterator<Item = (RangeInclusive<Slot>, BTreeSet<Pubkey>)> + use<'_, T> {
+        let certs = self
+            .tree
+            .iter()
+            .scan(
```
carllin: might be less nesting here to use a `filter_map`:

```
 let mut accumulated = 0f64;
 let mut current_contributors = BTreeSet::new();
 let mut cert: Option<(Slot, BTreeSet<Pubkey>)> = None;

let certs = self.tree.iter().filter_map(move |(slot, (starts, ends))| {

   // Once we hit threshold return the cert
    if accumulated < threshold_stake {
            let (start_slot, contributors)) = cert.take().expect("must have some contributors");
             return Some((start_slot..=*slot, contributors));
        }
}
```

core/src/alpenglow_consensus/skip_pool.rs
carllin: > generally I think we need to have protections around ingesting all-to-all / gossip when catching up as it could overwrite votes from replay that are necessary to skip certify "old" slots leading to catch up stalling.

Each block in replay should have the skip/notarization certificates necessary to prove block replay, so checking the vote states should be sufficient to prove the block is valid, shouldn't require checking the skip pool here I think.

As long as we're getting valid blocks, catch up shouldn't stall.

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -42,9 +43,54 @@ impl HasStake for (Pubkey, Stake) {
...
+        let ((start, mut end), mut pubkeys) = self.iter.next()?;
+
+        while let Some(((next_start, next_end), next_pubkeys)) = self.iter.peek_mut() {
+            if end + 1 == *next_start {
+                end = *next_end;
+                pubkeys.append(next_pubkeys);
+                self.iter.next();
+            } else {
+                break;
+            }
```
carllin: seems cleaner to handle this merge in the `scan_certificates()` function so we don't need a special iterator. The query() function does something similar: https://github.com/anza-xyz/alpenglow/blob/41936baf943a5c340f4b9c963a8c591f9355d1c2/core/src/alpenglow_consensus/skip_pool.rs#L191-L194

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -207,6 +320,8 @@ impl SkipPool {
             self.max_skip_certificate_range = max_range; // Update to the full range
         }
 
+        self.up_to_date = false;
```
carllin, line 235: can probably  get rid of the `query` method altogether and just set up_to_date to false anytime a new skip is added that doesn't entirely fall within an existing certificate range in `self.certificate_ranges`

carllin: maybe make an issue as well to make a batch insert_vote() function before running scan_certificates() have a feeling it will be much better  for perf lol

core/src/alpenglow_consensus/skip_pool.rs
AshwinSekar: Yeah that's true, just trying to figure out the best way to query.

If we're catching up, we should *only* look at the skip certificates from replayed votes, however if we're at tip then we should ingest all-to-all / gossip as well. We can't really tell which scenario we're in so I think we could:
- Maintain a separate skip pool with only votes from replay, check the replay skip pool first and then this general skip pool
- Tag votes with their source here and allow storage of 1 vote per source per validator
Open to any better suggestions haha

core/src/alpenglow_consensus/skip_pool.rs
AshwinSekar: yeah you're right that's cleaner 815e35d576047730d66ba6929952bf9eaf638a37

core/src/alpenglow_consensus/skip_pool.rs
AshwinSekar: makes sense, got too eager  trying to skyline 81b4b02d480885d4ed3e62a422926ff5c26bdf88

core/src/alpenglow_consensus/skip_pool.rs
AshwinSekar: removed `query` and handled the `up_to_date` logic here 5dafa50e0a34d54e3249d7c9c13c2c0ff487b5f5

AshwinSekar: made an issue here https://github.com/anza-xyz/alpenglow/issues/79

core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -57,58 +58,83 @@ impl<T: Ord + Clone + Debug + HasStake> DynamicSegmentTree<T> {
...
+            for item in ends {
+                current_contributors.remove(&item.pubkey());
+                accumulated -= item.stake_value() as f64
+            }
+
+            // Return cert if it has ended
+            if accumulated <= threshold_stake {
+                if let Some((start_slot, contributors)) = &current_cert {
+                    // Skip certificate has ended, reset and publish
+                    certs.push(((*start_slot, *slot), contributors.clone()));
```
carllin, line 131: If the threshold never falls below threshold_stake before the loop ends, we should still push the current cert if it exists

core/src/alpenglow_consensus/skip_pool.rs
carllin: That makes sense

1. For new blocks, check vote state. A separate skip pool could work, or we could run a simple algorithm that scans across each vote state and checks if the most recent skip vote covers the entire range (M, N) where M is the parent and N is this replayed block. If so, increment the stake counter.
2. We stream the latest votes during replay into this main skip pool

core/src/alpenglow_consensus/skip_pool.rs
carllin: lmao, write all your code in one line

core/src/alpenglow_consensus/skip_pool.rs
AshwinSekar: I don't think this can actually happen right, since every vote has to end `accumulated` is guaranteed to be `<= threshold_stake` before the loop ends. 

Added a debug assert for this case 4c08dbe

carllin: ah yeah, we separate out the ends, makes sense

PR #87, opened 2025-03-14
Allow Alpenglow accounts in StakesCache

wen-coding:
Allow Alpenglow accounts in StakesCache

Currently rewards don't work until we update epoch_stakes calculation.

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -822,7 +822,7 @@ pub fn commit_certificate(
 
     if summary.reached_max_poh_height {
         datapoint_error!(
-            "vote_certiificate_commit_failure",
+            "vote_certificate_commit_failure",
             ("error", "slot took too long to ingest votes", String),
             ("slot", bank.slot(), i64)
         );
@@ -832,7 +832,7 @@ pub fn commit_certificate(
 
     if summary.error_counters.total.0 != 0 {
         datapoint_error!(
-            "vote_certiificate_commit_failure",
+            "vote_certificate_commit_failure",
             (
                 "error",
                 format!("{} errors occurred", summary.error_counters.total.0),
--- a/core/src/cluster_info_vote_listener.rs
+++ b/core/src/cluster_info_vote_listener.rs
@@ -39,10 +39,7 @@ use {
         timing::AtomicInterval,
         transaction::Transaction,
     },
-    solana_vote::{
-        vote_parser::{self, ParsedVote},
-        vote_transaction::VoteTransaction,
-    },
+    solana_vote::vote_parser::{self, ParsedVote, ParsedVoteTransaction},
     std::{
         cmp::max,
         collections::HashMap,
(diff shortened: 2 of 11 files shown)


runtime/src/stakes.rs
```diff
@@ -120,6 +120,35 @@ impl StakesCache {
                     stakes.remove_vote_account(pubkey)
                 };
             };
+        } else if alpenglow_vote::check_id(owner) {
+            match VoteAccount::try_from(account.to_account_shared_data()) {
+                Ok(vote_account) => {
+                    if vote_account.alpenglow_vote_state().unwrap().version() > 0 {
```
carllin: when would this version not be greater than 0?

runtime/src/stakes.rs
```diff
@@ -230,7 +259,8 @@ impl Stakes<StakeAccount> {
                 let voter_pubkey = &delegation.voter_pubkey;
                 if stakes.vote_accounts.get(voter_pubkey).is_none() {
                     if let Some(account) = get_account(voter_pubkey) {
-                        if VoteStateVersions::is_correct_size_and_initialized(account.data())
+                        if (alpenglow_vote::check_id(account.owner())
```
carllin: do we need an equivalent version of `is_correct_size_and_initialized()` for alpenglow accounts?

runtime/src/stakes.rs
wen-coding: When the vote account has never been initialized

runtime/src/stakes.rs
wen-coding: I originally wanted to add that, but now I'm confused. Why do we need to check is_correct_size_and_initialized() separately instead of waiting to see if deserialize succeeded? I suppose old VoteState wanted to do it because deserialize is super expensive there, do we still need this check?

runtime/src/stakes.rs
carllin: can `alpenglow_vote_state()` return `Some()` if the vote account has not been initialized?

runtime/src/stakes.rs
carllin: if we can/already have a test that shows that uninitialized alpenglow vote accounts don't successfully deserialize, then we don't need this check

runtime/src/stakes.rs
wen-coding: Hmm, I think yes, currently as long as recast works it returns Some()

runtime/src/stakes.rs
wen-coding: So we have a choice, either check in deserialize, or give initialized to check outside. What do you think? Any value in returning a VoteState::default() which is not initialized?

wen-coding: I was writing other tests and it's kind of convenient to be able to use VoteState::default(), let me make the `is_initialized` of VoteState public, sending you a PR soon.

wen-coding: https://github.com/solana-program/alpenglow-vote/pull/49

runtime/src/stakes.rs
wen-coding: Changed to is_initialized()

runtime/src/stakes.rs
wen-coding: Changed to is_initialized

PR #88, opened 2025-03-16
Fix some panicking code

carllin:
#### Problem
1. `block_id()` is None on your own leader blocks, so `highest_frozen_bank.block_id().unwrap()` was panicking
2.  `assert!(poh_start_slot < highest_frozen_bank.slot());` was panicking because it's possible that poh finishes the last tick for leader bank `X+1`, calls flush_tick_cache(), which will set the poh_start_slot to bank `X+1`. This means it's possible for the start_slot to be `X+1` before replay_stage detects the bank is finished and freezes the bank, which makes it possible that start_slot = `X+1` and the highest_frozen_slot = `X`. 

#### Summary of Changes
1. Make block id default for now, not sure how to handle this for leader blocks
2. Removed the assert in reset poh logic

Fixes #

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -822,7 +822,7 @@ pub fn commit_certificate(
 
     if summary.reached_max_poh_height {
         datapoint_error!(
-            "vote_certiificate_commit_failure",
+            "vote_certificate_commit_failure",
             ("error", "slot took too long to ingest votes", String),
             ("slot", bank.slot(), i64)
         );
@@ -832,7 +832,7 @@ pub fn commit_certificate(
 
     if summary.error_counters.total.0 != 0 {
         datapoint_error!(
-            "vote_certiificate_commit_failure",
+            "vote_certificate_commit_failure",
             (
                 "error",
                 format!("{} errors occurred", summary.error_counters.total.0),


core/src/replay_stage.rs
```diff
@@ -3359,11 +3359,7 @@ impl ReplayStage {
         assert!(highest_frozen_bank.slot() >= first_alpenglow_slot);
 
         let poh_start_slot = poh_recorder.read().unwrap().start_slot();
-        if poh_start_slot != highest_frozen_bank.slot() {
-            // It's impossible for start poh_start_slot > highest_frozen_bank
-            // because we only ever start leader banks from parents that are
-            // frozen.
-            assert!(poh_start_slot < highest_frozen_bank.slot());
+        if poh_start_slot < highest_frozen_bank.slot() {
```
wen-coding, line 3362: Hmm, I thought you said you wanted to remove this code, why do we need it in Alpenglow?

core/src/replay_stage.rs
```diff
@@ -3445,7 +3441,7 @@ impl ReplayStage {
                         // TODO(ashwin): fixup when separating vote loop, check block_id for non leader slots
                         AlpenglowVote::new_finalization_vote(
                             highest_frozen_bank.slot(),
-                            highest_frozen_bank.block_id().unwrap(),
+                            Hash::default(),
```
wen-coding: Do we have block_id now? I thought you want block_id when you are not the leader?

core/src/replay_stage.rs
carllin: I think it's necessary to keep the poh up to date for banking stage would_be_leader() checks

Ideally those get replaced eventually by the separate skip loop timer which tracks the current slot

wen-coding: Please add comment saying why we want this now and when we plan to remove it.

core/src/replay_stage.rs
carllin: yeah it exists right now, i'll replicate that behavior

wen-coding: Then I think you want block_id().unwrap_or() rather than Hash::default()? And add it to comment we need to fix it for leader block in the future (maybe file an issue)

carllin: https://github.com/anza-xyz/alpenglow/blob/6770619b7c1cc2e59e623c50e14a15919f90d805/core/src/consensus.rs#L627. looks like leaders will be Hash::default()

wen-coding: Hmm why? The leaders should know block_id when the block is complete and it wants to endorse its own block?

carllin: I've forgotten the reason 👻 , I think it's hinted at here: https://github.com/anza-xyz/alpenglow/blob/6770619b7c1cc2e59e623c50e14a15919f90d805/core/src/replay_stage.rs#L3898-L3899

wen-coding: Doesn't totally make sense to me, let's talk with Ashwin when he's back.

carllin: oh and this PR description: https://github.com/anza-xyz/agave/pull/2776#issue-2492880606

carllin: ah i remember now, it's because

1. Bank tick is registered in Poh
2. Replay detects the bank ticks are full, freezes the bank
3. Shred is created by broadcast

There is no guarantee 3 happens before 2, so it's not necessarily possible for the leader to compute the block id

wen-coding: I would prefer leader votes with its own block-id if possible, I would still say chat with Ashwin to see if we can do that.

carllin: "or to simply resend the vote once the last shred is finished (this would involve messing with the timestamp or deduplication code)." 

I think this was the original plan

wen-coding: I think we can also hold the vote, I don't think we need to hold the vote for that long, shred in broadcast should be fast?

PR #92, opened 2025-03-16
Add alpenglow to rewards calculation

wen-coding:
Add alpenglow VoteState to rewards calculation.

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -20,7 +20,6 @@ use {
     },
     thiserror::Error,
 };
-
 // The following imports are only needed for dev-context-only-utils.
 #[cfg(feature = "dev-context-only-utils")]
 use {


vote/src/vote_account.rs
```diff
@@ -119,6 +125,53 @@ impl VoteAccount {
...
     }
 
+    pub fn commission(&self) -> u8 {
+        match &self.0.vote_state_view {
+            VoteAccountState::TowerBFT(vote_state_view) => vote_state_view.commission(),
+            VoteAccountState::Alpenglow => AlpenglowVoteState::deserialize(self.0.account.data())
+                .unwrap()
+                .commission(),
+        }
+    }
```
carllin, line 135: can we implement these member functions on `enum VoteAccountState` if we defined it as 

```
enum VoteAccountState {
    TowerBFT(VoteStateView),
    Alpenglow(AlpenglowVoteState),
}
```

then we could pass around `VoteAccountState` instead of the vote_account

wen-coding: Err, actually can't do that, because we are doing the black magic of deserializing in place so VoteAccountState::Alpenglow doesn't contain any data now. Do you prefer actually making a copy in the enum instead? This enum should go away very soon after Alpenglow launch.

Do you dislike passing around reference to vote_account because it's expensive? I thought passing reference should be cheap?

carllin: it's fine then, more of a style thing

PR #94, opened 2025-03-17
Update commitment cache in alpenglow

carllin:
#### Problem
Some RPC calls like `get_transaction_status()` still rely on an updated commitment cache https://github.com/anza-xyz/alpenglow/blob/efc48d52ca6c16092147c7ed13d4ffede8e7d13c/rpc/src/rpc.rs#L1726, so we need to update it

#### Summary of Changes
Update commitment cache with roots/supermajority roots/latest slot

Fixes #

### DIFF
--- a/runtime/src/commitment.rs
+++ b/runtime/src/commitment.rs
@@ -347,4 +347,28 @@ mod tests {
 
         assert_eq!(block_commitment_cache.calculate_highest_confirmed_slot(), 0);
     }
+
+    #[test]
+    fn test_setters_getters() {
+        let mut block_commitment_cache = BlockCommitmentCache::default();
+        // Setting bigger slots should be ok
+        block_commitment_cache.set_slot(1);
+        assert_eq!(block_commitment_cache.slot(), 1);
+        block_commitment_cache.set_highest_confirmed_slot(2);
+        assert_eq!(block_commitment_cache.highest_confirmed_slot(), 2);
+        block_commitment_cache.set_root(3);
+        assert_eq!(block_commitment_cache.root(), 3);
+        block_commitment_cache.set_highest_super_majority_root(4);
+        assert_eq!(block_commitment_cache.highest_super_majority_root(), 4);
+
+        // Setting smaller slots shuold be ignored
+        block_commitment_cache.set_slot(0);
+        assert_eq!(block_commitment_cache.slot(), 1);
+        block_commitment_cache.set_highest_confirmed_slot(1);
+        assert_eq!(block_commitment_cache.highest_confirmed_slot(), 2);
+        block_commitment_cache.set_root(2);
+        assert_eq!(block_commitment_cache.root(), 3);
+        block_commitment_cache.set_highest_super_majority_root(3);
+        assert_eq!(block_commitment_cache.highest_super_majority_root(), 4);
+    }
 }


core/src/commitment_service.rs
```diff
@@ -134,17 +161,52 @@ impl AggregateCommitmentService {
                     i64
                 ),
```
wen-coding, line 162: Should we add a field whether we are using Alpenglow here?

core/src/replay_stage.rs
```diff
@@ -4762,7 +4779,13 @@ impl ReplayStage {
             voted_signatures,
             epoch_slots_frozen_slots,
             drop_bank_sender,
-        )
+        );
+        Self::alpenglow_update_commitment_cache(
+            AlpenglowCommitmentType::Root,
+            new_root,
+            lockouts_sender,
+        );
```
wen-coding, line 4787: hmm, check_and_handle_new_root can fail right? Do you want to notify commitment cache if it failed?

runtime/src/commitment.rs
```diff
@@ -45,7 +45,7 @@ pub struct BlockCommitmentCache {
     block_commitment: HashMap<Slot, BlockCommitment>,
     /// Cache slot details. Cluster data is calculated from the block_commitment map, and cached in
     /// the struct to avoid the expense of recalculating on every call.
-    commitment_slots: CommitmentSlots,
+    pub commitment_slots: CommitmentSlots,
```
wen-coding: Why do we need this when we have `commitment_slots()` and `set_highest_confirmed_slot()` etc?

core/src/replay_stage.rs
carllin: ah yeah this is a good catch, fixed

runtime/src/commitment.rs
carllin: there were only these two accessors: 
image
and I got annoyed with them, I think in order to do the max() computation for each of the 4 fields, we would have to add 4 different getters for root, slot, supermajority and OC, and then 2 more setters for root and slot

core/src/commitment_service.rs
carllin: hmmm in Alpenglow the supermajority root and highest confirmed slot will be evident if they are greater than the first alpenglow activation slot

runtime/src/commitment.rs
wen-coding: I'd say add accessors for all your most common use cases.
In Alpenglow we only do:
1. set slot for Notarize
2. set all others for Finalize
Then add these two setters, you can do all the max() checks inside the setters.
Once you expose internal data structure, people would do all kinds of crazy stuff to it in code, then next time we want to change internal data structure, we need to change tons of places. I'd rather add more setters(). Just my 2c.

carllin: ok you're right, I updated!

core/src/commitment_service.rs
```diff
@@ -134,17 +161,37 @@ impl AggregateCommitmentService {
...
+    fn alpenglow_update_commitment_cache(
+        block_commitment_cache: &RwLock<BlockCommitmentCache>,
+        update_type: AlpenglowCommitmentType,
+        slot: Slot,
+    ) -> CommitmentSlots {
+        let mut w_block_commitment_cache = block_commitment_cache.write().unwrap();
+
+        match update_type {
+            AlpenglowCommitmentType::Notarize => {
+                w_block_commitment_cache.commitment_slots.slot =
```
wen-coding: nit: we could use set_slot() you just added?

PR #95, opened 2025-03-17
Parse out alpenglow votes in vote_parser.rs

carllin:
#### Problem
1. Gossip doesn't properly parse out alpenglow votes to pass to downstream listeners in repair and replay
2. Push vote services doesn't properly parse alpenglow votes

#### Summary of Changes
Parse out votes

Fixes #

### DIFF
--- a/core/src/cluster_info_vote_listener.rs
+++ b/core/src/cluster_info_vote_listener.rs
@@ -301,14 +301,8 @@ impl ClusterInfoVoteListener {
                 let (vote_account_key, vote, ..) = vote_parser::parse_vote_transaction(&tx)
                     .or_else(|| vote_parser::parse_alpenglow_vote_transaction(&tx))?;
                 let slot = vote.last_voted_slot()?;
-                if slot >= first_alpenglow_slot {
-                    if !vote.is_alpenglow_vote() {
-                        return None;
-                    }
-                } else {
-                    if vote.is_alpenglow_vote() {
-                        return None;
-                    }
+                if (slot >= first_alpenglow_slot) ^ vote.is_alpenglow_vote() {
+                    return None;
                 }
                 let epoch = epoch_schedule.get_epoch(slot);
                 let authorized_voter = root_bank
@@ -780,7 +774,7 @@ mod tests {
             pubkey::Pubkey,
             signature::{Keypair, Signature, Signer},
         },
-        solana_vote::vote_transaction,
+        solana_vote::vote_transaction::{self, VoteTransaction},
         solana_vote_program::vote_state::{TowerSync, Vote, MAX_LOCKOUT_HISTORY},
         std::{
             collections::BTreeSet,
@@ -1368,6 +1362,7 @@ mod tests {
         run_test_process_votes3(Some(Hash::default()));
     }
 
+    // TODO: Add Alpenglow equivalent tests
     #[test]
     fn test_vote_tracker_references() {
         // Create some voters at genesis
--- a/vote/src/vote_parser.rs
+++ b/vote/src/vote_parser.rs
@@ -138,6 +138,7 @@ pub fn is_alpenglow_vote_transaction(tx: &Transaction) -> bool {
     program_id == &alpenglow_vote::id()
 }
 
+// TODO: add tests
 pub fn parse_alpenglow_vote_transaction(tx: &Transaction) -> Option<ParsedVote> {
     // Check first instruction for a vote
     let message = tx.message();
@@ -198,7 +199,7 @@ fn parse_vote_instruction_data(
 }
 
 fn parse_alpenglow_vote_instruction_data(vote_instruction_data: &[u8]) -> Option<AlpenglowVote> {
-    alpenglow_vote::vote::Vote::deserialize_simple_vote(vote_instruction_data).ok()
+    AlpenglowVote::deserialize_simple_vote(vote_instruction_data).ok()
 }
 
 #[cfg(test)]


core/src/cluster_info_vote_listener.rs
```diff
@@ -297,8 +301,18 @@ impl ClusterInfoVoteListener {
...
                 let slot = vote.last_voted_slot()?;
+                if slot >= first_alpenglow_slot {
+                    if !vote.is_alpenglow_vote() {
+                        return None;
+                    }
+                } else {
+                    if vote.is_alpenglow_vote() {
+                        return None;
+                    }
+                }
```
wen-coding, line 306: nit: is this equivalent to
if (slot >= first_alpenglow_slot) ^ vote.is_alpenglow_vote() {
    return None;
}
?

vote/src/vote_parser.rs
```diff
@@ -1,11 +1,70 @@
...
+        }
+    }
+
+    pub fn last_voted_slot_hash(&self) -> Option<(Slot, Hash)> {
+        match self {
+            ParsedVoteTransaction::Tower(tx) => tx.last_voted_slot_hash(),
+            ParsedVoteTransaction::Alpenglow(tx) => match tx {
+                AlpenglowVote::Notarize(vote) => Some((vote.slot(), *vote.replayed_bank_hash())),
+                AlpenglowVote::Finalize(vote) => Some((vote.slot(), *vote.replayed_bank_hash())),
+                AlpenglowVote::Skip(_vote) => None,
```
wen-coding, line 40: We can give Skip `Some((vote.end_slot(), Hash::default()))` to make the vote sending work? I don't think we care that much about the start_slot changing

If we do that, then I think this function can return (Slot, Hash), since it will always return Some(...)

vote/src/vote_parser.rs
```diff
@@ -1,11 +1,70 @@
...
+    pub fn is_alpenglow_vote(&self) -> bool {
+        match self {
+            ParsedVoteTransaction::Tower(_tx) => false,
+            ParsedVoteTransaction::Alpenglow(_tx) => true,
+        }
+    }
+
+    pub fn is_full_tower_vote(&self) -> bool {
+        match self {
+            ParsedVoteTransaction::Tower(tx) => tx.is_full_tower_vote(),
```
wen-coding, line 54: Would we still have anything other than full tower vote in TowerBFT side when Alpenglow launches?

vote/src/vote_parser.rs
```diff
@@ -18,7 +77,12 @@ pub fn parse_sanitized_vote_transaction(tx: &impl SVMTransaction) -> Option<Pars
     let key = tx.account_keys().get(first_account)?;
     let (vote, switch_proof_hash) = parse_vote_instruction_data(first_instruction.data)?;
     let signature = tx.signatures().first().cloned().unwrap_or_default();
-    Some((*key, vote, switch_proof_hash, signature))
+    Some((
+        *key,
+        ParsedVoteTransaction::Tower(vote),
```
wen-coding, line 82: Hmm why don't we need this for Alpenglow?

vote/src/vote_parser.rs
```diff
@@ -35,7 +99,52 @@ pub fn parse_vote_transaction(tx: &Transaction) -> Option<ParsedVote> {
...
+    let signature = tx.signatures().first().cloned().unwrap_or_default();
+    Some((
+        *key,
+        ParsedVoteTransaction::Alpenglow(alpenglow_vote),
+        None,
+        signature,
+    ))
+}
+
+pub fn parse_alpenglow_vote_transaction(tx: &Transaction) -> Option<ParsedVote> {
```
wen-coding, line 142: nit: can this guy and `parse_sanitized_alpenglow_vote_transaction` share code?

vote/src/vote_parser.rs
```diff
@@ -76,6 +185,10 @@ fn parse_vote_instruction_data(
     }
 }
 
+fn parse_alpenglow_vote_instruction_data(vote_instruction_data: &[u8]) -> Option<AlpenglowVote> {
+    alpenglow_vote::vote::Vote::deserialize_simple_vote(vote_instruction_data).ok()
+}
```
wen-coding, line 203: nit:: add TODO to add test cases.

Also can we use alpenglow_vote::vote::Vote as AlpenglowVote?

core/src/cluster_info_vote_listener.rs
```diff
@@ -1388,7 +1425,10 @@ mod tests {
             // Add gossip vote for same slot, should not affect outcome
             vec![(
                 validator0_keypairs.vote_keypair.pubkey(),
-                VoteTransaction::from(Vote::new(vec![voted_slot], Hash::default())),
+                ParsedVoteTransaction::Tower(VoteTransaction::from(Vote::new(
```
wen-coding, line 1420: nit: add TODO for alpenglow test

core/src/cluster_info_vote_listener.rs
carllin: yeah it's equivalent, changed

vote/src/vote_parser.rs
carllin: I don't think you want to send skip votes to repair for fork choice weighting, so that logic should probably be handled separately

vote/src/vote_parser.rs
carllin: no I think we should be ignoring all alpenglow votes for slots < the first alpenglow slot, hence this check https://github.com/anza-xyz/alpenglow/pulls#discussion_r1999224815

vote/src/vote_parser.rs
carllin: could probably pass some closures to handle custom parsing, but code overlap is small enough where it doesn't seem worth it

vote/src/vote_parser.rs
carllin: what's "this"? this is the equivalent for alpenglow: https://github.com/anza-xyz/alpenglow/blob/15583607b40dcc96eff784601dd025a955cad2ad/vote/src/vote_parser.rs#L123

vote/src/vote_parser.rs
carllin: done!

core/src/cluster_info_vote_listener.rs
carllin: yeah going to need to go through and write some detailed tests, might hand off to @ksn6

vote/src/vote_parser.rs
wen-coding: Okay, let's add a comment here saying why Skip should return None then.

vote/src/vote_parser.rs
```diff
@@ -76,6 +198,10 @@ fn parse_vote_instruction_data(
     }
 }
 
+fn parse_alpenglow_vote_instruction_data(vote_instruction_data: &[u8]) -> Option<AlpenglowVote> {
```
AshwinSekar, line 201: nit: we could also check ```AlpenglowVote::is_simple_vote``` if we want to distinguish between accounting instructions here.
Also fine to keep it as is, accounting instructions and invalid packets will both turn into None

vote/src/vote_parser.rs
```diff
@@ -35,7 +99,65 @@ pub fn parse_vote_transaction(tx: &Transaction) -> Option<ParsedVote> {
...
+    let program_id_index = usize::from(first_instruction.program_id_index);
+    let Some(program_id) = message.account_keys.get(program_id_index) else {
+        return false;
+    };
+    program_id == &alpenglow_vote::id()
+}
+
+// TODO: add tests
+pub fn parse_alpenglow_vote_transaction(tx: &Transaction) -> Option<ParsedVote> {
+    // Check first instruction for a vote
```
AshwinSekar, line 143: Since we are doing this from scratch here, we could formalize the requirement that votes must be a single instruction transaction (fail to parse here and in banking stage).

PR #97, opened 2025-03-17
Verify certs in bank

wen-coding:
Verify we have valid certs in Alpenglow block:
- notarization certs for parent
- skip certs for slots between parent and current bank.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -4235,22 +4235,14 @@ impl ReplayStage {
             .iter()
             .for_each(|(vote_account_pubkey, (stake, account))| {
                 let Some(vote_state) = account.alpenglow_vote_state() else {
-                    info!("Vote account does not have vote state");
                     return;
                 };
                 let stake_in_parent_epoch =
                     parent_epoch_stakes.vote_account_stake(vote_account_pubkey);
-                info!(
-                    "Stake in parent epoch: {} {:?} {:?}",
-                    stake_in_parent_epoch,
-                    vote_state.latest_notarized_slot(),
-                    vote_state.latest_notarized_bank_hash()
-                );
                 if vote_state.latest_notarized_slot() == Some(parent_slot)
                     && vote_state.latest_notarized_bank_hash() == &parent_hash
                 {
                     notarization_stake += stake_in_parent_epoch;
-                    info!("Adding notarization stake: {}", stake_in_parent_epoch);
                 }
                 // TODO(wen): the stake here might be incorrect for different epoch.
                 let _ = skip_pool.add_vote(
@@ -10743,23 +10735,8 @@ pub(crate) mod tests {
         if let Some(slot) = latest_skip_end_slot {
             vote_state.set_latest_skip_end_slot(slot);
         }
-        info!(
-            "Store account {} {:?} {:?}",
-            pubkey,
-            vote_state.latest_notarized_slot(),
-            vote_state.latest_notarized_bank_hash()
-        );
         vote_state.serialize_into(account.data_as_mut_slice());
         let vote_account = VoteAccount::try_from(account).unwrap();
-        info!(
-            "Store account {} {} {:?}",
-            pubkey,
-            vote_account.owner(),
-            vote_account
-                .alpenglow_vote_state()
-                .unwrap()
-                .latest_notarized_slot()
-        );
         bank.store_account(pubkey, vote_account.account());
     }
 


core/src/replay_stage.rs
```diff
@@ -3892,6 +3893,35 @@ impl ReplayStage {
                     }
                 }
 
+                if let Some(first_alpenglow_slot) = first_alpenglow_slot {
+                    if *is_alpenglow_migration_complete {
```
carllin: I think this check is unnecessary, if `bank.slot() > first_alpenglow_slot` then we need to make the cert check

core/src/replay_stage.rs
```diff
@@ -3892,6 +3893,35 @@ impl ReplayStage {
...
+                        if let Some(parent_bank) = bank.parent() {
+                            if parent_bank.slot() >= first_alpenglow_slot {
```
carllin: We also need to make the alpenglow check if the parent is less than the first alpenglow slot. 

This is because we might have skipped some alpenglow slots in the current alpenglow epoch, even if the parent was from the previous epoch, and we need skip certs for those skipped alpenglow slots

carllin: I think basically the only check we need is `if bank.slot() > first_alpenglow_slot then we need to make the cert check`, then run the cert check. 

Note strictly greater because i think the first alpenglow slot doesn't need any certs

ledger/src/blockstore_processor.rs
```diff
@@ -885,6 +885,9 @@ pub enum BlockstoreProcessorError {
 
     #[error("invalid retransmitter signature final fec set")]
     InvalidRetransmitterSignatureFinalFecSet,
+
+    #[error("invalid cert in bank {0} for slot {1}: {2}")]
+    InvalidCert(Slot, Slot, String),
```
carllin, line 890: nit: we can spell out the full "certificate"

runtime/src/genesis_utils.rs
```diff
@@ -99,6 +100,21 @@ pub fn create_genesis_config_with_vote_accounts(
...
+    mint_lamports: u64,
+    voting_keypairs: &[impl Borrow<ValidatorVoteKeypairs>],
+    stakes: Vec<u64>,
+) -> GenesisConfigInfo {
+    create_genesis_config_with_vote_accounts_and_cluster_type(
+        mint_lamports,
+        voting_keypairs,
+        stakes,
+        ClusterType::Development,
+        true,
```
carllin, line 117: nice was going to need this for local cluster tests

carllin: @ksn6 can use this to setup the vote accounts too for primordial stakes in the gce cluster genesis setup scripts

core/src/replay_stage.rs
```diff
@@ -4073,6 +4103,97 @@ impl ReplayStage {
         new_frozen_slots
     }
 
+    // The bank must contain notarization cert for parent bank and skip cert for all slots between parent and current bank.
+    fn alpenglow_check_cert_in_bank(bank: &Bank) -> Result<(), BlockstoreProcessorError> {
+        if let Some(parent_bank) = bank.parent() {
```
carllin: can do a `let Some() else {return}` branch to avoid nesting

carllin: also annoying if a bank doesn't have a parent, should only happen on rooted banks, so i guess we should just return Ok(()) here

core/src/replay_stage.rs
```diff
@@ -4073,6 +4103,97 @@ impl ReplayStage {
...
+            let parent_slot = parent_bank.slot();
+            let parent_epoch_stakes = parent_bank
+                .epoch_stakes(parent_bank.epoch())
+                .expect("epoch stakes must exist");
+            let parent_hash = parent_bank.hash();
+            let mut notarization_stake = 0;
+            let mut skip_stake_map: HashMap<Slot, Stake> = HashMap::new();
```
carllin: for the skip check I think it's  more consistent to create a `skip_pool` and plug the skip ranges into there so that we use the same logic everywhere for generation/checking.

core/src/replay_stage.rs
```diff
@@ -4073,6 +4103,97 @@ impl ReplayStage {
...
+                    }
+                });
+            // Alpenglow VoteState can't store any vote for slot 0. This is okay because bank 0
+            // is genesis bank and doesn't need to be notarized.
+            if parent_slot > 0
+                && (notarization_stake as f64
+                    / parent_bank
+                        .epoch_total_stake(parent_bank.epoch())
+                        .expect("stake must exist") as f64)
+                    < SUPERMAJORITY
```
carllin: usually we check for > SUPERMAJORITY for passing, so <= is a fail

In skip pool we calculate a threshold like this: https://github.com/anza-xyz/alpenglow/blob/bd1b54cb68bb97e5222335afe7a1182424d1939c/core/src/alpenglow_consensus/skip_pool.rs#L255-L257

Let's also use the same threshold check everywhere to prevent rounding errors, shuold be a utility function exposed in the alpenglow_consensus.rs module i think

core/src/replay_stage.rs
wen-coding: Done.

ledger/src/blockstore_processor.rs
wen-coding: Done

core/src/replay_stage.rs
wen-coding: Hmm, that's a good point. I think we only need to check certs for any Alpenglow block, but no notarization check for the first one.

So if first_alpenglow_slot is 100:
1. for block 100 doesn't need to check notarization
2. for block 101 which links to block 99, no need to check notarization, but need to check skip

wen-coding: Changed

core/src/replay_stage.rs
wen-coding: Changed

core/src/replay_stage.rs
wen-coding: Moved to utils.rs.

core/src/replay_stage.rs
carllin: yup! thats how the tick verification works as well

core/src/replay_stage.rs
wen-coding: Thought about that, but skip_pool currently requires adding vote through passing in transactions, while I'm currently using VoteState. It seems more hassle to construct arbitrary Transaction rather than do simple bean counting here.
The benefit of using VoteState is: I only care about the transactions which landed in VoteState, don't care about anything which got rejected for whatever reason. And you can have multiple transactions in the same block, VoteState will nicely decouple everything and does its own check for you.
I can also use Skip transactions which landed in previous blocks.
I also don't need all extension checks in skip_pool.

carllin: https://github.com/anza-xyz/alpenglow/pull/99 what about this to make it generic, now you can construct a `SkipPool<()>` and no need to pass a transaction

wen-coding: Changed, but I had to make update() public so we can use the new utility function in both places. How does this look?

core/src/replay_stage.rs
```diff
@@ -4073,6 +4106,101 @@ impl ReplayStage {
...
+                parent_slot
+            );
+            return Err(BlockstoreProcessorError::InvalidCert(
+                bank.slot(),
+                parent_slot,
+                "Notarization".to_string(),
+            ));
+        }
+        if must_skip_start <= must_skip_end {
+            // TODO(wen): the stake can be incorrect.
```
carllin, line 4281: created an issue for this for my PR as well https://github.com/anza-xyz/alpenglow/issues/103

PR #98, opened 2025-03-18
Modify ticks and hashes_per_tick check for alpenglow blocks

carllin:
#### Problem
1. Alpenglow reduces the number of ticks to one per slot, and no ticks for skip slots, but the tick verification code still expects ticks_per_slot ticks for every slot between the slot `X` and its parent `Y`
2. Current PoH verification expects hashes_per_tick hashes between each tick, but alpenglow doesn't actually hash, so we should remove that check for alpenglow ticks

#### Summary of Changes

Fixes #

### DIFF
--- a/core/src/consensus/progress_map.rs
+++ b/core/src/consensus/progress_map.rs
@@ -140,6 +140,31 @@ impl ForkProgress {
         }
     }
 
+    fn calculate_alpenglow_ticks(
+        slot: Slot,
+        first_alpenglow_slot: Slot,
+        parent_slot: Slot,
+        ticks_per_slot: u64,
+    ) -> Option<u64> {
+        // Slots before alpenglow shouldn't have alpenglow ticks
+        if slot < first_alpenglow_slot {
+            return None;
+        }
+
+        let alpenglow_ticks = if parent_slot < first_alpenglow_slot && slot >= first_alpenglow_slot
+        {
+            // 1. All slots between the parent and the first alpenglow slot need to
+            // have `ticks_per_slot` ticks
+            // 2. One extra tick for the actual alpenglow slot
+            // 3. There are no ticks for any skipped alpenglow slots
+            (first_alpenglow_slot - parent_slot - 1) * ticks_per_slot + 1
+        } else {
+            1
+        };
+
+        Some(alpenglow_ticks)
+    }
+
     pub fn new_from_bank(
         bank: &Bank,
         validator_identity: &Pubkey,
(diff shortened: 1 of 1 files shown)


core/src/consensus/progress_map.rs
```diff
@@ -168,6 +168,24 @@ impl ForkProgress {
             num_dropped_blocks_on_fork,
         );
 
+        if let Some(first_alpenglow_slot) = bank
+            .feature_set
+            .activated_slot(&solana_feature_set::secp256k1_program_enabled::id())
```
wen-coding, line 198: nit: I think I keep seeing this check everywhere, is it possible we do it in one place and pass around?

core/src/consensus/progress_map.rs
```diff
@@ -168,6 +168,24 @@ impl ForkProgress {
...
+        {
+            let num_expected_ticks = {
+                let parent_slot = bank.parent_slot();
+                if parent_slot < first_alpenglow_slot && bank.slot() >= first_alpenglow_slot {
+                    // 1. All slots between the parent and the first alpenglow slot need to
+                    // have `ticks_per_slot` ticks
+                    // 2. One extra tick for the actual alpenglow slot
+                    (first_alpenglow_slot - parent_slot - 1) * bank.ticks_per_slot() + 1
+                } else {
+                    1
```
wen-coding: Add comment that in Alpenglow world we have no tick for skipped slots?

core/src/consensus/progress_map.rs
```diff
@@ -168,6 +168,24 @@ impl ForkProgress {
...
+                }
+            };
+            bank.set_tick_height(bank.max_tick_height() - num_expected_ticks);
```
wen-coding: Do you think we need to add assertion here this value is something reasonable?

ledger/src/blockstore_processor.rs
```diff
@@ -1197,6 +1197,28 @@ fn verify_ticks(
...
+            if entries.is_empty() {
+                // This shouldn't happen, but good to double check
+                error!("Processing empty entries in verify_ticks()");
+                return Ok(());
+            }
+            // last entry must be a tick, as verified by the `has_trailing_entry`
+            // check above. Because in Alpenglow the last tick does not have any
+            // hashing guarantees, we pass everything but that last tick to the
+            // entry verification.
+            entries = &entries[..entries.len() - 1];
```
wen-coding, line 1218: Err, do we have to change entries inside verify_ticks? Can we do skip_verification for all Alpenglow blocks instead in the caller?

core/src/consensus/progress_map.rs
carllin: i think it's uglier to do that, you would have to add arguments to all these functions that already take a bank

core/src/consensus/progress_map.rs
carllin: it's strictly less than max_tick_height() which is something reasonable

ledger/src/blockstore_processor.rs
carllin: we need to verify alpenglow blocks that have parents from before the epoch boundary because those parents have ticks, otherwise anybody could make a new alpenglow block with a very early parent from the previous epoch

ledger/src/blockstore_processor.rs
```diff
@@ -1197,6 +1197,28 @@ fn verify_ticks(
...
+        .activated_slot(&solana_feature_set::secp256k1_program_enabled::id())
+    {
+        if bank.parent_slot() >= first_alpenglow_slot {
+            return Ok(());
+        }
+
+        if bank.slot() >= first_alpenglow_slot && next_bank_tick_height == max_bank_tick_height {
+            if entries.is_empty() {
+                // This shouldn't happen, but good to double check
+                error!("Processing empty entries in verify_ticks()");
```
wen-coding, line 1211: If this shouldn't happen, should we return Ok(()) here?

carllin: It shouldn't happen because I'm assuming the callers will not pass this function an empty set of entries to verify, i.e. blockstore shouldn't deserialize empty entries

But an empty set of entries always passes the hashes_per_tick() check, so returning Ok(()) is fine here

ledger/src/blockstore_processor.rs
wen-coding: Can we someday do without that one tick in Alpenglow block?

PR #100, opened 2025-03-20
Fix some panics

carllin:
#### Problem
1. When the leader creates the first alpenglow slot during migration, it's possible that they finish migration but the alpenglow slot is not frozen yet. This means the highest frozen bank is still less than the first alpenglow slot
2. In alpenglow_start_leader(), it's possible that after migration no slot has been notarized yet

#### Summary of Changes
Remove panicking asserts

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -2484,7 +2484,7 @@ impl ReplayStage {
                 // We are in full alpenglow mode
                 let highest_certificate_slot = cert_pool.highest_certificate_slot();
                 if highest_certificate_slot < first_alpenglow_slot.unwrap() {
-                    // We haven't got a notarization yet for any of the first
+                    // We haven't got a notarization cert yet for any of the first
                     // alpenglow slots after the migration, wait for something to
                     // get notarized
                     info!(


core/src/replay_stage.rs
```diff
@@ -2404,6 +2404,12 @@ impl ReplayStage {
                 leader_schedule_cache,
             );
         }
+
+        info!(
```
wen-coding, line 2408: Do we still need info! on line 2376 if we log here?

core/src/replay_stage.rs
```diff
@@ -2476,6 +2482,17 @@ impl ReplayStage {
                 }
             } else {
                 // We are in full alpenglow mode
+                let highest_certificate_slot = cert_pool.highest_certificate_slot();
+                if highest_certificate_slot < first_alpenglow_slot.unwrap() {
+                    // We haven't got a notarization yet for any of the first
```
wen-coding: nit: notarization -> notarization cert

core/src/replay_stage.rs
carllin: We can exit before this line is hit

PR #101, opened 2025-03-20
Ingest gossip alpenglow votes into replay

carllin:
#### Summary of Changes

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -2077,7 +2077,6 @@ impl ReplayStage {
         let first_alpenglow_slot = first_alpenglow_slot?;
         let mut cached_root_bank = None;
 
-        // Find the next maximum finalization certificate
         alpenglow_vote_receiver
             .try_iter()
             .filter_map(|(vote, vote_account_pubkey, tx)| {
@@ -2093,31 +2092,38 @@ impl ReplayStage {
                 if validator_stake == 0 {
                     return None;
                 }
+
                 let total_stake = root_bank.epoch_total_stake(epoch)?;
-                cert_pool
-                    .add_vote(
-                        &vote,
-                        tx.into(),
-                        &vote_account_pubkey,
-                        validator_stake,
-                        total_stake,
-                    )
-                    .ok()
-                    .flatten()
-                    .filter(|cert| {
-                        info!(
-                            "{} got new highest gossip cert for {:?} from gossip vote",
-                            id, cert,
-                        );
+                match cert_pool.add_vote(
+                    &vote,
+                    tx.into(),
+                    &vote_account_pubkey,
+                    validator_stake,
+                    total_stake,
+                ) {
+                    Ok(Some(cert)) if cert.is_finalize() => {
                         let is_frozen = bank_forks
                             .read()
                             .unwrap()
                             .get(vote.slot())
-                            .map(|bank| bank.is_frozen())
-                            .unwrap_or(false);
-                        cert.is_finalize() && is_frozen
-                    })
-                    .map(|cert| cert.slot())
+                            .is_some_and(|bank| bank.is_frozen());
+
+                        if is_frozen {
+                            info!(
+                                "{} got new highest gossip cert for {:?} from gossip vote",
+                                id, cert
+                            );
+                            Some(cert.slot())
+                        } else {
+                            None
+                        }
+                    }
+                    Ok(_) => None,
+                    Err(e) => {
+                        error!("Adding vote {:?} errored with {:?}", vote, e);
+                        None
+                    }
+                }
             })
             .max()
     }


core/src/cluster_info_vote_listener.rs
```diff
@@ -465,17 +495,13 @@ impl ClusterInfoVoteListener {
...
         }
 
         // Hold lock for whole function to ensure hash consistency with bank_forks
         let mut slots_dumped = dumped_slot_subscription.lock().unwrap();
-        let Some((last_vote_slot, last_vote_hash)) = vote.last_voted_slot_hash() else {
-            // TODO: handle sending skip votes to replay because skp votes don't have a hash and will
-            // return
-            return;
-        };
+        let (last_vote_slot, last_vote_hash) = vote.last_voted_slot_hash().unwrap();
```
wen-coding, line 504: I forgot, why did we previously say don't send skip, and now why did it change?

core/src/cluster_info_vote_listener.rs
```diff
@@ -545,29 +571,20 @@ impl ClusterInfoVoteListener {
...
-                    .unwrap_or(Slot::MAX);
-
-                // Optimistic confirmation and duplicate confirmation are no longer relevant after
-                // alpenglow
-                if slot < first_alpenglow_slot {
-                    if reached_threshold_results[0] {
-                        if let Some(sender) = duplicate_confirmed_slot_sender {
-                            let _ = sender.send(vec![(slot, hash)]);
-                        }
+                if reached_threshold_results[0] {
```
wen-coding, line 574: I know you probably inherited this code, but this is so ugly, can we replace with a struct maybe?

core/src/replay_stage.rs
```diff
@@ -2054,6 +2065,41 @@ impl ReplayStage {
         }
     }
 
+    fn process_gossip_alpenglow_votes(
```
wen-coding: nit: from this name it's not obvious what it returns

core/src/replay_stage.rs
```diff
@@ -2054,6 +2065,41 @@ impl ReplayStage {
...
+                }
+
+                let root_bank =
+                    cached_root_bank.get_or_insert_with(|| bank_forks.read().unwrap().root_bank());
+                cert_pool
+                    .add_vote(
+                        &vote,
+                        tx.into(),
+                        &pubkey,
+                        root_bank.epoch_vote_account_stake(&pubkey),
```
wen-coding: Just a comment: since this epoch stake is wrong if skip range crosses epoch boundary, I bet we need to send the whole epoch stakes in some day.

core/src/cluster_info_vote_listener.rs
carllin: this is now reverted to nonalpenglow transactions only, so only tower votes should ever reach this function

core/src/replay_stage.rs
carllin: renamed to `ingest_gossip_alpenglow_votes_into_certificate_pool`

core/src/replay_stage.rs
carllin: ah yeah thats a good point, will change that

carllin: also have to check that the root slot is frozen, will change as well

core/src/replay_stage.rs
```diff
@@ -2054,6 +2067,61 @@ impl ReplayStage {
...
+                let total_stake = root_bank.epoch_total_stake(epoch)?;
+                cert_pool
+                    .add_vote(
+                        &vote,
+                        tx.into(),
+                        &vote_account_pubkey,
+                        validator_stake,
+                        total_stake,
+                    )
+                    .ok()
```
wen-coding: I think if you have an error here we should log it

core/src/replay_stage.rs
```diff
@@ -2054,6 +2067,61 @@ impl ReplayStage {
...
+                    .flatten()
+                    .filter(|cert| {
+                        info!(
+                            "{} got new highest gossip cert for {:?} from gossip vote",
+                            id, cert,
+                        );
+                        let is_frozen = bank_forks
```
wen-coding, line 2105: hmm, does our vote contain skip votes as well? Do you need to check is_frozen for skip votes?

core/src/replay_stage.rs
```diff
@@ -2054,6 +2067,61 @@ impl ReplayStage {
...
+                            .read()
+                            .unwrap()
+                            .get(vote.slot())
+                            .map(|bank| bank.is_frozen())
+                            .unwrap_or(false);
+                        cert.is_finalize() && is_frozen
```
wen-coding: I think you can do early return when cert is not finalize, no need to lock bank_forks.

Overall it feels weird to check is_frozen inside this function, can we maybe check it in the caller instead?

carllin: > Overall it feels weird to check is_frozen inside this function, can we maybe check it in the caller instead?

You can't because if you have new finalization certs: for 1, 2, 3, but you only froze up to 2, this function will return the max 3

core/src/cluster_info_vote_listener.rs
carllin: this function is exactly what exists in anza/master today, just reverted the previous changes

core/src/replay_stage.rs
carllin: one of the errors could be a duplicate vote, so might be a bit noisy, but I'll log for now

core/src/replay_stage.rs
carllin: we only care about frozen for finalization because we don't want to set_root on a nonfrozen slot. For skips we don't really care

wen-coding: Let's return early when the cert is not finalization then, save some bank_forks read lock

core/src/replay_stage.rs
wen-coding: I mean, we don't have to do it every time inside the loop, we can grab frozen_banks() outside the loop and just check against the list of frozen banks. Grabbing reading locks everywhere makes me a bit uncomfortable, even though performance wise it could be fine.

But I understand we don't have new finalization certs that often so this is probably fine for now as well.

carllin: > But I understand we don't have new finalization certs that often so this is probably fine for now as well.

Yeah, only triggers once per slot, so i think the lock should be fine

PR #104, opened 2025-03-21
feat: create alpenglow vote accounts in GCE test cluster

ksn6:
Support the ability for `net.sh` to create Alpenglow validators.

### DIFF
--- a/cli/src/stake.rs
+++ b/cli/src/stake.rs
@@ -1748,28 +1748,25 @@ pub fn process_deactivate_stake_account(
             rpc_client.commitment(),
         )?;
 
-        let efdd = match vote_state {
-            crate::vote::VoteStateWrapper::VoteState(vote_state) => {
+        let is_eligible_for_deactivate_delinquent = match vote_state {
+            crate::vote::VoteStateWrapper::VoteState(ref vote_state) => {
                 eligible_for_deactivate_delinquent(&vote_state.epoch_credits, current_epoch)
             }
-            // TODO: we're basically copying `eligible_for_deactivate_delinquent` from POH
-            // here. Ensure that this is correct.
-            crate::vote::VoteStateWrapper::AlpenglowVoteState(vote_state) => {
-                let epoch_credits = vote_state.epoch_credits();
+            crate::vote::VoteStateWrapper::AlpenglowVoteState(ref vote_state) => {
+                let credits = vote_state.epoch_credits().credits();
+                let prev_credits = vote_state.epoch_credits().prev_credits();
 
-                if epoch_credits.credits() == 0 && epoch_credits.prev_credits() == 0 {
-                    true
-                } else if let Some(minimum_epoch) = current_epoch
-                    .checked_sub(stake::MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION as Epoch)
-                {
-                    epoch_credits.epoch() <= minimum_epoch
-                } else {
-                    false
-                }
+                let mut epoch_credits = vec![];
+
+                if credits != 0 && prev_credits != 0 {
+                    epoch_credits.push((current_epoch, credits, prev_credits));
+                };
+
+                eligible_for_deactivate_delinquent(epoch_credits.as_slice(), current_epoch)
             }
         };
 
-        if !efdd {
+        if is_eligible_for_deactivate_delinquent {
             return Err(CliError::BadParameter(format!(
                 "Stake has not been delinquent for {} epochs",
                 stake::MINIMUM_DELINQUENT_EPOCHS_FOR_DEACTIVATION,
--- a/cli/src/vote.rs
+++ b/cli/src/vote.rs
@@ -866,14 +866,11 @@ pub fn process_create_vote_account(
         let node_pubkey = identity_pubkey;
         let authorized_voter = authorized_voter.unwrap_or(identity_pubkey);
 
-        let to = if seed.is_some() {
-            &vote_account_address
-        } else {
-            &vote_account_pubkey
-        };
+        let from_pubkey = &config.signers[0].pubkey();
+        let to_pubkey = &vote_account_address;
 
         let mut ixs = if is_alpenglow {
-            let iaid = InitializeAccountInstructionData {
+            let initialize_account_ixn_meta = InitializeAccountInstructionData {
                 node_pubkey,
                 authorized_voter,
                 authorized_withdrawer,
@@ -881,14 +878,17 @@ pub fn process_create_vote_account(
             };
 
             let create_ix = solana_system_interface::instruction::create_account(
-                &config.signers[0].pubkey(),
-                to,
+                from_pubkey,
+                to_pubkey,
                 lamports,
                 alpenglow_vote::state::VoteState::size() as u64,
                 &alpenglow_vote::id(),
             );
 
-            let init_ix = alpenglow_vote::instruction::initialize_account(*to, &iaid);
+            let init_ix = alpenglow_vote::instruction::initialize_account(
+                *to_pubkey,
+                &initialize_account_ixn_meta,
+            );
 
             vec![create_ix, init_ix]
         } else {
@@ -907,8 +907,8 @@ pub fn process_create_vote_account(
             }
 
             vote_instruction::create_account_with_config(
-                &config.signers[0].pubkey(),
-                to,
+                from_pubkey,
+                to_pubkey,
                 &vote_init,
                 lamports,
                 create_vote_account_config,
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -5365,7 +5365,6 @@ pub(crate) mod tests {
         blockstore_processor::{
             confirm_full_slot, fill_blockstore_slot_with_ticks, process_bank_0, ProcessOptions,
         },
-        build_alpenglow_vote::ALPENGLOW_VOTE_SO_PATH,
         crossbeam_channel::unbounded,
         itertools::Itertools,
         solana_account::{AccountSharedData, WritableAccount},
@@ -5388,7 +5387,7 @@ pub(crate) mod tests {
             accounts_background_service::AbsRequestSender,
             commitment::{BlockCommitment, VOTE_THRESHOLD_SIZE},
             genesis_utils::{
-                create_genesis_config_with_alpenglow_vote_accounts, GenesisConfigInfo,
+                create_genesis_config_with_alpenglow_vote_accounts_no_program, GenesisConfigInfo,
                 ValidatorVoteKeypairs,
             },
         },
@@ -10764,11 +10763,10 @@ pub(crate) mod tests {
         let my_keypairs = &validator_voting_keypairs[0];
         let lamports = 10_000;
         let GenesisConfigInfo { genesis_config, .. } =
-            create_genesis_config_with_alpenglow_vote_accounts(
+            create_genesis_config_with_alpenglow_vote_accounts_no_program(
                 lamports,
                 &validator_voting_keypairs,
                 vec![100; validator_voting_keypairs.len()],
-                ALPENGLOW_VOTE_SO_PATH,
             );
         let bank0 = Arc::new(Bank::new_for_tests(&genesis_config));
         bank0.freeze();
--- a/net/net.sh
+++ b/net/net.sh
@@ -295,16 +295,6 @@ deployBootstrapValidator() {
   remoteHome=$(remoteHomeDir "$ipAddress")
   local remoteCargoBin="${remoteHome}/.cargo/bin"
 
-  echo "!!!!! DEPLOYING BOOTSTRAP VALIDATOR !!!!!"
-  echo "sshOptions"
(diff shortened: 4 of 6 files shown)


net/remote/remote-node.sh
```diff
@@ -239,6 +240,13 @@ EOF
         done
       fi
 
+      if $alpenglow; then
+        echo "Consensus method: Alpenglow"
+        args+=(--alpenglow)
```
wen-coding: This generally looks fine to me, but I assume there is another PR to add --alpenglow? I'm thinking maybe approving that PR first.

ksn6: Yep the plan is to add the `--alpenglow` flag to this PR. Going to do this today.

We were trying to sort out the best way of addressing this PR in this thread: https://anza-xyz.slack.com/archives/C06GZ20BPQB/p1742593003630439.

ksn6: Turns out - I don't think we need to add an `--alpenglow` flag to `solana-genesis` after all. We're now addressing this by disabling the `secp256k1_program_enabled` feature when we want to do POH.

cli/src/vote.rs
```diff
@@ -842,35 +855,64 @@ pub fn process_create_vote_account(
...
+                node_pubkey,
+                authorized_voter,
+                authorized_withdrawer,
+                commission,
+            };
+
+            let create_ix = solana_system_interface::instruction::create_account(
+                &config.signers[0].pubkey(),
+                to,
+                lamports,
```
AshwinSekar, line 883: You'll need to update `space` and `amount` (above) to match the size of the new vote account

ksn6: good catch - done

cli/src/stake.rs
```diff
@@ -1747,7 +1747,29 @@ pub fn process_deactivate_stake_account(
             &vote_account_address,
             rpc_client.commitment(),
         )?;
-        if !eligible_for_deactivate_delinquent(&vote_state.epoch_credits, current_epoch) {
+
+        let efdd = match vote_state {
```
wen-coding: nit: it's really hard to tell what "efdd" stands for, longer variable name maybe?

cli/src/stake.rs
```diff
@@ -1747,7 +1747,29 @@ pub fn process_deactivate_stake_account(
...
+            crate::vote::VoteStateWrapper::VoteState(vote_state) => {
+                eligible_for_deactivate_delinquent(&vote_state.epoch_credits, current_epoch)
+            }
+            // TODO: we're basically copying `eligible_for_deactivate_delinquent` from POH
+            // here. Ensure that this is correct.
+            crate::vote::VoteStateWrapper::AlpenglowVoteState(vote_state) => {
```
wen-coding: nit: is it possible to convert to the old data structure here and call the same function?

cli/src/vote.rs
```diff
@@ -829,48 +846,81 @@ pub fn process_create_vote_account(
...
-        let mut create_vote_account_config = CreateVoteAccountConfig {
-            space,
-            ..CreateVoteAccountConfig::default()
-        };
-        let to = if let Some(seed) = seed {
-            create_vote_account_config.with_seed = Some((&vote_account_pubkey, seed));
+        let node_pubkey = identity_pubkey;
+        let authorized_voter = authorized_voter.unwrap_or(identity_pubkey);
+
+        let to = if seed.is_some() {
```
wen-coding: nit: to -> to_pubkey ?
And we can define from_pubkey to be &config.signers[0].pubkey() to share between TowerBFT and Alpenglow case.

cli/src/vote.rs
```diff
@@ -829,48 +846,81 @@ pub fn process_create_vote_account(
...
-        });
+        let mut ixs = if is_alpenglow {
+            let iaid = InitializeAccountInstructionData {
+                node_pubkey,
+                authorized_voter,
+                authorized_withdrawer,
+                commission,
+            };
+
+            let create_ix = solana_system_interface::instruction::create_account(
```
wen-coding, line 880: Just curious, do we support seed?

cli/src/vote.rs
```diff
@@ -829,48 +846,81 @@ pub fn process_create_vote_account(
...
-            lamports,
-            create_vote_account_config,
-        )
-        .with_memo(memo)
-        .with_compute_unit_config(&ComputeUnitConfig {
-            compute_unit_price,
-            compute_unit_limit,
-        });
+        let mut ixs = if is_alpenglow {
+            let iaid = InitializeAccountInstructionData {
```
wen-coding: nit: move variable closer to where it's used?

cli/src/vote.rs
```diff
@@ -1257,31 +1310,104 @@ pub fn process_vote_update_commission(
     }
 }
 
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum VoteStateWrapper {
```
wen-coding, line 1314: nit: explain why you need this wrapper here, and when/whether it will go away.

cli/src/vote.rs
```diff
@@ -1257,31 +1310,104 @@ pub fn process_vote_update_commission(
...
-            "{vote_account_pubkey:?} is not a vote account"
-        ))
-        .into());
-    }
-    let vote_state = VoteState::deserialize(&vote_account.data).map_err(|_| {
-        CliError::RpcRequestError(
-            "Account data could not be deserialized to vote state".to_string(),
-        )
-    })?;
+    let vote_state_wrapper = match vote_account.owner {
```
wen-coding, line 1385: nit: we could move this deserialize function into VoteStateWrapper so it's easier to add unittest.

cli/src/vote.rs
```diff
@@ -1303,55 +1429,73 @@ pub fn process_show_vote_account(
 
     let mut votes: Vec<CliLandedVote> = vec![];
     let mut epoch_voting_history: Vec<CliEpochVotingHistory> = vec![];
-    if !vote_state.votes.is_empty() {
-        for vote in &vote_state.votes {
-            votes.push(vote.into());
+    let mut epoch_rewards = None;
+
+    // TODO: handle Alpenglow case
```
wen-coding, line 1434: nit: if you don't handle Alpenglow now, could do early return so most of the code is intact?

cli/src/vote.rs
```diff
@@ -1853,6 +1997,7 @@ mod tests {
                     memo: None,
                     fee_payer: 0,
                     compute_unit_price: None,
+                    is_alpenglow: false,
```
wen-coding, line 2000: nit: add testcase for Alpenglow case?

cli/tests/vote.rs
```diff
@@ -59,6 +59,7 @@ fn test_vote_authorize_and_withdraw(compute_unit_price: Option<u64>) {
         memo: None,
         fee_payer: 0,
         compute_unit_price,
+        is_alpenglow: false,
```
wen-coding, line 62: Do we have test case for whatever is supported now?

genesis/src/main.rs
```diff
@@ -829,6 +867,10 @@ fn main() -> Result<(), Box<dyn error::Error>> {
         }
     }
 
+    if let Some(alpenglow_so_path) = alpenglow_so_path {
+        include_alpenglow_bpf_program(&mut genesis_config, alpenglow_so_path);
```
wen-coding, line 871: nit: specify why we need this and when/whether it will go away.

genesis/src/main.rs
```diff
@@ -1294,6 +1337,7 @@ mod tests {
             100,
             &Rent::default(),
             &mut genesis_config,
+            false,
```
wen-coding, line 1340: nit: test case for Alpenglow case?

net/net.sh
```diff
@@ -295,6 +295,16 @@ deployBootstrapValidator() {
...
 
+  echo "!!!!! DEPLOYING BOOTSTRAP VALIDATOR !!!!!"
+  echo "sshOptions"
+  echo ${sshOptions[*]}
+  echo "SOLANA_ROOT"
+  echo $SOLANA_ROOT
+  echo "ipAddress"
+  echo $ipAddress
+  echo "remoteCargoBin"
+  echo $remoteCargoBin
```
wen-coding: Do we need all of this?

net/net.sh
```diff
@@ -328,6 +339,16 @@ startBootstrapLeader() {
 
     deployBootstrapValidator "$ipAddress"
 
+    if $alpenglow; then
+      declare remoteHome
```
wen-coding, line 338: nit: add comments to say why we need this

rpc/src/rpc_health.rs
```diff
@@ -56,7 +56,9 @@ impl RpcHealth {
             return RpcHealthStatus::Ok;
         }
         if !self.startup_verification_complete.load(Ordering::Acquire) {
-            return RpcHealthStatus::Unknown;
+            // TODO: ¯\_(ツ)_/¯
```
wen-coding: I think Carl filed an issue, you can refer to the issue here.

rpc/src/rpc_health.rs
```diff
@@ -170,28 +178,38 @@ pub mod tests {
         // Override health check set to true - status is ok
         assert_eq!(health.check(), RpcHealthStatus::Ok);
 
+        // TODO(alpenglow): RpcHealthStatus::Ok should really be RpcHealthStatus::Unknown
```
wen-coding, line 181: Same, refer to the issue here.

core/src/replay_stage.rs
```diff
@@ -10767,6 +10768,7 @@ pub(crate) mod tests {
                 lamports,
                 &validator_voting_keypairs,
                 vec![100; validator_voting_keypairs.len()],
+                ALPENGLOW_VOTE_SO_PATH,
```
wen-coding: wait, I only need a genesis config with Alpenglow accounts here, why do I need a real path?

cli/src/stake.rs
ksn6: done!

cli/src/stake.rs
ksn6: good catch - done!

cli/src/vote.rs
ksn6: done

cli/src/vote.rs
ksn6: which variable should we move?

net/net.sh
ksn6: removed

net/net.sh
ksn6: done

rpc/src/rpc_health.rs
ksn6: done

rpc/src/rpc_health.rs
ksn6: done

core/src/replay_stage.rs
ksn6: `create_genesis_config_with_alpenglow_vote_accounts` eventually ends up invoking `create_genesis_config_with_leader_ex_no_features`, which then includes `spl_alpenglow-vote.so`

I agree that this isn't clean - we'd need to write the code in a way such that the following block of code isn't in `create_genesis_config_with_leader_ex_no_features`:

```
    if let Some(alpenglow_so_path) = alpenglow_so_path {
        include_alpenglow_bpf_program(&mut genesis_config, alpenglow_so_path);
    }
```

@carllin - thoughts on this?

carllin: Maybe creating a test-only (tag with `#[cfg(feature = "dev-context-only-utils")]`) `create_genesis_config_with_alpenglow_vote_accounts_no_program()` and passing an option for the alpenglow program path in the internals

cli/src/vote.rs
ksn6: Could you elaborate on what you mean by supporting seed?

We do support authorizing with seeds, etc.; e.g. see:

- https://github.com/solana-program/alpenglow-vote/blob/master/program/src/instruction.rs#L67-L77
- https://github.com/solana-program/alpenglow-vote/blob/master/program/src/processor.rs#L110-L137

core/src/replay_stage.rs
ksn6: done

cli/src/vote.rs
wen-coding: Hmm I forgot as well, never mind.

cli/src/vote.rs
wen-coding: I just saw in the TowerBFT case they did
```
if let Some(seed) = seed {
        create_vote_account_config.with_seed = Some((&vote_account_pubkey, seed));
}
```
Wondered whether you should check seed here as well.

ksn6: So I'm not sure why this code block is necessary for the `TowerBFT` case, as it doesn't seem to do anything. The code is currently:

image

But the `.with_seed` piece isn't used in `create_account_with_config`:

image

In fact, we could probably entirely get rid of `create_vote_account_config` and just pass in the `space` piece.

PR #108, opened 2025-03-23
feat: auto-build alpenglow-vote

ksn6:
#### Problem
Iterating on `alpenglow-vote` and viewing side effects on `alpenglow` is currently cumbersome. We have to manually re-compile `alpenglow-vote` and manually install the associated `.so` on test validators. This PR automates the process.

#### Summary of Changes
We introduce a Rust build script that detects whether `alpenglow-vote` either points to a (1) local checkout or (2) a github repo. We accordingly automatically build `alpenglow-vote` via `cargo-sbf` and copy the resulting `.so` to a directory local to the current `alpenglow` checkout.

### DIFF
--- a/runtime/src/genesis_utils.rs
+++ b/runtime/src/genesis_utils.rs
@@ -1,5 +1,6 @@
 use {
     alpenglow_vote::state::VoteState as AlpenglowVoteState,
+    build_alpenglow_vote::ALPENGLOW_VOTE_SO_PATH,
     log::*,
     solana_feature_set::{self, FeatureSet, FEATURE_NAMES},
     solana_loader_v3_interface::state::UpgradeableLoaderState,
@@ -424,8 +425,7 @@ pub fn create_genesis_config_with_leader_ex_no_features(
     solana_stake_program::add_genesis_accounts(&mut genesis_config);
 
     if is_alpenglow {
-        // TODO: uncomment this
-        // include_alpenglow_bpf_program(&mut genesis_config, ALPENGLOW_VOTE_SO_PATH);
+        include_alpenglow_bpf_program(&mut genesis_config, ALPENGLOW_VOTE_SO_PATH);
     }
 
     genesis_config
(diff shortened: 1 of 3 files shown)


runtime/src/genesis_utils.rs
```diff
@@ -286,6 +287,28 @@ pub fn activate_feature(genesis_config: &mut GenesisConfig, feature_id: Pubkey)
     );
 }
 
+pub fn include_alpenglow_bpf_program(genesis_config: &mut GenesisConfig) {
```
carllin: I think it might make sense to just call this function whenever alpenglow flag is specified https://github.com/anza-xyz/alpenglow/blob/1f4b6cae8d368a8ffe837157d6672fcae299a574/runtime/src/genesis_utils.rs#L127

genesis/src/main.rs
```diff
@@ -829,6 +830,13 @@ fn main() -> Result<(), Box<dyn error::Error>> {
         }
     }
 
+    if genesis_config
+        .accounts
+        .contains_key(&solana_feature_set::secp256k1_program_enabled::id())
+    {
+        include_alpenglow_bpf_program(&mut genesis_config);
+    }
+
```
carllin: hmmm I think we might need to just pass a new --alpenglow flag to this function anyways, because there are a few pieces in this file that need to be modified to support alpenglow

For instance right now it still creates Tower vote accounts: https://github.com/anza-xyz/alpenglow/blob/1f4b6cae8d368a8ffe837157d6672fcae299a574/genesis/src/main.rs#L265-L281

If we're going to add that flag anyways, might as well just condition on that flag

ksn6: fixed; we now create Alpenglow vote accounts if `is_alpenglow` is set to `true`

runtime/src/genesis_utils.rs
ksn6: fixed - right now, we're only calling this in `create_genesis_config_with_leader_ex_no_features`, since the other `create_genesis_config` functions all call into this one; lmk if you'd like me to explicitly invoke `include_alpenglow_bpf_program` everywhere

PR #109, opened 2025-03-25
Add utility to add alpenglow vote program to genesis

carllin:
#### Problem
Currently no convenient way to load the alpenglow vote program for local cluster tests

#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/genesis_utils.rs
+++ b/runtime/src/genesis_utils.rs
@@ -303,7 +303,7 @@ pub fn include_alpenglow_bpf_program(genesis_config: &mut GenesisConfig, alpengl
         solana_program::bpf_loader_upgradeable::get_program_data_address(&address);
 
     // Generate the data for the program data account
-    let upgrade_authority_address = Pubkey::default();
+    let upgrade_authority_address = system_program::id();
     let mut program_data = bincode::serialize(&UpgradeableLoaderState::ProgramData {
         slot: 0,
         upgrade_authority_address: Some(upgrade_authority_address),


runtime/src/genesis_utils.rs
```diff
@@ -288,6 +289,63 @@ pub fn activate_feature(genesis_config: &mut GenesisConfig, feature_id: Pubkey)
...
+        .unwrap_or_else(|err| panic!("Error: failed to read alpenglow-vote program: {}", err));
+
+    // Derive the address for the program data account
+    let address = alpenglow_vote::id();
+    let loader = solana_program::bpf_loader_upgradeable::id();
+    let programdata_address =
+        solana_program::bpf_loader_upgradeable::get_program_data_address(&address);
+
+    // Generate the data for the program data account
+    let upgrade_authority_address = Pubkey::default();
```
AshwinSekar: very minor nit: could use the system program `system_program::id()` as I've heard that's the convention for these "enshrined" programs

runtime/src/genesis_utils.rs
```diff
@@ -365,6 +423,11 @@ pub fn create_genesis_config_with_leader_ex_no_features(
 
     solana_stake_program::add_genesis_accounts(&mut genesis_config);
 
+    if is_alpenglow {
+        // TODO: enable this
```
AshwinSekar, line 427: there's no harm in just always enabling this right? having this random program shouldn't break tower bft?

carllin: the .so is not being built without Karthik's PR, so it would panic on trying to find the binaries

PR #110, opened 2025-03-25
Ingest all-to-all Alpenglow votes into the cert pool.

wen-coding:
Ingest all-to-all Alpenglow votes into the cert pool.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -633,31 +633,60 @@ impl ReplayStage {
             let _exit = Finalizer::new(exit.clone());
             let mut identity_keypair = cluster_info.keypair().clone();
             let mut my_pubkey = identity_keypair.pubkey();
+            // Alpenglow specific objects
+            // TODO: Properly initialize VoteHistory from storage
+            let mut vote_history = VoteHistory::new(my_pubkey, bank_forks.read().unwrap().root());
+            let mut first_alpenglow_slot = bank_forks
+                .read()
+                .unwrap()
+                .root_bank()
+                .feature_set
+                .activated_slot(&solana_feature_set::secp256k1_program_enabled::id());
+
+            let mut is_alpenglow_migration_complete = false;
+            if let Some(first_alpenglow_slot) = first_alpenglow_slot {
+                if bank_forks
+                    .read()
+                    .unwrap()
+                    .frozen_banks()
+                    .iter()
+                    .any(|(slot, _bank)| *slot >= first_alpenglow_slot)
+                {
+                    info!("initiating alpenglow migration on startup");
+                    Self::initiate_alpenglow_migration(
+                        &poh_recorder,
+                        &mut is_alpenglow_migration_complete,
+                    );
+                }
+            }
+
             if my_pubkey != tower.node_pubkey {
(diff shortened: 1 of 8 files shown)


core/src/banking_stage/packet_receiver.rs
```diff
@@ -64,6 +70,34 @@ impl PacketReceiver {
         result
     }
 
+    fn forward_alpenglow_votes(
```
apfitzge: Let's not call this `forward`, it's just sending over a channel to another thread. `forward` in banking stage has historically meant forwarding to another validator

wen-coding: Changed.

core/src/banking_stage/packet_receiver.rs
```diff
@@ -64,6 +70,23 @@ impl PacketReceiver {
...
+        }: &ReceivePacketResults,
+        alpenglow_vote_sender: &AlpenglowVoteSender,
+    ) {
+        for packet in deserialized_packets.iter() {
+            if let Some(result) =
+                parse_alpenglow_vote_transaction_from_sanitized(packet.transaction())
+            {
+                let _ = alpenglow_vote_sender.send(result);
+            }
+        }
```
apfitzge, line 87: should we collect these into a vec and send a single message? it might be faster than sending each vote individually.

runtime/src/bank_utils.rs
```diff
@@ -71,9 +71,7 @@ pub fn find_and_send_votes(
                             let _ = alpenglow_vote_sender.send((
                                 vote,
                                 pubkey,
-                                tx.to_versioned_transaction()
-                                    .into_legacy_transaction()
-                                    .unwrap(),
+                                tx.to_versioned_transaction(),
```
apfitzge, line 75: just FYI, the doc-comment on this function indicates we shouldn't use this unless strictly necessary.

Fine for now with the PoC, but we may need to revisit what we send over the channel in the future.

core/src/banking_stage/packet_receiver.rs
wen-coding: Right now we have so few nodes in test cluster it doesn't really matter. But this is a good point, I think we will reevaluate later on larger testbed and change if needed, is that okay?

runtime/src/bank_utils.rs
wen-coding: Err, why is to_versioned_transaction not suitable to use?

core/src/banking_stage/packet_receiver.rs
apfitzge: Yeah of course - I just mention things I see that could be bottlenecks, but most of the time I am fine with leaving it do to at a later time.

runtime/src/bank_utils.rs
apfitzge: It converts to the `sdk` type - which may do a ton of internal allocations. We have been moving away from using the `sdk` types at all in perf sensitive areas.

apfitzge: again for now, I think it's fine to use this.

wen-coding: Thanks for the reminder, let me put a TODO to remove it later.

PR #111, opened 2025-03-26
Fix Tower initialization panics before migration of vote accounts

carllin:
#### Problem
load_tower() https://github.com/anza-xyz/alpenglow/blob/b15b9b18057fb1beb1085c045007eccc4ced489c/core/src/replay_stage.rs#L640-L644 and  process_to_create_tower() https://github.com/anza-xyz/alpenglow/blob/b15b9b18057fb1beb1085c045007eccc4ced489c/core/src/validator.rs#L1460-L1472 both assume the validator's vote account is a tower account and will panic on trying to parse the vote account as a tower vote account if passed an alpenglow vote account.

#### Summary of Changes
Disable loading the tower on alpenglow from genesis/post-migration

Fixes #

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -633,31 +633,60 @@ impl ReplayStage {
             let _exit = Finalizer::new(exit.clone());
             let mut identity_keypair = cluster_info.keypair().clone();
             let mut my_pubkey = identity_keypair.pubkey();
+            // Alpenglow specific objects
+            // TODO: Properly initialize VoteHistory from storage
+            let mut vote_history = VoteHistory::new(my_pubkey, bank_forks.read().unwrap().root());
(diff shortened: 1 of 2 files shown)


core/src/replay_stage.rs
```diff
@@ -633,31 +633,60 @@ impl ReplayStage {
...
-                            my_old_pubkey, my_pubkey, err
-                        );
-                        // drop(_exit) will set the exit flag, eventually tearing down the entire process
-                        return;
-                    }
-                };
-                warn!(
-                    "Identity changed during startup from {} to {}",
-                    my_old_pubkey, my_pubkey
-                );
```
carllin, line 660: all i did here was move this code earlier so that the `is_alpenglow_migration_complete` flag is available before loading the tower

PR #115, opened 2025-03-27
Make get_timestamp_estimate work for Alpenglow

wen-coding:
Make get_timestamp_estimate work for Alpenglow

### DIFF
--- a/runtime/src/genesis_utils.rs
+++ b/runtime/src/genesis_utils.rs
@@ -215,6 +215,7 @@ pub fn create_genesis_config_with_leader(
     )
 }
 
+#[cfg(feature = "dev-context-only-utils")]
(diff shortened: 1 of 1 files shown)


runtime/src/genesis_utils.rs
```diff
@@ -211,6 +211,29 @@ pub fn create_genesis_config_with_leader(
         mint_lamports,
         validator_pubkey,
         validator_stake_lamports,
+        false,
+    )
+}
+
+pub fn create_genesis_config_with_leader_enable_alpenglow(
```
carllin, line 219: if this is only used in tests i think we should flag it with `#[cfg(feature = "dev-context-only-utils")]` and add that feature

wen-coding: Good point, added.

PR #117, opened 2025-03-27
Alpenglow local cluster test

carllin:
#### Problem
No local cluster tests for alpenglow

#### Summary of Changes
Add local cluster test for alpenglow

Fixes #


local-cluster/tests/local_cluster.rs
```diff
@@ -135,6 +135,53 @@ fn test_local_cluster_start_and_exit_with_config() {
     assert_eq!(cluster.validators.len(), NUM_NODES);
 }
 
+#[test]
+#[serial]
+fn test_2_nodes_alpenglow() {
```
wen-coding: This can be any number of nodes right? Just generate the validator keys below.

carllin: done

Issue #65, opened 2025-03-03
Remove unnecessary data structures like EpochSlotsFrozenSlots, HeaviestSubtreeForkChoice


carllin:
#### Problem

These structures EpochSlotsFrozenSlots, HeaviestSubtreeForkChoice can be removed, and purged from set_root () as well which assumes the new roots are in these data structures, but are not needed in alpenglow_set_root()

Issue #30, opened 2025-02-05
Handle duplicate block dump and repair on detecting a certificate, should be much simpler now


carllin:
#### Problem
Also note it's possible to timeout your own block while creating it, handle this

#### Proposed Solution

Dependent on this first: https://github.com/anza-xyz/alpenglow/issues/23

qkniep:
Reading #265 made me look at repair again. This should be drastically simpler post-Alpenglow:
- I see no reason at all repair would receive votes directly.
- Repair can simply get notifications of which blocks to repair, directly from Pool.
- Also makes #56 obsolete?

Repair would then probably do the following steps:
- Maybe check in Blokstor whether the block already exists (could be done at the trigger instead).
- Actually repair the block.
- Possibly trigger repair of the parent, if it doesn't exist (could be done at the trigger instead).

And the only three triggers for repair should be:
- A block newly got notarized (fallback).
- The votes we saw for a slot meet the safe-to-notar threshold but we are missing the block.
- We finalized a block and are missing an ancestor.

PR #118, opened 2025-03-28
replay: separate voting loop, leader, root, and add skip logic

### LINKED ISSUE #23: Track Certificates per Slot/Type, maybe_start_leader() needs to ingest this to know when to create a bank

### LINKED ISSUE #28: Add timers/timeouts for skip votes
#### Problem
Also note it's possible to timeout your own block while creating it, handle this

#### Proposed Solution

Dependent on this first: https://github.com/anza-xyz/alpenglow/issues/23

### LINKED ISSUE #42: Design better equivalent to Poh::would_be_leader based on skip slots and notarizations, essential for removing Poh

### LINKED ISSUE #91: handle block_id in votes by leaders (currently blockid is none on leader blocks)

AshwinSekar:
This breaks migration, we will need to test and fix the migration pathway first thing after this lands.

further TODO list:
1) Safeguard replay for set root happening at anytime during random unrooted old block replay https://github.com/anza-xyz/alpenglow/issues/135
2) wait_for_vote_to_start_leader communication and set identity #121
2) (dump &) repair on notarization certificate w/o replay finished / dead slot #30 
3) Produce leader blocks with no waiting (currently we wait for cert on each block) #120
4) Remove `ReplayCertificateTracker` and make sure it works with #123 instead
5) Share ProgressMap and cleanup the remaining consensus structs, once migration is complete don't need to track #65 
6) Add metrics #122
7) Test with partition tests #119 

Fixes #28 
Fixes #23 
Fixes #91 
Fixed #42

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -1264,7 +1264,6 @@ impl ReplayStage {
                             transaction_status_sender.is_some(),
                             &first_alpenglow_slot,
                             &mut is_alpenglow_migration_complete,
-                            &blockstore,
                         );
 
                         let poh_bank = poh_recorder.read().unwrap().bank();
@@ -2326,7 +2325,6 @@ impl ReplayStage {
         track_transaction_indexes: bool,
         first_alpenglow_slot: &Option<Slot>,
         is_alpenglow_migration_complete: &mut bool,
-        _blockstore: &Blockstore,
     ) -> bool {
         // all the individual calls to poh_recorder.read() are designed to
         // increase granularity, decrease contention
@@ -9167,7 +9165,6 @@ pub(crate) mod tests {
             track_transaction_indexes,
             &None,
             &mut false,
-            blockstore,
         ));
     }
 
@@ -9844,7 +9841,6 @@ pub(crate) mod tests {
             track_transaction_indexes,
             &None,
             &mut false,
-            &blockstore,
         ));
 
         // Register another slots worth of ticks  with PoH recorder
@@ -9874,7 +9870,6 @@ pub(crate) mod tests {
             track_transaction_indexes,
             &None,
             &mut false,
-            &blockstore,
         ));
         // Get the new working bank, which is also the new leader bank/slot
         let working_bank = bank_forks.read().unwrap().working_bank();


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -249,6 +251,7 @@ impl CertificatePool {
                 {
                     finalization_certificate
                 } else {
+                    error!("Missing notarization certificate {parent_slot}");
```
wen-coding, line 254: Do we need to keep this log?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -277,6 +280,7 @@ impl CertificatePool {
                         .1
                         .clone()
                 } else {
+                    error!("Missing skip certificate certificate for {begin_skip_slot} to {end_skip_slot} have {max_skip_range:?}");
```
wen-coding, line 283: same here, would be spammy.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
+//! The Alpenglow voting loop, handles all three types of votes as well as
+//! rooting, leader logic, and dumping and repairing the notarized versions.
```
wen-coding, line 2: nit: maybe also say what's not handled here  (replay?)

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+            atomic::{AtomicBool, Ordering},
+            Arc, RwLock,
+        },
+        thread::{self, Builder, JoinHandle},
+        time::Instant,
+    },
+};
+
+// Skip time for each block
+const BLOCKTIME: u128 = 800;
```
wen-coding: nit: I think this constant should go into alpenglow_consensus. How do we define the block time now in PoH?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+
+/// Inputs to the voting loop
+pub struct VotingLoopConfig {
+    pub exit: Arc<AtomicBool>,
```
wen-coding, line 69: nit: one blank line before each group to make it more readable? (e.g. between line 71 and 72, line 76 and 77)

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+    pub authorized_voter_keypairs: Arc<RwLock<Vec<Arc<Keypair>>>>,
+    pub blockstore: Arc<Blockstore>,
+    pub bank_forks: Arc<RwLock<BankForks>>,
+    pub cert_tracker: Arc<RwLock<ReplayCertificateTracker>>,
+    pub cluster_info: Arc<ClusterInfo>,
+    pub poh_recorder: Arc<RwLock<PohRecorder>>,
+    pub leader_schedule_cache: Arc<LeaderScheduleCache>,
+    pub rpc_subscriptions: Arc<RpcSubscriptions>,
+    pub banking_tracer: Arc<BankingTracer>,
+    // Senders
```
wen-coding, line 87: nit: is it possible to say in one word what the receiver is?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+    pub accounts_background_request_sender: AbsRequestSender,
+    pub voting_sender: Sender<VoteOp>,
+    pub lockouts_sender: Sender<CommitmentAggregationData>,
+    pub drop_bank_sender: Sender<Vec<BankWithScheduler>>,
+    pub bank_notification_sender: Option<BankNotificationSenderConfig>,
+    pub slot_status_notifier: Option<SlotStatusNotifier>,
+    // Receivers
+    pub vote_receiver: VoteReceiver,
```
wen-coding, line 96: nit: hard to tell what the difference is between voting_sender and vote_receiver

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+    pub poh_recorder: Arc<RwLock<PohRecorder>>,
+    pub leader_schedule_cache: Arc<LeaderScheduleCache>,
+    pub rpc_subscriptions: Arc<RpcSubscriptions>,
+    pub banking_tracer: Arc<BankingTracer>,
+    // Senders
+    pub accounts_background_request_sender: AbsRequestSender,
+    pub voting_sender: Sender<VoteOp>,
+    pub lockouts_sender: Sender<CommitmentAggregationData>,
+    pub drop_bank_sender: Sender<Vec<BankWithScheduler>>,
+    pub bank_notification_sender: Option<BankNotificationSenderConfig>,
```
wen-coding, line 92: Hmm, this one is not a sender, but a sender config?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+    identity_keypair: Arc<Keypair>,
+    authorized_voter_keypairs: Arc<RwLock<Vec<Arc<Keypair>>>>,
+    has_new_vote_been_rooted: bool,
+    voting_sender: Sender<VoteOp>,
+    wait_to_vote_slot: Option<Slot>,
+    voted_signatures: Vec<Signature>,
+}
+
+/// Context shared with replay, gossip, banking stage etc
+struct SharedContext {
```
wen-coding, line 112: Should we just have SharedContext in VoteLoopConfig?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+    leader_schedule_cache: Arc<LeaderScheduleCache>,
+    bank_forks: Arc<RwLock<BankForks>>,
+    poh_recorder: Arc<RwLock<PohRecorder>>,
+    cert_tracker: Arc<RwLock<ReplayCertificateTracker>>,
+    rpc_subscriptions: Arc<RpcSubscriptions>,
+    banking_tracer: Arc<BankingTracer>,
+    // TODO(ashwin): share this with replay (currently empty)
+    progress: ProgressMap,
+    // TODO(ashwin): integrate with gossip set-identity
+    my_pubkey: Pubkey,
```
wen-coding, line 123: nit: is this node_pubkey? We probably should be clear about this.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+}
+
+#[derive(Debug, Clone, Default)]
+// TODO(ashwin): Evaluate alternatives to this
```
wen-coding, line 127: Do you still need this if cert_pool supports multiple skip votes per pubkey?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1125 @@
...
+        self.skip_certificates.contains(&slot)
+    }
+
+    fn set_root(&mut self, root: Slot) {
+        self.notarization_certificates = self.notarization_certificates.split_off(&root);
+        self.skip_certificates = self.skip_certificates.split_off(&root);
+    }
+}
+
+pub(crate) enum GenerateVoteTxResult {
```
wen-coding, line 166: nit: can we use Result<...> here?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            vote_account_pubkey: vote_account,
+            identity_keypair,
+            authorized_voter_keypairs,
+            has_new_vote_been_rooted,
+            voting_sender,
+            wait_to_vote_slot,
+            voted_signatures: vec![],
+        };
+        let mut shared_context = SharedContext {
+            blockstore: blockstore.clone(),
```
wen-coding, line 276: I think all these were cloned before being put in config?
So from now on can you just access blockstore etc in shared_context so we don't have to clone again?
And it may make sense to just have SharedContext in the config?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            vote_history,
+            vote_account_pubkey: vote_account,
+            identity_keypair,
+            authorized_voter_keypairs,
+            has_new_vote_been_rooted,
+            voting_sender,
+            wait_to_vote_slot,
+            voted_signatures: vec![],
+        };
+        let mut shared_context = SharedContext {
```
wen-coding, line 275: nit: Can we split what needs to be changed and what doesn't change into different data structures so it's easy to know what will never change from now on? (e.g. you probably won't change blockstore)

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            bank_forks: bank_forks.clone(),
+            poh_recorder: poh_recorder.clone(),
+            cert_tracker: cert_tracker.clone(),
+            rpc_subscriptions: rpc_subscriptions.clone(),
+            banking_tracer,
+            progress: ProgressMap::default(),
+            my_pubkey,
+        };
+
+        ReplayStage::reset_poh_recorder(
```
wen-coding, line 288: nit: add comments what this is doing?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            &my_pubkey,
+            blockstore.as_ref(),
+            root_bank_cache.root_bank(),
+            poh_recorder.as_ref(),
+            &leader_schedule_cache,
+        );
+
+        // TODO(ashwin): Start loop once migration is complete current_slot from vote history
```
wen-coding, line 296: Not sure what this sentence means, do you mean this loop should recover current_slot from vote history?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        loop {
+            let leader_start_slot = first_of_consecutive_leader_slots(current_slot);
+            let leader_end_slot = leader_start_slot + NUM_CONSECUTIVE_LEADER_SLOTS - 1;
```
wen-coding, line 299: nit: does it make sense to make these two a utility function in leader_schedule_utils so we can add tests there?

Also these two don't change if current_slot doesn't change right?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            let mut skipped = false;
```
wen-coding, line 300: nit: I know the paper used skipped. Can we add comments here saying this means the whole leader window is skipped?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+
+            // TODO(ashwin): Do maybe_leader for all 4 blocks here, right now
+            // instead we build each block synchronously below
+            // if leader_schedule_cache.slot_leader_at(current_slot) {
+
+            // }
+
+            // Create timer for each slot in the leader window
+            let timers = [Instant::now(); NUM_CONSECUTIVE_LEADER_SLOTS as usize];
```
wen-coding: What if we exited in the middle of a block? What if we exited in the middle of a leader block?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let current_timer = timers[leader_slot_index];
+                let timeout = BLOCKTIME * ((leader_slot_index + 1) as u128);
+                let cert_timer = Instant::now();
+                let mut skip_refresh_timer = Instant::now();
+
+                let leader_pubkey = leader_schedule_cache.slot_leader_at(
+                    current_slot,
+                    Some(bank_forks.read().unwrap().root_bank().as_ref()),
+                );
```
wen-coding: Does leader_pubkey change for slots in [leader_start_slot, leader_end_slot]?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            while current_slot <= leader_end_slot {
+                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let current_timer = timers[leader_slot_index];
+                let timeout = BLOCKTIME * ((leader_slot_index + 1) as u128);
+                let cert_timer = Instant::now();
+                let mut skip_refresh_timer = Instant::now();
+
+                let leader_pubkey = leader_schedule_cache.slot_leader_at(
+                    current_slot,
+                    Some(bank_forks.read().unwrap().root_bank().as_ref()),
```
wen-coding: Can you use the cached root bank here?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+
+            // }
+
+            // Create timer for each slot in the leader window
+            let timers = [Instant::now(); NUM_CONSECUTIVE_LEADER_SLOTS as usize];
+
+            while current_slot <= leader_end_slot {
+                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let current_timer = timers[leader_slot_index];
+                let timeout = BLOCKTIME * ((leader_slot_index + 1) as u128);
```
wen-coding: nit: I think the paper calculates all the timeouts for all blocks in this window at the beginning of the loop (line 3 to 4).

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                let cert_timer = Instant::now();
```
wen-coding: nit: we have many certs, so maybe specify what this controls.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                let mut skip_refresh_timer = Instant::now();
```
wen-coding, line 318: Is this a bit different from the paper? Let's say my window is [4, 7], The paper says all timeout is calculated when you start 4. But let's say somehow I started block 4 at time 1000, I spent 700ms creating block 4, would my timer for 5 start from 1400 or 1700 now?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+
+                let leader_pubkey = leader_schedule_cache.slot_leader_at(
+                    current_slot,
+                    Some(bank_forks.read().unwrap().root_bank().as_ref()),
+                );
+                let is_leader = leader_pubkey.map(|pk| pk == my_pubkey).unwrap_or(false);
+                let mut notarized = false;
+
+                info!(
+                    "{my_pubkey}: Starting timer for slot: {current_slot}, is_leader: {is_leader}"
```
wen-coding: Should we also print the actual timeout we are using here for debugging? (cert_timer and skip_refresh_timer)

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                );
+
+                while !Self::branch_notarized(
+                    &my_pubkey,
+                    current_slot,
+                    &cert_pool,
+                    cert_tracker.as_ref(),
```
wen-coding, line 334: Does cert_tracker appear with cert_pool every time? Then does it make sense to offload this functionality into cert_pool maybe?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            // if leader_schedule_cache.slot_leader_at(current_slot) {
+
+            // }
+
+            // Create timer for each slot in the leader window
+            let timers = [Instant::now(); NUM_CONSECUTIVE_LEADER_SLOTS as usize];
+
+            while current_slot <= leader_end_slot {
+                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let current_timer = timers[leader_slot_index];
```
wen-coding: If values in timers never change, do we need a vector at all?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                            &mut root_bank_cache,
+                            &mut cert_pool,
+                            &mut voting_context,
+                        );
+                        skip_refresh_timer = Instant::now();
+                    }
+
+                    // TODO(ashwin): we could block here & use return value if we've skipped and are
+                    // waiting on a cert to avoid tight spin
+                    Self::ingest_votes_into_certificate_pool(
```
wen-coding, line 393: I'm a little bit worried that any of the steps in the current loop being slow can delay vote ingestion. How hard is it to move it into another thread? Just that we need to make cert_pool thread safe?

Also, should we do this once before the while loop to make sure we ingested all pending votes? Probably doesn't make too much a difference though.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                        &my_pubkey,
+                        &vote_receiver,
+                        &mut cert_pool,
+                        bank_forks.as_ref(),
+                    );
+
+                    if skipped || notarized {
```
wen-coding, line 400: Hmm this logic is a bit different from the paper. The paper says always sends NotarVote if skipped is not true entering the loop. Here I think you check the timeout first. Probably doesn't make much difference though.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                        Self::alpenglow_update_commitment_cache(
+                            AlpenglowCommitmentType::Notarize,
+                            current_slot,
+                            &lockouts_sender,
+                        );
+                    }
+                }
+
+                info!(
+                    "{my_pubkey}: Slot {current_slot} certificate observed in {} ms. Skip timer {} vs timeout {}",
```
wen-coding, line 432: nit: should we also say whether notar or skip cert was observed?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                    timeout,
+                );
+
+                if !skipped
+                    && Self::branch_notarized(
+                        &my_pubkey,
+                        current_slot,
+                        &cert_pool,
+                        cert_tracker.as_ref(),
+                    )
```
wen-coding, line 444: In the while loop we checked branch_notarized right? Can we reuse the result of that one?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                        is_leader,
+                        &bank_forks,
+                        &blockstore,
+                        &mut cert_pool,
+                        &mut voting_context,
+                    );
+                }
+
+                current_slot += 1;
+            }
```
wen-coding, line 467: nit: This loop is a bit long so it's hard to figure out where the start of loop is, can we add more functions to make the whole loop body less than 50 lines maybe?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+
+            // Set new root
+            if let Some(new_root) = Self::maybe_set_root(
```
wen-coding, line 470: Theoretically this can also happen in the middle of the window right? As soon as we get a finalization cert we can set root.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+    fn branch_notarized(
+        my_pubkey: &Pubkey,
+        slot: Slot,
+        cert_pool: &CertificatePool,
+        cert_tracker: &RwLock<ReplayCertificateTracker>,
+    ) -> bool {
+        if let Some(size) = cert_pool.is_notarization_certificate_complete(slot) {
+            info!(
+                "{my_pubkey}: Branch Notarized: Slot {} from {} validators",
+                slot, size
```
wen-coding, line 501: Why do we need to log size?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        my_pubkey: &Pubkey,
+        slot: Slot,
+        root_bank_cache: &mut RootBankCache,
+        cert_pool: &mut CertificatePool,
+        cert_tracker: &RwLock<ReplayCertificateTracker>,
+    ) -> bool {
+        let root_bank = root_bank_cache.root_bank();
+        let epoch = root_bank.epoch_schedule().get_epoch(slot);
+        // TODO(ashwin): once we figure out how to handle skip ranges across
+        // epoch boundaries, cache this accordingly
```
wen-coding, line 526: Nice, guess this can be simplified after we fix cert_pool.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            let timers = [Instant::now(); NUM_CONSECUTIVE_LEADER_SLOTS as usize];
+
+            while current_slot <= leader_end_slot {
+                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let current_timer = timers[leader_slot_index];
+                let timeout = BLOCKTIME * ((leader_slot_index + 1) as u128);
+                let cert_timer = Instant::now();
+                let mut skip_refresh_timer = Instant::now();
+
+                let leader_pubkey = leader_schedule_cache.slot_leader_at(
```
wen-coding: Is it worth a panic here if leader_pubkey is None. Under what case would it be a valid case that you don't know the leader schedule for the current epoch?

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: no, can remove this once your PR lands

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yes

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: not sure what you're asking, could you elaborate?

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: we plan to remove it

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah no need to though, set root expensive (write lock) so can evaluate once every leader window

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: cause I want to, helpful for debugging

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: Yeah this means we can't produce our leader block so something is wrong.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah good point I can just use one timer for the whole window

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: see line 307 all timers are started once at the first leader slot

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: it doesn't matter, the only times cert_pool is used is for branch notarized / skip certified which can only run after this function. 
Even if we put it in another thread we'd still have to wait for this loop to reach the end.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah it doesn't matter, just reduces indentation to do it this way

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        if *leader != ctx.my_pubkey {
+            // Not our leader slot
+            return false;
+        }
+
+        // TODO(ashwin): We max with root here, as the snapshot slot might not have a certificate.
+        // Think about this more and fix if necessary.
+        let parent_slot = cert_pool
+            .highest_not_skip_certificate_slot()
+            .max(ctx.bank_forks.read().unwrap().root());
```
wen-coding, line 581: I feel we need to have a function at the beginning of the loop to correctly set up the state, set up the correct current_slot etc from snapshots and certs.

Once we recovered what we could before the loop, we don't have to check whether parent_slot is higher than my intended slot every time in maybe_start_leader.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            .cert_tracker
+            .read()
+            .unwrap()
+            .highest_replay_notarized()
+            .unwrap_or(0)
+            > parent_slot
+        {
+            // A block has been frozen that has a notarization certificate for a slot
+            // higher than `parent_slot`. Certificate pool ingestion is lagging behind,
+            // do not attempt to build on `parent_slot` as we want to build off of the
```
wen-coding, line 592: Hmm, if we ingest votes from the replayed block, when would this happen? For a block to be valid and frozen, it should contain all votes which support it right? Then by the time it's frozen, we should have ingested all votes notarizing its parent.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            info!(
+                "{}: We want to produce {slot} off of notarized slot {parent_slot},
+                however {parent_slot} is not yet in bank forks,
+                we are running behind so delaying our leader block",
+                ctx.my_pubkey
+            );
+            return false;
+        };
+
+        if !parent_bank.is_frozen() {
```
wen-coding, line 622: We probably should compare parent_bank.bankhash with our cert on hand somewhere....

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            .epoch_total_stake(parent_bank.epoch_schedule().get_epoch(slot))
+            .expect("my_leader_slot - 1 got a certificate, so we must be in a known epoch range");
+        info!(
+            "{}: Checking leader decision slot {slot} parent {parent_slot}",
+            ctx.my_pubkey
+        );
+        let Some(StartLeaderCertificates {
+            notarization_certificate,
+            skip_certificate,
+        }) = cert_pool.make_start_leader_decision(
```
wen-coding, line 656: I should return some error to you so you know why this failed... how do you like to change the API?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                let r_cert_tracker = ctx.cert_tracker.read().unwrap();
+                if ((parent_slot + 1)..slot).all(|s| {
+                    r_cert_tracker.replay_skip_certified(s)
+                        || cert_pool.skip_certified(s, total_stake)
+                }) {
+                    // A block has been frozen that has the skip certificate for the required range
+                    // Certificate pool ingestion is lagging behind, we cannot build our block now
+                    // but reevaluate later.
+                    // TODO(ashwin): We can remove this once we solve the catchup problem w/o cert tracker
+                    warn!(
```
wen-coding, line 676: Let's change this to ERROR message returned from make_start_leader_decision()

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        }
+
+        let tpu_bank = ReplayStage::new_bank_from_parent_with_notify(
+            parent_bank.clone(),
+            slot,
+            root_slot,
+            &ctx.my_pubkey,
+            &ctx.rpc_subscriptions,
+            slot_status_notifier,
+            NewBankOptions::default(),
```
wen-coding, line 714: Would we need to pass vote_only_bank from caller somewhere someday?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        );
+        // make sure parent is frozen for finalized hashes via the above
+        // new()-ing of its child bank
```
wen-coding, line 717: Hmm, I think we checked parent is frozen, would something un-freeze it in between?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        );
+        if !alpenglow_update_bank_forks_and_poh_recorder_for_new_tpu_bank(
+            ctx.bank_forks.as_ref(),
+            ctx.poh_recorder.as_ref(),
+            tpu_bank,
+            track_transaction_indexes,
+            notarization_certificate,
+            skip_certificate,
+        ) {
+            error!(
```
wen-coding, line 731: Hmm I think we already log error inside that function, maybe we should just pass the error out?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                "{}: Unable to create bank and commit certificates for slot:{slot} parent:{parent_slot}.
+                This is likely because your block has already been skipped. Abandoning this block",
+                ctx.my_pubkey
+            );
+            // We return true here to indicate that we should not attempt to rebuild this block.
+            return true;
+        }
+
+        info!(
+            "{}: new fork:{} parent:{} (leader) root:{}",
```
wen-coding, line 741: Should we proudly put alpenglow somewhere in the log line so we know whether this block is generated by TowerBFT or Alpenglow?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+    fn maybe_set_root(
+        slot: Slot,
+        cert_pool: &mut CertificatePool,
+        accounts_background_request_sender: &AbsRequestSender,
+        bank_notification_sender: &Option<BankNotificationSenderConfig>,
+        drop_bank_sender: &Sender<Vec<BankWithScheduler>>,
+        ctx: &mut SharedContext,
+        vctx: &mut VotingContext,
+    ) -> Option<Slot> {
+        let old_root = vctx.vote_history.root;
```
wen-coding, line 761: If someone is stupid enough to wipe ledger, could the bank in old_root be not frozen at this point?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        info!(
+            "{}: Checking for finalization certificates between {old_root} and {slot}",
+            ctx.my_pubkey
+        );
+        let new_root = (old_root + 1..=slot).rev().find(|slot| {
+            cert_pool.is_finalized_slot(*slot) && ctx.bank_forks.read().unwrap().is_frozen(*slot)
+        })?;
+        info!("{}: Attempting to set new root {new_root}", ctx.my_pubkey);
```
wen-coding, line 769: I feel we don't need this many info logs maybe?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        let bank = root_bank_cache.root_bank();
+        info!(
+            "{my_pubkey}: Voting skip for slot range [{start},{end}] with blockhash from {}",
+            bank.slot()
+        );
+        Self::send_vote(vote, false, bank.as_ref(), cert_pool, voting_context);
+    }
+
+    /// Refresh the last skip vote
+    fn refresh_skip(
```
wen-coding, line 833: Do we only need to refresh skip but not notarize or finalize?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        start: Slot,
+        end: Slot,
+        root_bank_cache: &mut RootBankCache,
+        cert_pool: &mut CertificatePool,
+        voting_context: &mut VotingContext,
+    ) {
+        // Extend the previous range if possible
+        let start = voting_context
+            .vote_history
+            .prev_skip_start(start - 1)
```
wen-coding, line 820: Could this prev_skip_start be smaller than root? Do we care?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                    if !skipped && current_timer.elapsed().as_millis() > timeout {
+                        Self::vote_skip(
+                            &my_pubkey,
+                            current_slot,
+                            leader_end_slot,
+                            &mut root_bank_cache,
+                            &mut cert_pool,
+                            &mut voting_context,
+                        );
+                        skipped = true;
```
wen-coding: I'm paranoid and fear that in the future we return an error from vote_skip and skip this line and the following line. Can we set `skipped` to true before going into `Self::vote_skip`?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        slot: Slot,
+        is_leader: bool,
+        bank_forks: &RwLock<BankForks>,
+        blockstore: &Blockstore,
+        cert_pool: &mut CertificatePool,
+        voting_context: &mut VotingContext,
+    ) -> bool {
+        let Some(bank) = bank_forks.read().unwrap().get(slot) else {
+            if matches!(vote_type, BlockIdVote::Finalize) {
+                warn!("Replay is catching up skipping finalize vote on {slot}");
```
wen-coding: Hmm, I think we did check the bank we are finalizing is frozen before entering here? Maybe we should just pass in a bank and assert the bank is frozen in here?

Can a bank be frozen when block-id is missing?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                is_leader,
+                // TODO:(ashwin) ignore duplicate block checks (?)
+                &FeatureSet::all_enabled(),
+            ).unwrap_or_else(|e| {if !is_leader {
+                warn!("Unable to retrieve block_id, failed last fec set checks for slot {slot} hash {hash}: {e:?}")
+            }
+            None}
+            );
+
+        let Some(block_id) = block_id else {
```
wen-coding: is it possible to merge the logic here with line 869~880?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            if is_leader {
+                // For leader slots, shredding is asynchronous so block_id might not yet
+                // be available. In this case we want to retry our vote later
+                return false;
+            }
+            // At this point we could mark the bank as dead similar to TowerBFT, however
+            // for alpenglow this is not necessary
```
wen-coding, line 891: Would we break anything if we don't mark the bank dead here? Could we get stuck trying repeatedly on a duplicate block?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            warn!(
+                "Unable to retrieve block id or duplicate block checks have failed
+                for non leader slot {slot} {hash}, not voting {vote_type}"
+            );
+            return true;
```
wen-coding, line 896: Does this mean we lost hope in this bank so we don't want to retry any more? Should we perform some action on it?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            info!(
+                "Vote account node_pubkey mismatch: {} (expected: {}).  Unable to vote",
+                vote_state.node_pubkey(),
+                context.identity_keypair.pubkey()
+            );
+            return GenerateVoteTxResult::HotSpare;
+        }
+
+        let Some(authorized_voter_pubkey) = vote_state.get_authorized_voter(bank.epoch()) else {
+            warn!(
```
wen-coding, line 1024: I think no warning here, let's print out error in the caller?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        let vote_account = match bank.get_vote_account(&context.vote_account_pubkey) {
+            None => {
+                warn!("Vote account {vote_account_pubkey} does not exist.  Unable to vote");
+                return GenerateVoteTxResult::Failed;
+            }
+            Some(vote_account) => vote_account,
+        };
+        let vote_state = match vote_account.alpenglow_vote_state() {
+            None => {
+                warn!(
```
wen-coding, line 1007: print error in the caller instead?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        bank: &Bank,
+        cert_pool: &mut CertificatePool,
+        context: &mut VotingContext,
+    ) {
+        let mut generate_time = Measure::start("generate_alpenglow_vote");
+        let vote_tx_result = Self::generate_vote_tx(&vote, bank, context);
+        generate_time.stop();
+        // TODO(ashwin): add metrics struct here and throughout the whole file
+        // replay_timing.generate_vote_us += generate_time.as_us();
+        let GenerateVoteTxResult::Tx(vote_tx) = vote_tx_result else {
```
wen-coding, line 937: nit: maybe change to Result and log the error here?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            return;
+        };
+        if let Err(e) = cert_pool.add_vote(
```
wen-coding, line 940: if it's refresh, do we need to do add_vote at all?

Err, I remembered, we don't have cert_pool saved currently, is that why you are doing this?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+        let mut cached_root_bank = None;
+
+        vote_receiver
+            .try_iter()
+            .filter_map(|(vote, vote_account_pubkey, tx)| {
+                let root_bank =
+                    cached_root_bank.get_or_insert_with(|| bank_forks.read().unwrap().root_bank());
+                let epoch = root_bank.epoch_schedule().get_epoch(vote.slot());
+                let epoch_stakes = root_bank.epoch_stakes(epoch)?;
+                let validator_stake = epoch_stakes.vote_account_stake(&vote_account_pubkey);
```
wen-coding, line 1084: Luckily you don't need this very soon....

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                let total_stake = root_bank.epoch_total_stake(epoch)?;
+                match cert_pool.add_vote(
+                    &vote,
+                    tx,
+                    &vote_account_pubkey,
+                    validator_stake,
+                    total_stake,
+                ) {
+                    Ok(Some(cert)) if cert.is_notarization_or_skip() => {
+                        info!("{my_pubkey}: Ingested new certificate {cert:?}");
```
wen-coding, line 1098: Why don't we want to mention new finalization cert here?

ledger/src/blockstore.rs
```diff
@@ -3828,14 +3828,17 @@ impl Blockstore {
...
         slot: Slot,
         bank_hash: Hash,
+        is_leader: bool,
         feature_set: &FeatureSet,
     ) -> std::result::Result<Option<Hash>, BlockstoreProcessorError> {
         let results = self.check_last_fec_set(slot);
         let Ok(results) = results else {
-            warn!(
-                "Unable to check the last fec set for slot {slot} {bank_hash}, \
+            if !is_leader {
```
wen-coding, line 3836: Would we have no block_id for my own block forever?

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: it's not a blocking send, crossbeam can do whatever so we could finish replaying the block before this thread sees the votes.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: this will be removed with #123

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: it's already in the log line since it's in the name of the file

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yeah i will fix later there is no snapshot logic here yet, hence the "TODO"

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Certificate aggregation is not yet implemented for the certificate pool.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Doesn't matter, we panic on all errors once #123 is integrated + there is logging within this function. Feel free to return however you like.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: once we decide the conditions for vote only in aspenglow sure - not relevant to this PR.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: go for it.
that function does multiple things so nice to know which part fails

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: no

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Feel free to do log cleanup as you see fit once we are confident there are no bugs.
If I added a log it's because I found it useful when debugging the prototype

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: no, and we don't care even if it was

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: feel free to add other refreshes. Refreshing skip is the only one that matters for liveness (trying to keep this PR as small as possible).

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yes good point, can pass in the bank instead
> Can a bank be frozen when block-id is missing?

yes

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yeah good point, i'll merge it into one `let .. else {`

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: No, we return true below. If by some miracle the block gets notarized anyway, we can continue replay of the fork since it is not dead.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yes we don't vote on it. No we should not do anything with the block, it will eventually get cleaned up when setting a root in the future.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: If we're refreshing something went wrong with the initial transmission of the vote, so I want to do everything over again.

And yes if i don't know if we rebuild cert_pool currently so we should attempt to push it anyway

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: feel free to do it in a separate PR, this is a shared function so didn't want to add more to this PR refactoring replay stage

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: it'll be done separately as part of the TODO on line 467.
Wasn't necessary at this moment so i didn't add the fn to handle new finalize certificates

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: idk, looks like Lijun wrote it might want to ask him

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: nesting makes it hard to follow, prefer it to be separate

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yes, will refactor when I do set-identity

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: sure can refactor in a future pr when we add tests

AshwinSekar: and no, only changes when current_slot changes (every iteration of this loop)

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: no it's harder to follow with multiple levels of function nesting. Right now it's clear because each function only does 1 thing indicated by its name.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: not sure, looks like Ryo wrote this code I would ask him

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            .voting_sender
+            .send(VoteOp::PushAlpenglowVote {
+                tx: vote_tx,
+                slot: bank.slot(),
+                saved_vote_history: SavedVoteHistoryVersions::from(saved_vote_history),
+            })
+            .unwrap_or_else(|err| warn!("Error: {:?}", err));
+    }
+
+    fn generate_vote_tx(
```
carllin, line 978: Should probably check the vote history to make sure we can vote before the vote is sent, made an issue: https://github.com/anza-xyz/alpenglow/issues/130

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+            CommitmentAggregationData::AlpenglowCommitmentAggregationData(
+                AlpenglowCommitmentAggregationData {
+                    commitment_type,
+                    slot,
+                },
+            ),
+        ) {
+            trace!("lockouts_sender failed: {:?}", e);
+        }
+    }
```
carllin, line 1126: I know this is copied from master/replay, but might be good to fix this at some point, while we're here

all these send errors should cause the loop to exit and us setting the exit signal, can probably copy this https://github.com/anza-xyz/alpenglow/blob/6d04641dc345bd4bb68e927b65546b5cc4c1d0a0/core/src/replay_stage.rs#L172-L187

carllin: Some other places where this pops up:

Vote sender: 
https://github.com/anza-xyz/alpenglow/blob/0fee14d20e2a6dc1b4bda6d0625a316859abfb89/core/src/alpenglow_consensus/voting_loop.rs#L966-L973

Set root:
https://github.com/anza-xyz/alpenglow/blob/0fee14d20e2a6dc1b4bda6d0625a316859abfb89/core/src/alpenglow_consensus/voting_loop.rs#L787-L790

Created a separate issue: https://github.com/anza-xyz/alpenglow/issues/131

core/src/replay_stage.rs
```diff
@@ -1247,84 +1246,10 @@ impl ReplayStage {
...
-                                new_root_bank.slot()
-                            );
-                            return;
-                        }
-                    }
                 }
 
                 let mut start_leader_time = Measure::start("start_leader_time");
-                if !tpu_has_bank {
+                if !is_alpenglow_migration_complete && !tpu_has_bank {
```
carllin: I think we can just extend this check: https://github.com/anza-xyz/alpenglow/blob/6d04641dc345bd4bb68e927b65546b5cc4c1d0a0/core/src/replay_stage.rs#L832 since it ends at the line above anyways

core/src/replay_stage.rs
```diff
@@ -2540,10 +2319,9 @@ impl ReplayStage {
         banking_tracer: &Arc<BankingTracer>,
         has_new_vote_been_rooted: bool,
         track_transaction_indexes: bool,
-        cert_pool: &mut CertificatePool,
         first_alpenglow_slot: &Option<Slot>,
         is_alpenglow_migration_complete: &mut bool,
-        blockstore: &Blockstore,
+        _blockstore: &Blockstore,
```
carllin: think we can just delete this now

core/src/replay_stage.rs
```diff
@@ -3638,7 +3068,7 @@ impl ReplayStage {
         }
     }
 
-    fn reset_poh_recorder(
+    pub(crate) fn reset_poh_recorder(
```
carllin, line 3074: These pub(crate) functions might be good to start moving into a separate utility module

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -0,0 +1,1129 @@
...
+                        return;
+                    }
+                    // If we're the leader try to start building a block
+                    if last_leader_block < current_slot
+                        && !poh_recorder.read().unwrap().has_bank()
+                        && Self::maybe_start_leader(
+                            &leader_pubkey,
+                            current_slot,
+                            &mut cert_pool,
+                            &slot_status_notifier,
```
carllin, line 352: not sure if we want to take this in a tight loop since it hits locks like `bank_forks` quite often. Might be better to have a dirty flag in `cert_tracker` set by replay to signal we should check

core/src/replay_stage.rs
AshwinSekar: the metric is in the way haha, but ya can refactor it

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: It's logged in branch notarized and skip certified respectively

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Just following the paper, feel free to optimize later

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Let's say we were happily making blocks when we are the leader for [5, 8], we got to block 6, now the validator crashed here or operator restarted it. I think when the validator restarts you should come back and keep processing block 6. I think you might be setting the `current_slot` correctly, but should the timers restart from `Instant::now()`?

core/src/replay_stage.rs
AshwinSekar: 6aeddac349b77e770605a72ed9460f66919d8910

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 72e8dcbbb7bc4d6aaa126d5e4204c104d117f42e

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: eff0c95f2812dd23607d8484be142b5832a4a669

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 8891b83a2359fb4809205ae08be7f558b6eae6cc

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: a9fe2ecc38f0d1c883f23eab0f021387104cc94e

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 9a8471cd99e43fce5dab3939cd8974e0b29104b2

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 88b1747050782301141dbc209499be9a79bf1dde

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 0a32e29beb071fdf90449fa29541e3d4adf0a062

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: moved and split it into the components a9fe2ecc38f0d1c883f23eab0f021387104cc94e

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yeah good point, I plan to refactor `cert_tracker` after #123 lands, you mind if I tackle this in #133 when I refactor the leader portion of this loop?

core/src/replay_stage.rs
AshwinSekar: done 66d43d4d809c6afca5913d5aade443d75b2f7eb0

core/src/replay_stage.rs
AshwinSekar: Filed an issue https://github.com/anza-xyz/alpenglow/issues/134#issue-2970394062

PR #123, opened 2025-03-31
Move skip cert to per-slot pool and fix epoch stakes in cert pool

wen-coding:
- Move skip cert into per-slot pool, use Arc<VersionedTransaction> so we don't do too many copies
- Fix stake calculation by caching epoch_stakes_map and epoch_schedule
- Moved the original tests in skip_pool to certificate_pool and changed API accordingly
- This should now accept multiple skips per pubkey

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -3671,8 +3671,6 @@ impl ReplayStage {
         let parent_hash = parent_bank.hash();
         let mut notarization_stake = 0;
         let mut notarization_size = 0;
-        let must_skip_start = parent_slot + 1;
-        let must_skip_end = bank.slot() - 1;
         let mut certificate_pool = CertificatePool::new_from_root_bank(root_bank);
 
         let leader_slot_idx = leader_slot_index(bank.slot());
@@ -3767,10 +3765,10 @@ impl ReplayStage {
                     must_skip_start,
                     must_skip_end
                 );
-                return Err(BlockstoreProcessorError::InvalidCert(
+                return Err(BlockstoreProcessorError::InvalidSkipCertificate(
                     bank.slot(),
                     must_skip_start,
-                    "Skip".to_string(),
+                    must_skip_end,
                 ));
             }
         }
@@ -10238,23 +10236,28 @@ pub(crate) mod tests {
         // Test on bank2 should fail because it is not consectuive
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
-                    &bank2,
-                    &bank0,
-                    ReplayCertificateTracker::new_rw_arc().as_ref(),
-                    true,
-                )
-        {
-            assert_eq!(slot, 2);
-            assert_eq!(cert_slot, 1);
-            assert_eq!(cert, "Skip");
-        } else {
-            panic!("Expected InvalidCert error");
-        }
-        let bank3 = Bank::new_from_parent(bank2.clone(), &Pubkey::default(), 3);
+                &bank2,
+                &bank0,
+                ReplayCertificateTracker::new_rw_arc().as_ref(),
+                true,
+            )
+            .unwrap_err(),
+            BlockstoreProcessorError::NonConsecutiveLeaderSlot(2, 0)
+        );
+        let bank3 = Arc::new(Bank::new_from_parent(bank2.clone(), &Pubkey::default(), 3));
         // Test on bank3 should fail because it doesn't contain notarization for bank 2.
-        if let Err(BlockstoreProcessorError::InvalidCert(slot, cert_slot, cert)) =
+        assert!(ReplayStage::alpenglow_check_cert_in_bank(
+            &bank3,
+            &bank0,
+            ReplayCertificateTracker::new_rw_arc().as_ref(),
+            true,
+        )
+        .is_ok());
+        let bank4 = Bank::new_from_parent(bank3.clone(), bank3.collector_id(), 4);
+        // bank4 needs a certificate because it is a new leader window (even though it's consecutive and the same leader)
+        assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
-                &bank3,
+                &bank4,
                 &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true,
@@ -10274,7 +10277,7 @@ pub(crate) mod tests {
         );
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
-                &bank3,
+                &bank4,
                 &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true,
@@ -10296,7 +10299,7 @@ pub(crate) mod tests {
         }
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
-                &bank3,
+                &bank4,
                 &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true,
@@ -10476,6 +10479,7 @@ pub(crate) mod tests {
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
                 &bank4,
+                &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true
             )
@@ -10485,6 +10489,7 @@ pub(crate) mod tests {
         // 5 requires no certificate
         assert!(ReplayStage::alpenglow_check_cert_in_bank(
             &bank5,
+            &bank0,
             ReplayCertificateTracker::new_rw_arc().as_ref(),
             true
         )
@@ -10493,6 +10498,7 @@ pub(crate) mod tests {
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
                 &bad_bank6,
+                &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true
             )
@@ -10501,6 +10507,7 @@ pub(crate) mod tests {
         );
         assert!(ReplayStage::alpenglow_check_cert_in_bank(
             &good_bank6,
+            &bank0,
             ReplayCertificateTracker::new_rw_arc().as_ref(),
             true
         )
@@ -10509,6 +10516,7 @@ pub(crate) mod tests {
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
                 &bad_bank7,
+                &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true
             )
@@ -10517,6 +10525,7 @@ pub(crate) mod tests {
         );
         assert!(ReplayStage::alpenglow_check_cert_in_bank(
             &good_bank7,
+            &bank0,
             ReplayCertificateTracker::new_rw_arc().as_ref(),
             false
         )
@@ -10528,6 +10537,7 @@ pub(crate) mod tests {
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
                 &bank9,
+                &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 true
             )
@@ -10537,6 +10547,7 @@ pub(crate) mod tests {
         assert_matches!(
             ReplayStage::alpenglow_check_cert_in_bank(
                 &bank9,
+                &bank0,
                 ReplayCertificateTracker::new_rw_arc().as_ref(),
                 false
             )


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1,26 +1,37 @@
...
 
 #[derive(Debug, Error, PartialEq)]
 pub enum AddVoteError {
-    #[error("Add vote to skip pool failed: {0}")]
-    AddToSkipPoolFailed(#[from] skip_pool::AddVoteError),
-
     #[error("Add vote to vote certificate failed: {0}")]
     AddToCertificatePool(#[from] vote_certificate::AddVoteError),
+
+    #[error("Add vote to skip pool failed: {0}")]
```
AshwinSekar: update error msg

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -67,98 +78,160 @@ impl CertificateType {
...
-impl CertificatePool {
-    pub fn new() -> Self {
-        Self {
-            certificates: BTreeMap::default(),
-            skip_pool: SkipPool::new(),
-            highest_notarized_slot: 0,
-            highest_finalized_slot: 0,
+    fn update_epoch_stakes_map(&mut self, bank: &Bank) {
+        let epoch = bank.epoch();
+        if self.epoch_stakes_map.is_empty() || epoch != self.epoch_stakes_map_cached_at {
```
AshwinSekar:
```suggestion
        if self.epoch_stakes_map.is_empty() || epoch > self.epoch_stakes_map_cached_at {
```

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -39,19 +56,80 @@ impl VoteCertificate {
...
+                //    we use per-slot cert now, so if someone voted for (2, 5) using transaction t1,
+                //    then its local root becomes 3 and it now sends (4, 7) using transaction t2,
+                //    the slots 4 to 7 will save t2 in skip cert, while 2 to 3 will save t1 in skip
+                //    cert. Therefore, even if we allow new_start to be larger, we don't allow users
+                //    to un-vote.
+                //    However, in the per-slot cert world, we can forbid new_start to be smaller
+                //    than old_start. Someone may have sent (2, 5) and (3, 7), these two may
+                //    arrive out of order. (2, 5) will not be able to land in cert of slot 3 to 5,
+                //    but it can still land in cert of slot 2.
+                new_start >= old_start && new_end > old_end
```
AshwinSekar: We should be able to accept any ordering of votes.
e.g. catching up from snapshot but you receive an all-to-all skip vote from tip

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: We are able to accept votes in any order, but it's good to have definitive order of which one wins if votes can arrive in arbitrary order.
If someone voted (3, 5), then later (4, 7), previously we reject (4, 7), now we say for slot 3 we keep (3, 5) vote in the skip certificate, but for slot 4 and 5 we have two choices:
A. keep (3, 5) which arrives first
B. replace (3, 5) with (4, 7)
The problem of A is that we decide which one to keep by arriving order, so someone could keep sending us (3, 5) and (4, 7) and the transaction in slot 4 and 5 will keep changing. Not end of the world, it's still correct, but we have lots of unnecessary replacements.
So this code uses B, we decide an order, (4, 7) is always better than (3, 5) because 4 is greater than 3 and 7 is greater than 5.
But you are right in that it's a bit arbitrary why (4, 6) is worse than (3, 7). All we need is a deterministic order to reduce needless replacements, the code is still correct in that no one can un-vote skip.
So how about this order:
Because normal validators almost always extend the end, let's compare end first then compare start and leave the one with hopefully wider range so we can hopefully have fewer transactions in the cert:
1. new_end > old_end: (4, 8) > (3, 5), (2, 8) > (3, 5), (3, 8) > (3, 5) 
2. new_end == old_end, check if max(new_start, local_root) < max(old_start, local_root): if local_root is 6, reject (2, 8) when (3, 8) is available. But (7, 9) > (8, 9) when local_root is 6.
3. new_end < old_end: reject

wen-coding: Actually, thought about it more, changed to:
(new_end > old_end) || (new_end == old_end && new_start > old_start)

We can argue both ways and replace it with:
(new_end > old_end) || (new_end == old_end && new_start < old_start)

The algorithm just needs a deterministic order to be efficient.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
+    highest_slot_map: HashMap<CertificateType, Slot>,
+    // Cached epoch_schedule
+    epoch_schedule: EpochSchedule,
+    // Cached epoch_stakes_map
+    epoch_stakes_map: Arc<HashMap<Epoch, EpochStakes>>,
+    // The epoch at which the epoch_stakes_map is cached
+    epoch_stakes_map_cached_at: Epoch,
+    // The latest slot we did purge on, no need to save
+    // anything before this slot.
+    purged_at_slot: Slot,
```
carllin: nit: call this `root` for consistency

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
 }
 
-impl Default for CertificatePool {
-    fn default() -> Self {
-        Self::new()
+impl CertificatePool {
+    pub fn new_from_bank(bank: &Bank) -> Self {
```
carllin: nit: `new_from_root_bank`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
-    // Highest slot with a finalized certificate
-    highest_finalized_slot: Slot,
+    // Highest slot with each certificate
+    highest_slot_map: HashMap<CertificateType, Slot>,
+    // Cached epoch_schedule
+    epoch_schedule: EpochSchedule,
+    // Cached epoch_stakes_map
+    epoch_stakes_map: Arc<HashMap<Epoch, EpochStakes>>,
+    // The epoch at which the epoch_stakes_map is cached
+    epoch_stakes_map_cached_at: Epoch,
```
carllin: nit: `root_epoch`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
+            .highest_slot_map
+            .entry(certificate_type.clone())
+            .or_default();
+        if slot > *current_highest_slot {
+            self.highest_slot_map.insert(certificate_type, slot);
+            true
+        } else {
+            false
+        }
+    }
```
carllin: might be able to simplify this to:

```
fn set_highest_slot(&mut self, certificate_type: CertificateType, slot: Slot) -> bool {
    let current = self.highest_slot_map.entry(certificate_type)
        .and_modify(|s| {
            if slot > *s {
                *s = slot;
            }
        })
        .or_insert(slot);

    *current == slot
}
```

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
+                CertificateType::Skip => Ok(Some(NewHighestCertificate::Skip(highest_slot))),
+                CertificateType::Notarize => {
+                    Ok(Some(NewHighestCertificate::Notarize(highest_slot)))
+                }
+                CertificateType::Finalize => {
+                    Ok(Some(NewHighestCertificate::Finalize(highest_slot)))
                 }
             }
+        } else {
+            Ok(None)
```
carllin: can simplify this as:

```
 Ok(highest.map(|slot| match certificate_type {
        CertificateType::Skip => NewHighestCertificate::Skip(slot),
        CertificateType::Notarize => NewHighestCertificate::Notarize(slot),
        CertificateType::Finalize => NewHighestCertificate::Finalize(slot),
    }))
```

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
+                let validator_stake = epoch_stakes.vote_account_stake(validator_vote_key);
+                let total_stake = epoch_stakes.total_stake();
+                if validator_stake == 0 {
+                    continue;
+                }
+                found_non_zero_stake_slot = true;
+                if slot < self.purged_at_slot {
+                    continue;
+                }
+                if let Some(current_highest_slot) = self.update_certificate_pool(
```
carllin, line 202: nit: this `current_highest_slot` -> `new_highest_slot`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
-                    .entry((vote_slot, CertificateType::get_type(vote)))
-                    .or_insert_with(|| VoteCertificate::new(vote_slot));
-
-                certificate.add_vote(
+        let vote_range = match vote.voted_slots() {
+            Either::Left(vote_slot) => vote_slot..=vote_slot,
+            Either::Right(skip_range) => skip_range,
+        };
+        let certificate_type = CertificateType::get_type(vote);
+        let mut new_highest_slot = None;
```
carllin, line 187: nit: `new_highest_slot` -> `current_highest_slot`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -171,51 +244,73 @@ impl CertificatePool {
...
                     None
                 }
             })
     }
 
-    pub fn get_finalization_certificate(&self, slot: Slot) -> Option<Vec<VersionedTransaction>> {
-        self.certificates
-            .get(&(slot, CertificateType::Finalize))
-            .and_then(|certificate| {
-                if certificate.is_complete() {
```
carllin, line 190: 👍 ah this complete check was redundant, good catch

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -74,7 +150,80 @@ impl VoteCertificate {
         self.certificate.len()
     }
 
-    pub fn get_certificate(&self) -> Vec<VersionedTransaction> {
-        self.certificate.values().cloned().collect()
+    // Return an iterator of CertificateMap
+    pub(crate) fn get_certificate_iter(
```
carllin: nit `iter_certificate_entries()`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -70,98 +81,160 @@ impl CertificateType {
...
+        start_slot: Slot,
+        end_slot: Slot,
+        validator_vote_key: &Pubkey,
+    ) -> Result<Option<NewHighestCertificate>, AddVoteError> {
+        self.add_vote(
+            &Vote::new_skip_vote(start_slot, end_slot),
+            VersionedTransaction::default(),
+            validator_vote_key,
+        )
+    }
```
carllin, line 174: This function doesn't seem necessary, we can just pass a default() VersionedTransaction to `add_vote()`, or we can parameterize the VoteCertificates to take a generic T as certificate entries so we don't have to construct a `VersionedTransaction`, but that can be done later

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -654,15 +642,13 @@ impl VotingLoop {
...
         )
         else {
             if slot > parent_slot + 1 {
                 let r_cert_tracker = ctx.cert_tracker.read().unwrap();
-                if ((parent_slot + 1)..slot).all(|s| {
-                    r_cert_tracker.replay_skip_certified(s)
-                        || cert_pool.skip_certified(s, total_stake)
-                }) {
+                if ((parent_slot + 1)..slot)
+                    .all(|s| r_cert_tracker.replay_skip_certified(s) || cert_pool.skip_certified(s))
```
carllin, line 650: Do we a separate replay skip certification `r_cert_tracker.replay_skip_certified(s)`? Thought the cert_pool would be sufficient because it ingests replay votes as well

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -74,7 +150,80 @@ impl VoteCertificate {
...
+        &self,
+    ) -> std::collections::hash_map::Iter<'_, Pubkey, VoteCertificateEntry> {
+        self.certificate.iter()
+    }
+}
```
carllin, line 163: All calls to this function always check `is_complete()` beforehand, so we can just make this return an Option<> and make that completeness check in here

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -225,8 +320,33 @@ impl CertificatePool {
...
+    ) -> Option<Vec<Arc<VersionedTransaction>>> {
+        // Use a local hashmap to de-duplicate transactions
+        let mut tx_set = HashMap::new();
+        for slot in begin_skip_slot..=end_skip_slot {
+            let certificate = self.certificates.get(&(slot, CertificateType::Skip))?;
+            if certificate.is_complete() {
+                certificate
+                    .get_certificate_iter()
+                    .for_each(|(pubkey, entry)| {
+                        tx_set.insert((*pubkey, entry.skip_range()), entry.transaction().clone());
```
carllin: Isn't it possible we would need multiple disjoint skip votes for a specific validator? For instance a validator could vote (1, 4), and then (6, 9), and we might need both votes to make the skip certificate for (1, 9)

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -39,19 +56,78 @@ impl VoteCertificate {
...
+                panic!("New skip range should never be None");
+            }
+        } else {
+            // If both are None, this is notarization or finalization, do not accept
+            // We should always replace skip with a skip, so should never happen that old is
+            // None and new is Some
+            assert!(new_skip_range.is_none());
+            false
+        }
+    }
```
carllin, line 98: Overlapping skip votes should be slashable and we should reject them

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Yes it's possible, which is why the key to tx_set is (pubkey, skip_range). So we would have
(key1, (1, 4)), tx1
(key1, (6, 9)), tx2
So both tx1 and tx2 would be added to the certificate.

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Yup, it used to be we had a skip_pool here, so Ashwin needed a r_cert_tracker. Now we use one cert pool we can do that. But I want to do it in a separate PR.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Done.

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Actually I want return value to be true only if the entry didn't exist or the original value is smaller to the new value, changed.

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Done.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Now that we use per-slot skip pool, And we allow multiple overlapping skip votes per pubkey, the skip range in skip votes are basically abbreviation. What are we trying to achieve by checking for overlapping skip votes? What damage can it do? Do we care?
If we allow overlapping skip votes, then users can safely discard any slot older than local root, even though those slots might have been skipped by the current user. Otherwise users have to remember what he sent for last skip vote, otherwise his new votes might all get rejected. This is not impossible, but it's a bit annoying.

PR #127, opened 2025-04-02
[bls] add BLS crate for signing and verification

samkim-crypto:
#### Summary of Changes
Adding a bls crate with basic BLS signing, verification, and aggregation functionalities. 

I am not completely sure yet what the ideal API should be, so basic functionality like serialization is missing. These will be added incrementally as BLS is incorporated into the protocol.

Part of https://github.com/anza-xyz/alpenglow/issues/48.

Fixes #

### DIFF
--- a/bls/src/lib.rs
+++ b/bls/src/lib.rs
@@ -12,6 +12,10 @@ pub mod error;
 pub mod keypair;
 pub mod signature;
 
+/// Domain separation tag used for hashing messages to curve points to prevent
+/// potential conflicts between different BLS implementations. This is defined
+/// as the ciphersuite ID string as recommended in the standard
+/// https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2.1.
 pub const HASH_TO_POINT_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_";
 
 pub struct Bls;


bls/src/signature.rs
```diff
@@ -0,0 +1,70 @@
+use {
+    crate::{error::BlsError, keypair::BlsPubkey, Bls},
+    blstrs::G2Projective,
+};
+
+pub const BLS_SIGNATURE_SIZE: usize = 96;
```
0x0ece, line 6: I'd personally go for 96*2, i.e. uncompressed signatures. (but this is not used yet, so approving the PR)

bls/src/keypair.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    /// Constructs a new, random `BlsKeypair` using `OsRng`
+    pub fn new() -> Self {
+        let mut rng = OsRng;
+        Self::generate(&mut rng)
+    }
+
+    /// Derive a `BlsKeypair` from a seed (input key material)
+    pub fn derive(ikm: &[u8]) -> Result<Self, BlsError> {
```
wen-coding, line 148: naive question: I think we talked about deriving a BLS keypair using the user's current ed25519 keypair, is this how they do it? Can you add a full test case where we use ed25519 keypair to derive corresponding BLS keypair?

bls/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+    },
+    blstrs::{pairing, G1Affine, G2Projective},
+    group::prime::PrimeCurveAffine,
+};
+
+pub mod error;
+pub mod keypair;
+pub mod signature;
+
+pub const HASH_TO_POINT_DST: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_";
```
wen-coding, line 19: nit: explain a bit what this does maybe?

bls/src/lib.rs
```diff
@@ -0,0 +1,124 @@
...
+        message: &[u8],
+    ) -> Result<bool, BlsError>
+    where
+        I: IntoIterator<Item = &'a BlsPubkey>,
+        J: IntoIterator<Item = &'a BlsSignature>,
+    {
+        let aggregate_pubkey = BlsPubkey::aggregate(public_keys)?;
+        let aggregate_signature = BlsSignature::aggregate(signatures)?;
+
+        Ok(Self::verify(
```
wen-coding, line 52: naive question: so it's impossible to know which pubkey/signature cased the failure right? I guess we don't care that much, hmm

bls/src/signature.rs
samkim-crypto: Yeah this constant is not actually used since I was planning on adding a compressed and uncompressed version of the signature on a follow-up PR. The signature currently is defined as `G2Projective`, but I think we can change it to `G2Affine` on a follow-up as well.

bls/src/keypair.rs
samkim-crypto: Yeah I was thinking about adding a derivation function specifically for Solana signers, but I did not to prevent dependencies to the Solana sdk.

I ended up adding it under the `solana-signer-derive` feature in [029ec83](https://github.com/anza-xyz/alpenglow/pull/127/commits/029ec83fe2d1a1f9dda615ed20a8c4c9d8f20934) including the tests.

bls/src/lib.rs
samkim-crypto: Done in [fcc2f55](https://github.com/anza-xyz/alpenglow/pull/127/commits/fcc2f550476dbca49e4a416f1b225d4ababb4078)!

bls/src/lib.rs
samkim-crypto: Yeah it is not possible to know just from this function. I can add an additional function to spit out the failed signature index, but it probably makes sense as a separate follow-up PR.

bls/src/keypair.rs
```diff
@@ -48,6 +50,23 @@ impl BlsSecretKey {
...
+    /// Derive a `BlsSecretKey` from a Solana signer
+    #[cfg(feature = "solana-signer-derive")]
+    pub fn derive_from_signer(signer: &dyn Signer, public_seed: &[u8]) -> Result<Self, BlsError> {
+        let message = [b"bls-key-derive-", public_seed].concat();
+        let signature = signer
+            .try_sign_message(&message)
+            .map_err(|_| BlsError::KeyDerivation)?;
+
+        // Some `Signer` implementations return the default signature, which is not suitable for
+        // use as key material
```
wen-coding, line 62: Do I need to worry about some users not able to derive BLS key? Everyone should use our tool to generate key pair so they should all be covered right?

samkim-crypto: Yeah if users use the solana sdk for key management, then there shouldn't be any problems. Even if they use another crate like the dalek crate, they can generate an ikm and derive from it, so I don't think there should be a problem.

Issue #126, opened 2025-04-02
Rpc Health Check needs to migrate from OC slots


carllin:
#### Problem

Today we use the difference between OC slots in blockstore (the latest we've detected from the cluster) vs local replayed OC banks to detect how far we are behind the network https://github.com/anza-xyz/alpenglow/blob/4d70dc85a9924af8820dd2ecf608c42a2c34ff85/rpc/src/rpc_health.rs#L84-L96

In Alpenglow we cannot do this since OC is going away, we need to check the distance between our last known certificate slot and our local replayed slot

PR #137, opened 2025-04-04
Fixup rpc health check

carllin:
#### Problem
https://github.com/anza-xyz/alpenglow/issues/126

#### Summary of Changes
Write the alpenglow root as the OC slot to blockstore

Fixes #

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -802,6 +802,7 @@ impl VotingLoop {
             return None;
         }
 
+        // Distinguish between duplicate versions of same slot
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -801,6 +802,17 @@ impl VotingLoop {
             return None;
         }
 
+        let hash = ctx.bank_forks.read().unwrap().bank_hash(new_root).unwrap();
```
ksn6, line 806: Could we add a comment here that addresses why this is necessary?

carllin: not strictly necessary now since we never use OC in alpenglow, but right now to distinguish between duplicate versions of the same slot

Issue #120, opened 2025-03-29
Rework leader blocks to not require a certificate for each block


AshwinSekar:
#### Problem
Currently we check certificates for every single block.
Similarly when producing leader blocks we wait for the previous block to reach notarization (or skip).

#### Proposed Solution
Instead follow the paper and create all 4 blocks ASAP. We should only check certificates for the first block, or anytime the leader switches forks.

PR #138, opened 2025-04-04
replay: adjust cert reqs for consecutive and optimistic leader blocks

AshwinSekar:
#### Problem
Every block is checked for notarization and skip certificates. We should relax the requirements to allow for leaders to build as fast as possible.

#### Summary of Changes

Notarization:
* 1st slot in the leader block always requires a notarization certificate
* If the 2nd slot is not built off of the 1st slot, it requires a notarization certificate
* The 3rd and 4th slot  do not require notarization certificates, however they MUST be built consecutively

Skip:
* The 1st slot requires a skip certificate as normal
* If the 2nd slot is not built off the 1st slot, it requires a skip certificate for any range not including the 1st slot
* The 3rd and 4th slot do not require a skip certificate, however they MUST be built consecutively

As an example (4 leader slots per window)
```
         /- 7
        /        /- 9
 0 -> 2 -> 4 -> 5 -> 6 -> 8
```

| Block    | Notarization | Skip |
| -------- | ------- | -------|
| 2 | Invalid Block | Invalid Block |
| 4 | 2 | (3, 3) |
| 5 | None | None |
| 6 | None | None |
| 7 | Invalid Block | Invalid Block |
| 8  | 6 | (7, 7) |
| 9 | 5 | (6, 7) |

For reference here is the part of the paper detailing the block creation loop:
Screenshot 2025-04-04 at 2 29 22 PM
We could be even more strict, requiring that the second slot not build upon the first slot ONLY if the first slot's parent was skip certified, but I don't think that level of detail is necessary rn.

Fixes: https://github.com/anza-xyz/alpenglow/issues/120

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -28,8 +28,9 @@ use {
     solana_feature_set::FeatureSet,
     solana_gossip::cluster_info::ClusterInfo,
     solana_ledger::{
-        blockstore::Blockstore, leader_schedule_cache::LeaderScheduleCache,
-        leader_schedule_utils::first_of_consecutive_leader_slots,
+        blockstore::Blockstore,
+        leader_schedule_cache::LeaderScheduleCache,
+        leader_schedule_utils::{last_of_consecutive_leader_slots, leader_slot_index},
     },
     solana_measure::measure::Measure,
     solana_poh::poh_recorder::PohRecorder,
@@ -295,8 +296,7 @@ impl VotingLoop {
 
         // TODO(ashwin): Start loop once migration is complete current_slot from vote history
         loop {
-            let leader_start_slot = first_of_consecutive_leader_slots(current_slot);
-            let leader_end_slot = leader_start_slot + NUM_CONSECUTIVE_LEADER_SLOTS - 1;
+            let leader_end_slot = last_of_consecutive_leader_slots(current_slot);
             let mut skipped = false;
 
             // TODO(ashwin): Do maybe_leader for all 4 blocks here, right now
@@ -312,7 +312,7 @@ impl VotingLoop {
                 .collect();
 
             while current_slot <= leader_end_slot {
-                let leader_slot_index = (current_slot - leader_start_slot) as usize;
+                let leader_slot_index = leader_slot_index(current_slot);
                 let timeout = timeouts[leader_slot_index];
                 let cert_log_timer = Instant::now();
                 let mut skip_refresh_timer = Instant::now();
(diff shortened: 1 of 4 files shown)


core/src/replay_stage.rs
```diff
@@ -3702,6 +3708,10 @@ impl ReplayStage {
                     *stake,
                 );
             });
+        // We only need to check notarization certificates if we are building off a different window
+        let same_leader_window = (parent_bank.collector_id() == bank.collector_id())
+            && same_leader_window(parent_slot, bank.slot());
```
wen-coding: Can we just check whether parent and bank are in the same_leader_window here?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -295,8 +296,7 @@ impl VotingLoop {
 
         // TODO(ashwin): Start loop once migration is complete current_slot from vote history
         loop {
-            let leader_start_slot = first_of_consecutive_leader_slots(current_slot);
-            let leader_end_slot = leader_start_slot + NUM_CONSECUTIVE_LEADER_SLOTS - 1;
+            let leader_end_slot = last_of_consecutive_leader_slots(current_slot);
```
carllin, line 299: of note, can be fixed in another PR, i think the current_slot should be set to the  alpenglow slot that triggered the migration, not the current root slot as that can be an old tower slot from the non-alpenglow epoch > 32 slots ago

core/src/replay_stage.rs
```diff
@@ -3674,9 +3674,27 @@ impl ReplayStage {
         let parent_hash = parent_bank.hash();
         let mut notarization_stake = 0;
         let mut notarization_size = 0;
-        let must_skip_start = parent_slot + 1;
-        let must_skip_end = bank.slot() - 1;
         let mut skip_pool = SkipPool::new();
+
+        let leader_slot_idx = leader_slot_index(bank.slot());
+        if leader_slot_idx > 1 {
```
carllin, line 3680: nit: 1 -> `NUM_CONSECUTIVE_LEADER_SLOTS - 3`

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah good point, will add this when we reenable the migration

core/src/replay_stage.rs
AshwinSekar: I think 1 makes more sense here, because only the second leader block is special cased. If we increase `NUM_CONSECUTIVE_LEADER_SLOTS` in the future, then this will allow more blocks to skip the first leader block.

carllin: ah yeah sorry this was from before I understood what was going on

PR #141, opened 2025-04-04
replay: remove towerbft data structures from alpenglow set root

### LINKED ISSUE #65: Remove unnecessary data structures like EpochSlotsFrozenSlots, HeaviestSubtreeForkChoice

AshwinSekar:
#### Problem
We have a bunch of tower bft related data structures that are not needed in Alpenglow.

#### Summary of Changes
Remove them from set root, and instead manually clean them up when alpenglow is active

Fixes #65

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -180,6 +180,14 @@ struct SkippedSlotsInfo {
     last_skipped_slot: u64,
 }
 
+pub struct TowerBFTStructures<'a> {
+    pub heaviest_subtree_fork_choice: &'a mut HeaviestSubtreeForkChoice,
+    pub duplicate_slots_tracker: &'a mut DuplicateSlotsTracker,
+    pub duplicate_confirmed_slots: &'a mut DuplicateConfirmedSlots,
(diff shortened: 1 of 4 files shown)


core/src/replay_stage.rs
```diff
@@ -760,13 +768,30 @@ impl ReplayStage {
...
             loop {
                 // Stop getting entries if we get exit signal
                 if exit.load(Ordering::Relaxed) {
                     break;
                 }
 
+                // Rather than ripping these out immediately, we still populate the tower bft structures
+                // but ignore them. After migration is complete on all clusters we can remove the code
+                // paths that populate these
+                if is_alpenglow_migration_complete && tower_bft_cleanup.elapsed().as_secs() > 5 {
```
AshwinSekar, line 782: This is definitely a cop out, requires a decent amount of plumbing and updates to tests to stop populating these so I deferred for now.

With this change I believe we should be good enough to not OOM and we can rip out the code that populates these structs after migration is complete.

PR #143, opened 2025-04-05
Factor out tower-specific data structures

### LINKED ISSUE #65: Remove unnecessary data structures like EpochSlotsFrozenSlots, HeaviestSubtreeForkChoice

carllin:
#### Problem
We have a bunch of tower bft related data structures that are not needed in Alpenglow.

#### Summary of Changes
Built on https://github.com/anza-xyz/alpenglow/pull/141, factored out changes to all tower-specific data structures

Fixes #65

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -5076,37 +5076,35 @@ pub(crate) mod tests {
         let root_hash = root_bank.hash();
         bank_forks.write().unwrap().insert(root_bank);
 
-        let mut heaviest_subtree_fork_choice = HeaviestSubtreeForkChoice::new((root, root_hash));
-
         let mut progress = ProgressMap::default();
         for i in 0..=root {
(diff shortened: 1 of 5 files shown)


core/src/replay_stage.rs
```diff
@@ -822,6 +831,7 @@ impl ReplayStage {
                     &poh_recorder,
                     first_alpenglow_slot,
                     cert_tracker.as_ref(),
+                    (!is_alpenglow_migration_complete).then_some(&mut tbft_structs),
```
carllin, line 834: @AshwinSekar this is the only real change, so these structs no longer get touched after migration

PR #144, opened 2025-04-09
split out async block production loop, only need cert for first block

### LINKED ISSUE #133: Produce alpenglow leader blocks and maybe_start_leader checks smarter
Once we stop checking for certificates in each block, we should separate the `maybe_start_leader` slot from the timer slot. Currently we wait for each leader block to become certified before producing the next, however only the first leader block needs to wait for a certificate. 

We should also be smarter about grabbing the bank forks lock in such a tight loop.

AshwinSekar:
#### Problem
The current voting loop produces each leader block only after the certificate is obtained for the previous block. It also maintains a separate timer to determine when the block should end. There is some inefficient lock contention with poh recorder/bank forks here as the voting loop is responsible for adding the next block once replay/banking stage is complete.

Only the first leader block needs a certificate so in fact the rest of the window should be built independent of the voting loop.

#### Summary of Changes
Instead, similar to the paper we spread responsibility in two threads, the voting loop and the block creation loop (lives in poh_service):

Voting loop:
- At the beginning of the leader window, block on replay until we can either insert our first empty leader bank or our slot ends up getting skipped
- This requires communication from replay, as we could have seen certificates for the previous leader through all-to-all, however our replay of the previous leader's blocks have not yet completed.
- We use a mutex/condvar here to avoid tight spin and minimal contention
- The voting loop then continues as normal, it is not responsible for inserting empty banks for the rest of the leader window.

Poh service:
- After the first leader block is produced, the poh service keeps track of the same skip timer as the voting loop
- It uses this timer to control when the leader block should end
- The remaining three leader blocks are inserted here after the previous leader block has been cleared, without waiting for any certificates

For comparison, the reference implementation does something similar https://github.com/qkniep/alpenglow/blob/90f15866d0e067fc87b515a986710e383ee20022/src/consensus/mod.rs#L196 the difference is that:
- The block production loop is entirely separate from the voting loop, they share the certificate pool but loop through slots separately.
- We opt to keep this in one loop so we don't have to make the certificate pool thread safe. We simply slot in the voting logic here https://github.com/qkniep/alpenglow/blob/90f15866d0e067fc87b515a986710e383ee20022/src/consensus/mod.rs#L270
- The reference implementation produces all 4 blocks synchronously since they are fake blocks, however we don't have the luxury. After creating the first empty bank in the voting loop, the rest of the window is handled asynchronously , filled in by banking stage and inserted/ cleared by poh service.

Fixes #133

### DIFF
--- a/core/src/alpenglow_consensus.rs
+++ b/core/src/alpenglow_consensus.rs
@@ -60,14 +60,3 @@ pub const SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP: f64 = 0.2;
 pub const SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP: f64 = 0.6;
 
 pub const SAFE_TO_SKIP_THRESHOLD: f64 = 0.4;
-
-/// The amount of time a leader has to build their block in ms
-pub const BLOCKTIME: u128 = 400;
-
-/// The maximum message delay in ms
-pub const DELTA: u128 = 100;
-
-/// The Maximum delay a node can observe between entering the loop iteration
-/// for a window and receiving any shred of the first block of the leader.
-/// As a conservative global constant we set this to 3 * DELTA
-pub const DELTA_TIMEOUT: u128 = 300;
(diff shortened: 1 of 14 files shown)


core/src/alpenglow_consensus.rs
```diff
@@ -8,14 +8,3 @@ pub mod voting_loop;
 
 pub type Stake = u64;
 pub const SUPERMAJORITY: f64 = 2f64 / 3f64;
-
```
AshwinSekar: moved these to `poh` since `solana_poh` can't depend on `solana_core`

PR #149, opened 2025-04-10
[core] Add `VoteSignature` trait

samkim-crypto:
#### Problem
I am looking into incorporating BLS signatures to certs. Since the signatures for ed25519 and bls logic is largely overlapping, I was thinking of organizing a signature as a trait and then work with generically for any signature type.

#### Summary of Changes
The main addition is the `VoteSignature` trait, which should represent either a ed25519 or a bls signature.

It has an associated type `Aggregate`, which should ultimately be the certificate type:
- for ed25519 (currently represented as `VersionedTransaction`), it should be `Vec<VersionedTransaction>`
- for bls signatures, it should be the batch aggregated signature, so just `BlsSignature`

This trait has the `from_transaction(transaction: VersionedTransaction)`, which extracts the signature from the transaction:
- for ed25519 (represented as `VersionedTransaction`), it should just be a copy of the transaction
- for BLS sig, we would look at instruction data and extract the signature from there

Finally, the trait has the `aggregate(signatures: ...)` function which creates aggregate signature type, which should ultimately be the signature(s) in the cert.

I implemented this trait for ed25519 (represented as `VersionedTransaction`). For BLS, I will add it in a follow-up PR (assuming that this PR is a right approach).

I updated the `VoteCertificate`, `CertificatePool`, and `SkipPool` to use this generic type in place of `VersionedTransaction`. For the functions in the voting loop, I just hard-coded the generic with `VersionedTransaction` for now, but we can make these functions generic too (assuming that this PR is a right approach).

Please let me know what you think about this general direction. I am less familiar with this side of the codebase, so please let me know if you think this is not the right approach. I can always revert and think of a another approach.

Fixes #

### DIFF
--- a/core/src/alpenglow_consensus/skip_pool.rs
+++ b/core/src/alpenglow_consensus/skip_pool.rs
@@ -1,7 +1,7 @@
 use {
-    super::{utils::super_majority_threshold, Stake},
+    super::{utils::super_majority_threshold, vote_signature::VoteSignature, Stake},
     solana_pubkey::Pubkey,
-    solana_sdk::clock::Slot,
+    solana_sdk::{clock::Slot, transaction::VersionedTransaction},
     std::{
         collections::{BTreeMap, BTreeSet, HashMap},
         fmt::Debug,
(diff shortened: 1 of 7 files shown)


core/src/alpenglow_consensus/skip_pool.rs
```diff
@@ -141,28 +141,28 @@ impl<T: Ord + Clone + Debug + HasStake> DynamicSegmentTree<T> {
 }
 
 /// Structure to store a skip vote, including the range and transaction
-pub struct SkipVote<T> {
+pub struct SkipVote<S: VoteSignature> {
     skip_range: RangeInclusive<Slot>,
-    data: T,
+    signature: S,
```
samkim-crypto, line 146: Here, I am assuming that the transaction is used as just a signature. If the actual instruction data also needs to be kept around, then we would need to extract this data from `VersionedTransaction` and include it as a separate field in `SkipVote` or revert this and add a separate trait for transactions.

carllin: yeah we need to be able to process/pack these transactions later into a block later, so all the instruction data will be needed

samkim-crypto: Okay got it! I'll update this then. For the certificate itself that nodes (including non-leader nodes) broadcast out to other nodes once stake threshold is reached, it doesn't have to include the instruction data, correct? This can just be a vector of signatures (for the case of ed25519)?

carllin: > For the certificate itself that nodes (including non-leader nodes) broadcast out to other nodes once stake threshold is reached, i

Yeah those just need to be the smallest verifiable unit. We don't plant to broadcast the certificates right now for `ed25519`, too large to fit in a gossip packet.

PR #153, opened 2025-04-14
https://github.com/anza-xyz/agave/pull/7387

### LINKED ISSUE #135: Safeguard replay stage to allow for set root to happen at anytime
Since the voting loop now sets root, audit and make sure replay stage is resilient to set root happening at anytime

For example this check could potentially fail now:
https://github.com/anza-xyz/alpenglow/blob/66d43d4d809c6afca5913d5aade443d75b2f7eb0/core/src/replay_stage.rs#L3358-L3360

AshwinSekar:
#### Problem
Since `set_root` is handled by a different thread when alpenglow is active, we must audit some `unwrap` / `expect` assumptions.

#### Summary of Changes
We can no longer assume that banks exist if the lock was dropped in between

Fixes #135

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -3249,17 +3249,17 @@ impl ReplayStage {
             debug!("bank_slot {:?} is marked dead", bank_slot);
             replay_result.is_slot_dead = true;
         } else {
-            let bank = bank_forks
-                .read()
-                .unwrap()
-                .get_with_scheduler(bank_slot)
-                .unwrap();
+            let Some(bank) = bank_forks.read().unwrap().get_with_scheduler(bank_slot) else {
+                info!("Abandoning replay of unrooted slot {bank_slot}");
+                return replay_result;
+            };
             let parent_slot = bank.parent_slot();
             let prev_leader_slot = progress.get_bank_prev_leader_slot(&bank);
             let (num_blocks_on_fork, num_dropped_blocks_on_fork) = {
-                let stats = progress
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -4644,7 +4645,14 @@ impl ReplayStage {
         let mut generate_new_bank_forks_write_lock =
             Measure::start("generate_new_bank_forks_write_lock");
         let mut forks = bank_forks.write().unwrap();
-        for (_, bank) in new_banks {
+        let root = forks.root();
+        for (slot, bank) in new_banks {
+            if slot < root {
+                continue;
+            }
+            if forks.get(bank.parent_slot()).is_none() {
```
wen-coding, line 4653: When could this happen, do you need a log here?

core/src/replay_stage.rs
```diff
@@ -3249,17 +3249,17 @@ impl ReplayStage {
             debug!("bank_slot {:?} is marked dead", bank_slot);
             replay_result.is_slot_dead = true;
         } else {
-            let bank = bank_forks
-                .read()
-                .unwrap()
-                .get_with_scheduler(bank_slot)
-                .unwrap();
+            let Some(bank) = bank_forks.read().unwrap().get_with_scheduler(bank_slot) else {
+                info!("Abandoning replay of unrooted slot {bank_slot}");
```
wen-coding, line 3253: nice catch, I feel we probably have other places breaking as well, err.... Is it possible to add a test for this one?

Issue #156, opened 2025-04-16
Implement some way to populate the clock sysvar post Alpenglow


wen-coding:
#### Problem
We are removing timestamp from votes to enable BLS, we need some way to keep populating the clock sysvar

#### Proposed Solution
https://anza-xyz.slack.com/archives/C08AHR49WFQ/p1741202913717789?thread_ts=1741184985.307019&cid=C08AHR49WFQ

Let leader update clock sysvar within the allowed drift.

Other suggestions welcome.

PR #157, opened 2025-04-16
upgrade vote-program rev and rework certificate validation

AshwinSekar:
Upgrade vote program to https://github.com/solana-program/alpenglow-vote/pull/65
Rework cert pool to handle skip as single slot votes
Remove replay cert tracker
Move block validation code from replay_stage to voting_loop, use cert pool instead of vote account
Remove commit certificate logic in maybe_start_leader

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -135,15 +135,19 @@ impl CertificatePool {
             .certificates
             .entry((slot, certificate_type.clone()))
             .or_insert_with(|| VoteCertificate::new(slot));
-        certificate.add_vote(
+        if !certificate.add_vote(
             validator_vote_key,
             transaction,
             validator_stake,
             total_stake,
-        );
+        ) {
+            return None;
+        }
+
         if certificate.is_complete() && self.set_highest_slot(certificate_type, slot) {
             return Some(slot);
         }
+
         None
     }
 


runtime/src/bank.rs
```diff
@@ -7369,21 +7368,12 @@ pub mod test_utils {
         timestamp: BlockTimestamp,
         bank: &Bank,
         vote_pubkey: &Pubkey,
-        is_alpenglow: bool,
```
AshwinSekar, line 7372: we'll revisit this when designing the timestamp with BLS https://github.com/anza-xyz/alpenglow/issues/156
for now, clock timestamp will be incorrect in alpenglow blocks

core/src/replay_stage.rs
```diff
@@ -3435,35 +3424,6 @@ impl ReplayStage {
                     }
                 }
 
-                if let Some(first_alpenglow_slot) = first_alpenglow_slot {
-                    if let Some(parent_bank) = bank.parent() {
-                        if bank.slot() > first_alpenglow_slot {
-                            let root_bank = bank_forks.read().unwrap().root_bank();
-                            if let Err(e) = Self::alpenglow_check_cert_in_bank(
```
AshwinSekar, line 3442: this is now handled in the voting loop

vote/src/vote_parser.rs
```diff
@@ -39,8 +39,12 @@ impl ParsedVoteTransaction {
             ParsedVoteTransaction::Tower(tx) => tx.last_voted_slot_hash(),
             ParsedVoteTransaction::Alpenglow(tx) => match tx {
                 AlpenglowVote::Notarize(vote) => Some((vote.slot(), *vote.replayed_bank_hash())),
-                AlpenglowVote::Finalize(vote) => Some((vote.slot(), *vote.replayed_bank_hash())),
+                AlpenglowVote::Finalize(_vote) => None,
+                AlpenglowVote::NotarizeFallback(vote) => {
```
wen-coding, line 43: nit: now we have more vote types, can we sort alphabetically?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -137,40 +129,22 @@ impl CertificatePool {
...
-            _ => None,
-        };
         certificate.add_vote(
             validator_vote_key,
             transaction,
-            skip_range,
             validator_stake,
             total_stake,
-        )?;
+        );
```
wen-coding: If add_vote fails we can return immediately?

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -71,45 +71,27 @@ impl VoteHistory {
...
-        let new_skip_range = new_skip_vote.skip_range().unwrap();
-
         match self.skip_votes.last_mut() {
-            Some(last_vote)
-                if last_vote.skip_range().unwrap().start() == new_skip_range.start() =>
-            {
-                assert!(new_skip_range.end() >= last_vote.skip_range().unwrap().end());
-                // The new vote is a superset of the last vote, so just replace the last vote
+            Some(last_vote) if last_vote.slot() == new_skip_vote.slot() => {
+                // Refresh vote, replace the last vote
```
wen-coding, line 76: Refreshing only the last vote might not be enough any more, but guess we can wait for BLS here. Add TODO?

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -71,45 +71,27 @@ impl VoteHistory {
...
-            range.contains(&slot).then_some(*range.start())
-        })
+        self.skip_votes.iter().any(|vote| vote.slot() == slot)
     }
 
     pub fn set_root(&mut self, root: Slot) {
         self.root = root;
-        self.skip_votes
-            .retain(|skip_vote| *skip_vote.skip_range().expect("must be a skip vote").end() > root)
+        self.skip_votes.retain(|skip_vote| skip_vote.slot() > root)
```
wen-coding, line 94: Can this set_root happen from multiple threads? Do we need to protect the members now?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -482,46 +407,25 @@ impl VotingLoop {
...
     }
 
-    fn skip_certified(
-        my_pubkey: &Pubkey,
-        slot: Slot,
-        cert_pool: &mut CertificatePool,
-        cert_tracker: &RwLock<ReplayCertificateTracker>,
-    ) -> bool {
+    fn skip_certified(my_pubkey: &Pubkey, slot: Slot, cert_pool: &CertificatePool) -> bool {
         // TODO(ashwin): can include cert size for debugging
```
wen-coding, line 423: Do we still need this TODO?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -482,46 +407,25 @@ impl VotingLoop {
...
-    fn branch_notarized(
-        my_pubkey: &Pubkey,
-        slot: Slot,
-        cert_pool: &CertificatePool,
-        cert_tracker: &RwLock<ReplayCertificateTracker>,
-    ) -> bool {
+    fn branch_notarized(my_pubkey: &Pubkey, slot: Slot, cert_pool: &CertificatePool) -> bool {
         if let Some(size) = cert_pool.is_notarization_certificate_complete(slot) {
             info!(
                 "{my_pubkey}: Branch Notarized: Slot {} from {} validators",
```
wen-coding, line 413: Do we need this log when check succeeds? This might be spammy?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -56,82 +29,27 @@ impl VoteCertificate {
...
-                skip_range,
-            )
-        {
+        if self.certificate.contains_key(validator_key) {
             // Make duplicate vote fail silently, we may get votes from different resources and votes may arrive out of order.
             // This also needs to silently fail because the new skip vote might conflict with some old votes in old slots,
             // but perfectly fine for some other slots. E.g. old vote is (23, 23), (22, 24) will fail for slot 23, but it's
             // fine for slot 22 and 24.
-            return Ok(());
+            return false;
```
wen-coding, line 45: Hmm, we are blindly rejecting the new vote if some entry exists, should we compare the blockhash on vote transaction and update silently?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -818,29 +686,59 @@ impl VotingLoop {
...
+            skip_vote.slot(),
             bank.slot()
         );
         Self::send_vote(skip_vote, true, bank.as_ref(), cert_pool, voting_context);
     }
 
+    /// Determines if we can vote notarize at this time.
+    /// If this is the first leader block of the window, we check for certificates,
+    /// otherwise we ensure that the parent slot is consecutive.
+    fn can_vote_notarize(slot: Slot, parent_slot: Slot, cert_pool: &CertificatePool) -> bool {
```
wen-coding, line 705: I'm a bit lost, is this `safeToNotar` in the paper?

AshwinSekar: this is the block validation check that we talked about in the meeting

AshwinSekar: for example here's quientin's impl https://github.com/qkniep/alpenglow/blame/2334e84fb34e21c82229d381b8122579aa1824f9/src/consensus/votor.rs#L236

core/src/alpenglow_consensus/vote_history.rs
AshwinSekar: it's single threaded

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: i will take care of it when i update the voting loop

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: As I mentioned before, the logs I add are because I found it useful while debugging https://github.com/anza-xyz/alpenglow/pull/118#discussion_r2025848447

While I appreciate your reviews I would greatly appreciate it if you could focus on the actual logic of the PR and keep style comments to a minimum. It is important that we get the logic correct and it becomes hard to follow when there are many irrelevant or off topic comments on the PR.

If you feel strongly about style or logs, feel free to update them when we push to master, or in a separate pr.

core/src/alpenglow_consensus/vote_certificate.rs
AshwinSekar: I just maintained the behavior of the existing code.
To answer your question, it doesn't matter as the blockhash on refresh is not updated. Unlike TowerBFT blockhash expiration is not really a thing here

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: Yeah good catch, ba37274b8312cf9d1b5072e6b19df8ceeb950686

PR #162, opened 2025-04-20
feat: generalize `GossipVoter` to support Alpenglow

ksn6:
We're going to start writing local cluster tests for Alpenglow that use `GossipVoter`, so let's generalize `GossipVoter` to support Alpenglow votes.

### DIFF
--- a/local-cluster/src/cluster_tests.rs
+++ b/local-cluster/src/cluster_tests.rs
@@ -37,7 +37,10 @@ use {
     },
     solana_streamer::socket::SocketAddrSpace,
     solana_tpu_client::tpu_client::{TpuClient, TpuClientConfig, TpuSenderError},
-    solana_vote::vote_transaction::{self, VoteTransaction},
+    solana_vote::{
+        vote_parser::ParsedVoteTransaction,
+        vote_transaction::{self},
+    },
     solana_vote_program::vote_state::TowerSync,
     std::{
         collections::{HashMap, HashSet, VecDeque},
@@ -517,6 +520,9 @@ fn poll_all_nodes_for_signature(
     Ok(())
 }
 
+/// Represents a service that monitors the gossip network for votes, processes them according to
+/// provided filters and callbacks, and maintains a connection to the gossip network. Often used as
+/// a "spy" representing a Byzantine node in a cluster.
 pub struct GossipVoter {
     pub gossip_service: GossipService,
     pub tcp_listener: Option<TcpListener>,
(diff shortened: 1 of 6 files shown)


local-cluster/src/cluster_tests.rs
```diff
@@ -533,16 +536,91 @@ impl GossipVoter {
...
+            _ => None,
+        }
+    }
+
+    pub fn last_voted_slot(&self) -> Option<Slot> {
+        match self {
+            GossipVoterVoteTransaction::TowerVoteTransaction(vote_transaction) => {
+                vote_transaction.last_voted_slot()
+            }
+            GossipVoterVoteTransaction::AlpenglowVoteTransaction(vote) => match vote {
```
AshwinSekar: this can just be `vote.slot()`

local-cluster/src/cluster_tests.rs
```diff
@@ -533,16 +536,91 @@ impl GossipVoter {
...
+/// 7. Returns a GossipVoter struct that can be used to control and shut down the service
 pub fn start_gossip_voter(
     gossip_addr: &SocketAddr,
     node_keypair: &Keypair,
-    vote_filter: impl Fn((CrdsValueLabel, Transaction)) -> Option<(VoteTransaction, Transaction)>
+    vote_filter: impl Fn((CrdsValueLabel, Transaction)) -> Option<(GossipVoterVoteTransaction, Transaction)>
         + std::marker::Send
         + 'static,
-    mut process_vote_tx: impl FnMut(Slot, &Transaction, &VoteTransaction, &ClusterInfo)
+    mut process_vote_tx: impl FnMut(Slot, &Transaction, &GossipVoterVoteTransaction, &ClusterInfo)
```
AshwinSekar: Instead of creating a new enum can we just use `ParsedVoteTransaction` ? It supports alpenglow votes and is already the first step in these type of tests:
https://github.com/anza-xyz/agave/blob/61c1556de80ecf6de1001efa50f4a65841911164/local-cluster/tests/local_cluster.rs#L2805
https://github.com/anza-xyz/agave/blob/61c1556de80ecf6de1001efa50f4a65841911164/local-cluster/tests/local_cluster.rs#L4000

local-cluster/src/cluster_tests.rs
ksn6: fixed

local-cluster/src/cluster_tests.rs
ksn6: fixed

PR #164, opened 2025-04-21
Certificate Pool changes for 2-delta.

wen-coding:
- Support NotarizeFallback and SkipFallback
- Added safe_to_notar and safe_to_skip
- Added the limits according to the paper
- Added the conflicting votes check but commented them out, without cert retransmit this kills our local cluster tests.

TODO:
- branch_certified not added yet.

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -1,9 +1,8 @@
 use {
     super::{vote_pool::VotePool, Stake},
     crate::alpenglow_consensus::{
-        CertificateType, VoteType, CERTIFICATE_LIMITS, CONFLICTING_VOTETYPES,
-        MAX_ENTRIES_PER_PUBKEY_FOR_NOTARIZE_LITE, MAX_ENTRIES_PER_PUBKEY_FOR_OTHER_TYPES,
-        MAX_SLOT_AGE, SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP,
+        CertificateType, VoteType, CERTIFICATE_LIMITS, MAX_ENTRIES_PER_PUBKEY_FOR_NOTARIZE_LITE,
+        MAX_ENTRIES_PER_PUBKEY_FOR_OTHER_TYPES, MAX_SLOT_AGE, SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP,
         SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP, SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY,
         SAFE_TO_SKIP_THRESHOLD,
     },
@@ -119,7 +118,7 @@ pub struct CertificatePool {
     // Reverse lookup table of vote types to possible certificates it's affecting.
     vote_type_to_certificates: HashMap<VoteType, Vec<CertificateType>>,
     // Lookup table for checking conflicting vote types.
-    conflicting_vote_types: HashMap<VoteType, Vec<VoteType>>,
+    // conflicting_vote_types: HashMap<VoteType, Vec<VoteType>>,
     // Highest slot with each certificate
     highest_slot_map: HashMap<CertificateType, Slot>,
     // Cached epoch_schedule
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -54,29 +110,18 @@ impl NewHighestCertificate {
...
-}
-
 #[derive(Default)]
 pub struct CertificatePool {
-    // Notarization, finalization and skip vote certificates
+    // Vote pools to do bean counting for votes.
+    vote_pools: BTreeMap<PoolId, VotePool>,
+    // Certificate pools to keep track of the certs.
     certificates: BTreeMap<CertificateId, VoteCertificate>,
+    // Convert CERTIFICATE_LIMITS to a HashMap for faster lookup
```
AshwinSekar: can we just keep these as constants? We don't expect there to be multiple CertificatePool's with different limits or conflicting vote types

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -174,60 +314,42 @@ impl CertificatePool {
...
-        &self,
-        slot: Slot,
-    ) -> Option<Vec<Arc<VersionedTransaction>>> {
-        self.get_certificate(slot, CertificateType::Finalize)
-    }
-
-    pub fn highest_certificate_slot(&self) -> Slot {
-        self.highest_slot_map.values().max().copied().unwrap_or(0)
+        ) {
+            self.update_certificates(slot, vote_type, bank_hash, block_id, total_stake)?;
```
AshwinSekar, line 345: we can do this later, but it would be good to have a batched version of `add_vote` that can take a `Vec<Vote>` so we don't have to run `update_certificates` everytime.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -240,38 +362,148 @@ impl CertificatePool {
...
+            return true;
+        }
+        let Some(notarize_pool) = self.vote_pools.get(&(slot, VoteType::Notarize)) else {
+            return false;
+        };
+        notarize_pool.first_prev_vote_different(my_pubkey, bank_hash, block_id)
+    }
+
+    pub fn safe_to_notar(&self, my_pubkey: &Pubkey, bank: &Bank) -> bool {
+        if bank.slot() < self.root || !bank.is_frozen() {
```
AshwinSekar, line 440: just fyi this check and the `check_self_skip_or_notarized_different` doesn't need to be handled here, I will handle it in the voting loop with the `VoteHistory`.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -240,38 +362,148 @@ impl CertificatePool {
...
+        let Some(epoch_stakes) = self.epoch_stakes_map.get(&epoch) else {
+            return false;
+        };
+        let total_stake = epoch_stakes.total_stake();
+
+        let Some(notarize_pool) = self.vote_pools.get(&(slot, VoteType::Notarize)) else {
+            return false;
+        };
+        // Check if I voted notarize for some hash
+        if !notarize_pool.first_prev_vote_different(my_pubkey, None, None) {
```
AshwinSekar, line 492: same for here, i'll handle it with `VoteHistory`

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: I do need to make sure I have (block_id, hash) though, how do you want to change the API?

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: We can remove when your check is submitted.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Sure, let's do it in another PR though.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: I don't think we can have hashmap in constants, looking up entry in an array would be slower. But maybe not too bad, we only have 5 entries.
Changed.
Kept `conflicting_vote_types` because it's tiny and it's easier than going through the whole list every time to do reverse lookup.

PR #166, opened 2025-04-23
Add Bls transaction and certificate format

samkim-crypto:
I re-organized vote certificate and transactions into traits so that BLS can be incorporated without disrupting the existing logic.

I added the `AlpenglowVoteTransaction` trait , which attempts to capture vote transactions in Alpenglow. This is just a marker trait for now that is implemented by
- the traditional `VersionedTransaction`
- the new `BLSVoteTransaction`, which is a `VersionedMessage`, but just signed with BLS signatures instead
Assuming that we transition to BLS eventually, `BLSVoteTransaction` should be used for vote communication between validators. The `VersionedMessage` contains some data that is not really necessary in the context of votes, so we can consider creating a new variant of `VersionedMessage` in the future.

Then I added the `VoteCertificate` trait, which is implemented by:
- The `LegacyVoteCertificate`, which used to be the original `VoteCertificate` type
- The new `BLSCertificate`, which contains the BLS aggregate signature, bit-vector and some vote data that is currently empty. The `BLSCertificate` still does not implement the functions in the `VoteCertificate` trait yet, but I will do this in a follow-up.

I then updated `certificate_pool.rs and `vote_pool.rs` to work with generic traits. The `voting_loop.rs` is concretely instantiated with `LegacyVoteCertificate`. This will give me sufficient room to add the BLS certificate logic without disrupting the existing code.

I also added the logic for creating bit vector for the BLS pubkeys. I considered using the [bitvector](https://docs.rs/bitvector/latest/bitvector/) crate, but I decided to manually implement it for now since it is pretty simple and to also minimize dependencies. We can consider using an external crate in the future if we need more sophisticated logic.

I also noticed that the current BLS signature interface is quite clunky (we need to involve BLS `Pubkey`, `PubkeyProjective`, `Signature`, and `SignatureProjective`). This will be improved in the future.

### DIFF
--- a/core/src/alpenglow_consensus/bit_vector.rs
+++ b/core/src/alpenglow_consensus/bit_vector.rs
@@ -1,10 +1,14 @@
 use thiserror::Error;
 
 /// Maximum number of validators
+///
+/// There are around 1300 validators currently. For a clean power-of-two
+/// implementation, we should chosoe either 2048 or 4096. Choose a more
+/// conservative number 4096 for now.
 const MAXIMUM_VALIDATORS: usize = 4096;
 
 /// The number of bytes in a bit-vector to represent up to 4096 validators
-/// (4096 validators / 8)
+/// (`MAXIMUM_VALIDATORS` / 8)
 const VALIDATOR_BIT_MAP_U8_SIZE: usize = 512;
 
 #[derive(Debug, Error, PartialEq)]


core/src/alpenglow_consensus/bit_vector.rs
```diff
@@ -0,0 +1,91 @@
+use thiserror::Error;
+
+/// Maximum number of validators
+const MAXIMUM_VALIDATORS: usize = 4096;
+
+/// 4096 validators / 8
+const VALIDATOR_BIT_MAP_U8_SIZE: usize = 512;
```
wen-coding, line 12: Can you explain a bit how we arrived at the 512/4096 numbers?

core/src/alpenglow_consensus/bit_vector.rs
```diff
@@ -0,0 +1,91 @@
...
+
+#[derive(Debug, Error, PartialEq)]
+pub enum BitVectorError {
+    #[error("Index out of bounds")]
+    IndexOutOfBounds,
+}
+
+/// Validator bit map
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BitVector {
```
wen-coding, line 22: Can we use bv::BitVec?

core/src/alpenglow_consensus/transaction.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct BlsVoteTransaction {
+    /// BLS signature certifying the message
+    pub signature: Signature,
+    /// Message signed
+    pub message: VersionedMessage,
+}
+
+impl BlsVoteTransaction {
```
wen-coding, line 23: Maybe add some comments what this is for: is it for communication between validators (not going on chain), or the BLS aggregate (going on chain).

FYI, after BLS is working we plan to not use real Transactions for communication between validators any more, it will just be signed messages (e.g. we don't care about a bankhash etc).

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn new(stake: Stake, transactions: Vec<Arc<Self::VoteTransaction>>) -> Self;
+    fn size(&self) -> Option<usize>;
+    fn transactions(&self) -> Vec<Arc<Self::VoteTransaction>>;
+    fn stake(&self) -> Stake;
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct LegacyVoteCertificate {
+    // We don't need to send the actual vote transactions out for now.
+    transactions: Vec<Arc<VersionedTransaction>>,
```
wen-coding, line 26: We can add in comments: this will go away after BLS is done.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Vote message
+    pub vote_data: CertificateVoteData,
+    /// BLS aggregate signature
+    pub aggregate_signature: Signature,
+    /// Bit-vector indicating which votes are invluded in the aggregate signature
+    pub bit_vector: BitVector,
+}
+
+impl BlsCertificate {
+    pub fn new(
```
wen-coding, line 101: We haven't implemented this for now, but we might need to add late votes to the certificate later to be put in the block. Is it easy to support that?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -0,0 +1,131 @@
...
+        vote_data: CertificateVoteData,
+        validator_pubkey_map: HashMap<ValidatorPubkey, usize>,
+        transactions_map: HashMap<ValidatorPubkey, BlsVoteTransaction>,
+    ) -> Result<Self, BlsCertificateError> {
+        let mut aggregate_signature = SignatureProjective::default();
+        let mut bit_vector = BitVector::default();
+        let pubkey_transactions = transactions_map.iter();
+        for (pubkey, transaction) in pubkey_transactions {
```
wen-coding, line 112: Do you need these transactions to be in determined order?

core/src/alpenglow_consensus/transaction.rs
```diff
@@ -0,0 +1,38 @@
...
+};
+
+pub trait AlpenglowVoteTransaction: Clone + Default {}
+
+impl AlpenglowVoteTransaction for VersionedTransaction {}
+
+impl AlpenglowVoteTransaction for BlsVoteTransaction {}
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct BlsVoteTransaction {
```
AshwinSekar, line 16: can this be in the vote program?
As in, can we represent a bls agg vote transaction using the existing vote instructions with the signature in the instruction data? Or a new instruction variant?

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: They do not have to be in order. In fact, the signature aggregation can be parallelized as well. I added a comment regarding this for follow-up.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Yeah good point. I added a function to add new BLS transactions to an existing certificate. The implementation is a little messy due to a clunky BLS interface, but this will be improved in the future.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Done

core/src/alpenglow_consensus/transaction.rs
samkim-crypto: Done

core/src/alpenglow_consensus/bit_vector.rs
samkim-crypto: Yeah I implemented bit-vector manually to minimize dependency, but I realize now that `bv::BitVec` is already included as a dependency in `core`. I'll replace this with `bv::BitVec`, but if I can, let me do it in a separate follow-up PR 🙏

core/src/alpenglow_consensus/bit_vector.rs
samkim-crypto: This is the number of bytes in a bit-vector to represent up to 4096 validators.

There are currently 1300(?) validators. It is always clean to choose a power-of-two number as the maximum number of validators, so we can choose either 2048 or 4096. Once the protocol is set, it might be a headache to increase this bound later on, so I thought 4096 was a good conservative number (see https://anza-xyz.slack.com/archives/C08AHR49WFQ/p1745027120571029?thread_ts=1745025832.163619&cid=C08AHR49WFQ).

Then 4096 bits in bytes is 512.

Of course, this number can be changed at any time if we decide on a different number.

core/src/alpenglow_consensus/transaction.rs
samkim-crypto: Yeah the `BlsVoteTransaction` format is what would be used by validators to communicate votes with each other off-chain, so I think it makes sense to be its own thing (sorry for the confusion, I added a comment clarifying this now).

The `BlsCertificate` is the type that contains the bls agg sig. This should be wrapped inside an instruction as part of the instruction data to be verified on-chain.

- Leader receives `BlsVoteTransactions` from other validators
- Once it receives enough votes: 
  - it generates `BlsCertificate` from the list of `BlsVoteTransactions` it receives
  - wraps `BlsCertificate` in a vote instruction and executes it on-chain

Given this, `BlsCertificate` should live outside of `core` for programs to add as dependency, but I thought it would be simpler to have everything in one place for now for simplicity.

core/src/alpenglow_consensus/bit_vector.rs
wen-coding: I mean, maybe put it in comments?

samkim-crypto: Oh oops sorry. Done!

Issue #155, opened 2025-04-14
Add restarting from snapshot and WFSM functionality


AshwinSekar:
#### Problem

Allow for restarting from a snapshot or through WFSM in the middle of a leader window.

Specifically the snapshot slot will be auto notarized so the certificate checker needs to be able to handle this

PR #168, opened 2025-04-25
block creation loop: split from voting loop, cert only on first block

AshwinSekar:
#### Problem
The current voting loop produces each leader block only after the certificate is obtained for the previous block. It also maintains a separate timer to determine when the block should end. There is some inefficient lock contention with poh recorder/bank forks here as the voting loop is responsible for adding the next block once replay/banking stage is complete.

Only the first leader block needs a certificate so in fact the rest of the window should be built independent of the voting loop.

#### Summary of Changes

Instead, similar to the paper we spread responsibility in two threads, the voting loop and the block creation loop (lives in poh_service):

Voting loop:
1) At the beginning of our leader window, check that we have not missed our slot:
     - `parent_slot` = highest notarized slot < our leader window
     - our leader slot is not already skip certified
2) Notify the block creation loop of our window, `parent_slot`. and the `skip_timer`
3) Continue voting as normal

Block creation loop:
1) Wait until notified of a leader window by the voting loop
2) If the `parent_slot` sent by the voting loop has not finished replay, wait until the block timeout (via `skip_timer`) for replay to finish.
3) Basic checks that we can start producing blocks (startup verification, duplicate block, vote has landed)
4) Insert empty bank for the first slot
5) Wait until the block timeout (via `skip_timer`) is reached, or max poh tick height is reached
6) Clear the bank manually if needed
7) Wait until replay finishes
8) Insert the empty bank for the next slot (consecutive) immediately
9) Repeat (5) to (8) until the end of the window

Further TODO:
If during steps (5) to (9), we receive another notification from the voting loop because we were too slow, consider giving up on this window and going straight to the next window

Fixes https://github.com/anza-xyz/alpenglow/issues/133

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -232,13 +232,18 @@ impl VotingLoop {
 
             if is_leader {
                 // Let the block creation loop know it is time for it to produce the window
+                // TODO: We max with root here, as the snapshot slot might not have a certificate.
+                // Think about this more and fix if necessary.
+                let parent_slot = cert_pool
+                    .highest_not_skip_certificate_slot()
+                    .max(root_bank_cache.root_bank().slot());
                 Self::notify_block_creation_loop_of_leader_window(
                     &my_pubkey,
                     &cert_pool,
                     &leader_window_notifier,
-                    &mut root_bank_cache,
                     current_slot,
                     leader_end_slot,
+                    parent_slot,
                     skip_timer,
                 );
             }
@@ -810,26 +815,19 @@ impl VotingLoop {
     }
 
     /// Notifies the block creation loop of a new leader window to produce.
-    /// Uses the highest certified slot (not skip) as the parent for the first
-    /// leader slot.
+    /// Caller should use the highest certified slot (not skip) as the `parent_slot`
     ///
-    /// Fails to notify and returns false if the first leader slot has already
+    /// Fails to notify and returns false if the leader window has already
     /// been skipped, or if the parent is greater than or equal to the first leader slot
     fn notify_block_creation_loop_of_leader_window<VC: VoteCertificate>(
         my_pubkey: &Pubkey,
         cert_pool: &CertificatePool<VC>,
         leader_window_notifier: &LeaderWindowNotifier,
-        root_bank_cache: &mut RootBankCache,
         start_slot: Slot,
         end_slot: Slot,
+        parent_slot: Slot,
         skip_timer: Instant,
     ) -> bool {
-        // TODO: We max with root here, as the snapshot slot might not have a certificate.
-        // Think about this more and fix if necessary.
-        let parent_slot = cert_pool
-            .highest_not_skip_certificate_slot()
-            .max(root_bank_cache.root_bank().slot());
-
         // Check if we missed our window
         if (start_slot..=end_slot).any(|s| cert_pool.skip_certified(s)) {
             warn!(
@@ -846,7 +844,7 @@ impl VotingLoop {
             return false;
         }
 
-        // Sanity check
+        // Sanity check for certificates
         if !cert_pool.make_start_leader_decision(start_slot, parent_slot, 0) {
             panic!("Something has gone wrong with the voting loop: {start_slot} {parent_slot}");
         }


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -215,53 +205,61 @@ impl VotingLoop {
...
             let skip_timer = Instant::now();
-            let timeouts: Vec<_> = (1..=(NUM_CONSECUTIVE_LEADER_SLOTS as u128))
-                .map(|i| DELTA_TIMEOUT + i * BLOCKTIME + DELTA)
+            let timeouts: Vec<_> = (0..(NUM_CONSECUTIVE_LEADER_SLOTS as usize))
+                .map(crate::alpenglow_consensus::skip_timeout)
                 .collect();
 
+            if is_leader {
+                // Let the block creation loop know it is time for it to produce the window
+                Self::notify_block_creation_loop_of_leader_window(
```
AshwinSekar, line 240: this fn is the relevant change in this file, other than that some random optimizations that's not too important

core/src/replay_stage.rs
```diff
@@ -758,13 +763,16 @@ impl ReplayStage {
...
                 .expect("new rayon threadpool");
 
-            Self::reset_poh_recorder(
-                &my_pubkey,
-                &blockstore,
-                working_bank,
-                &poh_recorder,
-                &leader_schedule_cache,
-            );
+            if !is_alpenglow_migration_complete {
```
AshwinSekar, line 771: when the flag is active, `block_creation_loop` is the only thread that reset's poh, otherwise things can fail.

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -0,0 +1,443 @@
...
+            );
+
+            // Produce our next slot
+            slot += 1;
+            if slot > end_slot {
+                trace!("{my_pubkey}: finished leader window {start_slot}-{end_slot}");
+                break;
+            }
+
+            if let Err(e) = start_leader_retry_replay(slot, slot - 1, skip_timer, &ctx) {
```
AshwinSekar, line 263: I noticed in testing that sometimes when we get here the bank for `slot - 1` is not yet frozen, even though it always is cleared from the `poh_recorder`. That's why I use `start_leader_retry_replay` rather than just `maybe_start_leader`

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -0,0 +1,443 @@
...
+            error!(
+                "{my_pubkey}: Unable to produce first slot {start_slot}, skipping production of our entire leader window \
+                {start_slot}-{end_slot}: {e:?}"
+            );
+            continue;
+        }
+
+        // Produce our window
+        let mut slot = start_slot;
+        // TODO(ashwin): Handle preemption of leader window during this loop
```
AshwinSekar, line 198: What could happen is that our `skip_timer` is desynced from the rest of the cluster. After we start producing the leader window, we immediately see a skip certificate for our 4 leader blocks.

If we are the leader for the next window as well, perhaps we should abandon our leader window in favor of the next one. The current code will produce all 4 leader blocks before checking for the next leader window. 

I will plumb this the preemption logic in a future PR, opted to keep it simple for now.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -948,42 +775,106 @@ impl VotingLoop {
...
+        skip_timer: Instant,
+    ) -> bool {
+        // TODO: We max with root here, as the snapshot slot might not have a certificate.
+        // Think about this more and fix if necessary.
+        let parent_slot = cert_pool
+            .highest_not_skip_certificate_slot()
+            .max(root_bank_cache.root_bank().slot());
+
+        // Check if we missed our slot
+        if cert_pool.skip_certified(start_slot) {
```
carllin: should we expand this check to any slot in the range start_slot..end_slot

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -948,42 +775,106 @@ impl VotingLoop {
...
+            warn!(
+                "{my_pubkey}: Leader slot {start_slot} has a higher certified slot {parent_slot}, \
+                skipping production of {start_slot}-{end_slot}"
+            );
+            return false;
+        }
+
+        // Sanity check
+        if !cert_pool.make_start_leader_decision(start_slot, parent_slot, 0) {
+            panic!("Something has gone wrong with the voting loop: {start_slot} {parent_slot}");
```
carllin, line 849: is this generally safe because the voting loop cannot proceed to the next slot unless it sees a certificate? so we know that everything < start_slot must have had a certificate? 

Is this true even if we start from a snapshot with root slot thats not in the cert pool? So if root_slot == parent_slot == snapshot root, and that snapshot root is not in cert pool, would we not have a notarization cert for the root slot

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -948,42 +775,106 @@ impl VotingLoop {
...
+        root_bank_cache: &mut RootBankCache,
+        start_slot: Slot,
+        end_slot: Slot,
+        skip_timer: Instant,
+    ) -> bool {
+        // TODO: We max with root here, as the snapshot slot might not have a certificate.
+        // Think about this more and fix if necessary.
+        let parent_slot = cert_pool
+            .highest_not_skip_certificate_slot()
+            .max(root_bank_cache.root_bank().slot());
```
carllin: gut feeling is the `parent_slot` should also be passed in, and we should check that a cert for it exists in the cert pool before we even enter this function

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -948,42 +775,106 @@ impl VotingLoop {
...
+            panic!("Something has gone wrong with the voting loop: {start_slot} {parent_slot}");
+        }
+
+        // Notify the block creation loop.
+        let mut l_window_info = leader_window_notifier.window_info.lock().unwrap();
+        if let Some(window_info) = l_window_info.as_ref() {
+            error!(
+                "{my_pubkey}: Attempting to start leader window for {start_slot}, however there is \
+                already a pending window to produce {}-{}. Something has gone wrong, Overwriting in favor \
+                of the newer window",
```
carllin, line 858: might be worth logging the newer window start/end slot as well

poh/src/poh_service.rs
```diff
@@ -101,15 +94,23 @@ impl PohTiming {
...
         hashes_per_batch: u64,
         record_receiver: Receiver<Record>,
-    ) -> Self {
+        block_creation_loop: F,
+    ) -> Self
+    where
+        // TODO: this weirdness is because solana_poh can't depend on solana_core
+        // Once we cleanup the prototype and separate alpenglow into it's own crate this
+        // can be fixed.
+        F: FnOnce() + std::marker::Send + 'static,
```
carllin, line 112: lmao black magic

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -0,0 +1,443 @@
...
+                // We wait until either we finish replay of the parent or the skip timer finishes
+                let _unused = ctx
+                    .replay_highest_frozen
+                    .freeze_notification
+                    .wait_timeout_while(
+                        highest_frozen_slot,
+                        timeout.saturating_sub(skip_timer.elapsed()),
+                        |hfs| *hfs < parent_slot,
+                    )
+                    .unwrap();
```
carllin, line 332: Must the parent bank must always be eventually frozen if we have a notarization cert for it? Can we:

1. get a skip for the parent bank 
2. Give up on repairing the parent bank, move on to the other fork
3. The root moves past the parent slot, making the parent frozen condition impossible?

core/src/alpenglow_consensus/block_creation_loop.rs
carllin: yeah makes sense because replay_stage is the one that freezes the leader bank after detecting the tick has reached max tick height

Maybe a comment here would be good for memory

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -0,0 +1,443 @@
...
+                }
+            }
+
+            assert!(!poh_recorder.read().unwrap().has_bank());
+            // TODO(ashwin): Is this necesary?
+            PohService::read_record_receiver_and_process(
+                &poh_recorder,
+                &record_receiver,
+                remaining_slot_time, /* 0 */
+            );
```
carllin: This is a good assert, that last `read_record_receiver_and_process()` is only useful after the bank is cleared, because at that point we know there's nothing else to be written into the channel, and we can clear it via `read_record_receiver_and_process()`

As for that TODO, yeah that last read is necessary, otherwise banking_stage can get stuck

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -0,0 +1,443 @@
...
+    if ctx.bank_forks.read().unwrap().get(slot).is_some() {
+        return Err(StartLeaderError::AlreadyHaveBank(slot));
+    }
+
+    let Some(parent_bank) = ctx.bank_forks.read().unwrap().get(parent_slot) else {
+        return Err(StartLeaderError::ReplayIsBehind(parent_slot));
+    };
+
+    if !parent_bank.is_frozen() {
+        return Err(StartLeaderError::ReplayIsBehind(parent_slot));
```
carllin, line 369: wondering if parent slot can get pruned/we move past it because of skip before it ever gets frozen. Same comment as above

core/src/replay_stage.rs
```diff
@@ -267,6 +270,8 @@ pub struct ReplayStageConfig {
     pub log_messages_bytes_limit: Option<usize>,
     pub prioritization_fee_cache: Arc<PrioritizationFeeCache>,
     pub banking_tracer: Arc<BankingTracer>,
+    pub replay_highest_frozen: Arc<ReplayHighestFrozen>,
```
carllin, line 273: Should this highest frozen be initialized on startup with the highest frozen bank in bank forks? I think right now it's initialized to default, and the trigger is initialized to 0: https://github.com/anza-xyz/alpenglow/blob/c1d0c7f5786ae971f90a65569ae750ba26051232/core/src/replay_stage.rs#L641

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: makes sense!

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: To make your question concrete here's an example
We are set to produce slot 100, due to the sequential nature of the voting loop we MUST have at least one cert for all slots < 100:
- Slot 95 is the one we choose as a parent (assume 96 - 99 are offline leader)
- We wait for 95 to complete replay but we are very behind
- A skip certificate is ingested for 95
- We are so far behind on replay + our skip timer is so far out of sync that the cluster skip certifies 100 - 103
- The next leader produces 104 - 107 using skip cert for 95
- The cluster finalizes 104 - 107
- Voting loop roots 107

In such a scenario, as you said 95 would be pruned and there's no chance of it ever finishing replay here. What saves us here is the `skip_timer`. As long as `skip_timer.elapsed()` eventually ticks over `timeout`, this `fn` will exit and we will skip production of our leader window.

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: can definitely happen, eventually the `skip_timer` will timeout and we will give up on our leader window. There's no "dynamic reselection of the parent" logic here, but it's definitely something we could explore in the future?

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah exactly. I haven't figured out the snapshot slot stuff yet, there's a TODO at the top of this function for this exact case. We need to figure out how to allow exceptions for the snapshot slot / WFSM which might not be in the cert pool.

For GCE this is why `make_start_leader_decision` has that exception for slot 1, since GCE does WFSM on a snapshot from 1 😅

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Could you elaborate? Where would we get `parent_slot` if not from the `cert_pool`?

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: done 821526acad311893b51dbcc62aaa295a0fd98676

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: good call a0d73ab1eb18a0c6666b623780ab5a94afac66a3

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: 8606157b75cead1635a3f5c9dcea66eaffb0ea4b

core/src/replay_stage.rs
AshwinSekar: good call 25f3bff2cb5b50cef6183eeed810c09a39d91f1f

core/src/alpenglow_consensus/voting_loop.rs
carllin: ah yeah, I meant reading the  `cert_pool.highest_not_skip_certificate_slot().max(root_bank_cache.root_bank().slot());` outside of this function and passing it in.

Then I think it might be safer also to return if that parent is not notarized instead of panicking

core/src/alpenglow_consensus/block_creation_loop.rs
carllin: ah yeah the `while !timeout.saturating_sub(skip_timer.elapsed()).is_zero() {` will exit so it won't get stuck, nice

core/src/alpenglow_consensus/voting_loop.rs
carllin: Yeah I think you would have to either have the cert on disk or repair it. Seems to make sense then to add a case `if parent_slot is not notarized, return` so we don't panic then until this is resolved

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: done e2004d5d09b09e83ff47aed64e52f042a08f6a9a
I kept the panic for now because it's useful for my debugging of the voting loop/cert pool, we can consider making it less harsh if there's a case in the future.

Also I had to modify the `highest_frozen_slot` logic to account for no blocks frozen (which happens in tests). This also involved touching the migration flag code fyi a14acf3c9679c17bf0e6848c8b6f9fc0b2f0e272

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Lets demote the panic in #155 once I finish testing the April 18th voting loop. I want to make sure that it works correctly in the controlled tests where `make_start_leader_decision` should always succeed.

Issue #79, opened 2025-03-11
Figure out batching for add_vote when adding skip votes


AshwinSekar:
Currently this will rescan certs everytime, we can be smarter

PR #169, opened 2025-04-25
cert_pool: rework root notifications to improve local-cluster speed

AshwinSekar:
#### Problem
We only notify RPC of finalization once every leader window when we set root.
As a result, the local cluster tests which wait for 16 roots, end up waiting for 64 roots

However we can receive a finalization certificate at any point. We should propagate the notification as soon as it arrives.

#### Summary of Changes
- Refactor `cert_pool.add_vote(vote)` to return `Some(slot)`  if the result of adding `vote` to the pool caused `slot` to become the new highest finalized slot
- Notify rpc of the new finalization immediately
- Reenable the 4 node alpenglow local-cluster test

### DIFF
--- a/core/src/alpenglow_consensus.rs
+++ b/core/src/alpenglow_consensus.rs
@@ -43,6 +43,10 @@ pub const CONFLICTING_VOTETYPES: [(VoteType, VoteType); 5] = [
     (VoteType::Skip, VoteType::SkipFallback),
 ];
 
+/// Lookup from `CertificateType` to the `VoteType`s that contribute,
+/// as well as the stake fraction required for certificate completion.
+///
+/// Must be in sync with `vote_type_to_certificate_type`
 pub const fn certificate_limits_and_vote_types(
     cert_type: CertificateType,
 ) -> (f64, &'static [VoteType]) {
@@ -57,6 +61,9 @@ pub const fn certificate_limits_and_vote_types(
     }
 }
 
+/// Lookup from `VoteType` to the `CertificateType`s the vote accounts for
+///
+/// Must be in sync with `certificate_limits_and_vote_types`
 pub const fn vote_type_to_certificate_type(vote_type: VoteType) -> &'static [CertificateType] {
     match vote_type {
         VoteType::Notarize => &[


core/src/alpenglow_consensus.rs
```diff
@@ -37,19 +43,33 @@ pub const CONFLICTING_VOTETYPES: [(VoteType, VoteType); 5] = [
...
+            (0.6, &[VoteType::Notarize, VoteType::NotarizeFallback])
+        }
+        CertificateType::FinalizeFast => (0.8, &[VoteType::Notarize]),
+        CertificateType::Finalize => (0.6, &[VoteType::Finalize]),
+        CertificateType::Skip => (0.6, &[VoteType::Skip, VoteType::SkipFallback]),
+    }
+}
+
+pub const fn vote_type_to_certificate_type(vote_type: VoteType) -> &'static [CertificateType] {
+    match vote_type {
```
wen-coding, line 68: Hmm, I'm fine with adding code to this file (maybe we should move timout constants to here as well?).

However, vote_type_to_certificate_type is basically a reverse lookup table, I'm very worried that this may get out of sync with the table above. If we really can't dynamically build it based on some constants, at least put big red warning this is a reverse lookup table so it has to be kept in sync (and change name to make it really really obvious).

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -967,27 +969,53 @@ impl VotingLoop {
                     return;
                 }
 
-                if let Err(e) = cert_pool.add_vote(&vote, tx, &vote_account_pubkey) {
+                if let Err(e) = Self::add_vote_and_maybe_update_commitment(
```
wen-coding, line 972: I know we punt on https://github.com/anza-xyz/alpenglow/issues/79 for now. But some day we may want to add add_vote_batch.

Can we maybe check highest_finalized_slot upon entrance of this function, then check again when we are done with this batch of votes, and notify when the two are different? If we finalize slot 5 and 6 in the same batch, we only need one notification right?

This does leave the nice possibility of doing batch in the future.

local-cluster/tests/local_cluster.rs
```diff
@@ -182,7 +182,7 @@ fn test_alpenglow_nodes_basic(num_nodes: usize, num_offline_nodes: usize, num_ne
 
     // Check for new roots
     cluster.check_for_new_roots(
-        num_new_roots,
+        16,
```
wen-coding, line 185: Good catch on the notification problem. But why are we not allowing specification of num_new_roots any more?

core/src/alpenglow_consensus.rs
AshwinSekar: added a warning

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: doesn't need to be checked, add_vote_batch can return the highest finalized slot similar to add_vote

PR #172, opened 2025-04-26
[core] Implement `VoteCertificate` trait for `BlsCertificate`

samkim-crypto:
#### Summary of Changes
This is a small follow-up to https://github.com/anza-xyz/alpenglow/pull/166. The `BlsCertificate` doesn't implement all the functions in `VoteCertificate` yet, so I made the update:
- I added the aggregate Bls pubkey field in order to use the syscall to verify the cert [slack discussion](https://anza-xyz.slack.com/archives/C08AHR49WFQ/p1745449272051849?thread_ts=1745025832.163619&cid=C08AHR49WFQ) 
- I realized that the `validator_pubkey_map` should actually be defined w.r.t. Bls pubkeys, so I updated this
- I added a vote count field and renamed the `size` to `vote_count`
- It appeared to me that the `transactions()` function was not used, so I removed it since it doesn't make sense for Bls certificates

### DIFF
--- a/core/src/alpenglow_consensus/bls_vote_transaction.rs
+++ b/core/src/alpenglow_consensus/bls_vote_transaction.rs
@@ -11,6 +11,7 @@ impl AlpenglowVoteTransaction for BlsVoteTransaction {}
 /// real Solana program instructions to be processed on-chain.
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct BlsVoteTransaction {
+    /// Bls pubkey associated with the transaction
     pub pubkey: Pubkey,
     /// BLS signature certifying the message
     pub signature: Signature,


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -222,8 +222,12 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                     };
                     vote_pool.copy_out_transactions(bank_hash, block_id, &mut transactions);
                 }
-                self.certificates
-                    .insert((slot, cert_type), VC::new(accumulated_stake, transactions));
+                // TODO: use an empty hash map of pubkeys for now since it is not clear
+                // where to get the sorted list of validators yet
+                self.certificates.insert(
+                    (slot, cert_type),
+                    VC::new(accumulated_stake, transactions, &HashMap::default()),
```
wen-coding: CertificatePool has a cached epoch_stakes_map, would that work?

core/src/alpenglow_consensus/transaction.rs
```diff
@@ -14,6 +14,7 @@ impl AlpenglowVoteTransaction for BlsVoteTransaction {}
 /// real Solana program instructions to be processed on-chain.
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct BlsVoteTransaction {
+    pub pubkey: Pubkey,
```
wen-coding: Is this ed25519 pubkey or BLS pubkey?

core/src/alpenglow_consensus/transaction.rs
```diff
@@ -14,6 +14,7 @@ impl AlpenglowVoteTransaction for BlsVoteTransaction {}
 /// real Solana program instructions to be processed on-chain.
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
 pub struct BlsVoteTransaction {
```
wen-coding: Does it make sense to move BlsVoteTransaction into its own file?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -4,18 +4,21 @@ use {
...
 pub trait VoteCertificate: Default {
     type VoteTransaction: AlpenglowVoteTransaction;
 
-    fn new(stake: Stake, transactions: Vec<Arc<Self::VoteTransaction>>) -> Self;
-    fn size(&self) -> Option<usize>;
-    fn transactions(&self) -> Vec<Arc<Self::VoteTransaction>>;
+    fn new(
+        stake: Stake,
+        transactions: Vec<Arc<Self::VoteTransaction>>,
+        transactions: &HashMap<BlsPubkey, usize>,
```
wen-coding: Why do we have two `transactions` here?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -31,21 +34,21 @@ pub struct LegacyVoteCertificate {
...
+        _validator_bls_pubkey_map: &HashMap<BlsPubkey, usize>,
+    ) -> Self {
         Self {
             stake,
             transactions,
         }
     }
 
-    fn size(&self) -> Option<usize> {
+    fn vote_count(&self) -> Option<usize> {
```
wen-coding, line 59: I do like the new name.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -54,62 +57,73 @@ impl VoteCertificate for LegacyVoteCertificate {
...
-        unimplemented!()
-    }
-
-    fn size(&self) -> Option<usize> {
-        unimplemented!()
+    fn new(
+        stake: Stake,
+        transactions: Vec<Arc<BlsVoteTransaction>>,
+        validator_bls_pubkey_map: &HashMap<BlsPubkey, usize>,
+    ) -> Self {
```
wen-coding: We can change new() to return Result maybe?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -54,62 +57,73 @@ impl VoteCertificate for LegacyVoteCertificate {
...
 
         // TODO: signature aggregation can be done out-of-order;
         // consider aggregating signatures separately in parallel
-        for (pubkey, transaction) in pubkey_transactions {
+        for transaction in transactions {
+            // aggregate the pubkey
+            let bls_pubkey: PubkeyProjective = transaction
+                .pubkey
+                .try_into()
+                .map_err(|_| BlsCertificateError::InvalidPubkey)?;
```
wen-coding: Do we need to log error or put the actual pubkey into the Error message?

Also, does this mean one failed transaction can make us not able to continue processing other transaction any more?

We should definitely verify the transaction long before this step, because if we have anything failing here then the stake calculation of the CertificatePool is broken. If we check before this step we can just panic here maybe?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -118,27 +132,38 @@ impl BlsCertificate {
...
         transaction: &BlsVoteTransaction,
     ) -> Result<(), BlsCertificateError> {
+        let aggregate_pubkey: PubkeyProjective = self
+            .aggregate_pubkey
+            .try_into()
+            .map_err(|_| BlsCertificateError::InvalidPubkey)?;
+        let new_pubkey: PubkeyProjective = transaction
+            .pubkey
+            .try_into()
+            .map_err(|_| BlsCertificateError::InvalidPubkey)?;
```
wen-coding: Maybe we can share the code between `new` and `add`?

core/src/alpenglow_consensus/certificate_pool.rs
samkim-crypto: I think it would work, but we would need to sort the pubkeys (by stake weight?) and build the hashmap every time we construct a certificate or add a vote to it. So it would be best to construct the HashMap once and re-use it. I can take a stab at adding that logic in a follow-up PR.

core/src/alpenglow_consensus/transaction.rs
samkim-crypto: Sorry I missed writing the docs here. I added the docs and clarified that this is a bls pubkey.

core/src/alpenglow_consensus/transaction.rs
samkim-crypto: Yes, done!

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Sorry typo. Fixed!

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Yesh I changed `BlsCertificateError` to `CertificateError` in [239885e](https://github.com/anza-xyz/alpenglow/pull/172/commits/239885e4dffbfa6f7f8576f602eeb79bc1cc24bf). But to do this properly I needed to make non-trivial changes to all the caller functions like `update_certificate`, so I unwrapped it there for now. I'll think about proper propagation of the error in a follow-up.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Yeah we would only want to add a vote to the certificate if the signature is already valid. The signature verification logic should happen as soon as a vote is received. If the signature is valid, then the bls pubkey or the signature will never fail, so it is actually safe to unwrap here.

Let me figure out how to best handle this error as I think about https://github.com/anza-xyz/alpenglow/pull/172#discussion_r2065375548 in a follow-up.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Yeah this is a good point. Right now the Bls pubkey and signature verification is quite clunky, so I will need to clean that up. Let me address this refactoring when I update the final interface for Bls 🙏 .

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: We can construct and sort the list once for every epoch inside `update_epoch_stakes_map`, then you only need to do Arc::clone here.

PR #174, opened 2025-04-30
[core] Extend certificate pool tests for BLS certificates

samkim-crypto:
#### Summary of Changes
I extended the tests in `vote_pool` and `certificate_pool` to work with BLS certificates.

In the process, I realized that the BLS default (zeroed) pubkey and signatures are invalid, so I removed `Default` for `BlsTransaction` and added `new_for_test` constructor.

Also, I updated the `validator_bls_pubkey_map` to be an option parameter to the certificate constructors because providing a default version of this map makes the tests fail.

Finally, this is somewhat unrelated to tests, but I also added a function to aggregate BLS pubkeys according to a bit-vector for future use.

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -233,7 +233,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                 // TODO: remove unwrap and properly handle unwrap
                 self.certificates.insert(
                     (slot, cert_type),
-                    VC::new(accumulated_stake, transactions, &HashMap::default()).unwrap(),
+                    VC::new(accumulated_stake, transactions, None).unwrap(),
                 );
                 self.set_highest_slot(cert_type, slot)
                     .then_some((cert_type, slot))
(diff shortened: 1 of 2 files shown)


core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -192,4 +196,28 @@ impl BlsCertificate {
...
+    /// aggregate BLS pubkey.
+    ///
+    /// TODO: Defining this to be a static function for now, but it more
+    /// naturally belongs to where we keep the sorted list of BLS pubkeys
+    pub fn aggregate_pubkey(
+        bit_vector: &BitVector,
+        validator_bls_pubkeys: &[&BlsPubkey],
+    ) -> Result<BlsPubkey, CertificateError> {
+        let mut aggregate_pubkey = PubkeyProjective::default();
+        for (i, pubkey) in validator_bls_pubkeys.iter().enumerate() {
```
wen-coding, line 210: nit: we aggregate certificates as soon as they are available, and we have long-tail stake distribution, so maybe it is more effective to iterate over bit_vector instead?

But I guess it doesn't make too much difference if we only have thousands of pubkeys.

samkim-crypto: Yeah that is a good point. I'll merge this PR for now since we don't actually know the precise location where this function will be called yet. I'll add this to the list of things to do in a follow-up.

PR #175, opened 2025-04-30
[bls] Add serialization for bls

samkim-crypto:
#### Problem
The bls types don't yet implement serialization and deserialization.

#### Summary of Changes
I first refactored the signature, pubkey, and proof of possession types out to a separate module. I then implemented `Pod` for these types and then `Deserialize, Serialize`.

Fixes #

### DIFF
--- a/bls/Cargo.toml
+++ b/bls/Cargo.toml
@@ -27,8 +27,8 @@ subtle = { workspace = true, optional = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
-solana-keypair = { workspace = true }
 bincode = { workspace = true }
+solana-keypair = { workspace = true }
 
 [features]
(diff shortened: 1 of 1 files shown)


bls/src/pod.rs
```diff
@@ -0,0 +1,129 @@
...
+pub struct ProofOfPossessionCompressed(
+    #[serde_as(as = "[_; BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE]")]
+    pub  [u8; BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE],
+);
+
+impl Default for ProofOfPossessionCompressed {
+    fn default() -> Self {
+        Self([0; BLS_PROOF_OF_POSSESSION_COMPRESSED_SIZE])
+    }
+}
```
wen-coding, line 60: Thanks for adding this, is it possible to add a test that serialize and deserialize to verify we get the same thing?

Issue #140, opened 2025-04-04
Aggregate certificates in tracker by block_id + bank hash


AshwinSekar:
Currently certificates are only denoted by slot #, we should check the hashes as well, as it becomes necessary for dump & repair.

PR #177, opened 2025-04-30
prepare vote history and cert pool for voting loop and repair changes

AshwinSekar:
Adds and fixes existing api for use in voting loop and certificate broadcast/repair:
- Updates `VoteHistory` to track the state elements from the April 18th paper
- Adds a general `add_vote` to `VoteHistory` for use in the voting loop
- Removes `my_pubkey` related tracking in cert pool / vote pool
- Adds `VoteHistory` input into `safe_to_notar` and `safe_to_skip`
- Modifies `safe_to_notar` api to actually return the block that is safe to notarize fallback
- Tracks completed certificates by appropriate hashes #140 for future use in multiple notarization block portion of voting loop
- Tracks highest notarized fallback certificate slot for use in `branch_certified` in updated voting loop

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -153,8 +153,9 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
     /// of the related certificates are newly complete.
     /// For each newly constructed certificate
     /// - Insert it into `self.certificates`
+    /// - Potentially update `self.highest_notarized_fallback`,
     /// - Potentially update `self.highest_finalized_slot`,
-    /// - If this the new highest finalized slot, return the certificate ID
+    /// - If we have a new highest finalized slot, return it
     fn update_certificates(
         &mut self,
         vote: &Vote,
@@ -205,18 +206,14 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                 if cert_id.is_notarize_fallback()
                     && self
                         .highest_notarized_fallback
-                        .map(|(s, _, _)| s < slot)
-                        .unwrap_or(true)
+                        .map_or(true, |(s, _, _)| s < slot)
                 {
                     self.highest_notarized_fallback =
                         Some((slot, block_id.unwrap(), bank_hash.unwrap()));
                 }
 
                 if cert_id.is_finalization_variant()
-                    && self
-                        .highest_finalized_slot
-                        .map(|s| s < slot)
-                        .unwrap_or(true)
+                    && self.highest_finalized_slot.map_or(true, |s| s < slot)
                 {
                     self.highest_finalized_slot = Some(slot);
                     Some(slot)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -63,40 +61,16 @@ pub enum AddVoteError {
...
-}
-
 #[derive(Default)]
 pub struct CertificatePool<VC: VoteCertificate> {
     // Vote pools to do bean counting for votes.
     vote_pools: BTreeMap<PoolId, VotePool<VC>>,
-    // Certificate pools to keep track of the certs.
-    certificates: BTreeMap<CertificateId, VC>,
+    /// Completed certificates
+    completed_certificates: BTreeMap<CertificateId, VC>,
```
AshwinSekar, line 69: eventually I will split out just this field into a thread safe struct for use by both the block creation loop, voting loop and repair

core/src/alpenglow_consensus.rs
```diff
@@ -13,17 +18,30 @@ pub type Stake = u64;
...
-    Finalize,
-    FinalizeFast,
-    Notarize,
-    NotarizeFallback,
-    Skip,
+pub enum CertificateId {
+    Finalize(Slot),
+    FinalizeFast(Slot, Hash, Hash),
+    Notarize(Slot, Hash, Hash),
+    NotarizeFallback(Slot, Hash, Hash),
```
AshwinSekar, line 27: This is because we can have multiple `NotarizeFallback` certs per block, figured it's easier to just move the identifiers into the enum itself

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +294,171 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-            .contains_key(&(slot, CertificateType::Skip))
+    /// Checks if the any block in slot `slot` has received a `NotarizeFallback` certificate, if so return
+    /// the size of the certificate
+    pub fn slot_notarized_fallback(&self, slot: Slot) -> Option<usize> {
+        // TODO(ashwin): when updating voting loop for duplicate blocks, add parent tracker to make this
+        // a true "branchCertified". For now this sufficies as branchCertified in the sequential loop
+        self.completed_certificates
+            .iter()
+            .find_map(|(cert_id, cert)| {
+                matches!(cert_id, CertificateId::NotarizeFallback(s,_,_) if *s == slot)
```
AshwinSekar, line 384: rn we don't care about there being multiple `NotarizeFallback` but in the future this will become relevant for figuring out `branchCertified` when there are duplicate blocks (and for dump & repair)

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -235,14 +196,35 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-                    (slot, cert_type),
+                self.completed_certificates.insert(
+                    cert_id,
                     VC::new(accumulated_stake, transactions, &HashMap::default()).unwrap(),
                 );
-                self.set_highest_slot(cert_type, slot)
-                    .then_some((cert_type, slot))
+                // TODO(ashwin): Send to blockstore so that repair can serve. Also broadcast to other nodes
+
+                if cert_id.is_notarize_fallback()
```
wen-coding, line 206: Hmm, I guess you are special casing notarize_fallback because Apr 18 draft says BranchCertified(hash(b)) depends on notar-fallback certificate, but they are also saying notarization or finalization cert implies notar-fallack. So finalization > notarize > notar-fallback. For BranchCertified we just need one of the 3. Is there any other reason you are specializing notarize_fallback here?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -235,14 +196,35 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
                 // TODO: remove unwrap and properly handle unwrap
-                self.certificates.insert(
-                    (slot, cert_type),
+                self.completed_certificates.insert(
+                    cert_id,
                     VC::new(accumulated_stake, transactions, &HashMap::default()).unwrap(),
                 );
-                self.set_highest_slot(cert_type, slot)
-                    .then_some((cert_type, slot))
+                // TODO(ashwin): Send to blockstore so that repair can serve. Also broadcast to other nodes
```
wen-coding, line 204: Just curious, why are we saving certs in blockstore? Compared to saving on local disk maybe?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-        let slot = bank.slot();
-        let Some(epoch_stakes) = self.epoch_stakes_map.get(&bank.epoch()) else {
-            return false;
-        };
+    /// Checks if we have voted to skip `slot` or notarize some block `b' = (block_id', bank_hash')` in `slot`
+    /// Additionally check that for some differernt block `b = (block_id, bank_hash)` in `slot` either:
+    /// (i) At least 40% of stake has voted to notarize `b`
+    /// (ii) At least 20% of stake voted to notarize `b` and at least 60% of stake voted to either notarize `b` or skip `slot`
+    /// If all the above hold, return `Some(block_id, bank_hash)` for the `b`
+    pub fn safe_to_notar(&self, slot: Slot, vote_history: &VoteHistory) -> Option<(Hash, Hash)> {
```
wen-coding, line 400: The safe_to_notar() in the whitepaper draft comes with bank_hash and block_id: `SafeToNotar(slot(b), hash(b))`, why we don't have hash(b) in the argument any more?

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: why reinvent the wheel when we already have a persistent key value store

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: finalized cert doesn't have the `block_id` or `bank_hash` so `maybe_start_leader` doesn't know which duplicate block its supposed to build on. And dump & repair doesn't have enough info to proceed. 
No reason to check `notarize` since `notarize_fallback` is a superset of the same votes.

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: well where am I supposed to magically get `hash(b)` from? It makes more sense as a return type

the other option is for the `voting_loop` to iterate over all votes to figure out possibilities for `hash(b)` but I don't want to expose `vote_pool` internals to the outside

wen-coding: We can confirm with the researchers, but I believe you call `safe_to_notar` only if you have a replayed block b on hand. If you have that block b replayed then you can get `hash(b)` from the block. I don't think it's right to send either `Notar` or `NotarFallback` when you don't actually have the block replayed on hand. Because the replay itself does sanity-checking, you shouldn't just copy other's answers.

AshwinSekar: even if that's the case, how do I figure out which block to fetch for replay? How do you envision us finding `hash(b)`?

wen-coding: That's a good question, but a separate question than `safe_to_notar`. Are you asking about:
1. a validator just came online and it needs to find out which block to fetch:
    Since >60% agreed on all notarized block, random sampling should give you a good answer with high probability
2. a validator happened to have a minor version:
    In this case you should have a cert, I'm guessing we don't need to repair until there is a cert, because if the cluster can't decide on a specific block in slot s, then they will do skip and move on. With a BLS cert on hand you know exactly who to repair from
Or are you asking about something else?

AshwinSekar: The scenario is:
- We vote notarize for b' in slot s
- We are stuck waiting for SkipCertified(s) or BranchCertified(s, hash(b)) there is NO certificate for any block in s 
- safeToSkip(s) is false
What do you do? Try to dump & repair every combination of `Hash` for `b`? Where does the block magically come from? 
Screenshot 2025-05-01 at 11 49 23 AM

wen-coding: Hmm, okay, you are using v0 ref paper, the April 18 draft is defining `safeToNotar` differently. We can ask @kobi-lizard why the definition of `safeToNotar` changed in between. Since we still don't have the finalized white paper by now, I agree it's indeed better to use the v0 ref paper which is finalized. So my previous data structure all need to change. Go ahead and make changes, I'll optimize after you are done, so I don't block you. I think there might be better ways to organize it than repeatedly going through all `votes` on every `safeToNotar`.

wen-coding: I agree your implementation conforms with v0 ref paper, however I'm suspicious why the behavior described in v0 ref paper is correct:
Let's say we have the same block `b` but somehow my code calculates `hash1` while you calculate `hash2`, but I don't have the block `b` on hand yet, so I saw 40% voted `hash2` and I sent out `NotarizeFallback(slot(b), hash2)`, while I can never replay `b` to get `hash2` no matter how I dump and replay, how can I proceed going forward?

wen-coding: This is a possible scenario if Jump has 40% and we have 60%, then somehow our code calculated something in the block slower than Jump and reached a different conclusion, but we already sent NotarFallback on the bankhash Jump agreed on, then there is no going back.

core/src/alpenglow_consensus.rs
wen-coding: Wait, how can we have multiple `NotarizeFallback` certs per block? And they notarize different `Hash`?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -63,40 +61,18 @@ pub enum AddVoteError {
...
     vote_pools: BTreeMap<PoolId, VotePool<VC>>,
-    // Certificate pools to keep track of the certs.
-    certificates: BTreeMap<CertificateId, VC>,
+    /// Completed certificates
+    completed_certificates: BTreeMap<CertificateId, VC>,
     // Lookup table for checking conflicting vote types.
     // conflicting_vote_types: HashMap<VoteType, Vec<VoteType>>,
-    // Highest slot with each certificate
-    highest_slot_map: HashMap<CertificateType, Slot>,
+    /// Highest block that has a NotarizeFallback certificate, for use in producing our leader window
```
wen-coding, line 72: nit: We can add comment here NotarizeFallback implied Notarize, so what we really mean here is notarized. In the future they can invent new Notarize types, we just need to know they all mean Notarized maybe?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -192,25 +153,25 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
     /// of the related certificates are newly complete.
     /// For each newly constructed certificate
     /// - Insert it into `self.certificates`
-    /// - Potentially update the new highest certificate slot
-    /// - If this the new highest certificate, return the type and slot
+    /// - Potentially update `self.highest_finalized_slot`,
+    /// - If this the new highest finalized slot, return the certificate ID
```
wen-coding: nit: your return type is a Slot.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Got it, so what we really meant here is it got either Notarize or NotarizeFallback cert.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -235,14 +196,35 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
                 );
-                self.set_highest_slot(cert_type, slot)
-                    .then_some((cert_type, slot))
+                // TODO(ashwin): Send to blockstore so that repair can serve. Also broadcast to other nodes
+
+                if cert_id.is_notarize_fallback()
+                    && self
+                        .highest_notarized_fallback
+                        .map(|(s, _, _)| s < slot)
+                        .unwrap_or(true)
```
wen-coding: nit: .map_or(true, |(s, _, _)| s < slot) ?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+        // make this the true "highest branchCertified block". For now this sufficies.
+        self.highest_notarized_fallback
     }
 
-    pub fn highest_notarized_slot(&self) -> Slot {
+    #[cfg(test)]
+    fn highest_notarized_slot(&self) -> Slot {
         // Return the max of CertificateType::Notarize and CertificateType::NotarizeFallback
-        [CertificateType::Notarize, CertificateType::NotarizeFallback]
+        self.completed_certificates
```
wen-coding, line 320: nit: can we just use highest_notarized_fallback above? (Still feel highest_notarized is less confusing but I can accept this name)

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Hmm, I believe I read somewhere that there could only be 1 notarized block in each slot.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+                    .then_some(cert.vote_count())
+            })
     }
 
-    pub fn skip_certified(&self, slot: Slot) -> bool {
-        self.certificates
-            .contains_key(&(slot, CertificateType::Skip))
+    /// Checks if the any block in slot `slot` has received a `NotarizeFallback` certificate, if so return
+    /// the size of the certificate
+    pub fn slot_notarized_fallback(&self, slot: Slot) -> Option<usize> {
```
wen-coding, line 378: nit: could the name somehow implies the return type is certificate size?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-            self.vote_pools
-                .get(&(slot, VoteType::Notarize))
-                .map_or(0, |pool| pool.total_stake_by_key(bank_hash, block_id)) as f64
-                / total_stake as f64;
-        if notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY {
-            return true;
+        if !vote_history.voted(slot) {
+            return None;
         }
+        let b_prime = vote_history.voted_notar(slot);
```
wen-coding, line 409: nit: We could also check these in the caller and pass the result in, so that certificate_pool doesn't need to know about vote_history?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-        if slot < self.root {
-            return false;
+        let notarize_pool = self.vote_pools.get(&(slot, VoteType::Notarize))?;
+        for (
+            VoteKey {
+                bank_hash,
+                block_id,
+            },
+            votes,
+        ) in notarize_pool.votes.iter()
```
wen-coding, line 424: Sorry about this mess, this is me not strictly following v0, lessons learned.
We can check this in and I can think about better ways to organize vote_pool so certificate_pool doesn't need to look deep into vote_pool every time.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -325,159 +307,167 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+    /// - votedStake(s) is the cumulative stake of all nodes who voted notarize or skip on s
+    /// - topNotarStake(s) the highest of cumulative notarize stake per block in s
+    pub fn safe_to_skip(&self, slot: Slot, vote_history: &VoteHistory) -> bool {
         let epoch = self.epoch_schedule.get_epoch(slot);
         let Some(epoch_stakes) = self.epoch_stakes_map.get(&epoch) else {
             return false;
         };
         let total_stake = epoch_stakes.total_stake();
 
+        if vote_history.voted_notar(slot).is_none() {
```
wen-coding, line 461: nit: if we only need this bool then we can check in the caller, so our data structures are less tangled together.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -498,8 +488,8 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         let needs_notarization_certificate = parent_slot >= first_alpenglow_slot && parent_slot > 1;
 
         if needs_notarization_certificate
-            && self.get_notarization_cert_size(parent_slot).is_none()
-            && self.get_finalization_cert_size(parent_slot).is_none()
+            && self.block_notarized(parent_slot).is_none()
```
wen-coding, line 488: I know you inherited this, but do we need to check notarized bankhash actually agrees with our parent?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -527,13 +517,18 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
 
-    /// Cleanup old finalized slots from the certificate pool
+    /// Cleanup any old slots from the certificate pool
     pub fn handle_new_root(&mut self, bank: Arc<Bank>) {
         self.root = bank.slot();
-        // `certificates`` now only contains entries >= `finalized_slot`
-        self.certificates = self
-            .certificates
-            .split_off(&(bank.slot(), CertificateType::Finalize));
+        // `completed_certificates`` now only contains entries >= `slot`
```
wen-coding, line 520: nit: this comment looks wrong, should roughly be: // get rid of entries < slot in completed_certificates
?

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -39,59 +40,118 @@ impl VoteHistoryVersions {
...
 
 impl VoteHistory {
     pub fn new(node_pubkey: Pubkey, root: Slot) -> Self {
         Self {
             node_pubkey,
-            latest_notarize_vote: Vote::from(NotarizationVote::default()),
-            skip_votes: vec![],
-            latest_finalize_vote: Vote::from(FinalizationVote::default()),
             root,
+            ..Self::default()
```
wen-coding, line 80: Hmm, I know we save it to local file. But if someone sets hot standby the file may be stale. I really do wish we implement "find out my own vote from the repaired cert pool if I'm starting up" like TowerBFT.

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: NotarizeFallback does not imply Notarize

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: this will be handled when I actually incorporate  duplicate blocks in the voting loop, rn everything is just based on slot

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: April 18th draft also returns the hash:
Screenshot 2025-05-01 at 8 16 28 PM
Screenshot 2025-05-01 at 8 17 05 PM

do you understand my point though? you can't magically come up with a hash, it needs to be scrapped from the votes in the pool.

Regardless of if we go with the April 18th version that does not replay the new block before casting the vote, or the most up to date version which does replay the new block, we still need the hash in order to figure out which block to fetch and which block to cast NotarizeFallback for

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: I meant Notarize in the general case, okay, bad word choice. I should say: Notarize implies NotarizeFallback.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: If we decided to follow v0 ref paper that's fine, we can ignore what April 18 version is doing.

You are using `SafeToNotar` to decide both how to do repair and whether to cast NotarizeFallback, we don't have to do that. If we only use `SafeToNotar` to decide whether to cast `NotarizeFallback`, we are trying to notarize the block `b` we have on hand, no? Then there is nothing magical, the `hash(b)` comes from replayed block b on my hand. I'm suspicious we should cast `NotarizeFallback` just because 40% people said so.

It's a different story if you are trying to decide how to do repair. Do we have a write up for that already? I don't see it either in v0 or in April 18 version. I don't understand why we try to reuse `SafeToNotar`, we could use another function to decide that.

core/src/alpenglow_consensus.rs
AshwinSekar: @kobi-lizard  would know best, but one possible scenario is something like this:
- 40% of nodes (group A) vote notarize for b1
- 40% of nodes (group B) vote notarize for b2
- 20% of nodes (group C) vote notarize for b3
- A malicious 20% of nodes (group M) from group A or B double vote notarize for b3

Then
- Group A's votes for b1 are only observed by Group B
- Group B's votes for b2 are only observed by Group C
- Group C's votes + Group M's votes for b3  are only observed by Group A

Neither SkipCertified or BranchCertified is true so due to SafeToNotar:
- Group A sees 40% votes for b3 they cast a notarize fallback 40% for b3
- Group B sees 40% votes for b1 they cast a notarize fallback 40% for b1
- Group C sees 40% votes for b2 they cast a notarize fallback 20% for b2

Finally:
b1 has 40% notarize (group A) + 40% notarize fallback (group B) >= 60%
b2 has 40% notarize (group B) + 20% notarize fallback (group C) >= 60%
b3 has 40% notarize (group C) + 40% notarize fallback (group A) >= 60%

So we have 3 notarize fallback certificates in the same slot

wen-coding: Err, that sounds like a failure of the protocol, we can't uniquely notarize a block any more. Does that mean we should trust notarize fallback cert less?

However, this does assume the attacker can precisely control the network partition right? @kobi-lizard how do we plan to proceed in this case?

core/src/alpenglow_consensus/certificate_pool.rs
steviez: Got a ping about this piece, so a couple questions. I'll probably be able to answer the second and third questions after the first one is answered, but I'll post them anyways:
- Can you share a link to the type that will be written to the Blockstore ?
- How large (bytes) will the type be ?
- Is the size of the type variable ?
- Will the type be written to the Blockstore once per slot OR is it something similar to `SlotMeta` where we'll read/modify/write the type (which involves (de)serialization) multiple times per slot ?

AshwinSekar: I'll add you to the blockstore change so that it's makes more sense, this is just a TODO comment

core/src/alpenglow_consensus.rs
AshwinSekar: Why is it a failure of the protocol? I think you're conflating notarize and notarize fallback, at most one block can be notarized in a given slot (lemma 25), and the safety properties still hold if multiple blocks are notarized fallback in a slot

Issue #39, opened 2025-02-11
Save certificates in storage


carllin:
#### Proposed Solution
 saving latest finalization/skip/notarization certificates is a good idea

PR #178, opened 2025-05-01
blockstore: store critical certs to enable certificate repair

AshwinSekar:
#### Problem
Need to be able to serve repair and re initialize certificate pool on restart

#### Summary of Changes
Completed certificates that are critical (NotarizeFallback + Skip) are sent from cert pool to certificate service
certificate service inserts these in blockstore

#### Further TODO:
- #180 
- #181 
- #182 
- #148

Fixes https://github.com/anza-xyz/alpenglow/issues/39

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -282,13 +282,17 @@ impl VotingLoop {
 
                     // We use a blocking receive if skipped is true,
                     // as we can only progress on a certificate
-                    Self::ingest_votes_into_certificate_pool(
+                    if let Err(e) = Self::ingest_votes_into_certificate_pool(
                         &my_pubkey,
                         &shared_context.vote_receiver,
                         /* block */ skipped,
                         &mut cert_pool,
                         &voting_context.commitment_sender,
-                    );
+                    ) {
+                        error!("{my_pubkey}: error ingesting votes into certificate pool, exiting: {e:?}");
+                        // Finalizer will set exit flag
+                        return;
+                    }
 
                     if !skipped && skip_timer.elapsed().as_millis() > timeout.as_millis() {
                         skipped = true;
@@ -785,7 +789,7 @@ impl VotingLoop {
         block: bool,
         cert_pool: &mut CertificatePool<LegacyVoteCertificate>,
         commitment_sender: &Sender<CommitmentAggregationData>,
-    ) {
+    ) -> Result<(), AddVoteError> {
         let add_to_cert_pool =
             |(vote, vote_account_pubkey, tx): (Vote, Pubkey, VersionedTransaction)| {
                 match Self::add_vote_and_maybe_update_commitment(
@@ -796,27 +800,26 @@ impl VotingLoop {
                     cert_pool,
                     commitment_sender,
                 ) {
-                    Err(AddVoteError::CertificateSenderError) => panic!(
-                    "{my_pubkey}: Something has gone wrong with the certificate_service, exiting"
-                ),
+                    err @ Err(AddVoteError::CertificateSenderError) => err,
                     Err(e) => {
                         // TODO(ashwin): increment metrics on non duplicate failures
                         trace!("{my_pubkey}: unable to push vote into the pool {}", e);
+                        Ok(())
                     }
-                    Ok(()) => (),
+                    Ok(()) => Ok(()),
                 }
             };
 
         if block {
             let Ok(first) = vote_receiver.recv_timeout(Duration::from_secs(1)) else {
                 // Either timeout or sender disconnected, return so we can check exit
-                return;
+                return Ok(());
             };
             std::iter::once(first)
                 .chain(vote_receiver.try_iter())
-                .for_each(add_to_cert_pool)
+                .try_for_each(add_to_cert_pool)
         } else {
-            vote_receiver.try_iter().for_each(add_to_cert_pool)
+            vote_receiver.try_iter().try_for_each(add_to_cert_pool)
         }
     }
 


core/src/replay_stage.rs
```diff
@@ -287,6 +291,7 @@ pub struct ReplaySenders {
     pub block_metadata_notifier: Option<BlockMetadataNotifierArc>,
     pub dumped_slots_sender: Sender<Vec<(u64, Hash)>>,
     pub alpenglow_vote_sender: AlpenglowVoteSender,
+    pub certificate_sender: Sender<(CertificateId, LegacyVoteCertificate)>,
```
AshwinSekar, line 297: We will update this to the BLS cert once ready

ledger/src/blockstore_meta.rs
```diff
@@ -872,6 +873,44 @@ impl OptimisticSlotMetaVersioned {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+/// Holds the certificates for this slot in blockstore
```
AshwinSekar, line 877: @steviez added some color here, let me know if this answers your questions or you see any issues

carllin: Each slot can potentially hold up to 6 certificates, each about 1k in size?

core/src/alpenglow_consensus.rs
```diff
@@ -46,6 +46,10 @@ impl CertificateId {
     pub(crate) fn is_notarize_fallback(&self) -> bool {
         matches!(self, Self::NotarizeFallback(_, _, _))
     }
+
+    pub(crate) fn is_critical(&self) -> bool {
+        matches!(self, Self::NotarizeFallback(_, _, _) | Self::Skip(_))
+    }
```
carllin, line 59: Maybe a quick comment to define critical.

For instance, why is "Notarize" not considered critical?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -197,11 +208,16 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
-                    VC::new(accumulated_stake, transactions, None).unwrap(),
-                );
-                // TODO(ashwin): Send to blockstore so that repair can serve. Also broadcast to other nodes
+                let vote_certificate = VC::new(accumulated_stake, transactions, None).unwrap();
+                self.completed_certificates
+                    .insert(cert_id, vote_certificate.clone());
+                if let Some(sender) = &self.certificate_sender {
+                    if cert_id.is_critical() {
+                        if let Err(e) = sender.try_send((cert_id, vote_certificate)) {
+                            error!("Unable to send certificate {cert_id:?}: {e:?}");
```
carllin, line 220: Inability to send the cert probably means the cert storage thread panicked. 

Seems like a critical enough error where it might be worth returning an error here and having the voting loop exit?

core/src/alpenglow_consensus.rs
AshwinSekar: added a comment here [6ca0f5a](https://github.com/anza-xyz/alpenglow/pull/178/commits/6ca0f5a8e8b3b51bb9584def1eeba3da264d0365)

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: good idea, i passed it up and panicked in the voting loop [316694a](https://github.com/anza-xyz/alpenglow/pull/178/commits/316694a0d4ff97c06790e179eea69bf72beb9e71)

ledger/src/blockstore_meta.rs
AshwinSekar: it's only 4 certificates, I haven't stored any Finalization or Notarization certificates (just the 3 NotarizeFallback + 1 Skip). And according to Sam, it's roughly 800 bytes per certificate, should be reflected in the comment

core/src/repair/certificate_service.rs
```diff
@@ -0,0 +1,131 @@
...
+            .name("solCertInsertBCast".to_string())
+            .spawn(move || {
+                while !exit.load(Ordering::Relaxed) {
+                    let certs = match Self::receive_new_certificates(&certificate_receiver) {
+                        Ok(certs) => certs,
+                        Err(e) if Self::should_exit_on_error(&e, &handle_error) => break,
+                        Err(_e) => continue,
+                    };
+
+                    // TODO: broadcast to gossip / all-2-all
```
AshwinSekar, line 58: @wen-coding here we can send the new cert to all-to-all as part of #181

core/src/alpenglow_consensus/certificate_pool.rs
carllin: oh wait, maybe not good to panic, since out of order shut down of various services on detecting exit flag could cause panic

Maybe just error log, return, and set exit flag

AshwinSekar: I think it should be fine since you implemented this 
https://github.com/anza-xyz/alpenglow/blob/ced6af51de9a5d63e56482421263ec6698a178e7/core/src/replay_stage.rs#L143-L145

https://github.com/anza-xyz/alpenglow/blob/ced6af51de9a5d63e56482421263ec6698a178e7/core/src/alpenglow_consensus/voting_loop.rs#L173

So when voting loop panics it will automatically set the exit flag?

carllin: oh i meant on normal validator shut down, say for instance when we run local cluster, we might get a panic, which is bad

AshwinSekar: ah yeah makes sense [1eaf252](https://github.com/anza-xyz/alpenglow/pull/178/commits/1eaf25255e67f348c0d519e7ba4439bbe9b2a476)

we should definitely audit all our errors and panics later when we merge to master

carllin: yeah I thnk this is fine now because if a panic is the source of the CertififcateService exiting, we will bubble up the panic in the thread join in WindowService shutdown

`self.certificate_service.join()`

PR #185, opened 2025-05-05
Give BLS certificate a fake sorted validator list for now.

wen-coding:
In the future we will have real bls_pubkey in the vote account, put the sorting logic there for now with a fake list so we can continue implementing verification logic.

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -83,7 +83,8 @@ pub struct CertificatePool<VC: VoteCertificate> {
     // Cached epoch_stakes_map
     epoch_stakes_map: Arc<HashMap<Epoch, EpochStakes>>,
     // Calculate a stake weighted list of pubkey and BLS pubkey based on epoch_stakes_map
-    bls_key_to_stake_map: HashMap<Epoch, HashMap<BLSPubkey, usize>>,
+    // This maps BLS pubkey to its unique rank.
+    bls_key_to_rank_map: HashMap<Epoch, HashMap<BLSPubkey, usize>>,
     // The current root, no need to save anything before this slot.
     root: Slot,
     // The epoch of current root.
@@ -126,9 +127,9 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         self.root
     }
 
-    fn update_bls_key_to_stake_map(&mut self, new_root_epoch: Epoch) {
+    fn update_bls_key_to_rank_map(&mut self, new_root_epoch: Epoch) {
         for (epoch, stakes) in self.epoch_stakes_map.iter() {
-            if self.bls_key_to_stake_map.contains_key(epoch) || epoch < &new_root_epoch {
+            if self.bls_key_to_rank_map.contains_key(epoch) || epoch < &new_root_epoch {
                 continue;
             }
             let stakes_data = stakes.stakes().vote_accounts().staked_nodes();
@@ -147,9 +148,9 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                 .enumerate()
                 .map(|(index, _)| (BLSPubkey::default(), index))
                 .collect();
-            self.bls_key_to_stake_map.insert(*epoch, validator_map);
+            self.bls_key_to_rank_map.insert(*epoch, validator_map);
         }
-        self.bls_key_to_stake_map
+        self.bls_key_to_rank_map
             .retain(|epoch, _| *epoch >= new_root_epoch);
     }
 
@@ -161,7 +162,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         epoch: Epoch,
         validator_bls_pubkey_map: HashMap<BLSPubkey, usize>,
     ) {
-        self.bls_key_to_stake_map
+        self.bls_key_to_rank_map
             .insert(epoch, validator_bls_pubkey_map);
     }
 
@@ -169,7 +170,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         let epoch = bank.epoch();
         if self.epoch_stakes_map.is_empty() || epoch > self.root_epoch {
             self.epoch_stakes_map = Arc::new(bank.epoch_stakes_map().clone());
-            self.update_bls_key_to_stake_map(epoch);
+            self.update_bls_key_to_rank_map(epoch);
             self.root_epoch = epoch;
             self.epoch_schedule = bank.epoch_schedule().clone();
         }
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -77,6 +78,8 @@ pub struct CertificatePool<VC: VoteCertificate> {
     epoch_schedule: EpochSchedule,
     // Cached epoch_stakes_map
     epoch_stakes_map: Arc<HashMap<Epoch, EpochStakes>>,
+    // Calculate a stake weighted list of pubkey and BLS pubkey based on epoch_stakes_map
+    epoch_validators_map: HashMap<Epoch, HashMap<BLSPubkey, usize>>,
```
carllin: Seems like this should actually be constructed in bank.rs, and treated as an equivalent to the existing epoch stakes/be an extra field in `EpochStakes`

Will a lot of components be using BLSPubkeys instead of the existing vote pubkeys post migration? What's the current design for how we're mapping existing vote keys to the new BLS keys?

wen-coding: Nope, most of the code will stick with ed25519 pubkey, because it's too much work to change that. This is only used by BLS aggregation in certificates and later BLS verification. Even BLS signing of single votes doesn't need this bitmap.

carllin: do we need a separate map? can we just lookup in the epoch stakes by the vote key and convert to the bls key

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -81,6 +82,8 @@ pub struct CertificatePool<VC: VoteCertificate> {
...
```
carllin: nit: rename to bls_key_to_stake

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -123,10 +126,49 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+                    b_pubkey.cmp(a_pubkey)
+                } else {
+                    b_stake.cmp(a_stake)
+                }
+            });
+            let mut validator_map = HashMap::new();
+            for (index, _) in pubkey_with_stake.iter().enumerate() {
+                // TODO(wen): replace this with the real BLS pubkey from vote account
+                validator_map.insert(BLSPubkey::default(), index);
+            }
```
carllin:
```
 let validator_map = pubkey_with_stake
            .into_iter()
            .enumerate()
            .map(|(index, _)| (BLSPubkey::default(), index))
            .collect();
```

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: I was thinking maybe sorted by the stake might be a better choice, so we can hopefully discard the long-tail zeroes in the bitmap. The bitmap occupies the most space in certs.

wen-coding: I think you are right, we eventually might need to move this list to bank, because we need it in sigverify. But maybe let's do that in the sigverify PR?

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/certificate_pool.rs
carllin: seems better to do it in this PR in isolation to make that PR smaller

wen-coding: Sure, what do you feel about sort by pubkey vs sort by stake?

carllin: confused by the purpose and implementation here. The ordering index is only in the value of the hashmap, so the hashmap is not sorted by key.  In order to find the largeset N stakes you would still have to traverse the entire hashmap.

How are you planning on using this hashmap?

wen-coding: When you compose the cert you have no choice, you just look up index on whatever BLS pubkey for the votes you have on hand. I'm just saying, if we sort this list by stake, then chances are we have a lot of zeros in the end of the vec, so we can maybe later compress the bitmap better. Because the higher staked validators are likely to get the block earlier and playing it earlier, so they will probably send the votes faster. Does that make sense?

samkim-crypto: > How are you planning on using this hashmap?

An example of how the hashmap is used is [here](https://github.com/anza-xyz/alpenglow/blob/master/core/src/alpenglow_consensus/vote_certificate.rs#L145). When the signatures are aggregated, we are going to iterate through a list of (BLS) vote transactions, look up the public key, and then record that a signature of the owner of that public key has been aggregated using a bitvector. Since we are using a bitvector, we need a sorted list of public keys that everyone can agree on.

Technically, it doesn't matter how the public keys are ordered. There just needs some standard order that everyone has access to. But @wen-coding's reasoning makes sense in that if we sort this by stake, then since higher stake voters' signatures are more likely to be included in certs, we can potentially optimize the size of the bitvector.

samkim-crypto: Maybe a sorted list of public keys is actually a misnomer. What we need is a map that assigns a unique index to each public key and this is what the implemented hash map gives us.

carllin: yup that makes sense, thanks!

core/src/alpenglow_consensus/certificate_pool.rs
carllin: my bad i misunderstood what this was, maybe bls_key_to_rank

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: @carllin Checked the code more, I think without putting it in VoteAccount it's too many hacks to put it in epoch_stakes.

So @samkim-crypto how soon do you need this validator list to make verification work? We have two options:
1. Put in this, then later when you add bls_pubkey to alpenglow-vote again after moving bls crate, we add it to vote account, then move this list to epoch_stakes
Meanwhile this means BLS verification and sigverify can happen while your bls crate is getting approved, and we would be constantly moving towards making all-to-all work for cert broadcast
2. If the bls crate is getting approved soon, we can wait on that and make this work after it's checked in.

samkim-crypto: I think bls crate should probably be approved EOD today Paris time. If it simplifies to have the pubkey checked into alpenglow-vote, then maybe this can wait a day?

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: renamed.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Sure, if it's just one day we can wait.

PR #186, opened 2025-05-06
local-cluster: ensure that all nodes issue notar/finalize when a single node has lots of stake

ksn6:
Introducing the ability to pause a validator before its voting loop commences. This makes it easy to create setups where validators all voting at/very close to the same time. It's a bit of a hack that's required for certain tests until we get certificate repair implemented.

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -92,7 +92,6 @@ use {
         fs,
         io::Read,
         iter,
-        net::UdpSocket,
         path::Path,
         sync::{
             atomic::{AtomicBool, AtomicUsize, Ordering},
@@ -6112,7 +6111,7 @@ fn test_alpenglow_node_all_nodes_notarize_and_finalize() {
 
     // Create leader schedule
     let (leader_schedule, validator_keys) =
-        create_custom_leader_schedule_with_random_keys(&vec![0, 0, 100]);
+        create_custom_leader_schedule_with_random_keys(&[0, 0, 100]);
     assert_eq!(num_nodes, validator_keys.len());
 
     let leader_schedule = FixedSchedule {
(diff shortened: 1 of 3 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -6095,3 +6096,185 @@ fn test_invalid_forks_persisted_on_restart() {
...
+    }
+
+    // Let B and C start issuing skips for a few slots before A starts voting
+    sleep(Duration::from_secs(2));
+
+    pen_voting_a.store(false, Ordering::Relaxed);
+
+    vote_observer.join().unwrap();
+
+    // Out of the last 10 slots, ensure that all nodes have voted notarize and finalize on at least
```
AshwinSekar, line 6256: this requirement is a bit strict and can cause this test to be flaky. Since A has significant stake, `greatest_slot_with_finalize_vote` can increase without votes from B or C. Essentially we are at the whim of the scheduler to give enough time for B and C's threads to cast votes.

It could be worse with the repair / fast forward logic. Since A can finalize slots by itself, if B and C don't vote in time (e.g. if A was the leader), they can potentially see the newer finalization certs and increment their current slots without casting votes in between.

AshwinSekar: what is the exact scenario we are trying to test here? Would it be sufficient to check if B and C make roots?

AshwinSekar: or is it necessary for A to have so much stake here? If not we could setup stakes such that it's required for certain nodes to vote in order to make progress and then check if we make enough roots within a certain time frame.

ksn6: gotcha - I'm trying to ensure that the cert repair case doesn't get triggered here; i.e., B or C aren't just repeatedly issuing Skip(3) over and over again.

what's the best way to test this?

AshwinSekar: something simple could be to have 4 nodes, with 25% each. Node D will start with voting disabled.

Then we just check if the cluster makes roots, because that can only happen if A B and C are voting notarize + final

PR #188, opened 2025-05-08
Add bls_pubkey to VoteAccount and EpochStakes

wen-coding:


vote/src/vote_account.rs
```diff
@@ -193,6 +194,17 @@ impl VoteAccount {
...
 
+    pub fn bls_pubkey(&self) -> Option<&BLSPubkey> {
+        match &self.0.vote_state_view {
+            VoteAccountState::TowerBFT(_) => None,
+            VoteAccountState::Alpenglow => Some(
+                AlpenglowVoteState::deserialize(self.0.account.data())
+                    .unwrap()
+                    .bls_pubkey(),
+            ),
+        }
```
carllin, line 205: is the plan every validator operator creates a separate alpenglow account with this BLS pubkey?

wen-coding: Yes, everyone needs an alpenglow vote account in the new vote program, and we made BLS non-optional, meaning you must supply it to create the new account.

runtime/src/epoch_stakes.rs
```diff
@@ -126,6 +136,30 @@ impl EpochStakes {
...
+            .filter_map(|(_, (stake, account))| {
+                if *stake > 0 {
+                    account.bls_pubkey().map(|bls_pubkey| (*bls_pubkey, *stake))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        pubkey_stake_pair_vec.sort_by(|(_a_pubkey, a_stake), (_b_pubkey, b_stake)| {
+            // TODO(wen): should compare pubkey if stake is equal
```
carllin: seems easy to do here

wen-coding: Finally used a hack (converted BLSPubkey to string for now) to make ABI test pass. Put the map in its own struct now so I can put this construction logic into the constructor.

runtime/Cargo.toml
```diff
@@ -48,6 +48,7 @@ serde_derive = { workspace = true }
 serde_json = { workspace = true }
 serde_with = { workspace = true }
 solana-accounts-db = { workspace = true }
+solana-bls = { workspace = true, features = ["serde"] }
```
samkim-crypto, line 51: I think we should add the `frozen-abi` feature here to get `./test-abi.sh` working.

runtime/src/bank/serde_snapshot.rs
```diff
@@ -568,7 +568,7 @@ mod tests {
         #[cfg_attr(
             feature = "frozen-abi",
             derive(AbiExample),
-            frozen_abi(digest = "3PsrjAtyWBU3KPopGoM1UK1sa8HjVzehjBi7M2v6wW1Q")
+            frozen_abi(digest = "G21Xw3ZG6Z2v6R9J2qbFGAkJP8SxmQjtyNKy2GoBc4nq")
```
samkim-crypto: and update the hash here after adding the `frozen-abi` feature above.

wen-coding: Done.

runtime/Cargo.toml
wen-coding: Woohoo, that worked, thanks.

wen-coding: Hmm, actually that doesn't work,

error[E0554]: `#![feature]` may not be used on the stable release channel
--
  | --> /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-frozen-abi-2.2.1/src/lib.rs:2:37
  | \|
  | 2 \| #![cfg_attr(feature = "frozen-abi", feature(specialization))]
  | \|                                     ^^^^^^^^^^^^^^^^^^^^^^^

Let me try to compile without it. Now we've bumped the bls version maybe it will just work.

PR #189, opened 2025-05-09
voting loop: add fallback vote variants

### LINKED ISSUE #47: Fast path 2-delta implementation

AshwinSekar:
Updates voting loop to send fallback variants.

Further PRs will:
Handle duplicate blocks,
Handle fast forward on receival of new finalization certificate

Fixes #47

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -604,7 +604,7 @@ impl VotingLoop {
             if voting_context.vote_history.voted_notar(parent_slot)
                 != Some((parent_block_id, parent_bank_hash))
             {
-                // Must have voted for the parent block
+                // Voted skip or notarize on a different version of the parent
                 return false;
             }
         }


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -455,25 +465,34 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
             if notarized_ratio < SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP {
                 continue;
             }
             if notarized_ratio + skip_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP {
-                return Some((b_block_id, b_bank_hash));
+                safe_to_notar.push((b_block_id, b_bank_hash));
             }
         }
-        None
+        safe_to_notar
```
carllin: nit: might be cleaner to dump this into one block over

```
let qualifies = notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY
        || (notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP
            && notarized_ratio + skip_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP);

    if qualifies {
        safe_to_notar.push((*b_block_id, *b_bank_hash));
    }
```

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -447,6 +454,9 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         {
             let b_block_id = block_id.unwrap();
             let b_bank_hash = bank_hash.unwrap();
+            if vote_history.voted_notar_fallback(slot, b_block_id, b_bank_hash) {
+                continue;
+            }
```
carllin, line 459: nit: might be good to extend unit tests to cover these early exits

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -415,18 +411,26 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
     /// Checks if we have voted to skip `slot` or notarize some block `b' = (block_id', bank_hash')` in `slot`
-    /// Additionally check that for some differernt block `b = (block_id, bank_hash)` in `slot` either:
+    /// Additionally check that for some different block `b = (block_id, bank_hash)` in `slot` either:
     /// (i) At least 40% of stake has voted to notarize `b`
     /// (ii) At least 20% of stake voted to notarize `b` and at least 60% of stake voted to either notarize `b` or skip `slot`
+    /// and we have not already cast a notarize fallback for this `b`
     /// If all the above hold, return `Some(block_id, bank_hash)` for the `b`
-    pub fn safe_to_notar(&self, slot: Slot, vote_history: &VoteHistory) -> Option<(Hash, Hash)> {
-        let epoch_stakes = self
+    pub fn safe_to_notar(&self, slot: Slot, vote_history: &VoteHistory) -> Vec<(Hash, Hash)> {
```
carllin, line 419: nit: might be good to add a unit test where more than one (Hash, Hash) is returned

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -563,15 +591,37 @@ impl VotingLoop {
...
                 return false;
             }
-            // TODO: fix WFSM hack for 1
-        } else if parent_slot > 1 {
-            // Only vote notarize if it is a consecutive slot
+        } else {
+            let parent_bank_hash = bank.parent_hash();
+            let parent_block_id = bank
+                .parent_block_id()
+                .expect("Synchronous replay cannot have a missing parent block id");
```
carllin, line 598: Are we confident at this point that `bank` is not the root (i.e. has not been squashed and the parent removed: https://github.com/anza-xyz/alpenglow/blob/c6ab69b7b9c728eded2ff9d859773d480d6250ab/runtime/src/bank.rs#L2671)? I think this expect is functionally equivalent to asserting the root < bank.slot()

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -91,6 +98,18 @@ impl VoteHistory {
...
+        self.voted_notar_fallback
+            .get(&slot)
+            .is_some_and(|v| v.contains(&(block_id, bank_hash)))
+    }
+
+    /// Whether we voted skip fallback for `slot`
+    pub fn voted_skip_fallback(&self, slot: Slot) -> bool {
+        self.voted_skip_fallback.contains(&slot)
+    }
+
```
carllin: just realized we prune everything before the root, especially in calls like `voted_notar`, wonder if we should assert in these functions that we never ask for `slot < self.root`, seems like an easy bug if we ever do.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -563,15 +591,37 @@ impl VotingLoop {
...
+
             if parent_slot + 1 != slot {
-                // TODO(ashwin): can probably mark notarize here as we'll never be able to fix this
+                // Non consecutive
+                return false;
+            }
+            if voting_context.vote_history.voted_notar(parent_slot)
+                != Some((parent_block_id, parent_bank_hash))
```
carllin, line 605: interesting, reading the tryNotar function and this seems right 
image

This means if you voted skip on the parent so you didn't vote notarize, you can't vote notarize on its descendant?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -563,15 +591,37 @@ impl VotingLoop {
...
+            {
+                // Must have voted for the parent block
```
carllin: nit: slightly confusing because at this point we definitely *didn't* vote for the parent block.

maybe move this comment outside above the if check

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -265,86 +264,102 @@ impl VotingLoop {
...
+                    // Refresh votes if no progress is made
+                    if refresh_timer.elapsed() > Duration::from_secs(1) {
+                        Self::refresh_votes(
+                            &my_pubkey,
+                            current_slot,
+                            &mut root_bank_cache,
+                            &mut cert_pool,
+                            &mut voting_context,
+                        );
+                        refresh_timer = Instant::now();
```
carllin, line 362: this is only necessary when skips/fallbacks are lost right?

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yes, because of the way we set root, we cap the maximum as the last slot in the leader window. Then the next time we vote notarize the slot has been incremented.

The next PR #190 also handles the case where we restart with a non zero root, in which case the current slot has to  be set to be greater than the root (actually ran into this `expect` in the wild while testing restart from blockstore/vote history storage)

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah exactly voting skip on the parent or voting notarize for a different duplicate version of the parent precludes you from voting notarize on the child.

The skip vote is already covered by the fact that we skip the entire window and check if we've voted before `try_notar`, so this specific condition is for the duplicate case.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah basically if no progress has been made. The next PR finishes this function to the spec that Kobi mentioned by refreshing votes for all slots since the last finalization

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: yeah that's cleaner updated 86597cbab9777c1a6ce70f6796a512c5b571c2e9

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: added tests here 8d74eb907488dfcc8209087afa1cd456d6706862

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: added tests here 8d74eb907488dfcc8209087afa1cd456d6706862

core/src/alpenglow_consensus/vote_history.rs
AshwinSekar: yeah good call, added here 142be7564774839e3facec6f946d8f80c244e7ee

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: good point, updated the comment 3af88f59580c04a9dbaea0480b1de76622ad656a

PR #190, opened 2025-05-13
add restart from blockstore and catchup when behind logic

### LINKED ISSUE #139: Ensure VoteHistory root >= bank forks root on startup
#### Problem
It's assumed vote history root >= bank forks root in voting loop https://github.com/anza-xyz/alpenglow/blob/800ee316151b5c2d8db2bec27d4c6b2f0e20755e/core/src/alpenglow_consensus/voting_loop.rs#L761-L768

### LINKED ISSUE #180: Reinitialize certificate pool from blockstore on startup

AshwinSekar:
#### Problem
In order to test more interesting GCE scenarios and long running partition tests we need the ability to:
- cleanly restart a node and catchup from ledger
- catchup if a node joins / rejoins the network or recovers from a network issue

#### Summary of Changes
- Adjust blockstore process logic to account for alpenglow ticks
- Initialize certificate pool from blockstore
- On a finalization certificate, fast forward current slot to the new slot
- Adjust refresh logic, rebroadcast votes from the last finalization certificate onwards
- Add test for single node restart
- Add test for 2 nodes where one node restarts while the other node makes progress

### Further TODO
- Handle slow finalization certificates by looking at the corresponding notarize fallback. I think this will be easier if we refactor the cert pool to track everything per slot rather than having multiple vote pools etc.
- Rebroadcast the latest finalization certificate if no progress has been made (can do this once we have bls)
- When fast forwarding, ensure the block id chain is correct so replay can also catchup, this can be done with duplicate block impl

Fixes #180 
Fixes #139

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -16,10 +16,7 @@ use {
     crossbeam_channel::Sender,
     solana_ledger::blockstore::Blockstore,
     solana_pubkey::Pubkey,
-    solana_runtime::{
-        bank::Bank,
-        epoch_stakes::{BLSPubkeyToRankMap, EpochStakes},
-    },
+    solana_runtime::{bank::Bank, epoch_stakes::EpochStakes},
     solana_sdk::{
         clock::{Epoch, Slot},
         epoch_schedule::EpochSchedule,
@@ -585,30 +582,25 @@ pub(crate) fn load_from_blockstore(
         .slot_certificates_iterator(root_bank.slot())
         .unwrap()
     {
-        for ((block_id, bank_hash), cert) in slot_cert.notarize_fallback_certificates.into_iter() {
-            // TODO: this will be migrated to BLS
-            let cert_id = CertificateId::NotarizeFallback(slot, block_id, bank_hash);
+        let certs = slot_cert
+            .notarize_fallback_certificates
+            .into_iter()
+            .map(|((block_id, bank_hash), cert)| {
+                let cert_id = CertificateId::NotarizeFallback(slot, block_id, bank_hash);
+                (cert_id, cert)
+            })
+            .chain(slot_cert.skip_certificate.map(|cert| {
+                let cert_id = CertificateId::Skip(slot);
+                (cert_id, cert)
+            }));
+
+        for (cert_id, cert) in certs {
             let cert = cert.into_iter().map(Arc::from).collect();
             trace!("{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool");
-            cert_pool.insert_certificate(
-                cert_id,
-                LegacyVoteCertificate::new(0, cert, &BLSPubkeyToRankMap::default()).unwrap(),
-            );
+            let legacy_cert = LegacyVoteCertificate::new(cert_id, cert)
+                .expect("Certificate construction must not fail");
+            cert_pool.insert_certificate(cert_id, legacy_cert);
         }
-
-        let Some(cert) = slot_cert.skip_certificate else {
-            continue;
-        };
-        // TODO: this will be migrated to BLS
-        let cert_id = CertificateId::Skip(slot);
-        let cert = cert.into_iter().map(Arc::from).collect();
-        trace!(
-            "{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool"
-        );
-        cert_pool.insert_certificate(
-            cert_id,
-            LegacyVoteCertificate::new(0, cert, &BLSPubkeyToRankMap::default()).unwrap(),
-        );
     }
     cert_pool
 }


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -336,6 +336,15 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         self.highest_notarized_fallback
     }
 
+    /// The highest fast finalized block, for use in catchup
```
AshwinSekar, line 329: when refactoring, we can use a field here to track the highest block as result of slow or fast finalization certificate

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -175,25 +180,43 @@ impl VotingLoop {
...
+            slot
+        };
+
+        let mut cert_pool = certificate_pool::load_from_blockstore(
+            &my_pubkey,
+            &root_bank_cache.root_bank(),
+            blockstore.as_ref(),
+            Some(certificate_sender),
+        ).inspect_err(|err| {
+            // TODO: should we just panic
```
AshwinSekar: if we're not building a leader block soon this is usually fine, we'll end up fast forwarding on a new finalization certificate.

If the cluster can't make progress or we need to build a leader block soon, then we'll be in trouble if `load_from_blockstore` fails so maybe better to panic

core/src/consensus/progress_map.rs
```diff
@@ -140,31 +140,6 @@ impl ForkProgress {
         }
     }
 
-    fn calculate_alpenglow_ticks(
```
AshwinSekar, line 143: i just moved this and the related test to `blockstore_processor.rs`

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -175,25 +180,43 @@ impl VotingLoop {
...
         }
 
+        let mut current_leader = None;
+        let mut current_slot = {
+            let bank_forks_slot = root_bank_cache.root_bank().slot() + 1;
+            let vote_history_slot = vote_history.root() + 1;
+            let slot = bank_forks_slot.max(vote_history_slot);
+            if bank_forks_slot != vote_history_slot {
+                panic!(
+                    "{my_pubkey}: Mismatch bank forks root {bank_forks_slot} vs vote history root {vote_history_slot}"
```
AshwinSekar, line 199: I haven't been able to reproduce this panic, but i'm thinking we should support `vote_history.root() > bank_forks.root()` in case there's some weird blockstore issues and `bank_forks.root()` lags behind, wdyt?

`bank_forks.root() > vote_history.root()` is already handled when loading `vote_history`

core/src/alpenglow_consensus/vote_history.rs
```diff
@@ -188,6 +191,26 @@ impl VoteHistory {
...
     }
+
+    #[allow(dead_code)]
+    /// Save the vote history to `vote_history_storage` signed by `node_keypair`
+    pub fn save(
+        &self,
+        vote_history_storage: &dyn VoteHistoryStorage,
+        node_keypair: &Keypair,
+    ) -> Result<()> {
+        let saved_vote_history = SavedVoteHistory::new(self, node_keypair)?;
```
carllin, line 202: Is this needed? I think we already save in the voting_service.rs https://github.com/anza-xyz/alpenglow/blob/bb5f82fb405e2364008868ff3a801c50efcce3c7/core/src/voting_service.rs#L227

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -175,25 +180,43 @@ impl VotingLoop {
...
         // reminder prev 3 need to be mutable
         if my_pubkey != vote_history.node_pubkey {
             todo!();
         }
 
+        let mut current_leader = None;
+        let mut current_slot = {
+            let bank_forks_slot = root_bank_cache.root_bank().slot() + 1;
+            let vote_history_slot = vote_history.root() + 1;
+            let slot = bank_forks_slot.max(vote_history_slot);
```
carllin: 1. we panic if bank_forks_slot != vote_history_slot anyways
2. think the panic log will be off by 1 because we have a +1

so this entire block might be simpler as:

```
if root_bank_cache.root_bank().slot() != vote_history.root() {
panic!(
                    "{my_pubkey}: Mismatch bank forks root {bank_forks_slot} vs vote history root {vote_history_slot}"
                );
}
let slot = root_bank_cache.root_bank().slot() + 1;
info!(
                "{my_pubkey}: Starting voting loop from {slot}: root {}",
                root_bank_cache.root_bank().slot()
            );
slot
```

core/src/alpenglow_consensus/voting_loop.rs
carllin: In the current code we support tower.root > bank_forks.root when we reconcile on startup: https://github.com/anza-xyz/alpenglow/blob/995622392b8a405ce30316dafdd1cd91b25e009b/core/src/consensus.rs#L1735-L1736

Namely if the vote history/tower storage has a higher root `X`, we:

1. Check that `X` exists in blockstore (we don't check if it's complete and thus replayable up to that point though, which might be a problem....)
2. If `X` exists we mark those slots as rooted in blockstore
3. Then on startup, we should replay at least up to `X`. 

If 1. fails i.e. the slot doesn't even exist in blockstore, we will panic before we even finish startup, so I think it's fine to also panic here since it should never happen.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -557,6 +574,42 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+        // TODO: this will be migrated to BLS
+        let cert_id = CertificateId::Skip(slot);
+        let cert = cert.into_iter().map(Arc::from).collect();
+        trace!(
+            "{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool"
+        );
+        cert_pool.insert_certificate(
+            cert_id,
+            LegacyVoteCertificate::new(0, cert, &BLSPubkeyToRankMap::default()).unwrap(),
+        );
```
carllin: Maybe something like this to consolidate the logic:

```
pub(crate) fn load_from_blockstore(
    my_pubkey: &Pubkey,
    root_bank: &Bank,
    blockstore: &Blockstore,
    certificate_sender: Option<Sender<(CertificateId, LegacyVoteCertificate)>>,
) -> Result<CertificatePool<LegacyVoteCertificate>, BlockstoreError> {
    let mut cert_pool = CertificatePool::new_from_root_bank(root_bank, certificate_sender);

    for (slot, slot_cert) in blockstore.slot_certificates_iterator(root_bank.slot())? {
        let certs = slot_cert
            .notarize_fallback_certificates
            .into_iter()
            .map(|((block_id, bank_hash), cert)| {
                let cert_id = CertificateId::NotarizeFallback(slot, block_id, bank_hash);
                (cert_id, cert)
            })
            .chain(slot_cert.skip_certificate.map(|cert| {
                let cert_id = CertificateId::Skip(slot);
                (cert_id, cert)
            }));

        for (cert_id, cert) in certs {
            let cert = cert.into_iter().map(Arc::from).collect();
            trace!("{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool");
            let legacy_cert = LegacyVoteCertificate::new(0, cert, &BLSPubkeyToRankMap::default())
                .expect("Certificate construction must not fail");
            cert_pool.insert_certificate(cert_id, legacy_cert);
        }
    }

    Ok(cert_pool)
}
```

core/src/alpenglow_consensus/voting_loop.rs
carllin: Hmm seems like it should be impossible for load_from_blockstore to even panic, the only early return is here: https://github.com/anza-xyz/alpenglow/blob/bb5f82fb405e2364008868ff3a801c50efcce3c7/core/src/alpenglow_consensus/certificate_pool.rs#L584, which can't even happen because we hard unwrap all the deserialization logic inside the `slot_certificates_iterator()` function lol. So we would panic if Rocks was corrupted anyways.

Seems like we should just hard unwrap https://github.com/anza-xyz/alpenglow/blob/bb5f82fb405e2364008868ff3a801c50efcce3c7/core/src/alpenglow_consensus/certificate_pool.rs#L58 as well and just not even return a `Result`.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -361,6 +385,26 @@ impl VotingLoop {
                         );
                         refresh_timer = Instant::now();
                     }
+
+                    // Check if we can fast forward
+                    // TODO(ashwin): for duplicate blocks, ensure that we have the correct chain up to this
+                    let Some((slot, _, _)) = cert_pool.highest_fast_finalized() else {
```
carllin, line 388: can we also fast forward on regular finalizations?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -361,6 +385,26 @@ impl VotingLoop {
...
+                }
+
+                if let Some(slot) = fast_forward {
+                    info!(
+                        "{my_pubkey}: While waiting for certificate on {current_slot} we observed fast finalization certificate for {slot}. \
+                        Fast forwarding to {}",
+                        slot + 1,
+                    );
+                    current_slot = slot + 1;
+                    break;
```
carllin, line 404: Does the new local cluster test cover this scenario where we jump ahead? Seem like a good case to cover because it's out of band setting current_slot, where before it could only go up 1 at a time. 

For instance now we can jump into the middle of a 4 leader window, have to think through the consequences there 🤔 

EDIT: I see the tests, dope.

Things to think through, lemme work through it.

On a big jump/fast forward into the middle of a leader window, let's say the 3rd slot out of 4.

1. Most likely because repair/replay is so far behind we will vote skip here right? https://github.com/anza-xyz/alpenglow/blob/bb5f82fb405e2364008868ff3a801c50efcce3c7/core/src/alpenglow_consensus/voting_loop.rs#L304-L313
2. Will we then need cert repair in order to pass the branch notarized check? https://github.com/anza-xyz/alpenglow/blob/bb5f82fb405e2364008868ff3a801c50efcce3c7/core/src/alpenglow_consensus/voting_loop.rs#L341-L342

local-cluster/tests/local_cluster.rs
```diff
@@ -6095,3 +6095,133 @@ fn test_invalid_forks_persisted_on_restart() {
...
+
+    // Let A make roots by itself
+    cluster.check_for_new_roots(
+        8,
+        "test_alpenglow_imbalanced_stakes_catchup",
+        SocketAddrSpace::Unspecified,
+    );
+
+    info!("restarting node B");
+    cluster.restart_node(&node_pubkeys[1], b_info, SocketAddrSpace::Unspecified);
```
carllin, line 6219: Just to make sure node B has to fast forward here, can we delete the slots from `A` so `B` can't repair them and then check that `B` doesn't panic/its vote history has the new finalization certs/will send out skip votes?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -361,6 +385,26 @@ impl VotingLoop {
...
+
+                    // Check if we can fast forward
+                    // TODO(ashwin): for duplicate blocks, ensure that we have the correct chain up to this
+                    let Some((slot, _, _)) = cert_pool.highest_fast_finalized() else {
+                        continue;
+                    };
+                    if slot > current_slot {
+                        fast_forward = Some(slot);
+                        break;
+                    }
```
carllin, line 394: might be a good time to start factoring out these various nested while loops into separate functions, beginning to get hard to read the more nested we get

core/src/alpenglow_consensus/vote_history.rs
AshwinSekar: Just kept it here as a utility for local-cluster tests, similar to the tower version

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: Yep. wen is adding `cert_pool.highest_finalized()` to get `(Slot, Hash, Hash)` for either fast / slow into the cert pool api and then we can plug it in here

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: So when we jump through a big gap, regardless of which slot in the window it is, we will most likely vote skip:
- Either we have not finished replay
- Even if we have finished replay, if we're in the middle of a leader window, we didn't vote on the parent so vote skip

Then to pass through the branch notarized check, either we are close enough to the tip of the cluster that the votes/certificates have been received through all-to-all OR we are recovering from network issues and are missing the votes.

There's 2 cases:
1) Our stake is insignificant, cluster will continue and once our network recovers we can fast forward again and hopefully now the votes/certificates near tip will be received
2) Our stake is significant for certificate on this slot or some future slot and the whole cluster is stuck. The refresh votes case will refresh ALL votes since the last finalization certificate + rebroadcast the last finalization certificate. So eventually the cluster will converge.

(2) is a little tricky, but after talking to Kobi I believe this is enough to cover our bases.

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah that's fair, was planning to clean this up when I refactor this to the parentReady version

local-cluster/tests/local_cluster.rs
AshwinSekar: what's the relevance of repair here? Even if we delete the slots since A has made 8 roots, B doesn't have certificates from any of those slots since it missed the all-to-all broadcast (and gossip is too slow/wonky with small cluster) so it has no choice but to fast forward.

Since B has only 10% of the stake, even if it ends up voting notarize it will still have to fast forward for the missing certificates:
```
solana_core::replay_stage] Starting alpenglow voting loop
solana_core::alpenglow_consensus::voting_loop]: Starting voting loop from 36: root 35
solana_core::alpenglow_consensus::certificate_pool]: loading certificate NotarizeFallback(35, 9ZgDaDZpTJZEBXunXSJjjPLkeipmpFuHuxHn6RokxqDL, 4QZY8EMZHTEXtjTLry8bbBgt2u5fcd8o1UQY9eDiCEE1) from blockstore into certificate pool
solana_core::alpenglow_consensus::certificate_pool]: loading certificate NotarizeFallback(36, 8jeV8oY5KjmKWqN3FDoMcYFb2483Aos8Y2enhRhjQyw2, 7x9tocQf4KqjxFtMnnBNLir6PqVPhF9M7SWYV2xJLWWN) from blockstore into certificate pool
solana_core::alpenglow_consensus::certificate_pool]: loading certificate NotarizeFallback(37, C5BY4qaxUAitUNGzrEyorxueG5qYQLC9Wh3Xh3o2XEFt, 2SjrEFC37tzmc3fwop45nSKybaeamzGZWCr4cGLivn73) from blockstore into certificate pool

solana_core::alpenglow_consensus::voting_loop]: Entering voting loop for slot: 36, is_leader: false. Skip timer 0 vs timeout 800
solana_core::alpenglow_consensus::voting_loop]: Branch Certified: Slot 36 from 2 validators
solana_core::alpenglow_consensus::voting_loop]: Slot 36 certificate observed in 0 ms. Skip timer 0 vs timeout 800
solana_core::alpenglow_consensus::voting_loop]: Entering voting loop for slot: 37, is_leader: false. Skip timer 0 vs timeout 1200
solana_core::alpenglow_consensus::voting_loop]: Branch Certified: Slot 37 from 2 validators
solana_core::alpenglow_consensus::voting_loop]: Slot 37 certificate observed in 0 ms. Skip timer 0 vs timeout 1200
solana_core::alpenglow_consensus::voting_loop]: Entering voting loop for slot: 38, is_leader: false. Skip timer 0 vs timeout 1600
solana_core::alpenglow_consensus::voting_loop]: Voting notarize for slot 38 hash BVr82sqJkH7m9Ankofpvn1z5zypfNQKwD6dUZ3WexE5W block_id AZeSDjH1WobBQkaGQYW3UJPMnCcE8LKk7HqpTf565ZRb

solana_core::alpenglow_consensus::voting_loop]: new finalization certificate for 56
solana_core::alpenglow_consensus::voting_loop]: new finalization certificate for 59
solana_core::alpenglow_consensus::voting_loop]: new finalization certificate for 60
solana_core::alpenglow_consensus::voting_loop]: While waiting for certificate on 38 we observed fast finalization certificate for 60. Fast forwarding to 61
solana_core::alpenglow_consensus::voting_loop]: Entering voting loop for slot: 61, is_leader: false. Skip timer 0 vs timeout 1200
```

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: good call I just hard unwrap here [983d49a](https://github.com/anza-xyz/alpenglow/pull/190/commits/983d49acd2721b5554bf69c9bd833c47c9af8127)

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: nice that's less confusing and simpler 😄 [cd39785](https://github.com/anza-xyz/alpenglow/pull/190/commits/cd397854dd75e843041974886981fcdcc4395446)

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: good call [0cdafb6](https://github.com/anza-xyz/alpenglow/pull/190/commits/0cdafb6857fd9e308a0fca3d5b0c5a380b630b3b)

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: this also means that we don't need certificate repair, so i'll certificate service can just broadcast + insert into blockstore. Will want to extend it to finalize certificates as well at least for the broadcast

carllin: `rebroadcast the last finalization certificate` ah yeah that should do the trick

PR #192, opened 2025-05-14
Re-enable conflicting vote type check.

wen-coding:
Detect and reject conflicting vote type from the same pubkey.

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -1652,18 +1652,28 @@ mod tests {
 
     fn test_reject_conflicting_vote<VC: VoteCertificate>(
         pool: &mut CertificatePool<VC>,
-        pubkey: &Pubkey,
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -72,7 +73,7 @@ pub struct CertificatePool<VC: VoteCertificate> {
     /// Completed certificates
     completed_certificates: BTreeMap<CertificateId, VC>,
     // Lookup table for checking conflicting vote types.
-    // conflicting_vote_types: HashMap<VoteType, Vec<VoteType>>,
```
AshwinSekar, line 75: can we just make this a constant, via a static function or lazy_static?

wen-coding: The original plan was to have all data and no code in alpenglow_consensus.rs. But since we already have quite some code in there, I think it makes sense to move this there as well. Changed.

PR #194, opened 2025-05-17
Add sorted_pubkeys to BLSPubkeyToRankMap.

wen-coding:
Because we only send a rank in the BLS vote right now, we need to translate that into the actual ed25519 and BLS Pubkey when we verify the message and when we add the BLS vote message to certificate pool.

We can probably remove ed25519 pubkey from this map later when the certificate pool only uses BLS pubkey.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1182,7 +1182,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4114279215a005bc675e386011e594e1d9b800918cea18fcadadcce864a2046b"
 dependencies = [
  "borsh-derive 0.10.3",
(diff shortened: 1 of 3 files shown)


runtime/src/epoch_stakes.rs
```diff
@@ -52,6 +61,10 @@ impl BLSPubkeyToRankMap {
     pub fn get(&self, bls_pubkey: &BLSPubkey) -> Option<&u16> {
```
carllin: should probably rename this to `get_rank()` or `rank()` to distinguish from the new `get_pubkey()`

wen-coding: Good point, changed.

PR #196, opened 2025-05-17
Add BLSSigverifier.

wen-coding:
Add BLS Sigverifier, it's just a place holder for now without any actual verification.

We will plug it into TPU port when send and receive code is ready.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1184,7 +1184,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4114279215a005bc675e386011e594e1d9b800918cea18fcadadcce864a2046b"
 dependencies = [
  "borsh-derive 0.10.3",
- "hashbrown 0.13.2",
+ "hashbrown 0.12.3",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Default, Clone)]
+pub struct BLSSigVerifier {
+    sender: Option<Sender<BLSMessage>>,
+}
+
+impl SigVerifier for BLSSigVerifier {
+    type SendType = ();
+    fn verify_batches(&self, batches: Vec<PacketBatch>, _valid_packets: usize) -> Vec<PacketBatch> {
+        batches
```
samkim-crypto, line 20: Nit: Maybe we can add a TODO here just so that it shows up when we grep for TODOs.

wen-coding: Done.

wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -0,0 +1,45 @@
...
+        // TODO(wen): just a placeholder without any batching.
+        if let Some(sender) = &self.sender {
+            packet_batches.iter().for_each(|batch| {
+                batch.iter().for_each(|packet| {
+                    if let Ok(message) = packet.deserialize_slice::<BLSMessage, _>(..) {
+                        sender.send(message).unwrap();
+                    }
+                });
+            });
+        }
```
carllin, line 36: why not just  `self.banking_stage_sender.send(packet_batches)?;` like we do in today's sigverify?

wen-coding: I looked at that, but banking_stage_sender is mostly related to banking_stage, not totally relevant to us. We are not sending packets to banking_stage anyways.

PR #197, opened 2025-05-22
[core] Implement `AlpenglowVoteTransaction` and `VoteCertificate` traits for the new BLS message formats

samkim-crypto:
#### Summary of Changes
Implemented `AlpenglowVoteTransaction` and `VoteCertificate` traits for the new BLS message formats. I made some small changes to the traits themselves to cope with the new format. The most noticeable change is probably the  removal of the hashmap from the cert generation function inputs. We still need the hashmap for vote verification and whenever we need the aggregated public key, but for cert generation, since the vote messages already contain the rank index of the public key, the cert constructor doesn't necessarily need access to the public key index any more.

Fixes #

### DIFF
--- a/core/src/alpenglow_consensus/vote_certificate.rs
+++ b/core/src/alpenglow_consensus/vote_certificate.rs
@@ -132,6 +132,7 @@ fn aggregate_vote_signatures(
         bitmap.set(transaction.rank as usize, true);
     }
 
+    // TODO: truncate trailing zeros in bitmap
     Ok((aggregate_signature.into(), bitmap))
 }
 


core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -1,15 +1,26 @@
...
+/// Maximum number of validators in a certificate
+///
+/// There are around 1500 validators currently. For a clean power-of-two
+/// implementation, we should chosoe either 2048 or 4096. Choose a more
+/// conservative number 4096 for now.
+const MAXIMUM_VALIDATORS: usize = 4096;
+
+/// The number of bytes in a bitmap to represent up to 4096 validators
+/// (`MAXIMUM_VALIDATORS` / 8)
+const VALIDATOR_BITMAP_U8_SIZE: usize = 512;
```
wen-coding, line 25: Hmm, I thought we use base 3 encoding now so it's more than 512?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -58,174 +67,86 @@ impl VoteCertificate for LegacyVoteCertificate {
...
-            aggregate_pubkey: aggregate_pubkey.into(),
-            aggregate_signature: aggregate_signature.into(),
-            bit_vector,
-            stake,
-            vote_count,
-        })
+fn aggregate_vote_signatures(
+    transactions: Vec<Arc<VoteMessage>>,
+) -> Result<(Signature, BitVec<u8, Lsb0>), CertificateError> {
+    if transactions.len() > MAXIMUM_VALIDATORS {
```
wen-coding, line 108: We will have fewer than 4096 validators for a long time, maybe we should pass down the actual number of validators.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -58,174 +67,86 @@ impl VoteCertificate for LegacyVoteCertificate {
...
-            .set_bit(*validator_index as usize, true)
-            .map_err(|_| CertificateError::IndexOutOfBound)?;
-
-        self.stake += stake;
-        self.vote_count += 1;
-        Ok(())
+        //
+        // TODO: This only accounts for one type of vote. Update this after
+        // we have a base3 encoding implementation.
+        bitmap.set(transaction.rank as usize, true);
```
wen-coding, line 132: Do you think we need to double check transaction.rank is not out of bound here?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -58,174 +67,86 @@ impl VoteCertificate for LegacyVoteCertificate {
...
+pub fn aggregate_pubkey(
+    bitmap: &BitVec<u8, Lsb0>,
+    bls_pubkey_to_rank_map: &BLSPubkeyToRankMap,
+) -> Result<BlsPubkey, CertificateError> {
+    let mut aggregate_pubkey = PubkeyProjective::default();
+    for (i, included) in bitmap.iter().enumerate() {
+        if *included {
+            let bls_pubkey: PubkeyProjective = bls_pubkey_to_rank_map
+                .get_pubkey(i)
+                .ok_or(CertificateError::IndexOutOfBound)?
```
wen-coding, line 150: Hmm, now I think we should actually pass the out of bound index in the error message.

core/src/alpenglow_consensus.rs
```diff
@@ -59,6 +60,43 @@ impl CertificateId {
     }
 }
 
+impl From<CertificateId> for Certificate {
```
wen-coding: Hmm, this file was supposed to hold constants originally, I'm wondering whether there is a better file to put this function?

wen-coding: Hmm, this file was supposed to hold constants originally, I'm wondering whether there is a better file to put this function?

core/src/alpenglow_consensus/transaction.rs
```diff
@@ -9,3 +16,19 @@ impl AlpenglowVoteTransaction for VersionedTransaction {
         Self::default()
     }
 }
+
+impl AlpenglowVoteTransaction for VoteMessage {
+    fn new_for_test(bls_keypair: BLSKeypair) -> Self {
+        // use notarization vote since this is just for tests
+        let vote = Vote::Notarize(NotarizationVote::default());
```
wen-coding: We have new_notarization_vote you can use. Better give it a real slot and some hashes, otherwise we might be scratching our heads why we couldn't get the slot out of VoteMessage?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -58,174 +67,86 @@ impl VoteCertificate for LegacyVoteCertificate {
...
-            if bit_vector
-                .get_bit(i)
-                .map_err(|_| CertificateError::IndexOutOfBound)?
-            {
-                let pubkey_projective: PubkeyProjective = (*pubkey)
-                    .try_into()
-                    .map_err(|_| CertificateError::InvalidPubkey)?;
-                aggregate_pubkey.aggregate_with([&pubkey_projective]);
-            }
+    Ok((aggregate_signature.into(), bitmap))
```
wen-coding, line 136: Hmm, we won't need 512 bytes for many years, and in test clusters there are only 10 nodes, maybe we can remember the highest bit set and shrink the trailing zeroes?

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Oh yeah we don't have a proper base3 encoding implementation yet, so let's update this number in the PR that adds the base3 encoding.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: This is a good idea. It seems like we don't have logic to determine what this number should be initially should be so let me add a TODO here for now and then address this in a follow-up.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Good point. Fixed.

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Okay yep. I still need to clean up the error handling in a follow-up, so I'll add the index in the error message in that PR.

core/src/alpenglow_consensus.rs
samkim-crypto: Okay, I moved it to `vote_certificate.rs`.

core/src/alpenglow_consensus/transaction.rs
samkim-crypto: Yep, done!

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Yeah we should definitely do that. Let me implement the base3 encoding first and then we can add a serialization logic there to maybe shrink down the trailing zeros.

wen-coding: Sounds good, let's add a TODO here then.

PR #198, opened 2025-05-27
voting_loop: migrate to parent ready, fix WFSM / snapshot hack

### LINKED ISSUE #155: Add restarting from snapshot and WFSM functionality

AshwinSekar:
Ignore first commit built off of #190 

#### Problem
We still use the branch certified / skip certified naming from an earlier draft of the paper. Additionally we don't actually verify the hash for notarize fallback certificates.
Finally we hack in slot 1 for WFSM to enable GCE tests.

#### Summary of Changes

- Add parent ready tracker to explicitly track notarize fallback / skip certificates by block hashes
- Update voting loop to use parent ready instead of old checks (Note this shouldn't actually change any behavior at the moment, but will once we have duplicate blocks)
- Instead of hacking in WFSM / snapshot slots, implicitely notarize fallback the root bank used in startup

Fixes #155

### DIFF
--- a/core/src/alpenglow_consensus/parent_ready_tracker.rs
+++ b/core/src/alpenglow_consensus/parent_ready_tracker.rs
@@ -1,0 +1,265 @@
+//! Tracks the parent-ready condition
+//!
+//! The parent-ready condition pertains to a slot `s` and a block hash `hash(b)`,
+//! where `s` is the first slot of a leader window and `s > slot(b)`.
+//! Specifically, it is defined as the following:
+//!   - Block `b` is notarized or notarized-fallback, and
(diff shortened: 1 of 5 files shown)


core/src/alpenglow_consensus/parent_ready_tracker.rs
```diff
@@ -0,0 +1,254 @@
+//! Tracks the parent-ready condition
```
AshwinSekar, line 1: Mostly plagiarized from reference impl, ty quentin
https://github.com/qkniep/alpenglow/blob/main/src/consensus/pool/parent_ready_tracker.rs

PR #199, opened 2025-05-28
voting_loop: add parent ready tracker, fix WFSM / snapshot hack

AshwinSekar:
Reuploaded from #198 

### Problem
We still use the branch certified / skip certified naming from an earlier draft of the paper. Additionally we don't actually verify the hash for notarize fallback certificates.
Finally we hack in slot 1 for WFSM to enable GCE tests.

### Summary of Changes

- Add parent ready tracker to explicitly track notarize fallback / skip certificates by block hashes
- Update voting loop to use parent ready instead of old checks (Note this shouldn't actually change any behavior at the moment, but will once we have duplicate blocks)
- Instead of hacking in WFSM / snapshot slots, implicitely notarize fallback the root bank used in startup

Fixes https://github.com/anza-xyz/alpenglow/issues/155

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -74,7 +74,9 @@ pub struct CertificatePool<VC: VoteCertificate> {
     vote_pools: BTreeMap<PoolId, VotePool<VC>>,
     /// Completed certificates
     completed_certificates: BTreeMap<CertificateId, VC>,
-    /// Parent ready tracker
+    /// Tracks slots which have reached the parent ready condition:
+    /// - They have a potential parent block with a NotarizeFallback certificate
+    /// - All slots from the parent have a Skip certificate
     pub(crate) parent_ready_tracker: ParentReadyTracker,
     /// Highest block that has a NotarizeFallback certificate, for use in producing our leader window
     highest_notarized_fallback: Option<(Slot, Hash, Hash)>,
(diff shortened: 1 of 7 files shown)


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -382,26 +378,6 @@ impl VotingLoop {
                         );
                         refresh_timer = Instant::now();
                     }
-
-                    // Check if we can fast forward
```
AshwinSekar, line 386: i realized parent ready gives us this for free, no need to special case this

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -73,6 +74,8 @@ pub struct CertificatePool<VC: VoteCertificate> {
     vote_pools: BTreeMap<PoolId, VotePool<VC>>,
     /// Completed certificates
     completed_certificates: BTreeMap<CertificateId, VC>,
+    /// Parent ready tracker
```
bw-solana: pretty worthless comment. Might be nice to indicate what ready means in this context. Maybe something like `"Tracks valid parent blocks each slot can build off of"`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -221,6 +235,15 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
                     self.highest_notarized_fallback =
                         Some((slot, block_id.unwrap(), bank_hash.unwrap()));
+                    self.parent_ready_tracker.add_new_notar_fallback((
+                        slot,
+                        block_id.unwrap(),
+                        bank_hash.unwrap(),
+                    ));
+                }
+
+                if cert_id.is_skip() {
```
bw-solana: might be time for a match on `cert_id`

core/src/alpenglow_consensus/parent_ready_tracker.rs
```diff
@@ -0,0 +1,265 @@
...
+        let status = self.slot_statuses.entry(slot).or_default();
+        if status.notar_fallbacks.contains(&block) {
+            return;
+        }
+        trace!(
+            "{}: Adding new notar fallback for {block:?}",
+            self.my_pubkey
+        );
+        status.notar_fallbacks.push(block);
+        assert!(status.notar_fallbacks.len() <= 3);
```
bw-solana: might as well create the constant for max duplicates

core/src/alpenglow_consensus/parent_ready_tracker.rs
```diff
@@ -0,0 +1,265 @@
...
+}
+
+#[derive(Clone, Default, Debug)]
+struct ParentReadyStatus {
+    /// Whether this slot has a skip certificate
+    skip: bool,
+    /// The blocks that have been notar fallbacked in this slot
+    notar_fallbacks: Vec<Block>,
+    /// The parent blocks that achieve parent ready in this slot
+    parents_ready: Vec<Block>,
```
bw-solana, line 46: is this intended to be the full suite of possible parents for this slot/block to build off of?

core/src/alpenglow_consensus/parent_ready_tracker.rs
```diff
@@ -0,0 +1,265 @@
...
+    slot_statuses: HashMap<Slot, ParentReadyStatus>,
+
+    /// Root
+    root: Slot,
+
+    /// Highest parent ready slot
+    // TODO: While the voting loop is sequential we track every slot (not just the first in window)
+    // However once we handle all slots concurrently we will update this to only count first leader
+    // slot in window
+    highest_parent_ready: Slot,
```
bw-solana: I think we should rename to `highest_with_parent_ready`

core/src/alpenglow_consensus/voting_loop.rs
bw-solana: I can see how parent ready allows us to skip the inner loop (because `current_slot` presumably < highest w/ a parent ready), but we still need to iterate through all the looping and such to walk up to the highest finalized block, right?

Is there a reason we can't jump current slot up to the highest with parent ready?

bw-solana (commented):
Left a few suggestions and questions.

I'm not completely following everything yet, but everything I can grok appears correct

core/src/alpenglow_consensus/parent_ready_tracker.rs
AshwinSekar: good idea, I reused the one for how many notarize fallback votes we can store

core/src/alpenglow_consensus/parent_ready_tracker.rs
AshwinSekar: yes exactly, updated the comment

`block_production_parent()` will choose the highest possible parent slot, but as detailed in the comment there I want to change it to pick the parent with the least amount of duplicate blocks available

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah I just wanted to simplify the logic here, either way we'll need to replay all the blocks to get to this newest highest finalized block so didn't think it was a big deal to have us loop iteratively. 

Once things have stabilized and I can make this an event driven loop, we'll instantly move up to the highest finalized block timeoutwise, and pending blocks will take care of voting notarize for the blocks in between

core/src/alpenglow_consensus/certificate_pool.rs
AshwinSekar: good call, also noted that I was only adding higher notarize fallback certificates, but changed it to always add the notarize fallback certificate as it could be relevant for earlier duplicate blocks.

PR #200, opened 2025-05-30
local-cluster: ensure liveness after single notar fallback

ksn6:
#### Problem

The introduced local cluster test validates Alpenglow's ability to maintain liveness when a node needs to issue a NotarizeFallback vote. The test sets up a two-node cluster with a specific stake distribution to create a scenario where:

- Node A has 60% of stake minus a small amount (epsilon)
- Node B has 40% of stake plus a small amount (epsilon)

The test simulates the following sequence:
1. Node B (as leader) proposes a block for slot 32
2. Node A is unable to receive the block (simulated via turbine disconnection)
3. Node A sends Skip votes to both nodes for slot 32
4. Node B sends Notarize votes to both nodes for slot 32
5. Node A receives both votes and its certificate pool determines:
   - Skip has (60% - epsilon) votes
   - Notarize has (40% + epsilon) votes
   - Protocol determines it's "SafeToNotar" and issues a NotarizeFallback vote
6. Node B doesn't issue NotarizeFallback because it already submitted a Notarize
7. Node B receives Node A's NotarizeFallback vote
8. Network progresses and maintains liveness after this fallback scenario

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -95,7 +95,7 @@ use {
         iter,
         path::Path,
         sync::{
-            atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
+            atomic::{AtomicBool, AtomicUsize, Ordering},
             Arc, Mutex,
         },
         thread::{sleep, Builder, JoinHandle},
@@ -6344,7 +6344,6 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
 
     // Track Node A's votes and when the test can conclude
     let node_a_filtered_votes = Arc::new(Mutex::new(HashMap::new()));
-    let node_a_two_vote_slot = Arc::new(AtomicU64::new(0));
     let mut post_experiment_votes = HashMap::new();
     let mut post_experiment_roots = HashSet::new();
 
@@ -6352,7 +6351,7 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
     let vote_listener = std::thread::spawn({
         let mut buf = [0_u8; 65_535];
         let node_a_filtered_votes = node_a_filtered_votes.clone();
-        let node_a_two_vote_slot = node_a_two_vote_slot.clone();
+        let mut check_for_roots = false;
 
         move || loop {
             let n_bytes = vote_listener.recv(&mut buf).unwrap();
@@ -6379,16 +6378,16 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
                     .or_insert(HashSet::new());
                 cur_filtered_votes.insert(vote_tuple.1);
 
-                if node_a_two_vote_slot.load(Ordering::Acquire) == 0
-                    && cur_filtered_votes.len() == 2
-                {
-                    node_a_two_vote_slot.store(vote.slot(), Ordering::Release);
+                if !check_for_roots && cur_filtered_votes.len() == 2 {
+                    check_for_roots = true;
+                    assert!(cur_filtered_votes.contains(&2)); // skip on slot 32
+                    assert!(cur_filtered_votes.contains(&3)); // notar fallback on slot 32
                 }
             }
 
             // We should see a skip followed by a notar fallback. Once we do, the experiment is
             // complete.
-            if node_a_two_vote_slot.load(Ordering::Acquire) > 0 {
+            if check_for_roots {
                 node_a_turbine_disabled.store(false, Ordering::Relaxed);
 
                 if vote.is_finalize() {
@@ -6409,15 +6408,4 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
     });
 
     vote_listener.join().unwrap();
-
-    // Verify that Node A issued the expected sequence of votes
-    {
-        let node_a_filtered_votes = node_a_filtered_votes.lock().unwrap();
-
-        let slot = node_a_two_vote_slot.load(Ordering::Relaxed);
-
-        assert_eq!(2, node_a_filtered_votes[&slot].len());
-        assert!(node_a_filtered_votes[&slot].contains(&2)); // skip on slot 32
-        assert!(node_a_filtered_votes[&slot].contains(&3)); // notar fallback on slot 32
-    }
 }


local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6226,164 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+            // Gather all votes issued by node A on slot 32
+            if txn.slot() == 32 && node_name == "A" {
+                node_a_filtered_votes.lock().unwrap().push(*txn);
+            };
+
+            let num_votes = { node_a_filtered_votes.lock().unwrap().len() };
+
+            // We should see a skip followed by a notar fallback. Once we do, the experiment is
+            // complete.
+            if num_votes == 2 {
```
wen-coding: Could we have more than 2 votes here? e.g. We are somehow refreshing an earlier vote?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6226,164 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+                node_a_turbine_disabled.store(false, Ordering::Relaxed);
+
+                // Once we've observed >= 10 roots upon the experiment completing, we're all set.
+                if txn.is_finalize() {
```
wen-coding: Could we be tricked by 10 finalize votes here instead of actually root making progress?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6226,164 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+
+    vote_listener.join().unwrap();
+
+    // Verify that Node A issued the expected sequence of votes
+    {
+        let node_a_filtered_votes = node_a_filtered_votes.lock().unwrap();
+
+        assert_eq!(2, node_a_filtered_votes.len());
+        assert!(node_a_filtered_votes[0].is_skip());
+        assert!(node_a_filtered_votes[1].is_notarize_fallback());
```
wen-coding: Should you check the slot on votes here?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+    };
+
+    // Create our UDP socket to listen to votes
+    let vote_listener = solana_net_utils::bind_to_localhost().unwrap();
+
+    // Create validator configs
+    let mut validator_config = ValidatorConfig::default_for_test();
+    validator_config.fixed_leader_schedule = Some(leader_schedule);
+    validator_config.voting_service_additional_listeners =
+        Some(vec![vote_listener.local_addr().unwrap()]);
```
wen-coding, line 6302: (Don't have to do it in current PR)
I guess if these are used by many tests, we should someday put it into common utility functions.

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+        move || loop {
+            let n_bytes = vote_listener.recv(&mut buf).unwrap();
+            let vote_txn = bincode::deserialize::<Transaction>(&buf[0..n_bytes]).unwrap();
+
+            let (vote_pubkey, parsed_vote, ..) =
+                vote_parser::parse_alpenglow_vote_transaction(&vote_txn).unwrap();
+
+            let node_name = match vote_pubkeys
+                .iter()
+                .position(|pk| pk == &vote_pubkey)
```
wen-coding: You only used vote_pubkeys here, so we don't need to iterate through vote_pubkeys every time, make vote_pubkeys HashMap<Pubkey, usize> and do a lookup is also fine.

Of course, since we only have two nodes, this is not important.

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+                .unwrap()
+            {
+                0 => "A",
+                1 => "B",
+                _ => "Unknown",
```
wen-coding: Can we have any value other than 0 and 1? We can panic here?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+            };
+
+            let vote = parsed_vote.as_alpenglow_transaction_ref().unwrap();
+
+            // Once we've received a vote from node B at slot 31, we can start the experiment.
+            if vote.slot() == 31 && node_name == "B" {
```
wen-coding: 31 is just a random number where we decided the cluster is stable enough for experiment right?

We can make it vote.slot() >= 31 and get the real slot for latter operations.

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+                node_a_turbine_disabled.store(true, Ordering::Relaxed);
+            }
+
+            // Gather all votes issued by node A on slot 32
+            if vote.slot() == 32 && node_name == "A" {
```
wen-coding: Could it be that you didn't disable Turbine fast enough for A to have a problem with slot 32? Should we listen for longer and just check for one slot with both votes?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
         SocketAddrSpace::Unspecified,
     );
 }
+
+fn _vote_to_tuple(vote: &Vote) -> (u64, u8) {
+    let discriminant = if vote.is_notarization() {
```
wen-coding, line 6232: Can we just do a match here?

local-cluster/tests/local_cluster.rs
```diff
@@ -6225,3 +6227,195 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
         SocketAddrSpace::Unspecified,
     );
 }
+
+fn _vote_to_tuple(vote: &Vote) -> (u64, u8) {
```
wen-coding, line 6231: We can make this an utility function, other tests will likely use it.

Do you think it should return a local enum to be more readable?

local-cluster/tests/local_cluster.rs
ksn6: Yep, makes things cleaner.

local-cluster/tests/local_cluster.rs
ksn6: No longer needed b/c of the HashMap change.

local-cluster/tests/local_cluster.rs
ksn6: There's no explicit check to see which slot is "stable"; we'll need to commit to some random number where the cluster is stable enough for experimentation. 31 should be fine for now - if this starts getting flaky (seems unlikely?) we can change this number.

I'll add a comment here for now.

local-cluster/tests/local_cluster.rs
ksn6: Fixed.

local-cluster/tests/local_cluster.rs
ksn6: I don't believe we're able to directly access the underlying enum variants, sadly.

local-cluster/tests/local_cluster.rs
ksn6: Something like this, although - will refactor in a future commit once we have more tests to see what really ends up being commonly used.

local-cluster/tests/local_cluster.rs
ksn6: Yep, will get this in a future PR.

local-cluster/tests/local_cluster.rs
wen-coding: I mean something like:
match vote {
   Vote::Notarize(...) => LocalVoteType::Notarize,
   Vote::NotarizeFallback(...) => LocalVoteType::NotarizeFallback,
   ...
}

wen-coding (approved):

PR #202, opened 2025-05-30
Add alpenglow port and alpenglow sigverify stage.

wen-coding:
- Add Alpenglow port
- Plug BLS sigverify stage into Alpenglow port

Note:

- this port is currently UDP, but I believe we're expecting this to be QUIC or SNP based in the future
- Alpenglow sigverify stage is not currently hooked up to anything. It will be linked up to the voting loop in a future PR
- We are not currently sending any votes to this port (still sending to "normal" TPU vote port)


alexpyattaev:
Just a nit, but maybe we should not name things "alpenglow_receiver" and such? Today it makes 100% sense, but after a year alpenglow will be the norm, tower will be evicted, and one would be looking for a more descriptive name, along the lines of "vote_receiver" or "consensus_msg_receiver".

wen-coding:
> Just a nit, but maybe we should not name things "alpenglow_receiver" and such? Today it makes 100% sense, but after a year alpenglow will be the norm, tower will be evicted, and one would be looking for a more descriptive name, along the lines of "vote_receiver" or "consensus_msg_receiver".

Totally agreed. However, we already have vote_receiver everywhere for TowerBFT, which makes the code confusing during the migration. What do you feel if we just keep it alpenglow_receiver for now, then rename when we remove TowerBFT code?

core/src/tpu.rs
```diff
@@ -137,18 +141,24 @@ impl Tpu {
...
             transactions_forwards_quic: transactions_forwards_quic_sockets,
             vote_quic: tpu_vote_quic_sockets,
+            alpenglow: alpenglow_socket,
         } = sockets;
 
         let (packet_sender, packet_receiver) = unbounded();
         let (vote_packet_sender, vote_packet_receiver) = unbounded();
         let (forwarded_packet_sender, forwarded_packet_receiver) = unbounded();
+        //TODO(wen): we should actually send the messages to voting loop.
+        let (bls_packet_sender, bls_packet_receiver) = unbounded();
```
alexpyattaev: Do we have to use an unbounded channel here? 

Unbounded queues can hide nasty performance regressions, and are generally far slower than their bounded equivalents, especially under heavy load. In this particular case, a bounded channel of capacity 2048 is perfectly sufficient to buffer all packets from an entire voting round. If we can not keep up with consensus packet ingress, so many things have gone wrong it makes no difference if we drop a few packets due to queue overflow.

core/src/tpu.rs
```diff
@@ -137,18 +141,24 @@ impl Tpu {
...
+        let (alpenglow_bls_message_sender, _) = unbounded();
```
alexpyattaev: If we are defining a sender and no receiver for an unbounded channel, are we not just creating a memory leak (messages get sent -> get stuck in channel -> can never get evicted)?

wen-coding: Yes, the receiver will be defined in another PR later, this is just to make the PR smaller to review.

core/src/tpu.rs
wen-coding: What you said makes sense. We've traditionally used bounded at the packet ingestion level, I guess mostly due to fear of dropping important packets if someone floods us. I would prefer to keep the packet_receiver unbounded, without sigverify you don't even know which packets to discard. We can maybe define a bound later on alpenglow_bls_message_sender.

alexpyattaev: If this is before sigverify this MUST be bounded, else we are open to OOM attack by just flooding a port with packets. If you are concerned about drops, make capacity larger, buffering even 10000 packets is just a 10MB of RAM. But there has to be a reasonable upper bound for security reasons. If the buffer is full while packets are still coming, it is a good idea to have a counter of overflowing packets and report that via metrics, this way we'd know if we are getting issues (or are under attack).

core/src/tpu.rs
alexpyattaev: Well if you just make this channel bounded there would be no issue - it would just fill up and stop eating RAM at that point => perfectly safe to keep running for weeks. Just make sure writer side uses `try_send()` rather than blocking.

core/src/tpu.rs
wen-coding: Hmm okay, I change both to be bounded buffer of 10000.

core/src/tpu.rs
wen-coding: Changed.

core/src/tpu.rs
alexpyattaev: Thank you!

core/src/tpu.rs
alexpyattaev: Thank you!

core/src/fetch_stage.rs
```diff
@@ -34,21 +34,30 @@ impl FetchStage {
...
-    ) -> (Self, PacketBatchReceiver, PacketBatchReceiver) {
+    ) -> (
+        Self,
+        PacketBatchReceiver,
+        PacketBatchReceiver,
+        PacketBatchReceiver,
+    ) {
         let (sender, receiver) = unbounded();
         let (vote_sender, vote_receiver) = unbounded();
+        let (alpenglow_sender, alpenglow_receiver) = unbounded();
```
alexpyattaev: This channel should be bounded (it is fed by streamer => can grow out of control if not emptied in time)

wen-coding: Changed.

wen-coding:
We originally thought they would be transactions and packed them into
blocks. That will go away in the following BLS PRs.

alexpyattaev:
Thank you, I think from the packet handling PoV this is good to go!

validator/src/commands/run/execute.rs
```diff
@@ -1210,6 +1210,7 @@ pub fn execute(
         node.info.remove_tpu_forwards();
         node.info.remove_tvu();
         node.info.remove_serve_repair();
+        node.info.remove_alpenglow();
```
gregcusack, line 1213: thank you for adding this in!

core/src/tpu.rs
```diff
@@ -60,6 +62,9 @@ use {
     tokio::sync::mpsc::Sender as AsyncSender,
 };
 
+// The maximum number of alpenglow packets that can be processed in a single batch
+pub const MAX_ALPENGLOW_PACKET_NUM: usize = 10000;
```
gregcusack, line 66: how was this constant calculated? just want to make sure that 1) it's not too small and 2) if it is too small and we start dropping packets, how bad is that?

wen-coding: 1. Assuming we have 4000 validators, actually we should have mostly 1 packet from everyone (1 vote after block reaches everywhere, then 1 cert after vote reaches everywhere) each time. So in the normal cases we should have at most 4000 packets. 10000 > 4000 * 2.
However, I admit this is very brittle protection. In the future we should have symmetric auth connections between any pair of validators and rate limit on the connection instead.
2. I think the hope is that even if packets are dropped because of DoS, some good validators can get enough votes to create a cert, and then some good validator will succeed to send the cert to the other good validator. We currently don't use a protocol that will re-send lost packets, we probably should. This is just a stop-gap for now, definitely not the full solution.

alexpyattaev: SNP will become a thing before alpenglow is ready to hit mainnet. So symmetric auth will be available.

gossip/src/contact_info.rs
```diff
@@ -1021,7 +1030,7 @@ mod tests {
         let socket = repeat_with(|| new_rand_socket(&mut rng))
             .filter(|socket| matches!(sanitize_socket(socket), Ok(())))
             .find(|socket| socket.port().checked_add(11).is_some())
-            .unwrap();
+            .expect("Failed to find a valid socket with port + 11 in test_new_with_socketaddr");
```
ksn6: Why are we changing this test?

core/src/tpu.rs
```diff
@@ -68,12 +73,14 @@ pub struct TpuSockets {
     pub transactions_quic: Vec<UdpSocket>,
     pub transactions_forwards_quic: Vec<UdpSocket>,
     pub vote_quic: Vec<UdpSocket>,
+    pub alpenglow: UdpSocket,
```
ksn6, line 76: Unless we use this socket for multiple reasons, could we call this socket `alpenglow_bls` rather than just `alpenglow`?

core/src/tpu.rs
```diff
@@ -68,12 +73,14 @@ pub struct TpuSockets {
...
 }
 
 pub struct Tpu {
     fetch_stage: FetchStage,
     sigverify_stage: SigVerifyStage,
     vote_sigverify_stage: SigVerifyStage,
+    alpenglow_sigverify_stage: SigVerifyStage,
```
ksn6, line 83: Same here - how about `alpenglow_bls_sigverify_stage`?

core/src/tpu.rs
```diff
@@ -137,18 +144,24 @@ impl Tpu {
...
             transactions_forwards_quic: transactions_forwards_quic_sockets,
             vote_quic: tpu_vote_quic_sockets,
+            alpenglow: alpenglow_socket,
         } = sockets;
 
         let (packet_sender, packet_receiver) = unbounded();
         let (vote_packet_sender, vote_packet_receiver) = unbounded();
         let (forwarded_packet_sender, forwarded_packet_receiver) = unbounded();
+        //TODO(wen): we should actually send the messages to voting loop.
+        let (bls_packet_sender, bls_packet_receiver) = bounded(MAX_ALPENGLOW_PACKET_NUM);
```
ksn6, line 153: Why are we okay with the channels on L150-152 with being unbounded while the alpenglow channels need to be bounded to 10_000?

core/src/tpu.rs
alexpyattaev: As I understand it, a single alpenglow socket would be sufficient. We already have sockets for repair protocol.

core/src/tpu.rs
alexpyattaev: We are not ok with those, but bounding them requires extensive testing that we are not dropping anything important, and validators are extremely sensitive to losses of vote packets (due to how rewards are structured). We just did not have enough time to find appropriate bound for those.

gossip/src/contact_info.rs
wen-coding: It's just more descriptive because I ran into it. But it's rare enough, I'll revert this change.

core/src/tpu.rs
gregcusack: > 1. Assuming we have 4000 validators, actually we should have mostly 1 packet from everyone (1 vote after block reaches everywhere, then 1 cert after vote reaches everywhere) each time. So in the normal cases we should have at most 4000 packets. 10000 > 4000 * 2.

This feels good to me. Assume this will be tested pretty heavily before prod so feels like a reasonable size for now. 

>    However, I admit this is very brittle protection. In the future we should have symmetric auth connections between any pair of validators and rate limit on the connection instead.

ahh ya this will be good

gossip/src/cluster_info.rs
```diff
@@ -4400,15 +4415,15 @@ mod tests {
 
         let trace = cluster_info44.contact_info_trace();
         info!("cluster:\n{}", trace);
-        assert_eq!(trace.len(), 431);
+        assert_eq!(trace.len(), 452);
```
bw-solana: what's the point of verifying this? Do we actually rely on some hard coded size anywhere? Or just that it's less than MTU?

If we care, I would think this should live as a constant with some comment indicating why the size matters.

Sort of outside the scope of this PR, but it caught my eye

core/src/tpu.rs
```diff
@@ -137,18 +144,24 @@ impl Tpu {
             transactions_quic: transactions_quic_sockets,
             transactions_forwards_quic: transactions_forwards_quic_sockets,
             vote_quic: tpu_vote_quic_sockets,
+            alpenglow: alpenglow_socket,
         } = sockets;
 
         let (packet_sender, packet_receiver) = unbounded();
         let (vote_packet_sender, vote_packet_receiver) = unbounded();
         let (forwarded_packet_sender, forwarded_packet_receiver) = unbounded();
+        //TODO(wen): we should actually send the messages to voting loop.
```
bw-solana: this TODO applies to the next line, right? We would expect voting loop to hook up to `alpenglow_bls_message_receiver`

Any reason to not do that with this PR? Or should we split out the fetching from Alpenglow port into its own PR entirely?

core/src/tpu.rs
bw-solana: What about the non-typical case where we do something like initial vote --> fallback --> cert?

Seems like it would be good to either:

1. Bump this up to something higher (e.g. 100k), add metrics to record max channel length every N second interval, and look at the metrics to understand any outliers where channels grows very large. We could then ratchet this down after testing informed by what we see in practice.
2. Report error (maybe even panic for now) if we ever see packets fail to send because channel is full. Treat channel full as a bug that needs to be solved w/ haste.
3. Commit to using a vote sending protocol that will retry (I think this means we would have to ack at the very end to ensure we don't drop at any point in the pipeline, which seems bad)

bw-solana (commented):
Left a few suggestions, but overall looks good to me.

I think it would be good to add more context to the PR description for posterity. For example:

- this port is currently UDP, but I believe we're expecting this to be QUIC or SNP based in the future
- might also want to comment on how many receive ports/threads we think we should support
- the fact that we're still doing nothing with these votes and still need to hook up sigverify to vote loop

gossip/src/cluster_info.rs
wen-coding: I think Stephen spent some time because someone broke contact_info_trace so the two lines don't align any more, so he added this test as a reminder: https://github.com/anza-xyz/agave/commit/b3fd87fe8189e7e278d882eb2cacafe1f9503d89

That said, I agree this is brittle protection.

core/src/tpu.rs
wen-coding: Oh, good catch.

I was trying to make the PR really small, because it touches Gossip I want Alex and Greg to review it. I think it would be good to show how the new socket is being used, linking in fetching should be less controversial and hopefully won't drag this PR for too long.

Linking into voting loop would take weeks to review.... please check https://github.com/anza-xyz/alpenglow/pull/203 for all the places touching voting loop.

That said, I can remove fetching if you feel strongly about it. But we already had long conversations about setting a cap on the channels, so would be nice to keep them here.

bw-solana: I'm okay with what you have 👍

core/src/tpu.rs
wen-coding: As discussed off-line, this number is a batch, we should definitely process one batch within 10ms, so initial vote -> fallback -> cert is fine, they should be in different batches. Guess my comment is a bit misleading.

We will add per-connection limit which is the real protection. Under normal cases we should never get as close to 10k. If we have a bug like Turbine shred forwarder where a packet is repeatedly sent back to the channel, then even 100k is not enough.

And this number should change after we implement actual BLS verification and have more experience with this part.

Yes, switching to Quic or SNP is scheduled to happen before this thing even lands in production.

I'll add more error reporting here.

wen-coding: Okay, in this PR we added three channels with bound:
alpenglow_sender/receiver: line 52 in fetch_stage.rs
bls_packet_sender/receiver: line 153 in tpu.rs

The above two go into the same location through different code path:
FetchStage::new() or FetchStage::new_with_sender(), they all end in recv_loop of streamer.rs (line 163). FetchStage already has pretty good statistics and error reporting, so we are covered.

The other one we are adding is alpenglow_bls_message_sender/receiver: line 156 in tpu.rs.

The error reporting for this actually is in BLSSigverifier. If I add error reporting now I need to add that to this PR, which makes this PR larger.

Added TODO, I actually have a plan to send another PR on BLSSigverifier maybe later today or tomorrow, because I needed to handle BLSMessage deserialize error there and add more error reporting and maybe add some unittests. Can I roll the error reporting on alpenglow_bls_message_sender into that PR instead?

wen-coding:
@carllin @AshwinSekar how does this PR look?

core/src/tpu.rs
wen-coding: @bw-solana https://github.com/anza-xyz/alpenglow/pull/206 created.

bw-solana:
> 1. I would advise against buffering 100K packets, there is no legitimate condition where this would make sense. Even if we concatenate 3 rounds of voting that is < 10K packets. And if we are under DoS attack there is no point buffering thousands of packets to process possible votes that will be too stale by the time we get to them through the stream of junk
> 2. Panic on a full channel is asking for someone to DoS the validator. Reporting error and/or metrics is reasonable though.
> 3. We have design that will retry sending votes periodically if there is no progress made. Any protocol with explicit ACK packets for votes will nearly double the vote bandwidth since vote packets are tiny, and acks will have to be sent immediately upon receiving the vote packet.

I haven't heard an argument I trust enough to place a bet on the liveness of mainnet re: how high this channel can spike and/or how dropping vote packets is okay.

Because of this, I think load shedding and praying without really good observability into what is happening at this stage of maturity of Alpenglow is a really bad idea.

There are several ways we can get good observability:

1. Make it unbounded and capture max size periodically
2. Panic when exceeding some size we think is large enough
3. Print obnoxious error messages and/or metrics when try_send fails

I'm not partial to any particular solution, but we need something.

To be clear, I am not advocating obnoxiously high channel size or panicking when full in a production system. That's clearly an awful idea.

core/src/tpu.rs
bw-solana: To be clear, my understanding is we expect individual alpenglow votes in addition to BLS aggregated votes to be sent to this port

core/src/tpu.rs
bw-solana: I believe this stage will verify individual AG votes as well

core/src/tpu.rs
wen-coding: Yes, both votes and certs to the same port.

core/src/tpu.rs
wen-coding: Yes, that's right.

bw-solana:
> > > 1. I would advise against buffering 100K packets, there is no legitimate condition where this would make sense. Even if we concatenate 3 rounds of voting that is < 10K packets. And if we are under DoS attack there is no point buffering thousands of packets to process possible votes that will be too stale by the time we get to them through the stream of junk
> > > 2. Panic on a full channel is asking for someone to DoS the validator. Reporting error and/or metrics is reasonable though.
> > > 3. We have design that will retry sending votes periodically if there is no progress made. Any protocol with explicit ACK packets for votes will nearly double the vote bandwidth since vote packets are tiny, and acks will have to be sent immediately upon receiving the vote packet.
> > 
> > 
> > I haven't heard an argument I trust enough to place a bet on the liveness of mainnet re: how high this channel can spike and/or how dropping vote packets is okay.
> > Because of this, I think load shedding and praying without really good observability into what is happening at this stage of maturity of Alpenglow is a really bad idea.
> > There are several ways we can get good observability:
> > 
> > 1. Make it unbounded and capture max size periodically
> > 2. Panic when exceeding some size we think is large enough
> > 3. Print obnoxious error messages and/or metrics when try_send fails
> > 
> > I'm not partial to any particular solution, but we need something.
> > To be clear, I am not advocating obnoxiously high channel size or panicking when full in a production system. That's clearly an awful idea.
> 
> Let me state again:
> 
> 1. Alpenglow will _not_ go into production (not even testnet) as is, this is just an initial number we use for load testing. We will add per-channel rate limits and adjust this number after load testing. Because this number may change depending on how we design and implement BLS batch verification. No one is doing load shedding and praying here.
> 2. Obnoxious error messages and metrics are added in [Add error handling to BLS Sigverifier. #206](https://github.com/anza-xyz/alpenglow/pull/206)

This was a response to Alex. The key is that you added observability with #206 . We're on the same page

core/src/tpu.rs
```diff
@@ -137,18 +144,25 @@ impl Tpu {
             transactions_quic: transactions_quic_sockets,
             transactions_forwards_quic: transactions_forwards_quic_sockets,
             vote_quic: tpu_vote_quic_sockets,
+            alpenglow: alpenglow_socket,
         } = sockets;
 
         let (packet_sender, packet_receiver) = unbounded();
         let (vote_packet_sender, vote_packet_receiver) = unbounded();
         let (forwarded_packet_sender, forwarded_packet_receiver) = unbounded();
+        let (bls_packet_sender, bls_packet_receiver) = bounded(MAX_ALPENGLOW_PACKET_NUM);
```
bw-solana, line 153: `bls_packet_*` naming is a little deceiving given we accept non-BLS votes as well. Maybe `alpenglow_vote_sanitized_*` ?

core/src/tpu.rs
```diff
@@ -137,18 +144,25 @@ impl Tpu {
...
+        //TODO(wen): we should actually send the messages to voting loop.
+        //TODO(wen): actually add error reporing to BLS sigverifier.
+        let (alpenglow_bls_message_sender, _) = bounded(MAX_ALPENGLOW_PACKET_NUM);
```
bw-solana: similarly here, maybe `alpenglow_vote_verified_*`?

gossip/src/cluster_info.rs
```diff
@@ -4400,15 +4415,15 @@ mod tests {
...
 
         let trace = cluster_info44.rpc_info_trace();
         info!("rpc:\n{}", trace);
         assert_eq!(trace.len(), 335);
 
         let trace = cluster_info43.contact_info_trace();
         info!("cluster:\n{}", trace);
-        assert_eq!(trace.len(), 431);
+        assert_eq!(trace.len(), 452);
```
bw-solana: would be nice to add a local constant to ensure it is clear that these must match (the whole intent of the test)

core/src/tpu.rs
wen-coding: The votes will be BLS messages as well, we need to verify the BLS signatures in votes to make sure they are from authorized party, we also need to aggregate the BLS signatures in votes into the BLS signatures in certificates.

That said, do you still want to rename this?

gossip/src/cluster_info.rs
wen-coding: Done.

core/src/tpu.rs
bw-solana: ah, makes sense. `bls_` prefix sounds fine to me in this case.

We have two channels to think about now:

Fetch --> sigverify --> voting loop

We currently call these:
`bls_packet_*`
`alpenglow_bls_message_*`

How should we differentiate what is being sent in each case?

Seems like the important piece is that we are sending lightly sanitized packets between fetch/sigverify and we are sending deduped/signature verified packets between sigverify/voting loop.

Current naming convention w/ Packet vs. Message doesn't make this distinction abundantly clear to me. But this is a bit of a bikeshed, so if everyone else is happy w/ the naming, I'll get over it

wen-coding: bls_packet_* are currently for raw BLS packets (pre-sigverify)
alpenglow_bls_message_* are for deserialized and sigverified BLS messages, does it make sense to rename them:
bls_raw_packet_*
bls_message_sanitized_*
?

wen-coding: I'll keep the bls_packet_* names for now to be the same as other packet_sender/receiver, renaming alpenglow_bls_message_* to bls_verified_message_*

wen-coding: Changed.

bw-solana: that sounds reasonable to me

bw-solana (approved):

PR #204, opened 2025-06-07
feat: optionally remove self from staked validators cache

ksn6:
#### Problem
`StakedValidatorsCache` always has the locally running validator's TPU vote socket in its cache entry.

#### Summary of Changes
Add the ability for the locally running validator to opt out of adding its own TPU vote socket to the cache entry. This avoids the validator from sending a vote to itself.

### DIFF
--- a/core/src/staked_validators_cache.rs
+++ b/core/src/staked_validators_cache.rs
@@ -40,7 +40,7 @@ pub struct StakedValidatorsCache {
     /// Protocol
     protocol: Protocol,
 
-    /// Whether to include the running validator's socket address in the list of staked validators
+    /// Whether to include the running validator's socket address in cache entries
     include_self: bool,
 }
 
@@ -96,12 +96,9 @@ impl StakedValidatorsCache {
             .iter()
             .filter(|(pubkey, stake)| {
                 let positive_stake = **stake > 0;
+                let not_self = pubkey != &&cluster_info.id();
 
-                if self.include_self {
-                    positive_stake
-                } else {
-                    positive_stake && pubkey != &&cluster_info.id()
-                }
+                positive_stake && (self.include_self || not_self)
             })
             .filter_map(|(pubkey, stake)| {
                 cluster_info
@@ -550,18 +547,15 @@ mod tests {
         assert!(refreshed);
     }
 
-    #[test_case(325_000_000_u64, 1_usize, 10_usize, 123_u64, Protocol::UDP)]
-    #[test_case(325_000_000_u64, 3_usize, 10_usize, 123_u64, Protocol::QUIC)]
-    #[test_case(325_000_000_u64, 10_usize, 10_usize, 123_u64, Protocol::UDP)]
-    #[test_case(325_000_000_u64, 10_usize, 10_usize, 123_u64, Protocol::QUIC)]
-    #[test_case(325_000_000_u64, 50_usize, 60_usize, 123_u64, Protocol::UDP)]
-    fn test_exclude_self_from_cache(
-        slot_num: u64,
-        num_nodes: usize,
-        num_vote_accounts: usize,
-        genesis_lamports: u64,
-        protocol: Protocol,
-    ) {
+    #[test_case(1_usize, Protocol::UDP)]
+    #[test_case(1_usize, Protocol::QUIC)]
+    #[test_case(10_usize, Protocol::UDP)]
+    #[test_case(10_usize, Protocol::QUIC)]
+    fn test_exclude_self_from_cache(num_nodes: usize, protocol: Protocol) {
+        let slot_num = 325_000_000_u64;
+        let num_vote_accounts = 10_usize;
+        let genesis_lamports = 123_u64;
+
         // Create our harness
         let (keypair_map, vahm) = build_epoch_stakes(num_nodes, 0, num_vote_accounts);
 


core/src/staked_validators_cache.rs
```diff
@@ -39,6 +39,9 @@ pub struct StakedValidatorsCache {
 
     /// Protocol
     protocol: Protocol,
+
+    /// Whether to include the running validator's socket address in the list of staked validators
```
wen-coding:
```suggestion
    /// Whether to include the running validator's socket address
```

core/src/staked_validators_cache.rs
```diff
@@ -89,7 +94,15 @@ impl StakedValidatorsCache {
...
             .iter()
-            .filter(|(_, stake)| **stake > 0)
+            .filter(|(pubkey, stake)| {
+                let positive_stake = **stake > 0;
+
+                if self.include_self {
+                    positive_stake
+                } else {
+                    positive_stake && pubkey != &&cluster_info.id()
+                }
```
wen-coding: nit: we can also do
```
let positive_stake = **stake > 0;
let not_self = pubkey != &cluster_info.id();
is_positive && (self.include_self || not_self)
```

I think there is a slight chance cluster_info.id() doesn't match pubkey if you are doing set-identity, but it's not end of the world if I sometimes send a few packets to myself if I do set-identity, so I think this is good enough.

core/src/staked_validators_cache.rs
```diff
@@ -378,12 +390,14 @@ mod tests {
...
-            .bank_forks();
+        let (bank_forks, cluster_info) =
+            StakedValidatorsCacheHarness::new(&genesis_config, Keypair::new())
+                .with_vote_accounts(slot_num, keypair_map, vahm, protocol)
+                .bank_forks();
 
         // Create our staked validators cache
-        let mut svc = StakedValidatorsCache::new(bank_forks, protocol, Duration::from_secs(5), 5);
+        let mut svc =
+            StakedValidatorsCache::new(bank_forks, protocol, Duration::from_secs(5), 5, true);
```
wen-coding, line 397: nit: this is a test, I guess we don't need 5 seconds here? Making it 1 second can make the test run faster?

core/src/staked_validators_cache.rs
```diff
@@ -532,4 +549,62 @@ mod tests {
         let (_, refreshed) = svc.get_staked_validators_by_slot(2 * slot_num, &cluster_info, now);
         assert!(refreshed);
     }
+
+    #[test_case(325_000_000_u64, 1_usize, 10_usize, 123_u64, Protocol::UDP)]
+    #[test_case(325_000_000_u64, 3_usize, 10_usize, 123_u64, Protocol::QUIC)]
+    #[test_case(325_000_000_u64, 10_usize, 10_usize, 123_u64, Protocol::UDP)]
+    #[test_case(325_000_000_u64, 10_usize, 10_usize, 123_u64, Protocol::QUIC)]
+    #[test_case(325_000_000_u64, 50_usize, 60_usize, 123_u64, Protocol::UDP)]
+    fn test_exclude_self_from_cache(
```
wen-coding: nit: This new test is a function test, I guess we don't need to test all combinations, just testing one should be more than enough.

core/src/staked_validators_cache.rs
ksn6: Would rather leave the more detailed comment. Updated it for clarity.

core/src/staked_validators_cache.rs
ksn6: I originally had these three lines haha. I think it's a bit more clear the current way by expanding the code out more explicitly.

core/src/staked_validators_cache.rs
ksn6: All tests in `staked_validators_cache.rs` use simulated time, so e.g., even increasing 5 seconds to 60 seconds wouldn't make a difference; most of these tests take at most a few milliseconds to run.

core/src/staked_validators_cache.rs
ksn6: I'm testing that this works for n = 1, 3, 10, and 50 validators - probably a good idea to test for edge cases, just in case.

EDIT: these tests collectively take ~0.12 seconds to run; not too much overhead.

core/src/staked_validators_cache.rs
wen-coding: Sure, my comment was mostly about this line being a bit long.

core/src/staked_validators_cache.rs
wen-coding: The current logic is very short, so it's probably fine either way. But I think your original three lines are much clearer, because "let not_self = pubkey != &cluster_info.id();" clearly explains your intention on doing the test, so if some day a newbie comes in and reads the code, this is totally self explanatory.
The current code, on the other hand, needs the code reader to dig in and see why the logic is different when self.include_self is false, which IMHO is a bit less self explanatory.

core/src/staked_validators_cache.rs
wen-coding: I don't feel strongly about this one case, probably fine either way.

That said, tests are the supplementary documents of code. If you list this many combinations, I would wonder whether they make a difference in your logic. (That is, why don't we test more combinations, why these numbers.)
I think testing 1 and multiple validators totally makes sense, say 1 and 10. But I don't think the code path changes a lot on 3 vs 10 validators. If I need to add more tests, it probably makes more sense to add some contact_info with no tpu_vote port than to test 3 vs 10 validators here. Because that adds more test coverage. Just my 2c.

core/src/staked_validators_cache.rs
ksn6: Sure - no strong opinions on this from me.

core/src/staked_validators_cache.rs
ksn6: Cool - keeping 1 and 10, but changing between UDP and QUIC for now.

wen-coding (approved):

PR #205, opened 2025-06-08
Add option to push BLS message to Alpenglow sockets in voting service.

wen-coding:
Cache Alpenglow sockets and use them when pushing BLS messages in voting service.

Note that this depends on https://github.com/anza-xyz/alpenglow/pull/202.


core/src/staked_validators_cache.rs
```diff
@@ -101,24 +105,40 @@ impl StakedValidatorsCache {
...
                         stake: *stake,
-                        tpu_socket: socket_addr,
+                        tpu_socket,
+                        alpenglow_socket,
                     })
+                })?
             })
             .collect();
 
         nodes.dedup_by_key(|node| node.tpu_socket);
```
bw-solana: should we dedup by ag socket as well?

core/src/staked_validators_cache.rs
```diff
@@ -368,18 +403,77 @@ mod tests {
...
+        Protocol::UDP,
+        true
+    )]
+    #[test_case(
+        325_000_000_u64,
+        50_usize,
+        7_usize,
+        60_usize,
+        123_u64,
+        Protocol::UDP,
```
bw-solana: Is the plan to move to QUIC for AG votes at some point? We will likely move to QUIC for tower BFT votes before AG goes live

bw-solana (commented):
Looks reasonable to me, but it would be good to add some more context to this PR. Especially around:

- UDP vs. QUIC for these sockets. This may have related more to #202, but still.. would be good to leave a paper trail on the thought process there.
- Noting that we are not actually sending any BLS votes outside of unit tests and that hooking this up is a TODO item
- Plans to retire Alpenglow votes once BLS messages supplant them

core/src/staked_validators_cache.rs
wen-coding: Yes, move to QUIC or SNP. Alex thought we should move to SNP because it allows upgrade better, but guess we will wait a bit to see how fast SNP moves.

wen-coding:
> Looks reasonable to me, but it would be good to add some more context to this PR. Especially around:
> 
> * UDP vs. QUIC for these sockets. This may have related more to [Add alpenglow port and alpenglow sigverify stage. #202](https://github.com/anza-xyz/alpenglow/pull/202), but still.. would be good to leave a paper trail on the thought process there.
> * Noting that we are not actually sending any BLS votes outside of unit tests and that hooking this up is a TODO item
> * Plans to retire Alpenglow votes once BLS messages supplant them

Sure, added to the PR description.

core/src/staked_validators_cache.rs
wen-coding: When we get rid of pushing of Alpenglow vote transactions we can do that, because at that time tpu_socket here can totally be removed. TowerBFT votes are only pushed to next leaders.

I'll add TODO here.

PR #206, opened 2025-06-10
Add error handling to BLS Sigverifier.

wen-coding:
Add error handling to BLS Sigverifier.

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -7,8 +7,9 @@ use {
     crossbeam_channel::{Sender, TrySendError},
     solana_streamer::packet::PacketBatch,
     std::{
-        sync::atomic::{AtomicI64, Ordering},
-        time::Instant,
+        mem,
+        sync::{Arc, Mutex},
+        time::{Duration, Instant},
     },
 };
 
@@ -16,17 +17,17 @@ const STATS_INTERVAL_SECONDS: u64 = 10; // Log stats every 10 second
 
 // We are adding our own stats because we do BLS decoding in batch verification,
 // and we send one BLS message at a time. So it makes sense to have finer-grained stats
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
 pub(crate) struct BLSSigVerifierStats {
-    pub bls_messages_sent: AtomicI64,
-    pub bls_messages_malformed: AtomicI64,
-    pub sent_failed: AtomicI64,
-    pub packets_received: AtomicI64,
+    pub bls_messages_sent: u64,
+    pub bls_messages_malformed: u64,
+    pub sent_failed: u64,
+    pub packets_received: u64,
 }
 
 pub struct BLSSigVerifier {
-    sender: Option<Sender<BLSMessage>>,
-    stats: BLSSigVerifierStats,
+    sender: Sender<BLSMessage>,
+    stats: Arc<Mutex<BLSSigVerifierStats>>,
     last_stats_logged: Instant,
 }
 
@@ -42,33 +43,40 @@ impl SigVerifier for BLSSigVerifier {
         packet_batches: Vec<PacketBatch>,
     ) -> Result<(), SigVerifyServiceError<Self::SendType>> {
         // TODO(wen): just a placeholder without any batching.
-        if let Some(sender) = &self.sender {
-            packet_batches.iter().for_each(|batch| {
-                batch.iter().for_each(|packet| {
-                    self.stats.packets_received.fetch_add(1, Ordering::Relaxed);
-                    match packet.deserialize_slice::<BLSMessage, _>(..) {
-                        Ok(message) => match sender.try_send(message) {
-                            Ok(()) => {
-                                self.stats.bls_messages_sent.fetch_add(1, Ordering::Relaxed);
-                            }
-                            Err(TrySendError::Full(_)) => {
-                                warn!("BLS message channel is full, dropping message");
-                                self.stats.sent_failed.fetch_add(1, Ordering::Relaxed);
-                            }
-                            Err(TrySendError::Disconnected(_)) => {
-                                // There is no hope to recover, restart the validator.
-                                panic!("BLS message channel is disconnected");
-                            }
-                        },
-                        Err(e) => {
-                            trace!("Failed to deserialize BLS message: {}", e);
-                            self.stats
-                                .bls_messages_malformed
-                                .fetch_add(1, Ordering::Relaxed);
+        let mut packet_received = 0;
+        let mut bls_messages_sent = 0;
+        let mut bls_messages_malformed = 0;
+        let mut sent_failed = 0;
+        packet_batches.iter().for_each(|batch| {
+            batch.iter().for_each(|packet| {
+                packet_received += 1;
+                match packet.deserialize_slice::<BLSMessage, _>(..) {
+                    Ok(message) => match self.sender.try_send(message) {
+                        Ok(()) => {
+                            bls_messages_sent += 1;
                         }
+                        Err(TrySendError::Full(_)) => {
+                            warn!("BLS message channel is full, dropping message");
+                            sent_failed += 1;
+                        }
+                        Err(TrySendError::Disconnected(_)) => {
+                            // There is no hope to recover, restart the validator.
+                            panic!("BLS message channel is disconnected");
+                        }
+                    },
+                    Err(e) => {
+                        trace!("Failed to deserialize BLS message: {}", e);
+                        bls_messages_malformed += 1;
                     }
-                });
+                }
             });
+        });
+        if packet_received > 0 {
+            let mut stats = self.stats.lock().unwrap();
+            stats.packets_received += packet_received;
+            stats.bls_messages_sent += bls_messages_sent;
+            stats.bls_messages_malformed += bls_messages_malformed;
+            stats.sent_failed += sent_failed;
         }
         self.report_stats();
         Ok(())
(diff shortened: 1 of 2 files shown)


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -28,18 +45,223 @@ impl SigVerifier for BLSSigVerifier {
         if let Some(sender) = &self.sender {
```
ksn6: Would avoid deep nesting + usage of `.iter().for_each()` unless using a builder pattern or something. How about instead:

```
    fn send_packets(
        &mut self,
        packet_batches: Vec<PacketBatch>,
        let Some(sender) = &self.sender else {
            self.report_stats();
            return Ok(());
        };

        for batch in &packet_batches {
            for packet in batch {
                self.stats.packets_received.fetch_add(1, Ordering::Relaxed);

                let message = match packet.deserialize_slice::<BLSMessage, _>(..) {
                    Ok(message) => message,
                    Err(e) => {
                        trace!("Failed to deserialize BLS message: {}", e);
                        self.stats.bls_messages_malformed.fetch_add(1, Ordering::Relaxed);
                        continue;
                    }
                };

                match sender.try_send(message) {
                    Ok(()) => {
                        self.stats.bls_messages_sent.fetch_add(1, Ordering::Relaxed);
                    }
                    Err(TrySendError::Full(_)) => {
                        warn!("BLS message channel is full, dropping message");
                        self.stats.sent_failed.fetch_add(1, Ordering::Relaxed);
                    }
                    Err(TrySendError::Disconnected(_)) => {
                        // There is no hope to recover, restart the validator.
                        panic!("BLS message channel is disconnected");
                    }
                }
            }
        }
        self.report_stats();
        Ok(())
    }
```

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -4,13 +4,30 @@
...
+    },
 };
 
-#[derive(Default, Clone)]
+const STATS_INTERVAL_SECONDS: u64 = 10; // Log stats every 10 second
+
+// We are adding our own stats because we do BLS decoding in batch verification,
+// and we send one BLS message at a time. So it makes sense to have finer-grained stats
+#[derive(Debug, Default)]
+pub(crate) struct BLSSigVerifierStats {
```
ksn6, line 17: s/AtomicI64/AtomicU64

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -28,18 +45,223 @@ impl SigVerifier for BLSSigVerifier {
...
             });
         }
+        self.report_stats();
         Ok(())
     }
 }
 
 impl BLSSigVerifier {
+    fn report_stats(&mut self) {
+        if Instant::now()
```
ksn6: To make this more readable:

```
let now = Instant::now();
let time_since_last_stats_logged = now.duration_since(self.last_stats_logged).as_secs();

if time_since_last_stats_logged < STATS_INTERVAL_SECONDS {
   return;
}

...

self.last_stats_logged = now;
```

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -28,18 +45,223 @@ impl SigVerifier for BLSSigVerifier {
...
+            >= STATS_INTERVAL_SECONDS
+        {
+            let stats = &mut self.stats;
+            let (sent, sent_failed, malformed, received) = (
+                stats.bls_messages_sent.swap(0, Ordering::Relaxed),
+                stats.sent_failed.swap(0, Ordering::Relaxed),
+                stats.bls_messages_malformed.swap(0, Ordering::Relaxed),
+                stats.packets_received.swap(0, Ordering::Relaxed),
+            );
+            datapoint_info!(
```
ksn6: Because we're using `Ordering::Relaxed`, the reads below could happen before the swaps above.

We should throw the above swaps behind a Mutex / ensure such that either all or none of them are run.

wen-coding: Changed.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Changed.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: This is now u64.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: That's a good point, however, now I wonder why we need to make sender Option, it's probably to make unittest easy, but giving an actual sender isn't that bad, I'm changing this to not be an option.

core/src/sigverifier/bls_sigverifier.rs
ksn6: wdym? we don't need atomics anymore?

wen-coding: Now it's a big mutex lock on the whole struct, do we still need atomics?

wen-coding:
@AshwinSekar @carllin @samkim-crypto Any more comments on this PR?

wen-coding:
@carllin @AshwinSekar @samkim-crypto @bw-solana I'm submitting this now so I can rebase other PRs. If you have more concerns, feel free to comment on this PR and I'll write another PR to address your concerns.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -4,13 +4,27 @@
...
+    pub bls_messages_malformed: u64,
+    pub sent_failed: u64,
+    pub packets_received: u64,
+}
+
 pub struct BLSSigVerifier {
-    sender: Option<Sender<BLSMessage>>,
+    sender: Sender<BLSMessage>,
+    stats: BLSSigVerifierStats,
+    last_stats_logged: Instant,
```
bw-solana, line 27: I think it would be better if this was part of the `BLSSigVerifierStats` struct

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
+        // We don't need lock on stats because stats are read and write in a single thread.
+        self.report_stats();
         Ok(())
     }
 }
 
 impl BLSSigVerifier {
-    pub fn new(sender: Option<Sender<BLSMessage>>) -> Self {
-        Self { sender }
+    fn report_stats(&mut self) {
```
bw-solana, line 83: I think we conventionally use something like `maybe_report_stats` when there are conditionals inside. But I'm okay with whatever

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
+        let now = Instant::now();
+        let time_since_last_log = now.duration_since(self.last_stats_logged);
+        if time_since_last_log < Duration::from_secs(STATS_INTERVAL_SECONDS) {
+            return;
+        }
+        datapoint_info!(
+            "bls_sig_verifier_stats",
+            ("sent", self.stats.bls_messages_sent, u64),
+            ("sent_failed", self.stats.sent_failed, u64),
+            ("malformed", self.stats.bls_messages_malformed, u64),
```
bw-solana, line 93: maybe `received_malformed`?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
+    pub fn new(sender: Sender<BLSMessage>) -> Self {
+        Self {
+            sender,
+            stats: BLSSigVerifierStats::default(),
+            last_stats_logged: Instant::now(),
+        }
+    }
+
+    #[cfg(feature = "dev-context-only-utils")]
+    #[allow(dead_code)]
```
bw-solana, line 108: could we use `#[cfg(test)]` instead?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
+                    Err(e) => {
+                        trace!("Failed to deserialize BLS message: {}", e);
+                        bls_messages_malformed += 1;
                     }
-                });
+                }
             });
+        });
+        if packet_received > 0 {
+            self.stats.packets_received += packet_received;
```
bw-solana, line 71: might be nice to add a `self.stats.accumulate` function

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
+                        Ok(()) => {
+                            bls_messages_sent += 1;
+                        }
+                        Err(TrySendError::Full(_)) => {
+                            warn!("BLS message channel is full, dropping message");
+                            sent_failed += 1;
+                        }
+                        Err(TrySendError::Disconnected(_)) => {
+                            // There is no hope to recover, restart the validator.
+                            panic!("BLS message channel is disconnected");
```
bw-solana, line 60: how does the shutdown sequencing work? Are we careful to shut down this service before the receiver?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -4,13 +4,27 @@
...
+const STATS_INTERVAL_SECONDS: u64 = 10; // Log stats every 10 second
+
+// We are adding our own stats because we do BLS decoding in batch verification,
+// and we send one BLS message at a time. So it makes sense to have finer-grained stats
+#[derive(Debug, Default, Clone)]
+pub(crate) struct BLSSigVerifierStats {
+    pub bls_messages_sent: u64,
+    pub bls_messages_malformed: u64,
+    pub sent_failed: u64,
+    pub packets_received: u64,
```
bw-solana, line 21: I think we could rename these to match the reported metrics values you used - the bls_messages prefix seems obviated by the Struct name:
```
    pub sent: u64,
    pub received_malformed: u64,
    pub sent_failed: u64,
    pub received: u64,
```

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -4,13 +4,27 @@
...
     crate::sigverify_stage::{SigVerifier, SigVerifyServiceError},
     alpenglow_vote::bls_message::BLSMessage,
-    crossbeam_channel::Sender,
+    crossbeam_channel::{Sender, TrySendError},
     solana_streamer::packet::PacketBatch,
+    std::time::{Duration, Instant},
 };
 
-#[derive(Default, Clone)]
+const STATS_INTERVAL_SECONDS: u64 = 10; // Log stats every 10 second
```
bw-solana, line 12: 10 seconds seems like an eternity. Maybe 1 second?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -25,21 +39,224 @@ impl SigVerifier for BLSSigVerifier {
...
         // TODO(wen): just a placeholder without any batching.
-        if let Some(sender) = &self.sender {
-            packet_batches.iter().for_each(|batch| {
-                batch.iter().for_each(|packet| {
-                    if let Ok(message) = packet.deserialize_slice::<BLSMessage, _>(..) {
-                        sender.send(message).unwrap();
+        let mut packet_received = 0;
+        let mut bls_messages_sent = 0;
+        let mut bls_messages_malformed = 0;
+        let mut sent_failed = 0;
```
bw-solana, line 45: if we go the accumulate route, could replace these with a local `BLSSigVerifierStats`

bw-solana (commented):
Left a few suggestions. Mostly naming. Nothing fundamentally wrong with what you merged (unless the shutdown/panic is going to cause problems)

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Thanks for the review, since this one is already merged, I'll address your comments in https://github.com/anza-xyz/alpenglow/pull/210

wen-coding: Yeah this is me being lazy and wanted to use Default(), changed.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Good catch, I found out this can't be a panic, changed to error! in https://github.com/anza-xyz/alpenglow/pull/210

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
AshwinSekar: sorry for the late review, we should bubble up the error to sigverify stage so that send errors are properly handled here for shutdown:
https://github.com/anza-xyz/agave/blob/6b566ba9cd58a4cb08015744f9db1236ff2f2458/core/src/sigverify_stage.rs#L415

PR #207, opened 2025-06-12
Add option to push BLS message to Alpenglow sockets in voting service.

wen-coding:
Cache Alpenglow sockets and use them when pushing BLS messages in voting service.

- We will switch to QUIC or SNP in a later PR, this is just using UDP for the prototype testing
- Actually switching to BLS messages will happen in later PRs
- Once we switch to BLS messages, the original pushAlpenglowVote which sends vote transactions can be decommed

### DIFF
--- a/core/src/staked_validators_cache.rs
+++ b/core/src/staked_validators_cache.rs
@@ -240,7 +240,7 @@ mod tests {
             sync::{Arc, RwLock},
             time::{Duration, Instant},
         },
-        test_case::test_case,
+        test_case::{test_case, test_matrix},
     };
 
     fn new_rand_vote_account<R: rand::Rng>(
@@ -616,7 +616,7 @@ mod tests {
         assert!(refreshed);
     }
 
-#[test_matrix(
+    #[test_matrix(
     [1_usize, 10_usize],
     [Protocol::UDP, Protocol::QUIC],
     [false, true]
--- a/core/src/staked_validators_cache.rs
+++ b/core/src/staked_validators_cache.rs
@@ -47,6 +47,11 @@ pub struct StakedValidatorsCache {
     include_self: bool,
 }
 
+enum PortsToUse {
+    TpuVote,
+    Alpenglow,
+}
+
 impl StakedValidatorsCache {
     pub fn new(
         bank_forks: Arc<RwLock<BankForks>>,
@@ -93,6 +98,7 @@ impl StakedValidatorsCache {
         struct Node {
             stake: u64,
             tpu_socket: SocketAddr,
+            // TODO(wen): this should not be an Option after BLS all-to-all is submitted.
             alpenglow_socket: Option<SocketAddr>,
         }
 
@@ -144,27 +150,40 @@ impl StakedValidatorsCache {
         );
     }
 
-    pub fn get_staked_validators_by_slot(
+    pub fn get_staked_validators_by_slot_with_tpu_vote_ports(
         &mut self,
         slot: Slot,
         cluster_info: &ClusterInfo,
         access_time: Instant,
-        use_alpenglow_socket: bool,
     ) -> (&[SocketAddr], bool) {
         self.get_staked_validators_by_epoch(
             self.cur_epoch(slot),
             cluster_info,
             access_time,
-            use_alpenglow_socket,
+            PortsToUse::TpuVote,
         )
     }
 
-    pub fn get_staked_validators_by_epoch(
+    pub fn get_staked_validators_by_slot_with_alpenglow_ports(
+        &mut self,
+        slot: Slot,
+        cluster_info: &ClusterInfo,
+        access_time: Instant,
+    ) -> (&[SocketAddr], bool) {
+        self.get_staked_validators_by_epoch(
+            self.cur_epoch(slot),
+            cluster_info,
+            access_time,
+            PortsToUse::Alpenglow,
+        )
+    }
+
+    fn get_staked_validators_by_epoch(
         &mut self,
         epoch: Epoch,
         cluster_info: &ClusterInfo,
         access_time: Instant,
-        use_alpenglow_socket: bool,
+        ports_to_use: PortsToUse,
     ) -> (&[SocketAddr], bool) {
         // For a given epoch, if we either:
         //
@@ -187,12 +206,9 @@ impl StakedValidatorsCache {
             // self.cache[epoch].
             self.cache
                 .get(&epoch)
-                .map(|v| {
-                    if use_alpenglow_socket {
-                        &*v.alpenglow_sockets
-                    } else {
-                        &*v.validator_sockets
-                    }
+                .map(|v| match ports_to_use {
+                    PortsToUse::TpuVote => &*v.validator_sockets,
+                    PortsToUse::Alpenglow => &*v.alpenglow_sockets,
                 })
                 .unwrap(),
             refresh_cache,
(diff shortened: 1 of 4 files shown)


core/src/staked_validators_cache.rs
```diff
@@ -101,24 +105,41 @@ impl StakedValidatorsCache {
...
+        let (validator_sockets, alpenglow_sockets) = nodes.into_iter().fold(
+            (Vec::new(), Vec::new()),
+            |(mut validator_sockets, mut alpenglow_sockets), node| {
+                validator_sockets.push(node.tpu_socket);
+                if let Some(alpenglow_socket) = node.alpenglow_socket {
+                    alpenglow_sockets.push(alpenglow_socket);
+                }
+                (validator_sockets, alpenglow_sockets)
+            },
+        );
```
ksn6:
```suggestion
        let mut validator_sockets = Vec::new();
        let mut alpenglow_sockets = Vec::new();

        for node in nodes {
            validator_sockets.push(node.tpu_socket);
            if let Some(alpenglow_socket) = node.alpenglow_socket {
                alpenglow_sockets.push(alpenglow_socket);
            }
        }
```

core/src/staked_validators_cache.rs
```diff
@@ -129,15 +150,22 @@ impl StakedValidatorsCache {
         slot: Slot,
         cluster_info: &ClusterInfo,
         access_time: Instant,
+        use_alpenglow_socket: bool,
```
ksn6: Rather than having `use_alpenglow_socket` be passed into function definitions like this, could we instead:

- Have `use_alpenglow_socket` be a member variable of `StakedValidatorsCache`
- Supply a mutator that allows `use_alpenglow_socket` to be mutated?

I imagine that `use_alpenglow_socket` won't constantly be changing during a validator's lifetime, so this should make things a bit more clear.

core/src/staked_validators_cache.rs
```diff
@@ -368,18 +406,86 @@ mod tests {
...
+    )]
+    #[test_case(
+        325_000_000_u64,
+        50_usize,
+        7_usize,
+        60_usize,
+        123_u64,
+        Protocol::UDP,
+        true
+    )]
```
ksn6: Could we explicitly set `let slot_num = 325_000_000_u64` in the code below? (I should've done that for the other tests...) That should get all these `test_case`'s onto a single lines each.

core/src/staked_validators_cache.rs
```diff
@@ -537,21 +664,34 @@ mod tests {
...
-    #[test_case(1_usize, Protocol::QUIC)]
-    #[test_case(10_usize, Protocol::UDP)]
-    #[test_case(10_usize, Protocol::QUIC)]
-    fn test_exclude_self_from_cache(num_nodes: usize, protocol: Protocol) {
+    #[test_case(1_usize, Protocol::UDP, false)]
+    #[test_case(1_usize, Protocol::UDP, true)]
+    #[test_case(1_usize, Protocol::QUIC, false)]
+    #[test_case(10_usize, Protocol::UDP, false)]
+    #[test_case(10_usize, Protocol::UDP, true)]
+    #[test_case(10_usize, Protocol::QUIC, false)]
```
ksn6:
```suggestion
#[test_matrix(
    [1_usize, 10_usize],
    [Protocol::UDP, Protocol::QUIC],
    [false, true]
)]
```

core/src/voting_service.rs
```diff
@@ -191,6 +210,39 @@ impl VotingService {
...
+                            "Failed to send alpenglow message to {}: {:?}",
+                            alpenglow_socket, e
+                        );
+                    }
+                }
+            }
+            Err(err) => {
+                error!("Failed to serialize alpenglow message: {:?}", err);
+            }
+        }
```
ksn6:
```suggestion
        let buf = match serialize(bls_message) {
            Ok(buf) => buf,
            Err(err) => {
                error!("Failed to serialize alpenglow message: {:?}", err);
                return;
            }
        };

        for alpenglow_socket in sockets {
            if let Err(e) = send_message(buf.clone(), alpenglow_socket, &connection_cache) {
                warn!(
                    "Failed to send alpenglow message to {}: {:?}",
                    alpenglow_socket, e
                );
            }
        }
```

core/src/voting_service.rs
```diff
@@ -257,3 +331,148 @@ impl VotingService {
...
+
+    #[allow(clippy::disallowed_methods)]
+    #[test]
+    fn test_send_bls_message() {
+        solana_logger::setup();
+        let (vote_sender, vote_receiver) = crossbeam_channel::unbounded();
+        // Create listener thread on a random port we allocated and return SocketAddr to create VotingService
+
+        // Bind to a random UDP port
+        let socket = UdpSocket::bind("127.0.0.1:0").expect("Failed to bind UDP socket");
```
ksn6:
```suggestion
        let socket = solana_net_utils::bind_to_localhost();
```

core/src/voting_service.rs
```diff
@@ -257,3 +331,148 @@ impl VotingService {
...
+                }
+                Err(err) => panic!(
+                    "Failed to deserialize BLSMessage: {:?} {:?}",
+                    size_of::<BLSMessage>(),
+                    err
+                ),
+            }
+        } else {
+            panic!("No packets received");
+        }
```
ksn6:
```suggestion
        let packet = packet_batch.iter().next().expect("No packets received");
        let received_bls_message = packet
            .deserialize_slice::<BLSMessage, _>(..)
            .unwrap_or_else(|err| {
                panic!(
                    "Failed to deserialize BLSMessage: {:?} {:?}",
                    size_of::<BLSMessage>(),
                    err
                )
            });
        assert_eq!(received_bls_message, bls_message);
```

core/src/staked_validators_cache.rs
wen-coding: All of this use_alpenglow_socket logic should go away after 1 or 2 weeks, when BLS all-to-all change is submitted we should no longer need tpu sockets in StakedValidatorsCache, because TowerBFT uses next leader, and this class is for Alpenglow only, which is sent to alpenglow sockets only.

Do you still want me to change this?

core/src/staked_validators_cache.rs
wen-coding: Done.

core/src/staked_validators_cache.rs
ksn6: Yes - I'd very strongly prefer changing this to have `use_alpenglow_socket` as a member variable.

The way it's done now is not ideal, since the same flag is showing up in a number of places rather than one; it gets hard to track what's going on + makes things a bit more error prone. In the off-chance this logic sticks around a bit longer, probably better to have it just be a member variable.

wen-coding: Okay then, but at least I don't think we need to change the variable. I can't think of any way that would be useful.

wen-coding: Hmm, now I remember why I had to do it this way. If we need to support PushAlpenglowVote and PushAlpenglowBLSMessage at the same time, we have to use TPU sockets for PushAlpenglowVote and BLS sockets for PushAlpenglowBLSMessage. Can't select one or the other at construction time, otherwise we will instantly break local-cluster tests.

core/src/voting_service.rs
```diff
@@ -191,6 +210,40 @@ impl VotingService {
...
+        let buf = match serialize(bls_message) {
+            Ok(buf) => buf,
+            Err(err) => {
+                error!("Failed to serialize alpenglow message: {:?}", err);
+                return;
+            }
+        };
+
+        for alpenglow_socket in sockets {
+            if let Err(e) = send_message(buf.clone(), alpenglow_socket, &connection_cache) {
```
bw-solana, line 241: I'm wondering if we should use something like `multi_target_send` to avoid overhead of individual send calls

core/src/staked_validators_cache.rs
```diff
@@ -90,6 +93,7 @@ impl StakedValidatorsCache {
         struct Node {
             stake: u64,
             tpu_socket: SocketAddr,
+            alpenglow_socket: Option<SocketAddr>,
```
bw-solana, line 102: might drop a todo that this will no longer be an option at some point

core/src/staked_validators_cache.rs
bw-solana: yep, unfortunately, the `use_alpenglow_socket` would be thrashing all the time.

An alternative to consider:

- make `get_staked_validators_by_slot` private
- expose pub APIs `get_staked_validators_by_slot_tpu_vote` & `get_staked_validators_by_slot_alpenglow`
- These new pub APIs pass an enum into `get_staked_validators_by_slot` under the hood

core/src/voting_service.rs
wen-coding: Actually Kevin (from Jump)'s worry is that we send too fast, not too slow. He was worried that by sending N^2 messages too fast, we can cause a packet spike and overwhelm some routers in between. I'll add a TODO here that we can change to multi_target_send if all-to-all test proves this is not an issue.

core/src/staked_validators_cache.rs
wen-coding: Done.

core/src/staked_validators_cache.rs
wen-coding: Done.

AshwinSekar (approved):

PR #208, opened 2025-06-13
Switch from copy_out_transactions to add_to_certificate.

wen-coding:
When we switch to BLS certificate, we will never need to copy transactions out.

We just need to iterate through the current transactions saved in vote_pool and aggregate them to the final certificates.

This saves us a lot of unnecessary copy.

Also simplify the tests a bit.

### DIFF
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -677,7 +677,7 @@ mod tests {
         validator_keypairs: &[ValidatorVoteKeypairs],
         vote: Vote,
     ) {
-        for rank in 0..7 {
+        for rank in 0..6 {
             assert!(pool
                 .add_vote(
                     &vote,
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -139,7 +139,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         vote_type: VoteType,
         bank_hash: Option<Hash>,
         block_id: Option<Hash>,
-        transaction: Arc<VC::VoteTransaction>,
+        transaction: VC::VoteTransaction,
         validator_vote_key: &Pubkey,
         validator_stake: Stake,
     ) -> bool {
@@ -194,15 +194,15 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                 if accumulated_stake as f64 / (total_stake as f64) < limit {
                     return Ok(highest);
                 }
-                let mut transactions = Vec::new();
+                let mut vote_certificate = VC::new(cert_id);
                 for vote_type in vote_types {
                     let Some(vote_pool) = self.vote_pools.get(&(slot, *vote_type)) else {
                         continue;
                     };
-                    vote_pool.copy_out_transactions(bank_hash, block_id, &mut transactions);
+                    vote_pool
+                        .add_to_certificate(bank_hash, block_id, &mut vote_certificate)
+                        .map_err(AddVoteError::Certificate)?;
                 }
-                // TODO: remove unwrap and properly handle unwrap
-                let vote_certificate = VC::new(cert_id, transactions).unwrap();
                 self.completed_certificates
                     .insert(cert_id, vote_certificate.clone());
                 if let Some(sender) = &self.certificate_sender {
@@ -268,7 +268,6 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         validator_vote_key: &Pubkey,
     ) -> Result<Option<Slot>, AddVoteError> {
         let slot = vote.slot();
-        let transaction = Arc::new(transaction);
         let epoch = self.epoch_schedule.get_epoch(slot);
         let Some(epoch_stakes) = self.epoch_stakes_map.get(&epoch) else {
             return Err(AddVoteError::EpochStakesNotFound(epoch));
@@ -310,7 +309,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
             vote_type,
             bank_hash,
             block_id,
-            transaction.clone(),
+            transaction,
             validator_vote_key,
             validator_stake,
         ) {
@@ -595,10 +594,14 @@ pub(crate) fn load_from_blockstore(
             }));
 
         for (cert_id, cert) in certs {
-            let cert = cert.into_iter().map(Arc::from).collect();
             trace!("{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool");
-            let legacy_cert = LegacyVoteCertificate::new(cert_id, cert)
-                .expect("Certificate construction must not fail");
+            let mut legacy_cert = LegacyVoteCertificate::new(cert_id);
+            if let Err(e) = legacy_cert.aggregate(cert.iter()) {
+                error!(
+                    "{my_pubkey}: Error aggregating certificate {cert_id:?} from blockstore: {e:?}"
+                );
+                continue;
+            }
             cert_pool.insert_certificate(cert_id, legacy_cert);
         }
     }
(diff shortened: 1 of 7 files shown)


core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -595,10 +594,14 @@ pub(crate) fn load_from_blockstore(
...
 
         for (cert_id, cert) in certs {
-            let cert = cert.into_iter().map(Arc::from).collect();
             trace!("{my_pubkey}: loading certificate {cert_id:?} from blockstore into certificate pool");
-            let legacy_cert = LegacyVoteCertificate::new(cert_id, cert)
-                .expect("Certificate construction must not fail");
+            let mut legacy_cert = LegacyVoteCertificate::new(cert_id);
+            if let Err(e) = legacy_cert.aggregate(cert.iter()) {
+                error!(
+                    "{my_pubkey}: Error aggregating certificate {cert_id:?} from blockstore: {e:?}"
```
AshwinSekar, line 601: nit: I would suggest panicking here as it means our blockstore is corrupted

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -70,70 +72,78 @@ impl LegacyVoteCertificate {
...
+        for vote_message in messages {
+            // set bit-vector for the validator
+            //
+            // TODO: This only accounts for one type of vote. Update this after
+            // we have a base3 encoding implementation.
+            if self.bitmap.len() < vote_message.rank as usize {
+                return Err(CertificateError::IndexOutOfBound);
+            }
+            if self.bitmap.get(vote_message.rank as usize).as_deref() == Some(&true) {
+                return Ok(());
```
AshwinSekar: Wdyt about omitting this check or panicking? It seems `VotePool::add_vote` already checks for duplicate votes, so we should never reach this right?

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Done.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: That's a good point, since we are still changing certificate pool frequently, I'll leave a panic here to see if we missed anything.

AshwinSekar (approved):

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1009,249 +1023,62 @@ mod tests {
...
+    #[test_case(Vote::new_skip_fallback_vote(9))]
+    fn test_add_vote_and_create_new_certificate_with_types(vote: Vote) {
+        test_add_vote_and_create_new_certificate_with_type::<LegacyVoteCertificate>(vote);
+        test_add_vote_and_create_new_certificate_with_type::<CertificateMessage>(vote);
     }
 
-    fn test_add_vote_new_finalize_certificate_with_type<VC: VoteCertificate>() {
+    fn test_add_vote_and_create_new_certificate_with_type<VC: VoteCertificate>(vote: Vote) {
         let (validator_keypairs, mut pool) = create_keypairs_and_pool::<VC>();
         let pubkey = validator_keypairs[5].vote_keypair.pubkey();
```
ksn6: This is hard to understand - why is this `validator_keypairs[5]`?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1009,249 +1023,62 @@ mod tests {
...
+    #[test_case(Vote::new_skip_vote(8))]
+    #[test_case(Vote::new_skip_fallback_vote(9))]
+    fn test_add_vote_and_create_new_certificate_with_types(vote: Vote) {
+        test_add_vote_and_create_new_certificate_with_type::<LegacyVoteCertificate>(vote);
+        test_add_vote_and_create_new_certificate_with_type::<CertificateMessage>(vote);
     }
 
-    fn test_add_vote_new_finalize_certificate_with_type<VC: VoteCertificate>() {
+    fn test_add_vote_and_create_new_certificate_with_type<VC: VoteCertificate>(vote: Vote) {
         let (validator_keypairs, mut pool) = create_keypairs_and_pool::<VC>();
```
ksn6, line 1037: Could we have `create_keypairs_and_pool` take in a `usize` and then explicitly pass `10` in here?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1009,249 +1023,62 @@ mod tests {
...
-            )
-            .is_ok());
-        assert_eq!(pool.highest_skip_slot(), 0);
-        // Same key voting again shouldn't make a certificate
-        assert!(pool
-            .add_vote(
-                &Vote::new_skip_vote(5),
-                dummy_transaction::<VC>(validator_keypairs[5].bls_keypair.clone()),
+                &vote,
+                dummy_transaction::<VC>(&validator_keypairs, &vote, 5),
```
ksn6: Could we set `let validator_ix = 5;` at the top of this test and then use `validator_ix` when appropriate throughout this test?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1009,249 +1023,62 @@ mod tests {
...
                 &pubkey,
             )
             .is_ok());
-        assert_eq!(pool.highest_skip_slot(), 0);
-        // 20% voted for skip 5, 20% voted for skip_fallback 5
-        for keys in validator_keypairs.iter().take(2) {
+        assert!(highest_slot_fn(&pool) < slot);
```
ksn6, line 1064: Could we make a stronger check? here to see that e.g.:

- the duplicate vote was thrown out if not a notar-fallback
- the duplicate vote was kept if it was a notar-fallback?

core/src/alpenglow_consensus/certificate_pool.rs
ksn6: Come to think of it, we should generalize this test for `N` validators and run a matrix of tests here. Can do this in a future PR.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -1320,16 +1148,16 @@ mod tests {
...
-        for pubkeys in validator_keypairs.iter().skip(4).take(3) {
-            add_skip_vote_range::<VC>(&mut pool, 5, 8, pubkeys);
+        for rank in 4..7 {
+            add_skip_vote_range::<VC>(&mut pool, 5, 8, &validator_keypairs, rank);
         }
         // The rest voted for (11, 15)
-        for pubkeys in validator_keypairs.iter().skip(7) {
-            add_skip_vote_range::<VC>(&mut pool, 11, 15, pubkeys);
+        for rank in 7..10 {
+            add_skip_vote_range::<VC>(&mut pool, 11, 15, &validator_keypairs, rank);
```
ksn6, line 1162: Could you break the for loop below (lines 1163 - 1170) into three separate for loops? Ideally, we avoid if-statements as much as possible in tests.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: It's just a randomly chosen validator in the middle I think, inherited from a test long ago....

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Sounds good, we can clean up all these tests after BLS all-to-all is done and certificate pool is much simpler.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Right now we don't make that distinction on the add_vote API. This is tested in vote_pool unittest. I don't think the caller of add_vote (voting_loop) cares whether we got a duplicate vote for now.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: Done.

core/src/alpenglow_consensus/certificate_pool.rs

PR #210, opened 2025-06-15
In BLSSigverifier, reset stats after reporting, do not panic when channel is disconnected.

wen-coding:
Two problems:
- After each datapoint_info! stats should be reset, also added tests.
- The channel to send BLS messages may be disconnected when exiting, do not panic, instead just log error.

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -9,7 +9,7 @@ use {
     std::time::{Duration, Instant},
 };
 
-const STATS_INTERVAL_SECONDS: u64 = 1; // Log stats every second
+const STATS_INTERVAL_DURATION: Duration = Duration::from_secs(1); // Log stats every second
 
 // We are adding our own stats because we do BLS decoding in batch verification,
 // and we send one BLS message at a time. So it makes sense to have finer-grained stats
@@ -82,9 +82,7 @@ impl SigVerifier for BLSSigVerifier {
                 }
             }
         }
-        if stats.received > 0 {
-            self.stats.accumulate(stats);
-        }
+        self.stats.accumulate(stats);
         // We don't need lock on stats for now because stats are read and written in a single thread.
         self.maybe_report_stats();
         Ok(())
@@ -95,7 +93,7 @@ impl BLSSigVerifier {
     fn maybe_report_stats(&mut self) {
         let now = Instant::now();
         let time_since_last_log = now.duration_since(self.stats.last_stats_logged);
-        if time_since_last_log < Duration::from_secs(STATS_INTERVAL_SECONDS) {
+        if time_since_last_log < STATS_INTERVAL_DURATION {
             return;
         }
         datapoint_info!(
@@ -219,9 +217,7 @@ mod tests {
         assert_eq!(stats.received_malformed, 0);
 
         // Pretend 10 seconds have passed, make sure stats are reset
-        verifier.set_last_stats_logged(
-            Instant::now() - Duration::from_secs(STATS_INTERVAL_SECONDS + 1),
-        );
+        verifier.set_last_stats_logged(Instant::now() - STATS_INTERVAL_DURATION);
         let messages = vec![BLSMessage::Vote(VoteMessage {
             vote: Vote::new_finalization_vote(7),
             signature: Signature::default(),


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -56,8 +57,8 @@ impl SigVerifier for BLSSigVerifier {
                             sent_failed += 1;
                         }
                         Err(TrySendError::Disconnected(_)) => {
-                            // There is no hope to recover, restart the validator.
-                            panic!("BLS message channel is disconnected");
+                            error!("BLS message channel is disconnected");
+                            self.stats.channel_disconnected = true;
```
ksn6: Nit: to maintain symmetry, could we do a local variable called `let channel_disconnected = false` and then set `self.stats.channel_disconnected = channel_disconnected;` in the `packet_received > 0` condition below?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -91,8 +92,10 @@ impl BLSSigVerifier {
             ("sent", self.stats.bls_messages_sent, u64),
             ("sent_failed", self.stats.sent_failed, u64),
             ("malformed", self.stats.bls_messages_malformed, u64),
-            ("received", self.stats.packets_received, u64)
+            ("received", self.stats.packets_received, u64),
+            ("disconnected", self.stats.channel_disconnected, bool),
         );
+        self.stats = BLSSigVerifierStats::default();
```
ksn6: This resets stats each time we report stats; is this the intended behavior? Or are we to instead supply the cumulative stats?

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Yes this is intended behavior, so we report accumulated numbers every 10 seconds, then influxDB can do the averaging for us I think.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -39,76 +61,74 @@ impl SigVerifier for BLSSigVerifier {
...
-                packet_received += 1;
+                stats.received += 1;
                 match packet.deserialize_slice::<BLSMessage, _>(..) {
                     Ok(message) => match self.sender.try_send(message) {
                         Ok(()) => {
-                            bls_messages_sent += 1;
+                            stats.sent += 1;
                         }
                         Err(TrySendError::Full(_)) => {
                             warn!("BLS message channel is full, dropping message");
```
AshwinSekar: we should use the existing error handling by setting:
`SendType = String`
And then returning the `SigVerifyServiceError` variant for these by using `SendPktsError` for the transient errors, and the `Send` error for the disconnected situation.

wen-coding: Changed SendType. I'm confused, the code is not using `solana_streamer`, why should it return `SendPktsError`?

I checked the error handling currently in sigverifier, it doesn't do much.
https://github.com/anza-xyz/alpenglow/blob/8e3a696f163f6c876921b49e227b8bcc619f511d/core/src/sigverify_stage.rs#L336
One error just makes the whole send_packets() end,
https://github.com/anza-xyz/alpenglow/blob/8e3a696f163f6c876921b49e227b8bcc619f511d/core/src/sigverify_stage.rs#L411
Then it just exits the whole loop. Not much metrics collecting either, it just logs or ignores the returned error.

This makes sense when we used unbounded channel, since we would never see SendError::Full. Now we use bounded channel, we need to handle TrySendError::Full.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -9,26 +9,45 @@ use {
     std::time::{Duration, Instant},
 };
 
-const STATS_INTERVAL_SECONDS: u64 = 10; // Log stats every 10 second
+const STATS_INTERVAL_SECONDS: u64 = 1; // Log stats every second
```
bw-solana: Could make this of type Duration for convenience: `Duration::from_secs(1)`

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -39,76 +58,72 @@ impl SigVerifier for BLSSigVerifier {
...
-            });
-        });
-        if packet_received > 0 {
-            self.stats.packets_received += packet_received;
-            self.stats.bls_messages_sent += bls_messages_sent;
-            self.stats.bls_messages_malformed += bls_messages_malformed;
-            self.stats.sent_failed += sent_failed;
+            }
+        }
+        if stats.received > 0 {
```
bw-solana: could tuck this conditional inside the accumulate

bw-solana (approved):
left a couple nits, but this looks like an improvement to me

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: I feel we don't need this check at all, it's not too inefficient to add zeroes to stats.

bw-solana (approved):

PR #211, opened 2025-06-17
test(local-cluster): ensure liveness after double notar fallback

ksn6:
#### Problem
We'd like to ensure that a cluster remains live after having observed a node issue two notar fallbacks on the same slot.

Consider four nodes with the following stakes:

A: 20% - eps
B: 40%
C: 20%
D: 20% + eps

A is the leader and proposes / votes for b1; however, pretend that A is Byzantine and *only* sends node B a different block, b2. Everybody else sees block b1.

B votes for b2; to simulate this, initially start by having B copy vote A. Then, at some stage, have B send out voting transactions that sign for a different block.

D votes for b1.

Suppose that C's turbine is disabled, and it votes skip. But, it then observes that:
- A and D both voted for b1; this is 40% of stake; SafeToNotar triggers, resulting in C issuing a NotarFallback.
- B voted for b2; this is 40% of stake; SafeToNotar triggers once again, resulting in C issuing another NotarFallback.

As a result, C ends up issuing two notar fallback votes.

In this experiment A is the only Byzantine node, with < 20% stake. C has a liveness issue, but has <= 20% stake and still votes.

After confirming that C issued two notar fallback votes, have B just continue copy voting A and ensure that we continue seeing roots.

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -6515,27 +6515,26 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
 #[test]
 #[serial]
 fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
+    solana_logger::setup_with_default(AG_DEBUG_LOG_FILTER);
+
     // Configure total stake and stake distribution
-    let total_stake = 10 * DEFAULT_NODE_STAKE;
-    let slots_per_epoch = MINIMUM_SLOTS_PER_EPOCH;
+    const TOTAL_STAKE: u64 = 10 * DEFAULT_NODE_STAKE;
+    const SLOTS_PER_EPOCH: u64 = MINIMUM_SLOTS_PER_EPOCH;
 
-    let node_a_stake = total_stake * 2 / 10 - 1;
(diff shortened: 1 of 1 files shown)


local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6456,364 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+                .validators
+                .get(&keypair.pubkey())
+                .map(|validator| (validator.info.voting_keypair.pubkey(), index))
+        })
+        .collect();
+
+    assert_eq!(vote_pubkeys.len(), node_stakes.len());
+
+    // Collect node pubkeys and TPU addresses
+    let node_pubkeys: Vec<_> = validator_keys.iter().map(|key| key.pubkey()).collect();
```
bw-solana, line 6595: this all looks very similar to the other local cluster test you created. If there ends up being a 3rd, might be nice to have some helper function for setting up alpenglow local cluster :)

ksn6: yep, makes sense - the plan is to eventually factor this all out into common pieces; we'll probably want to write a few more of these first to ensure that there aren't corner cases that our soon-to-be abstractions won't cover

PR #212, opened 2025-06-17
gossip fallback

bw-solana:
#### Problem
When starting a boostrap node (no `entrypoint` args set) without `gossip_host` set, we will fall back to localhost IP. This causes contact info to be filtered out, and other nodes cannot join the cluster. `gossip_host` has been deprecated upstream.

This was discovered when trying to spin up Alpenglow repo on invalidator cluster (it has removed `gossip_host` arg inline with the upstream deprecation).

Note: This was fixed upstream with https://github.com/anza-xyz/agave/pull/6519

#### Summary of Changes
Fallback to the `bind_address` (presumably external facing IP) instead of localhost.

### DIFF
--- a/validator/src/commands/run/execute.rs
+++ b/validator/src/commands/run/execute.rs
@@ -1138,7 +1138,7 @@ pub fn execute(
                         )
                 })
             } else {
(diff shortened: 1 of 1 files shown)


bw-solana:
Closing in favor of #219

PR #214, opened 2025-06-18
Switch Alpenglow Vote from Transaction to BLS Message.

wen-coding:
Switch Alpenglow Vote from Transaction to BLS Message.


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -105,6 +108,9 @@ struct VotingContext {
     vote_account_pubkey: Pubkey,
     identity_keypair: Arc<Keypair>,
     authorized_voter_keypairs: Arc<RwLock<Vec<Arc<Keypair>>>>,
+    // The BLS keypair should always change with authorized_voter_keypairs.
+    derived_bls_keypairs: HashMap<Pubkey, Arc<BLSKeypair>>,
+    my_rank: HashMap<Epoch, u16>,
```
bw-solana: comment here to help people understand what rank means would be helpful

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -285,27 +293,50 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+        let vote = &vote_message.vote;
+        let rank = vote_message.rank;
+        let slot = vote.slot();
+        let (validator_vote_key, validator_stake, total_stake) =
+            self.get_key_and_stakes(slot, rank)?;
 
         if validator_stake == 0 {
-            return Err(AddVoteError::ZeroStake);
+            // Since we have a valid rank, this should never happen, there is no rank for zero stake.
+            panic!("Validator stake is zero for pubkey: {validator_vote_key}");
```
AshwinSekar: nit: can we remove `AddVoteError::ZeroStake` now?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,70 @@ pub enum CertificateError {
     InvalidVoteType,
 }
 
-pub trait VoteCertificate: Clone {
-    type VoteTransaction: AlpenglowVoteTransaction;
+#[derive(Clone)]
+pub struct VoteCertificate(CertificateMessage);
```
AshwinSekar, line 38: should we just rename `CertificateMessage` as `VoteCertificate` and have this be one impl? Or are there plans to add more to this struct in the future?

core/src/validator.rs
```diff
@@ -954,7 +954,10 @@ impl Validator {
         );
 
         let (replay_vote_sender, replay_vote_receiver) = unbounded();
-        let (alpenglow_vote_sender, alpenglow_vote_receiver) = unbounded();
+        // TODO(wen): Remove this later.
+        let (alpenglow_vote_sender, _) = unbounded();
```
AshwinSekar, line 960: (can be in a future pr), so now we can remove this from banking stage and gossip correct?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +943,97 @@ impl VotingLoop {
...
+                }
+                Some(authorized_voter_keypair) => authorized_voter_keypair.clone(),
+            };
+        }
+
+        let bls_keypair = match Self::get_bls_keypair(context, &authorized_voter_keypair) {
+            Ok(keypair) => keypair,
+            Err(e) => {
+                warn!("Failed to derive my own BLS keypair: {e:?}");
+                return GenerateVoteTxResult::FailedToDeriveBlsKeypair;
```
AshwinSekar: for my understanding how can the derivation fail? 
Basically when we enable set-identity we'll want to classify errors as transient, hot spare or non voting, would this indicate that this node will never be able to vote?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -1042,15 +1109,13 @@ impl VotingLoop {
...
-        vote: &Vote,
-        vote_account_pubkey: &Pubkey,
-        tx: VC::VoteTransaction,
-        cert_pool: &mut CertificatePool<VC>,
+        message: &BLSMessage,
+        cert_pool: &mut CertificatePool,
         commitment_sender: &Sender<CommitmentAggregationData>,
     ) -> Result<(), AddVoteError> {
-        let Some(new_finalized_slot) = cert_pool.add_vote(vote, tx, vote_account_pubkey)? else {
+        let Some(new_finalized_slot) = cert_pool.add_transaction(message)? else {
```
AshwinSekar, line 1125: So for here, we want to match on BLSMessage and call either `add_transaction` or `insert_completed_certificate` depending on the type right?

AshwinSekar (commented):
nice work! Just some small comments

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: also we'll want to certificate_sender to trigger as well for a completed certificate, might have to refactor that sender function

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +943,97 @@ impl VotingLoop {
...
-        let vote_ix =
-            vote.to_vote_instruction(vote_account_pubkey, authorized_voter_keypair.pubkey());
-        let vote_tx = Transaction::new_signed_with_payer(
-            &[vote_ix],
-            Some(&context.identity_keypair.pubkey()),
-            &[&context.identity_keypair, authorized_voter_keypair],
-            bank.last_blockhash(),
-        );
-
+        let my_bls_pubkey = bls_keypair.public.into();
```
AshwinSekar: Should we be comparing this pubkey to the one on the vote account?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -223,7 +232,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                     .insert(cert_id, vote_certificate.clone());
                 if let Some(sender) = &self.certificate_sender {
                     if cert_id.is_critical() {
-                        if let Err(e) = sender.try_send((cert_id, vote_certificate)) {
+                        if let Err(e) = sender.try_send((cert_id, vote_certificate.ceritifcate())) {
```
carllin: nit: `ceritifcate` -> `certificate`

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -285,27 +293,50 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+            .get_pubkey(rank as usize)
+        else {
+            return Err(AddVoteError::InvalidRank(rank));
+        };
+        let stake = epoch_stakes.vote_account_stake(vote_key);
+        if stake == 0 {
+            return Err(AddVoteError::ZeroStake);
+        }
+        Ok((*vote_key, stake, epoch_stakes.total_stake()))
+    }
```
carllin: getting the pubkey from the rank seems better implemented as a `the `bls_rank_to_pubkey()` method in the `EpochStakes` object, and then we can call the existing `vote_account_stake()` and `total_stake()` functions individually

local-cluster/tests/local_cluster.rs
```diff
@@ -6355,51 +6341,49 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
 
         move || loop {
             let n_bytes = vote_listener.recv(&mut buf).unwrap();
-            let vote_txn = bincode::deserialize::<Transaction>(&buf[0..n_bytes]).unwrap();
+            let bls_message = bincode::deserialize::<BLSMessage>(&buf[0..n_bytes]).unwrap();
+            if let BLSMessage::Vote(vote_message) = bls_message {
```
carllin: nit: `let/else` block to make the diff smaller

carllin: out of curiosity, what was breaking in the local cluster tests?

core/src/voting_service.rs
```diff
@@ -46,7 +46,8 @@ pub enum VoteOp {
     PushAlpenglowBLSMessage {
         bls_message: BLSMessage,
         slot: Slot,
-        saved_vote_history: SavedVoteHistoryVersions,
+        // For refresh votes, saved_vote_history would be None.
+        saved_vote_history: Option<SavedVoteHistoryVersions>,
```
carllin: Hmmm, for refresh we should probably just have a separate message like we have currently for `VoteOp::RefreshVote`, maybe a `VoteOp::AlpenglowRefreshVote`

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -830,53 +846,95 @@ impl VotingLoop {
...
+        authorized_voter_keypair: &Arc<Keypair>,
+    ) -> Result<Arc<BLSKeypair>, BlsError> {
+        let pubkey = authorized_voter_keypair.pubkey();
+        if let Some(existing) = context.derived_bls_keypairs.get(&pubkey) {
+            return Ok(existing.clone());
+        }
+
+        let bls_keypair = Arc::new(BLSKeypair::derive_from_signer(
+            authorized_voter_keypair,
+            b"alpenglow",
```
carllin: nit: let's make this a constant in the top level alpenglow module

core/src/alpenglow_consensus/voting_loop.rs
carllin: do we really need to cache this? Seems like unnecessary headache for cleanup/maintenance since its just a lookup

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +943,97 @@ impl VotingLoop {
...
             if context.voted_signatures.len() > MAX_VOTE_SIGNATURES {
                 context.voted_signatures.remove(0);
             }
         } else {
             context.voted_signatures.clear();
         }
 
-        GenerateVoteTxResult::Tx(vote_tx)
+        let Some(my_rank) = Self::get_my_rank(context, bank, &my_bls_pubkey) else {
+            return GenerateVoteTxResult::NoRankFound;
```
carllin: hmm this error seems like this would fail silently if say somebody misconfigured or never swapped over to BLS since the callers never log:

https://github.com/anza-xyz/alpenglow/blob/6e3110dd18d5b23c3f81fc364d867706d1cd6d2f/core/src/alpenglow_consensus/voting_loop.rs#L594

https://github.com/anza-xyz/alpenglow/blob/6e3110dd18d5b23c3f81fc364d867706d1cd6d2f/core/src/alpenglow_consensus/voting_loop.rs#L441-L447

https://github.com/anza-xyz/alpenglow/blob/6e3110dd18d5b23c3f81fc364d867706d1cd6d2f/core/src/alpenglow_consensus/voting_loop.rs#L360-L366

Could we add an error log here/in the caller `send_vote()` as well?

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: It's possible, but it's pretty ugly to put the aggregate logic into CertificateMessage. Put a TODO there so we can consider it later.

core/src/validator.rs
wen-coding: Yeah this PR is already too big, I'll clean up alpenglow_vote_sender/receiver in a future PR.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: @samkim-crypto I don't believe this will fail for ed25519 keypair, unless we somehow happen to get a default signature in the middle, then this keypair can never recover, changed to panic.

core/src/alpenglow_consensus/certificate_pool.rs

local-cluster/tests/local_cluster.rs
wen-coding: Done.

Previously this test was listening for transactions, now it listen for BLSMessages. Also, because we have rank directly in BLSMessage, we don't need to do lookup to figure out which vote message comes from A any more.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,71 @@ pub enum CertificateError {
...
 
         // aggregate the votes
+        let bitmap = &mut self.0.bitmap;
         for vote_message in messages {
             // set bit-vector for the validator
             //
             // TODO: This only accounts for one type of vote. Update this after
             // we have a base3 encoding implementation.
-            if self.bitmap.len() < vote_message.rank as usize {
+            if bitmap.len() < vote_message.rank as usize {
```
ksn6: If `rank` is 0-based, then I think this should be:

`if bitmap.len() <= vote_message.rank as usize {`

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,71 @@ pub enum CertificateError {
...
                 return Err(CertificateError::IndexOutOfBound);
             }
-            if self.bitmap.get(vote_message.rank as usize).as_deref() == Some(&true) {
+            if bitmap.get(vote_message.rank as usize).as_deref() == Some(&true) {
                 panic!("Conflicting vote check should make this unreachable {vote_message:?}");
```
ksn6: nit: s/panic/[unreachable](https://doc.rust-lang.org/std/macro.unreachable.html)

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Done.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,71 @@ pub enum CertificateError {
...
             // TODO(wen): put this into bls crate
             let uncompressed = SignatureProjective::try_from(vote_message.signature)?;
             current_signature_uncompressed.aggregate_with([&uncompressed]);
         }
-        self.signature = Signature::from(current_signature_uncompressed);
+        *signature = Signature::from(current_signature_uncompressed);
         Ok(())
     }
+
+    pub fn certificate(&self) -> CertificateMessage {
```
ksn6: Could we return a reference here, and have the invoker clone? It's usually not a good idea to have clones present in "getters."

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -575,7 +590,7 @@ impl VotingLoop {
                     bank.slot()
                 );
                 let vote = Vote::new_skip_vote(slot);
-                Self::send_vote(vote, false, bank.as_ref(), cert_pool, voting_context);
+                let _ = Self::send_vote(vote, false, bank.as_ref(), cert_pool, voting_context);
```
ksn6: nit: remove let _ =

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +942,96 @@ impl VotingLoop {
...
+                    return GenerateVoteTxResult::Failed;
+                }
+            }
+            let vote_account = match bank.get_vote_account(&context.vote_account_pubkey) {
+                None => {
+                    warn!("Vote account {vote_account_pubkey} does not exist.  Unable to vote");
+                    return GenerateVoteTxResult::Failed;
+                }
+                Some(vote_account) => vote_account,
+            };
```
ksn6:
```suggestion
            let vote_account = bank.get_vote_account(&context.vote_account_pubkey)
                .ok_or_else(|| {
                    warn!("Vote account {vote_account_pubkey} does not exist.  Unable to vote");
                    GenerateVoteTxResult::Failed
                })?;

```

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +942,96 @@ impl VotingLoop {
...
+            let vote_state = match vote_account.alpenglow_vote_state() {
+                None => {
+                    warn!(
                     "Vote account {vote_account_pubkey} does not have an Alpenglow vote state.  Unable to vote",
                 );
-                return GenerateVoteTxResult::Failed;
+                    return GenerateVoteTxResult::Failed;
+                }
+                Some(vote_state) => vote_state,
+            };
```
ksn6:
```
            let vote_state = vote_account.alpenglow_vote_state()
                .ok_or_else(|| {
                    warn!(
                        "Vote account {vote_account_pubkey} does not have an Alpenglow vote state.  Unable to vote"
                    );
                    GenerateVoteTxResult::Failed
                })?;
```

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +942,96 @@ impl VotingLoop {
...
+            let Some(authorized_voter_pubkey) = vote_state.get_authorized_voter(bank.epoch())
+            else {
+                warn!(
                 "Vote account {vote_account_pubkey} has no authorized voter for epoch {}.  Unable to vote",
                 bank.epoch()
             );
-            return GenerateVoteTxResult::Failed;
-        };
+                return GenerateVoteTxResult::Failed;
+            };
```
ksn6:
```
            let authorized_voter_pubkey = vote_state.get_authorized_voter(bank.epoch())
                .ok_or_else(|| {
                    warn!(
                        "Vote account {vote_account_pubkey} has no authorized voter for epoch {}.  Unable to vote",
                        bank.epoch()
                    );
                    GenerateVoteTxResult::Failed
                })?;
```

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +942,96 @@ impl VotingLoop {
...
+                        "The authorized keypair {authorized_voter_pubkey} for vote account \
                      {vote_account_pubkey} is not available.  Unable to vote"
-                );
-                return GenerateVoteTxResult::NonVoting;
+                    );
+                    return GenerateVoteTxResult::NonVoting;
+                }
+                Some(authorized_voter_keypair) => authorized_voter_keypair.clone(),
+            };
+        }
```
ksn6:
```
            authorized_voter_keypair = authorized_voter_keypairs
                .iter()
                .find(|keypair| keypair.pubkey() == authorized_voter_pubkey)
                .map(|keypair| keypair.clone())
                .unwrap_or_else(|| {
                    warn!(
                        "The authorized keypair {authorized_voter_pubkey} for vote account \
                     {vote_account_pubkey} is not available.  Unable to vote"
                    );
                    return GenerateVoteTxResult::NonVoting;
                });
```

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -885,86 +942,96 @@ impl VotingLoop {
...
+
+        let bls_keypair = match Self::get_bls_keypair(context, &authorized_voter_keypair) {
```
ksn6: use `.unwrap_or_else()` here

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -973,28 +1040,26 @@ impl VotingLoop {
...
-                    commitment_sender,
-                ) {
-                    err @ Err(AddVoteError::CertificateSenderError) => err,
-                    Err(e) => {
-                        // TODO(ashwin): increment metrics on non duplicate failures
-                        trace!("{my_pubkey}: unable to push vote into the pool {}", e);
-                        Ok(())
-                    }
-                    Ok(()) => Ok(()),
+        let add_to_cert_pool = |bls_message: BLSMessage| {
```
ksn6, line 1055:
```
        let add_to_cert_pool = |bls_message: BLSMessage| -> Result<(), AddVoteError> {
            trace!("{my_pubkey}: ingesting BLS Message: {bls_message:?}");
            Self::add_message_and_maybe_update_commitment(
                my_pubkey,
                &bls_message,
                cert_pool,
                commitment_sender,
            )
            .or_else(|e| match e {
                AddVoteError::CertificateSenderError => Err(e),
                _ => {
                    // TODO(ashwin): increment metrics on non duplicate failures
                    trace!("{my_pubkey}: unable to push vote into the pool {}", e);
                    Ok(())
                }
            })
        };

        let first = vote_receiver.recv_timeout(Duration::from_secs(1))?;
        std::iter::once(first)
            .chain(vote_receiver.try_iter())
            .try_for_each(add_to_cert_pool)
```

core/src/voting_service.rs
wen-coding: I originally wrote that, but the difference between PushVote and RefreshVote is much larger than PushBLSMessage and RefreshBLSMessage. In the latter case the only difference is that we are not saving the message into vote_history.
Why is a different voteop better here?

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: 3 lookups vs one: epoch_stakes -> bls_pubkey_to_rank_map -> get_rank

Maybe it's not that bad, changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Yeah, haven't totally decided where to put that logic yet, that will be in PR#203. For this PR we only have votes sent as BLS messages.

Also in this PR we don't send out completed certificate at all, let's discuss how to refactor in the following PR.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Done.

core/src/alpenglow_consensus/certificate_pool.rs
wen-coding: I'm confused, we already have get_pubkey() in bls_pubkey_to_rank_map(), and we are using vote_account_stake() and total_stake(), this method is mostly a wrapper of all these, what are you suggesting?

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Outside send_vote the error reason is lost, since only bool is returned. I added log inside send_vote, do we need duplicate log in the caller as well?

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Good catch, done.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Done

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Done.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Hmm I don't think that works, ok_or_else expect returning a VoteAccount, changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Changed.

core/src/alpenglow_consensus/voting_loop.rs
samkim-crypto: Yeah if we use a real ed25519 keypair, then the derivation should never fail. It is just that the Solana `Signer` trait  is very general and is implemented by non-keypair objects like the [public key](https://github.com/anza-xyz/agave/blob/master/clap-v3-utils/src/input_parsers/signer.rs#L47). If we sign with objects like these, then it will result in a default signature, which will cause the derivation to fail. Validators will never use such keys, so I think it is safe to panic.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,73 @@ pub enum CertificateError {
...
-    type VoteTransaction = VoteMessage;
-
-    fn new(certificate_id: CertificateId) -> Self {
-        CertificateMessage {
+impl VoteCertificate {
+    pub fn new(certificate_id: CertificateId) -> Self {
+        VoteCertificate(CertificateMessage {
             certificate: certificate_id.into(),
             signature: Signature::default(),
             bitmap: BitVec::<u8, Lsb0>::repeat(false, VALIDATOR_BITMAP_U8_SIZE),
```
samkim-crypto: Sorry, this is not related to the changes in this PR, but I think the `repeat` function takes in the number of bits rather than bytes. So in this case, should this be 4096 or `MAXIMUM_VALIDATORS`?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,73 @@ pub enum CertificateError {
...
         // aggregate the votes
+        let bitmap = &mut self.0.bitmap;
         for vote_message in messages {
             // set bit-vector for the validator
             //
             // TODO: This only accounts for one type of vote. Update this after
             // we have a base3 encoding implementation.
-            if self.bitmap.len() < vote_message.rank as usize {
+            if bitmap.len() <= vote_message.rank as usize {
                 return Err(CertificateError::IndexOutOfBound);
```
samkim-crypto: Nit: Is `CertificateError::ValidatorDoesNotExist` used anywhere? It seems like it overlaps with `IndexOutOfBound`, maybe we can use the more descriptive `ValidatorDoesNotExist` here and remove the `IndexOutOfBound`?

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -38,115 +35,73 @@ pub enum CertificateError {
...
             }
-            if self.bitmap.get(vote_message.rank as usize).as_deref() == Some(&true) {
-                panic!("Conflicting vote check should make this unreachable {vote_message:?}");
+            if bitmap.get(vote_message.rank as usize).as_deref() == Some(&true) {
+                unreachable!(
```
samkim-crypto: It is a little weird to have an if statement followed by an `unreachable` check that does not do anything. Seems like a suitable place to to use `assert!` (or `debug_assert` if we only plan to keep this check during development) in place of the entire if statement?

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -285,27 +292,51 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
+        let vote = &vote_message.vote;
+        let rank = vote_message.rank;
+        let slot = vote.slot();
+        let (validator_vote_key, validator_stake, total_stake) =
+            self.get_key_and_stakes(slot, rank)?;
 
         if validator_stake == 0 {
-            return Err(AddVoteError::ZeroStake);
+            // Since we have a valid rank, this should never happen, there is no rank for zero stake.
+            panic!("Validator stake is zero for pubkey: {validator_vote_key}");
```
samkim-crypto: Nit: could be more idiomatic to have
```
assert_ne!(stake, 0, "Validator stake is zero for pubkey: {validator_vote_key}");
```
instead of the if statement.

samkim-crypto (commented):
Thanks for cleaning this up! It looks good overall. Just some small comments below.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Oops, that's right, good catch.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: Done.

core/src/alpenglow_consensus/vote_certificate.rs
wen-coding: This should not happen during normal cases, changed to assert!

core/src/alpenglow_consensus/certificate_pool.rs

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -797,7 +819,7 @@ impl VotingLoop {
         my_pubkey: &Pubkey,
         slot: Slot,
         root_bank_cache: &mut RootBankCache,
-        cert_pool: &mut CertificatePool<LegacyVoteCertificate>,
+        cert_pool: &mut CertificatePool,
         voting_context: &mut VotingContext,
     ) {
```
qkniep, line 823: This function should also rebroadcast any observed certificates for slots after the finalized one. Also, the `TODO` comment mentions sending a NotarFallback certificate for the finalized slot (if slow finalized), this should be a Notar certificate instead.

AshwinSekar: noted, will fix this up once BLS stabilizes, currently we don't broadcast any certificates.

wen-coding: Rebroadcasting of certificates is intentionally left out from this PR, because this PR is already too big, it already took a lot of time to review.

And sending Notar certificate vs NotarFallback certificate doesn't belong to this PR either.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -938,86 +984,112 @@ impl VotingLoop {
...
-            info!(
-                "Vote account node_pubkey mismatch: {} (expected: {}).  Unable to vote",
-                vote_state.node_pubkey(),
-                context.identity_keypair.pubkey()
-            );
-            return GenerateVoteTxResult::HotSpare;
-        }
+            bls_pubkey_in_vote_account = match vote_account.bls_pubkey() {
+                None => {
+                    warn!(
```
AshwinSekar: Prefer to panic here, the operator must migrate their vote account. Can add instructions once we figure out Vote State v4

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -938,86 +984,112 @@ impl VotingLoop {
...
-        );
+            authorized_voter_keypair = keypair.clone();
+        }
 
+        let bls_keypair = Self::get_bls_keypair(context, &authorized_voter_keypair)
+            .unwrap_or_else(|e| panic!("Failed to derive my own BLS keypair: {e:?}"));
+        let my_bls_pubkey: BLSPubkey = bls_keypair.public.into();
+        if my_bls_pubkey != bls_pubkey_in_vote_account {
+            error!(
+                "Vote account bls_pubkey mismatch: {:?} (expected: {:?}).  Unable to vote",
```
AshwinSekar, line 1010: Same here, if authorized voter / identity matches but bls keypair is invalid we should panic with a detailed description as this is a voting node with a misconfiguration.

AshwinSekar (commented):
LGTM just some minor comments. Will defer to the rest of the team for final approval.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Thanks for the review, done.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: Done.

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -18,135 +15,88 @@ use {
 /// There are around 1500 validators currently. For a clean power-of-two
 /// implementation, we should chosoe either 2048 or 4096. Choose a more
 /// conservative number 4096 for now.
-/// The number of bytes in a bitmap to represent up to 4096 validators
-/// (`MAXIMUM_VALIDATORS` / 8)
-const VALIDATOR_BITMAP_U8_SIZE: usize = 512;
+const MAXIMUM_VALIDATORS: usize = 4096;
```
bw-solana, line 20: This is the type of change that would be great to fork out into its own PR in an effort to reduce how giant this PR is

(but don't worry about it this time)

core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -18,135 +15,88 @@ use {
...
     #[error("BLS error: {0}")]
     BlsError(#[from] BlsError),
-    #[error("Index out of bounds")]
-    IndexOutOfBound,
     #[error("Invalid pubkey")]
     InvalidPubkey,
     #[error("Invalid signature")]
     InvalidSignature,
-    #[error("Validator does not exist")]
+    #[error("Validator does not exist for given rank")]
```
bw-solana, line 30: same thing here - could be decoupled from this PR

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -627,7 +640,10 @@ impl VotingLoop {
                     bank.slot()
                 );
                 let vote = Vote::new_skip_vote(slot);
-                Self::send_vote(vote, false, bank.as_ref(), cert_pool, voting_context);
+                if !Self::send_vote(vote, false, bank.as_ref(), cert_pool, voting_context) {
+                    warn!("send_vote failed, break out early");
+                    break;
```
bw-solana: why do we want to break out early here?

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: This map has been removed.

core/src/alpenglow_consensus/voting_loop.rs
wen-coding: send_vote shouldn't normally fail, it will fail because:
1. our BLS pubkey is broken, no hope to generate a correct BLSMessage
2. adding to cert_pool failed, hmm this one could be different for different votes
I'll change it to not break out, just add a warning, good catch.

AshwinSekar:
> FYI - running a quick checkout with this code (https://buildkite.com/anza/infra-invalidator/builds/2266#0197eaa0-1a7e-49aa-aac6-3ca3cbfb3143), but it is failing because it is not observing any roots get created

Looks like some of the machines are failing because the BLSPubkey doesn't match:
```
thread 'solVotingLoop' panicked at core/src/alpenglow_consensus/voting_loop.rs:1054:13:
Vote account bls_pubkey mismatch: Pubkey([15, 83, 249, 230, 130, 27, 145, 51, 216, 42, 88, 227, 221, 46, 239, 43, 21, 150, 32, 59, 49, 187, 242, 187, 84, 146, 15, 89, 118, 85, 176, 0, 161, 192, 161, 10, 203, 117, 115, 1, 175, 138, 111, 85, 227, 216, 11, 252, 16, 152, 64, 16, 123, 78, 232, 144, 215, 228, 241, 54, 144, 156, 108, 101, 81, 19, 85, 132, 225, 97, 67, 250, 0, 27, 124, 219, 85, 113, 218, 21, 181, 163, 172, 103, 57, 82, 97, 191, 226, 56, 164, 137, 163, 190, 55, 83]) (expected: Pubkey([5, 173, 38, 22, 84, 152, 116, 151, 84, 120, 38, 215, 250, 2, 32, 243, 247, 122, 232, 145, 98, 201, 184, 5, 232, 221, 39, 54, 1, 65, 38, 15, 226, 251, 108, 69, 207, 201, 99, 204, 118, 82, 81, 174, 77, 187, 17, 124, 22, 23, 33, 132, 81, 69, 254, 3, 105, 197, 103, 73, 208, 82, 71, 119, 113, 82, 163, 171, 195, 7, 7, 126, 83, 165, 163, 215, 220, 84, 245, 98, 198, 161, 85, 67, 2, 172, 30, 91, 175, 90, 134, 72, 142, 70, 115, 94])).  Unable to vote
```
Might be something wrong with the derivation in genesis?

wen-coding:
> > FYI - running a quick checkout with this code (https://buildkite.com/anza/infra-invalidator/builds/2266#0197eaa0-1a7e-49aa-aac6-3ca3cbfb3143), but it is failing because it is not observing any roots get created
> 
> Looks like some of the machines are failing because the BLSPubkey doesn't match:
> 
> ```
> thread 'solVotingLoop' panicked at core/src/alpenglow_consensus/voting_loop.rs:1054:13:
> Vote account bls_pubkey mismatch: Pubkey([15, 83, 249, 230, 130, 27, 145, 51, 216, 42, 88, 227, 221, 46, 239, 43, 21, 150, 32, 59, 49, 187, 242, 187, 84, 146, 15, 89, 118, 85, 176, 0, 161, 192, 161, 10, 203, 117, 115, 1, 175, 138, 111, 85, 227, 216, 11, 252, 16, 152, 64, 16, 123, 78, 232, 144, 215, 228, 241, 54, 144, 156, 108, 101, 81, 19, 85, 132, 225, 97, 67, 250, 0, 27, 124, 219, 85, 113, 218, 21, 181, 163, 172, 103, 57, 82, 97, 191, 226, 56, 164, 137, 163, 190, 55, 83]) (expected: Pubkey([5, 173, 38, 22, 84, 152, 116, 151, 84, 120, 38, 215, 250, 2, 32, 243, 247, 122, 232, 145, 98, 201, 184, 5, 232, 221, 39, 54, 1, 65, 38, 15, 226, 251, 108, 69, 207, 201, 99, 204, 118, 82, 81, 174, 77, 187, 17, 124, 22, 23, 33, 132, 81, 69, 254, 3, 105, 197, 103, 73, 208, 82, 71, 119, 113, 82, 163, 171, 195, 7, 7, 126, 83, 165, 163, 215, 220, 84, 245, 98, 198, 161, 85, 67, 2, 172, 30, 91, 175, 90, 134, 72, 142, 70, 115, 94])).  Unable to vote
> ```
> 
> Might be something wrong with the derivation in genesis?

Yeah, very likely the case, checking now.

bw-solana:
Note that this now appears to be working well in testing: https://buildkite.com/anza/infra-invalidator/builds/2308

AshwinSekar (approved):

PR #215, opened 2025-06-18
voting_loop: use sender from replay rather than bank forks lock

AshwinSekar:
#### Problem
Spamming the bank forks lock to check if replay has finished is bad for performance

#### Summary of Changes
Instead have replay notify the voting loop, will also be useful when we refactor to a fully event driven loop

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -63,6 +63,7 @@ use {
     },
 };
 
+/// Banks that have completed replay, but are yet to be voted on
 type PendingBlocks = BTreeMap<Slot, Arc<Bank>>;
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -60,6 +63,8 @@ use {
     },
 };
 
+type PendingBlocks = BTreeMap<Slot, Arc<Bank>>;
```
bw-solana, line 67: Would be helpful to add a comment here indicating these are blocks that have completed replay but need to be voted on

bw-solana (approved):
LGTM. Left one nit

bw-solana (approved):

PR #216, opened 2025-06-19
block_creation_loop: split out record and receive to separate thread

AshwinSekar:
#### Problem
Trying to be fancy and doing the block creation in the same loop as the transaction receive and record causes all sorts of deadlocks as:
- Consumer holds the freeze lock
- TransactionRecorder sends the record through the crossbeam channel and waits for a response
- If the PoHService does not respond, the freeze lock is held indefinitely
- This blocks the voting loop, and when creating a child block in the block creation loop
- As the block creation loop is stuck waiting for the lock, it doesn't process the record from TransactionRecorder

#### Summary of Changes
Instead keep it simple and run record and receive in a separate loop to ensure that the crossbeam channels are always free.
The consume worker will take work iteratively, so there will be points where the freeze lock is released.

### DIFF
--- a/core/src/alpenglow_consensus/block_creation_loop.rs
+++ b/core/src/alpenglow_consensus/block_creation_loop.rs
@@ -8,16 +8,13 @@ use {
         banking_trace::BankingTracer,
         replay_stage::{Finalizer, ReplayStage},
     },
-    crossbeam_channel::Receiver,
+    crossbeam_channel::{Receiver, RecvTimeoutError},
     solana_gossip::cluster_info::ClusterInfo,
     solana_ledger::{
         blockstore::Blockstore, leader_schedule_cache::LeaderScheduleCache,
         leader_schedule_utils::leader_slot_index,
     },
-    solana_poh::{
-        poh_recorder::{PohRecorder, Record, GRACE_TICKS_FACTOR, MAX_GRACE_SLOTS},
-        poh_service::PohService,
-    },
+    solana_poh::poh_recorder::{PohRecorder, Record, GRACE_TICKS_FACTOR, MAX_GRACE_SLOTS},
     solana_rpc::{rpc_subscriptions::RpcSubscriptions, slot_status_notifier::SlotStatusNotifier},
     solana_runtime::{
         bank::{Bank, NewBankOptions},
@@ -29,6 +26,7 @@ use {
             atomic::{AtomicBool, Ordering},
             Arc, Condvar, Mutex, RwLock,
         },
+        thread,
         time::{Duration, Instant},
     },
     thiserror::Error,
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -110,6 +108,35 @@ enum StartLeaderError {
     VoteNotRooted,
 }
 
+fn start_receive_and_record_loop(
+    exit: Arc<AtomicBool>,
+    poh_recorder: Arc<RwLock<PohRecorder>>,
+    record_receiver: Receiver<Record>,
+) {
+    while !exit.load(Ordering::Relaxed) {
+        match record_receiver.recv_timeout(Duration::from_millis(100)) {
```
bw-solana: might be nice to comment why we chose 100ms. I'm assuming this could just be a `recv` if not for the need to check exit flag at some frequency

bw-solana (approved):
LGTM. Left one nit

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: yeah exactly, if we don't check the exit flag process won't exit cleanly. Added a comment and also extended this to 400ms.

bw-solana (approved):

bw-solana (approved):

carllin:
Just to be clear, "the block creation loop is stuck waiting for the lock" is because block creation was waiting on replay to finish?

bw-solana:
> > Block creation trying to acquire the bank hash lock as parent of freezing the parent.
> 
> https://github.com/anza-xyz/alpenglow/blob/832ceb68d38f24fae9c2a1a433bc76495084c61f/core/src/alpenglow_consensus/block_creation_loop.rs#L351-L359
> 
> block creation loop doesn't actually freeze the parent, just waits for the freeze to happen in replay right?

It actually calls freeze as part of _new_from_parent. We call freeze twice but skip the second one after checking the hash. 

But even if we skip the freeze in block creation, we still deadlocked because block creation loop would stop pulling PoH record records when waiting for leader slot to come around, which would end up never happening sometimes because part of being ready for leader is freezing parent, and sometimes replay would never freeze because it couldn’t get the freeze lock because consumer thread was holding it.

An alternative solution is hacking up block creation to always check PoH receiver periodically, but this is much messier.

carllin:
> It actually calls freeze as part of _new_from_parent. We call freeze twice but skip the second one after checking the hash.

Ah yeah, I see, thanks!

>  and sometimes replay would never freeze because it couldn’t get the freeze lock because consumer thread was holding it.

Ah ok this was more what I was alluding to above. Sneaky sneaky 🥷

PR #220, opened 2025-06-21
hack: prune progress map in replay

AshwinSekar:
ProgressMap is not plugged into the voting loop (as half the fields are not necessary anymore).
As a hack, prune it in replay every time we complete a bank for now, until we can handle this properly.

#### Summary of Changes

Fixes #


AshwinSekar:
Closing in favor of #230

PR #222, opened 2025-06-22
poh: populate next leader and current ticks for proper leader detection

AshwinSekar:
banking stage, forwarding stage, and transaction sending depend on leader detection which is derived from the leader schedule, leader tick height, and current tick height fields in poh recorder.

draft to see if we can populate these values accurately

#### Summary of Changes

Fixes #

### DIFF
--- a/core/src/alpenglow_consensus/block_creation_loop.rs
+++ b/core/src/alpenglow_consensus/block_creation_loop.rs
@@ -279,6 +279,7 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
             // Produce our next slot
             slot += 1;
             if slot > end_slot {
+                update_next_leader_slot(end_slot, &ctx);
                 trace!("{my_pubkey}: finished leader window {start_slot}-{end_slot}");
                 break;
             }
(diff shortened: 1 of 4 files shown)


core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -279,6 +279,7 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
             // Produce our next slot
             slot += 1;
             if slot > end_slot {
+                update_next_leader_slot(end_slot, &ctx);
```
AshwinSekar, line 282: IIUC since we don't update this, and we no longer reset poh (apart from below when creating a child bank), we'll be stuck with the current tick height as the max tick height of our latest 3rd leader slot (as a result of the poh reset), and the next leader tick height as the 4th leader slot.

Thus we will always think we are leader on the next tick, and would_be_leader will always be true.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -317,6 +320,16 @@ impl VotingLoop {
...
                 );
 
+                // If we're not the leader for this slot, we should adjust the poh
+                // tick height to the beginning of this slot to keep leader detection
+                // logic accurate
+                if !is_leader {
+                    poh_recorder
+                        .write()
+                        .unwrap()
+                        .set_tick_height_to_start_of_slot(current_slot);
```
AshwinSekar, line 330: don't know how accurate this current slot will be for transaction sending purposes, I think nozomi and triton use some sort of shred based approach?

AshwinSekar:
actually nvm I think I'm way off here, `clear_bank` should already handle updating the next leader tick height

PR #225, opened 2025-06-24
replay: stop sending to tower bft related crossbeam channels

AshwinSekar:
#### Problem
Unlikely that this is the cause of the memory leak, but these crossbeam channels are no longer received from once the feature flag is active. Since they are `unbounded` this could lead to memory growth:
```
ancestor_duplicate_slots_receiver
duplicate_confirmed_slots_receiver
gossip_verified_vote_hash_receiver
popular_pruned_forks_receiver
duplicate_slots_receiver
```

#### Summary of Changes
Stop sending on these channels when the feature flag is active. Later we can retire these services completely

### DIFF
--- a/core/src/window_service.rs
+++ b/core/src/window_service.rs
@@ -167,6 +167,13 @@ fn run_check_duplicate(
             }
         };
 
+        if root_bank
+            .feature_set
+            .is_active(&solana_feature_set::secp256k1_program_enabled::id())
+        {
+            return Ok(());
+        }
+
         // Propagate duplicate proof through gossip
         cluster_info.push_duplicate_shred(&shred1, &shred2)?;
         // Notify duplicate consensus state machine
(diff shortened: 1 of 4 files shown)


core/src/window_service.rs
```diff
@@ -167,6 +167,13 @@ fn run_check_duplicate(
...
 
+        if root_bank
+            .feature_set
+            .is_active(&solana_feature_set::secp256k1_program_enabled::id())
+        {
+            return Ok(());
+        }
+
         // Propagate duplicate proof through gossip
         cluster_info.push_duplicate_shred(&shred1, &shred2)?;
```
AshwinSekar, line 178: This also stops sending duplicate proofs through gossip, which takes care of the other `duplicate_slots_sender` on gossip proof receival from honest nodes.

we should have another PR to drop duplicate shred proof and epoch slots frozen messages in gossip itself when the feature flag is active

bw-solana (approved):

PR #226, opened 2025-06-24
replay: do not grab bank forks write lock if no banks to insert

AshwinSekar:
#### Problem
We grab the write lock every replay iteration even when there are no new banks to insert.

#### Summary of Changes
Only grab the lock if we're actually going to insert something.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -4535,16 +4535,18 @@ impl ReplayStage {
 
         let mut generate_new_bank_forks_write_lock =
             Measure::start("generate_new_bank_forks_write_lock");
-        let mut forks = bank_forks.write().unwrap();
-        let root = forks.root();
-        for (slot, bank) in new_banks {
-            if slot < root {
-                continue;
+        if !new_banks.is_empty() {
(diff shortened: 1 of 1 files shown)


bw-solana (approved):

core/src/replay_stage.rs
```diff
@@ -4535,16 +4535,18 @@ impl ReplayStage {
...
-            }
-            if forks.get(bank.parent_slot()).is_none() {
-                continue;
+        if !new_banks.is_empty() {
+            let mut forks = bank_forks.write().unwrap();
+            let root = forks.root();
+            for (slot, bank) in new_banks {
+                if slot < root {
+                    continue;
+                }
```
carllin, line 4544: could even filter this before grabbing the write lock, although realistically should never happen

carllin (approved):

core/src/replay_stage.rs
AshwinSekar: it's tricky because in the brief time between us dropping the bank forks read lock above and this loop, `handle_new_root` could have updated the root from the voting loop thread

PR #227, opened 2025-06-25
replay: remove reentrant locking in prune program cache

AshwinSekar:
#### Problem
We grab the bank forks read lock in order to call `prune_program_cache()` , which in turn grabs the read lock again. This can cause a deadlock if a writer requests the lock in between:
1) `handle_new_root` acquires the read lock in order to call prune_program_cache
2) `generate_new_bank_forks` asks for the write lock
3) `prune_program_cache` tries to acquire the read lock again during `fork_graph.read()`
4) Deadlock, as (3) will never complete due to the presence of (2)

#### Summary of Changes
Have `prune_program_cache` pass in `BankForks` to avoid the reentrant locking

### DIFF
--- a/program-runtime/src/loaded_programs.rs
+++ b/program-runtime/src/loaded_programs.rs
@@ -952,6 +952,10 @@ impl<FG: ForkGraph> ProgramCache<FG> {
             error!("Failed to lock fork graph for reading.");
             return;
         };
+        self.prune_locked(new_root_slot, new_root_epoch, &fork_graph);
+    }
+
+    pub fn prune_locked(&mut self, new_root_slot: Slot, new_root_epoch: Epoch, fork_graph: &FG) {
         let mut preparation_phase_ends = false;
         if self.latest_root_epoch != new_root_epoch {
             self.latest_root_epoch = new_root_epoch;
--- a/runtime/src/bank_forks.rs
+++ b/runtime/src/bank_forks.rs
@@ -542,7 +542,7 @@ impl BankForks {
 
     pub fn prune_program_cache(&self, root: Slot) {
         if let Some(root_bank) = self.banks.get(&root) {
-            root_bank.prune_program_cache(root, root_bank.epoch());
+            root_bank.prune_program_cache_locked(root, root_bank.epoch(), self);
         }
     }
 
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -1543,6 +1543,19 @@ impl Bank {
             .prune(new_root_slot, new_root_epoch);
     }
 
+    pub fn prune_program_cache_locked(
+        &self,
+        new_root_slot: Slot,
+        new_root_epoch: Epoch,
+        bank_forks: &BankForks,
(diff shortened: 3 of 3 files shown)


runtime/src/bank_forks.rs
```diff
@@ -542,7 +542,7 @@ impl BankForks {
 
     pub fn prune_program_cache(&self, root: Slot) {
         if let Some(root_bank) = self.banks.get(&root) {
-            root_bank.prune_program_cache(root, root_bank.epoch());
+            root_bank.prune_program_cache_locked(root, root_bank.epoch(), self);
```
bw-solana, line 545: is there any case where we could possibly come through here and root bank FG doesn't match the bank_forks we call in with? I don't really understand all the usages here well enough

program-runtime/src/loaded_programs.rs
```diff
@@ -952,6 +952,10 @@ impl<FG: ForkGraph> ProgramCache<FG> {
             error!("Failed to lock fork graph for reading.");
             return;
         };
+        self.prune_locked(new_root_slot, new_root_epoch, &fork_graph);
```
carllin, line 955: can we relegate this prune function to only tests now?

bw-solana (approved):
Looks reasonable to me.

Another possible solution would be switching `fn handle_new_root` to call:
```
let root_bank = bank_forks.read(requester_id).unwrap().get(new_root).expect("Root bank must exist");
root_bank.prune_program_cache(new_root, root_bank.epoch());
```

program-runtime/src/loaded_programs.rs
AshwinSekar: actually we can't, `blockstore_processor::load_frozen_forks` still uses the regular prune function, since it grabs the `root_bank` separately and drops the `bank_forks` lock in between

AshwinSekar: It's quite similar to brennan's solution below, i'm also open to using the same workflow in `handle_new_root`

runtime/src/bank_forks.rs
AshwinSekar: On snapshot load the root bank is initially deserialized with an empty program cache, however when initializing `BankForks::new_rw_arc()`, we fill in the forks graph as the `BankForks`.

Then `bank::_new_from_parent` will use the `transaction_processor` and thus the `program_cache / FG` from the parent bank so I believe this should be consistent. Also I don't think we should ever have multiple `BankForks` instances outside of tests.

carllin (approved):

runtime/src/bank_forks.rs
ksn6: thoughts on tossing in an assert, just in case root_bank and self don't match?

AshwinSekar: it's a bit tricky becauses it's not exactly the `root_bank` but the forks_graph inside the `root_bank`'s program cache. This would involve doing equality on `BankForks` which isn't trivial

PR #228, opened 2025-06-25
feat(vote): add bls pubkey to vote state view

ksn6:
#### Problem
We're slowly getting `alpenglow` in shape to merge into `agave`. In the process, we're moving voting structures from `alpenglow-vote` into `alpenglow`.

#### Summary of Changes
Add the `bls_pubkey` field to `VoteStateView`.

### DIFF
--- a/vote/src/vote_state_view/frame_v1_14_11.rs
+++ b/vote/src/vote_state_view/frame_v1_14_11.rs
@@ -95,8 +95,9 @@ impl VoteStateFrameV1_14_11 {
         self.epoch_credits_offset() + self.epoch_credits_frame.total_size()
     }
 
+    // TODO(ksn): add tests for this
     fn bls_pubkey_offset(&self) -> usize {
-        self.last_timestamp_offset() + core::mem::size_of::<BlsPubkey>()
+        self.last_timestamp_offset() + core::mem::size_of::<BlockTimestamp>()
     }
 }
 
--- a/vote/src/vote_state_view/frame_v3.rs
+++ b/vote/src/vote_state_view/frame_v3.rs
@@ -96,8 +96,9 @@ impl VoteStateFrameV3 {
         self.epoch_credits_offset() + self.epoch_credits_frame.total_size()
     }
 
+    // TODO(ksn): add tests for this
     fn bls_pubkey_offset(&self) -> usize {
-        self.last_timestamp_offset() + core::mem::size_of::<BlsPubkey>()
+        self.last_timestamp_offset() + core::mem::size_of::<BlockTimestamp>()
     }
 }
 


vote/src/vote_state_view/frame_v1_14_11.rs
```diff
@@ -92,6 +94,10 @@ impl VoteStateFrameV1_14_11 {
     fn last_timestamp_offset(&self) -> usize {
         self.epoch_credits_offset() + self.epoch_credits_frame.total_size()
     }
+
+    fn bls_pubkey_offset(&self) -> usize {
+        self.last_timestamp_offset() + core::mem::size_of::<BlsPubkey>()
```
wen-coding: I would guess here you should add size of last_timestamp?

vote/src/vote_state_view/frame_v3.rs
```diff
@@ -93,6 +95,10 @@ impl VoteStateFrameV3 {
     fn last_timestamp_offset(&self) -> usize {
         self.epoch_credits_offset() + self.epoch_credits_frame.total_size()
     }
+
+    fn bls_pubkey_offset(&self) -> usize {
+        self.last_timestamp_offset() + core::mem::size_of::<BlsPubkey>()
```
wen-coding: Same here.

vote/src/vote_state_view.rs
```diff
@@ -128,6 +130,12 @@ impl VoteStateView {
         }
     }
 
+    pub fn bls_pubkey(&self) -> &BlsPubkey {
+        let offset = self.frame.offset(Field::BlsPubkey);
+        // SAFETY: `frame` was created from `data`.
+        unsafe { &*(self.data.as_ptr().add(offset) as *const BlsPubkey) }
+    }
```
wen-coding, line 137: Can we add a test somewhere maybe?

vote/src/vote_state_view/frame_v1_14_11.rs
ksn6: Oh whoops, good catch.

vote/src/vote_state_view/frame_v3.rs
ksn6: Done

vote/src/vote_state_view.rs
ksn6: Added a TODO for now; will do this in a later PR - would like to get the overall logic in first and then add in tests.

jstarry:
Can we just add it to V4? I have to amend it anyways because I think we are getting rid of basis points for commission

AshwinSekar:
> Can we just add it to V4? I have to amend it anyways because I think we are getting rid of basis points for commission

I'm cool with this, saves us some code. 

V4 migration will happen automatically on any vote instruction correct? I wanted to have the migration happen via an instruction to ensure that all migrated vote states have a valid `BLSPubkey`, but it's not a huge deal. 
We can have it as an `Option<BLSPubkey>` and detect this on the agave side.

jstarry:
Added to the vote state v4 simd [here](https://github.com/solana-foundation/solana-improvement-documents/pull/185/commits/07edc0171931da9ab1225a36424684c286ff9541). Can you take a look? I assumed we wanted to store the compressed pubkey, please confirm if that's correct.

PR #234, opened 2025-06-30
voting_loop: keep poh tick height somewhat consistent

AshwinSekar:
#### Problem
Tick height is not updated except for during block production.

#### Summary of Changes
Simulate tick height based on voting loop slot to keep poh_current_slot somewhat accurate


AshwinSekar:
not really relevant at the moment. We can address the consumers of poh recorder's current_slot in the future

PR #235, opened 2025-07-01
test(local-cluster): liveness after intertwined skip-fallback and notar-fallback

ksn6:
#### Problem
Test to validate Alpenglow's ability to maintain liveness when nodes issue both NotarizeFallback and SkipFallback votes in an intertwined manner.

This test simulates a consensus scenario with four nodes having specific stake distributions:
- Node A: 40% + epsilon stake
- Node B: 40% - epsilon stake
- Node C: 20% - epsilon stake
- Node D: epsilon stake (minimal, acts as perpetual leader)

The test proceeds through two main stages:

#### Stage 1: Stable Network Operation
All nodes are voting normally for leader D's proposals, with notarization votes going through successfully and the network maintaining consensus.

#### Stage 2: Network Partition and Fallback Scenario
At slot 50, Node A's turbine is disabled, creating a network partition. This triggers the following sequence:
1. Node D (leader) proposes a block b1
2. Nodes B, C, and D can communicate and vote to notarize b1
3. Node A is partitioned and cannot receive b1, so it issues a skip vote
4. The vote distribution creates a complex fallback scenario:
   - Nodes B, C, D: Issue notarize votes initially, then skip fallback votes
   - Node A: Issues skip vote initially, then notarize fallback vote
5. This creates the specific vote pattern:
   - B, C, D: notarize + skip_fallback
   - A: skip + notarize_fallback

The test validates that:
- The network can handle intertwined fallback scenarios
- Consensus is maintained despite complex vote patterns
- The network continues to make progress and create new roots after the partition is resolved
- At least 10 new roots are created post-experiment to ensure sustained liveness

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -2,6 +2,7 @@
 use {
     alpenglow_vote::{
         bls_message::{BLSMessage, VoteMessage, BLS_KEYPAIR_DERIVE_SEED},
+        certificate::CertificateType,
         vote::Vote,
     },
     assert_matches::assert_matches,
@@ -6895,20 +6896,6 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
     let cluster = LocalCluster::new_alpenglow(&mut cluster_config, SocketAddrSpace::Unspecified);
     assert_eq!(NUM_NODES, cluster.validators.len());
 
-    // Create mapping from vote pubkeys to node indices
-    let vote_pubkey_to_node_index: HashMap<_, _> = validator_keys
-        .iter()
-        .enumerate()
-        .filter_map(|(index, keypair)| {
-            cluster
-                .validators
-                .get(&keypair.pubkey())
-                .map(|validator| (validator.info.voting_keypair.pubkey(), index))
-        })
-        .collect();
-
-    assert_eq!(NUM_NODES, vote_pubkey_to_node_index.len());
-
     /// Helper struct to manage experiment state and vote pattern tracking
     #[derive(Debug, PartialEq, Eq)]
     enum Stage {
@@ -6926,14 +6913,6 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
             }
         }
 
-        fn name(&self) -> &'static str {
-            match self {
-                Stage::Stability => "Stability",
-                Stage::ObserveSkipFallbacks => "ObserveSkipFallbacks",
-                Stage::ObserveLiveness => "ObserveLiveness",
-            }
-        }
-
         fn all() -> Vec<Stage> {
             vec![
                 Stage::Stability,
@@ -6948,7 +6927,6 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
         stage: Stage,
         vote_type_bitmap: HashMap<u64, [u8; 4]>, // slot -> [node_vote_pattern; 4]
         consecutive_pattern_matches: usize,
-        post_experiment_vote_weights: HashMap<u64, u64>, // slot -> total_weight
         post_experiment_roots: HashSet<u64>,
     }
 
@@ -6958,7 +6936,6 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
                 stage: Stage::Stability,
                 vote_type_bitmap: HashMap::new(),
                 consecutive_pattern_matches: 0,
-                post_experiment_vote_weights: HashMap::new(),
                 post_experiment_roots: HashSet::new(),
             }
         }
@@ -6993,23 +6970,12 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
             false
         }
 
-        fn record_finalization_vote(
-            &mut self,
-            slot: u64,
-            node_index: usize,
-            node_stakes: &[u64; 4],
-        ) {
-            let total_weight = self.post_experiment_vote_weights.entry(slot).or_insert(0);
-            *total_weight += node_stakes[node_index];
-
-            // Check if we have sufficient stake for finalization (60% threshold)
-            if *total_weight >= TOTAL_STAKE * 6 / 10 {
-                self.post_experiment_roots.insert(slot);
-            }
+        fn record_certificate(&mut self, slot: u64) {
+            self.post_experiment_roots.insert(slot);
         }
 
         fn sufficient_roots_created(&self) -> bool {
-            self.post_experiment_roots.len() >= 10
+            self.post_experiment_roots.len() >= 8
         }
     }
 
@@ -7031,55 +6997,55 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
                 let bls_message = bincode::deserialize::<BLSMessage>(&buffer[..bytes_received])
                     .expect("Failed to deserialize BLS message");
 
-                let BLSMessage::Vote(vote_message) = bls_message else {
-                    continue;
-                };
+                match bls_message {
+                    BLSMessage::Vote(vote_message) => {
+                        let vote = &vote_message.vote;
+                        let node_index = vote_message.rank as usize;
 
-                let vote = &vote_message.vote;
-                let node_index = vote_message.rank as usize;
+                        // Stage timeouts
+                        let elapsed_time = timer.elapsed();
 
-                // Stage timeouts
-                let elapsed_time = timer.elapsed();
+                        for stage in Stage::all() {
+                            if elapsed_time > stage.timeout() {
+                                panic!(
+                                    "Timeout during {:?}. node_c_turbine_disabled: {:#?}. Latest vote: {:#?}. Experiment state: {:#?}",
+                                    stage,
+                                    node_c_turbine_disabled.load(Ordering::Acquire),
+                                    vote,
+                                    experiment_state
+                                );
+                            }
+                        }
 
-                for stage in Stage::all() {
-                    if elapsed_time > stage.timeout() {
-                        panic!(
-                            "Timeout during {} stage. node_c_turbine_disabled: {:#?}. Latest vote: {:#?}. Experiment state: {:#?}",
-                            stage.name(),
-                            node_c_turbine_disabled.load(Ordering::Acquire),
-                            vote,
-                            experiment_state
-                        );
+                        // Stage 1: Wait for stability, then introduce partition at slot 20
+                        if vote.slot() == 20 && !node_c_turbine_disabled.load(Ordering::Acquire) {
+                            node_c_turbine_disabled.store(true, Ordering::Release);
+                            experiment_state.stage = Stage::ObserveSkipFallbacks;
+                        }
+
+                        // Stage 2: Monitor for expected fallback vote patterns
+                        if experiment_state.stage == Stage::ObserveSkipFallbacks {
+                            experiment_state.record_vote_bitmap(vote.slot(), node_index, vote);
+
+                            // Check if we've observed the expected pattern for 3 consecutive slots
+                            if experiment_state.matches_expected_pattern() {
+                                node_c_turbine_disabled.store(false, Ordering::Release);
+                                experiment_state.stage = Stage::ObserveLiveness;
+                            }
+                        }
                     }
-                }
+                    BLSMessage::Certificate(cert_message) => {
+                        // Stage 3: Verify continued liveness after partition resolution
+                        if experiment_state.stage == Stage::ObserveLiveness
+                            && [CertificateType::Finalize, CertificateType::FinalizeFast]
+                                .contains(&cert_message.certificate.certificate_type)
+                        {
+                            experiment_state.record_certificate(cert_message.certificate.slot);
 
-                // Stage 1: Wait for stability, then introduce partition at slot 50
-                if vote.slot() == 50 && !node_c_turbine_disabled.load(Ordering::Acquire) {
-                    node_c_turbine_disabled.store(true, Ordering::Release);
-                    experiment_state.stage = Stage::ObserveSkipFallbacks;
-                }
-
-                // Stage 2: Monitor for expected fallback vote patterns
-                if experiment_state.stage == Stage::ObserveSkipFallbacks {
-                    experiment_state.record_vote_bitmap(vote.slot(), node_index, vote);
-
-                    // Check if we've observed the expected pattern for 3 consecutive slots
-                    if experiment_state.matches_expected_pattern() {
-                        node_c_turbine_disabled.store(false, Ordering::Release);
-                        experiment_state.stage = Stage::ObserveLiveness;
-                    }
-                }
-
-                // Stage 3: Verify continued liveness after partition resolution
-                if experiment_state.stage == Stage::ObserveLiveness && vote.is_finalize() {
-                    experiment_state.record_finalization_vote(
-                        vote.slot(),
-                        node_index,
-                        &node_stakes,
-                    );
-
-                    if experiment_state.sufficient_roots_created() {
-                        break;
+                            if experiment_state.sufficient_roots_created() {
+                                break;
+                            }
+                        }
                     }
                 }
             }


programs/sbf/Cargo.lock
```diff
@@ -229,7 +229,7 @@ checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 [[package]]
 name = "alpenglow-vote"
 version = "0.1.0"
-source = "git+ssh://git@github.com/solana-program/alpenglow-vote.git?rev=e4f33e5#e4f33e57647f8ca15789d347e763d28478823b5d"
```
AshwinSekar: do we need this change?

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+    // Start vote monitoring thread
+    let vote_listener_thread = std::thread::spawn({
+        let vote_pubkey_to_node_index = vote_pubkey_to_node_index.clone();
+        let node_c_turbine_disabled = node_c_turbine_disabled.clone();
+
+        move || {
+            let mut buffer = [0u8; 65_535];
+            let mut experiment_state = ExperimentState::new();
+
+            loop {
```
AshwinSekar, line 6992: Lets add some slot timeouts here to panic and print `expirement_state` so that we have some debugging if this test fails

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+    // Configure stake distribution for the four-node cluster
+    const TOTAL_STAKE: u64 = 10 * DEFAULT_NODE_STAKE;
+    const SLOTS_PER_EPOCH: u64 = MINIMUM_SLOTS_PER_EPOCH;
+    const EPSILON: u64 = 1;
+
+    let node_stakes = [
+        EPSILON,                        // Node A: epsilon
+        TOTAL_STAKE * 2 / 10 - EPSILON, // Node B: 20% - epsilon
+        TOTAL_STAKE * 4 / 10 + EPSILON, // Node C: 40% + epsilon
+        TOTAL_STAKE * 4 / 10 - EPSILON, // Node D: 40% - epsilon
```
bw-solana: Do we have constants for these thresholds anywhere? I'm assuming these are derived specifically to be just above or below the important Alpenglow vote decision stake thresholds

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+/// - The network continues to make progress and create new roots after the partition is resolved
+/// - At least 10 new roots are created post-experiment to ensure sustained liveness
+#[test]
+#[serial]
+fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
+    solana_logger::setup_with_default(AG_DEBUG_LOG_FILTER);
+
+    // Configure stake distribution for the four-node cluster
+    const TOTAL_STAKE: u64 = 10 * DEFAULT_NODE_STAKE;
+    const SLOTS_PER_EPOCH: u64 = MINIMUM_SLOTS_PER_EPOCH;
```
bw-solana: Do we need short epochs for this test? Historically with tower BFT, they gave us fits because we needed to confirm every slot to root something and generate the leader schedule and not get stuck. If this is still set to 32, it probably doesn't actually make sense anymore in an AG world

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+        validator_keys: Some(
+            validator_keys
+                .iter()
+                .cloned()
+                .zip(std::iter::repeat(true))
+                .collect(),
+        ),
+        slots_per_epoch: SLOTS_PER_EPOCH,
+        stakers_slot_offset: SLOTS_PER_EPOCH,
+        ticks_per_slot: DEFAULT_TICKS_PER_SLOT,
```
bw-solana: this is already default, can drop it

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+                post_experiment_vote_weights: HashMap::new(),
+                post_experiment_roots: HashSet::new(),
+            }
+        }
+
+        fn record_vote_pattern(&mut self, slot: u64, node_index: usize, vote: &Vote) {
+            let (_, vote_type) = _vote_to_tuple(vote);
+            let slot_pattern = self.vote_patterns.entry(slot).or_insert([0u8; 4]);
+
+            assert!(node_index < 4, "Invalid node index: {}", node_index);
```
bw-solana: can use `num_nodes`

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+        })
+        .collect();
+
+    assert_eq!(vote_pubkey_to_node_index.len(), num_nodes);
+
+    /// Helper struct to manage experiment state and vote pattern tracking
+    struct ExperimentState {
+        observe_skip_fallbacks: bool,
+        check_for_roots: bool,
+        vote_patterns: HashMap<u64, [u8; 4]>, // slot -> [node_vote_pattern; 4]
```
bw-solana: what is a vote pattern in this context?

Edit: I see now. Maybe call it a vote_type_bitmap?

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+            &mut self,
+            slot: u64,
+            node_index: usize,
+            node_stakes: &[u64; 4],
+        ) {
+            let total_weight = self.post_experiment_vote_weights.entry(slot).or_insert(0);
+            *total_weight += node_stakes[node_index];
+
+            // Check if we have sufficient stake for finalization (60% threshold)
+            const TOTAL_STAKE: u64 = 10 * DEFAULT_NODE_STAKE;
```
bw-solana: I think we have `TOTAL_STAKE` defined above already

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+                    vote_parser::parse_alpenglow_vote_transaction(&vote_transaction)
+                        .expect("Failed to parse vote transaction");
+
+                let node_index = vote_pubkey_to_node_index[&vote_pubkey];
+                let vote = parsed_vote
+                    .as_alpenglow_transaction_ref()
+                    .expect("Failed to get alpenglow vote reference");
+
+                // Stage 1: Wait for stability, then introduce partition at slot 50
+                if vote.slot() == 50 && !node_c_turbine_disabled.load(Ordering::Acquire) {
```
bw-solana: `>= 50` to be safe?

local-cluster/tests/local_cluster.rs
```diff
@@ -6409,3 +6409,258 @@ fn test_alpenglow_ensure_liveness_after_single_notar_fallback() {
...
+                    experiment_state.record_vote_pattern(vote.slot(), node_index, vote);
+
+                    // Check if we've observed the expected pattern for 3 consecutive slots
+                    if experiment_state.matches_expected_pattern() {
+                        node_c_turbine_disabled.store(false, Ordering::Release);
+                        experiment_state.check_for_roots = true;
+                    }
+                }
+
+                // Stage 3: Verify continued liveness after partition resolution
```
bw-solana: These stages might be easier to follow if written like a state machine

programs/sbf/Cargo.lock
ksn6: yep - `alpenglow-vote` is public so we'll want s/ssh/https/ here

in any case, this all goes away once we merge `alpenglow-vote` in with `alpenglow`

local-cluster/tests/local_cluster.rs
ksn6: wdym by "slot timeouts"?

local-cluster/tests/local_cluster.rs
ksn6: The thresholds themselves don't really have constants associated with them. I selected the stakes such that we'd be able to trigger specific cases explored in this test.

Throughout these tests + upcoming tests, we'll frequently see stakes of the form `20 * C +/- epsilon` where `C \in 0..=5`. This naturally occurs because 20 is the GCD of the constants appearing in the inequalities present for the SafeToNotar and SafeToSkip conditions (see https://drive.google.com/file/d/1y_7ddr8oNOknTQYHzXeeMD2ProQ0WjMs/view, page 18 and 19).

`TOTAL_STAKE` represents the total stake in the cluster
`EPSILON` is just some small amount of stake; it's 1 lamport

local-cluster/tests/local_cluster.rs
ksn6: done

local-cluster/tests/local_cluster.rs
ksn6: done

local-cluster/tests/local_cluster.rs
ksn6: done

local-cluster/tests/local_cluster.rs
ksn6: done

local-cluster/tests/local_cluster.rs
ksn6: we're guaranteed to view at least one vote for slot 50, so we should be safe here

local-cluster/tests/local_cluster.rs
ksn6: done

local-cluster/tests/local_cluster.rs
ksn6: Makes sense - removed.

local-cluster/tests/local_cluster.rs
AshwinSekar: basically if this test fails it will run forever until it's killed by the CI runner and we won't have any output on why it failed. Would be nice to have a timeout for each stage and if it doesn't complete by then print some state information.

local-cluster/tests/local_cluster.rs
bw-solana: Yeah, I'm thinking the `2/10` and `4/10` are going to show up frequently. Not going to die on this hill tho

bw-solana (approved):
LGTM

local-cluster/tests/local_cluster.rs
ksn6: sure, done

local-cluster/tests/local_cluster.rs
ksn6: Yep, in general, integer multiples of 1/5 tend to show up a lot. E.g., also see: https://github.com/qkniep/alpenglow/blob/03ca0d204e6b5f47228b6d69dd6dea07509c9a22/src/consensus/pool/slot_state.rs#L189-L203

EDIT: closing this convo for now - feel free to reopen with any more questions.

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+                stage: Stage::Stability,
+                vote_type_bitmap: HashMap::new(),
+                consecutive_pattern_matches: 0,
+                post_experiment_vote_weights: HashMap::new(),
+                post_experiment_roots: HashSet::new(),
+            }
+        }
+
+        fn record_vote_bitmap(&mut self, slot: u64, node_index: usize, vote: &Vote) {
+            let (_, vote_type) = _vote_to_tuple(vote);
```
wen-coding, line 6944: Sorry I forgot, can we just use VoteType here?

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+        ),
+        ..ClusterConfig::default()
+    };
+
+    // Initialize the cluster
+    let cluster = LocalCluster::new_alpenglow(&mut cluster_config, SocketAddrSpace::Unspecified);
+    assert_eq!(NUM_NODES, cluster.validators.len());
+
+    // Create mapping from vote pubkeys to node indices
+    let vote_pubkey_to_node_index: HashMap<_, _> = validator_keys
```
wen-coding: Do we still need this?

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+        ObserveSkipFallbacks,
+        ObserveLiveness,
+    }
+
+    impl Stage {
+        fn timeout(&self) -> Duration {
+            match self {
+                Stage::Stability => Duration::from_secs(60),
+                Stage::ObserveSkipFallbacks => Duration::from_secs(120),
+                Stage::ObserveLiveness => Duration::from_secs(180),
```
wen-coding, line 6912: Err, this test will really run for 180 seconds? Can we make it shorter?

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+            }
+        }
+
+        fn name(&self) -> &'static str {
```
wen-coding: Debug is derived for your enum, so you don't really need this right?

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+            *total_weight += node_stakes[node_index];
+
+            // Check if we have sufficient stake for finalization (60% threshold)
+            if *total_weight >= TOTAL_STAKE * 6 / 10 {
+                self.post_experiment_roots.insert(slot);
+            }
+        }
+
+        fn sufficient_roots_created(&self) -> bool {
+            self.post_experiment_roots.len() >= 10
```
wen-coding: I guess this number can be smaller so our test ends faster?

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+                            "Timeout during {} stage. node_c_turbine_disabled: {:#?}. Latest vote: {:#?}. Experiment state: {:#?}",
+                            stage.name(),
+                            node_c_turbine_disabled.load(Ordering::Acquire),
+                            vote,
+                            experiment_state
+                        );
+                    }
+                }
+
+                // Stage 1: Wait for stability, then introduce partition at slot 50
```
wen-coding: Are we really only stable at slot 50? That sounds like a lot :)

local-cluster/tests/local_cluster.rs
```diff
@@ -6770,3 +6791,302 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
...
+
+                    // Check if we've observed the expected pattern for 3 consecutive slots
+                    if experiment_state.matches_expected_pattern() {
+                        node_c_turbine_disabled.store(false, Ordering::Release);
+                        experiment_state.stage = Stage::ObserveLiveness;
+                    }
+                }
+
+                // Stage 3: Verify continued liveness after partition resolution
+                if experiment_state.stage == Stage::ObserveLiveness && vote.is_finalize() {
```
wen-coding: In this stage we can also check finalization cert right? Then you don't need to record and check total stake for finalization votes yourself. Just count how many new finalization certs you have observed.

local-cluster/tests/local_cluster.rs
ksn6: Done

local-cluster/tests/local_cluster.rs
ksn6: The entire test should run in < 40 seconds; this is just a timeout, in the off chance things go off the rails.

local-cluster/tests/local_cluster.rs
ksn6: Done

local-cluster/tests/local_cluster.rs
ksn6: Should probably check a few leader runs just to ensure that things are really fine. I'll make this 8 for now.

local-cluster/tests/local_cluster.rs
ksn6: It's just an arbitrary number that's low-ish; will make this 20 instead.

local-cluster/tests/local_cluster.rs
ksn6: Nice, done.

local-cluster/tests/local_cluster.rs
ksn6: We could, though this is a bit more cumbersome to do. Ideally `vote_type` ends up being a `u8` that's a power of two, so that we can easily perform set equality checks in `matches_expected_pattern`.

wen-coding (approved):

Issue #121, opened 2025-03-30
Handle set identity in voting loop


AshwinSekar:
Validators use `set-identity` in order to perform quick failover between machines https://docs.anza.xyz/operations/guides/validator-failover

When the `set-identity` cli is run, the pubkey on `cluster_info.keypair()` will change. 
We need to update the voting and block creation loops to handle this pubkey changing.

As a guide here's the logic to handle this in agave:
* Check if set-identity has been called during the startup procedure and reload tower + update pubkey if so https://github.com/anza-xyz/agave/blob/e72266ca6ff6505a9b21d608d81becbba2c38db9/core/src/replay_stage.rs#L622
* Check if set-identity has been called in steady state, and reload tower + update pubkey if so https://github.com/anza-xyz/agave/blob/e72266ca6ff6505a9b21d608d81becbba2c38db9/core/src/replay_stage.rs#L1065

In alpenglow we will need to do something similar:
* If set-identity has been called during startup, reload the vote history + update pubkey https://github.com/anza-xyz/alpenglow/blob/34d86a1c8651b3455b440915a9b510e56126720d/core/src/alpenglow_consensus/voting_loop.rs#L198
* At some point in the main `loop`, check if set-identity has been called and update vote history + pubkey
* If set-identity has been called during startup, update pubkey in block creation loop https://github.com/anza-xyz/alpenglow/blob/34d86a1c8651b3455b440915a9b510e56126720d/core/src/alpenglow_consensus/block_creation_loop.rs#L169
* Also at each leader window start, check and update pubkey in block creation loop

fahimahmedx:
Taking this on!

fahimahmedx:
In `voting_loop.rs`, is this along the right track?

```
if my_pubkey != vote_history.node_pubkey {
            // set-identity has been called during startup
            let my_old_pubkey = vote_history.node_pubkey;
            vote_history = match VoteHistory::restore(vote_history_storage, &my_pubkey) {
                Ok(vote_history) => vote_history,
                    Err(err) => {
                        error!(
                            "Unable to load new vote history when attempting to change identity from {} \
                             to {} on VotingLoop startup, Exiting: {}",
                            my_old_pubkey, my_pubkey, err
                        );
                        return;
                    }
            }
        }
```

Some follow up questions:

[...]

PR #236, opened 2025-07-03
Handle set-identity in voting loop

fahimahmedx:
#### Problem
Implements #121

#### Summary of Changes
For the voting loop, I've added code that fetches the vote history for the new pubkey, if the `set-identity` was called during startup. If it was called during the main loop, I update the pubkey and fetch the vote history for the new pubkey, and update related contexts & certificate pool where the pubkey is used in.
* For the main loop, I check this before the pubkey is used, such as in the `is_leader` boolean.

For the block creation loop, I update the pubkey if `set-identity` was called during the main loop.
* For the startup of the block creation loop, the original issue mentions "If set-identity has been called during startup, update pubkey in block creation loop". I'm unsure if I'm misunderstanding, but isn't this already implemented in [line 170](https://github.com/anza-xyz/alpenglow/blob/34d86a1c8651b3455b440915a9b510e56126720d/core/src/alpenglow_consensus/block_creation_loop.rs#L170) where it fetches the latest pubkey during the function's startup?

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -275,7 +275,7 @@ impl VotingLoop {
                     Err(err) => {
                         error!(
                                 "Unable to load new vote history when attempting to change identity from {} \
-                                 to {} on VotingLoop startup, Exiting: {}",
+                                 to {} in voting loop, Exiting: {}",
                                 my_old_pubkey, my_pubkey, err
                             );
                         return;
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -118,7 +118,6 @@ struct SharedContext {
     bank_forks: Arc<RwLock<BankForks>>,
     rpc_subscriptions: Arc<RpcSubscriptions>,
     vote_receiver: VoteReceiver,
-    // TODO(ashwin): integrate with gossip set-identity
     my_pubkey: Pubkey,
 }
 
@@ -284,13 +283,8 @@ impl VotingLoop {
                 voting_context.identity_keypair = identity_keypair.clone();
                 shared_context.my_pubkey = my_pubkey;
                 
-                // Recreate certificate pool with new identity
-                cert_pool = certificate_pool::load_from_blockstore(
-                    &my_pubkey,
-                    &root_bank_cache.root_bank(),
-                    blockstore.as_ref(),
-                    Some(certificate_sender),
-                );
+                // Update the certificate pool's pubkey (which is used for logging purposes)
+                cert_pool.update_pubkey(my_pubkey);
                 
                 warn!("{my_pubkey}: Identity changed from {my_old_pubkey} to {my_pubkey}");
             }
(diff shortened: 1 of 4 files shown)


fahimahmedx:
As a side note, should I be splitting this PR into two PR's (one for the voting loop, another for block creation loop)?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -246,6 +262,39 @@ impl VotingLoop {
...
+                            );
+                            return;
+                        }
+                };
+                
+                // Update contexts with new identity
+                voting_context.identity_keypair = identity_keypair.clone();
+                shared_context.my_pubkey = my_pubkey;
+                
+                // Recreate certificate pool with new identity
```
AshwinSekar: this part is not actually necessary, the certificate pool is not distinguished by pubkey, it's only used for logging.
For logging purposes, let's have a function that just updates the pubkey on the cert pool and parent ready tracker

fahimahmedx: Thanks for the heads up! Fixed, let me know what you think!

fahimahmedx:
I assume that the two failing checks "buildkite/alpenglow" and "buildkite/alpenglow/checks" need to pass before my PR can be merged. 

@AshwinSekar it doesn't seem like I have access to the buildkite build log details (when clicking into the failing tests to see what I need to fix). Is it supposed to be private, and if so what is causing the build to fail?

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -246,6 +261,37 @@ impl VotingLoop {
...
+
+                voting_context.vote_history = match VoteHistory::restore(
+                    vote_history_storage.as_ref(),
+                    &my_pubkey,
+                ) {
+                    Ok(vote_history) => vote_history,
+                    Err(err) => {
+                        error!(
+                                "Unable to load new vote history when attempting to change identity from {} \
+                                 to {} on VotingLoop startup, Exiting: {}",
```
AshwinSekar:
```suggestion
                                 to {} in voting loop, Exiting: {}",
```

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -190,13 +189,29 @@ impl VotingLoop {
...
         if my_pubkey != vote_history.node_pubkey {
-            todo!();
+            // set-identity has been called during startup
+            let my_old_pubkey = vote_history.node_pubkey;
+            vote_history = match VoteHistory::restore(vote_history_storage.as_ref(), &my_pubkey) {
+                Ok(vote_history) => vote_history,
+                Err(err) => {
+                    error!(
+                            "Unable to load new vote history when attempting to change identity from {} \
+                             to {} on VotingLoop startup, Exiting: {}",
```
AshwinSekar:
```suggestion
                             to {} on voting loop startup, Exiting: {}",
```

AshwinSekar (approved):
looks good just a couple of nits about the logs

AshwinSekar:
Great work! Appreciate your contribution.

PR #238, opened 2025-07-04
voting_loop: send optimistically confirmed BankNotification

### LINKED ISSUE #240: Prioritization fee cache causes unbounded memory growth
Documenting the memory leak for here for posterity

## Prioritization fee cache

The prioritization fee cache receives updates for each executed transaction batch. 
https://github.com/anza-xyz/alpenglow/blob/5d7b7e206664a4d2f765cf561df44d3d501d9568/ledger/src/blockstore_processor.rs#L271

The cache then sends the fees and writable accounts for each transaction via crossbeam for consumption in another thread.
https://github.com/anza-xyz/alpenglow/blob/5d7b7e206664a4d2f765cf561df44d3d501d9568/runtime/src/prioritization_fee_cache.rs#L238-L244

The receiving thread adds these writable accounts to a `UnfinalizedPrioritizationFees: BTreeMap<Slot, HashMap<BankId, PrioritizationFee>>`
https://github.com/anza-xyz/alpenglow/blob/5d7b7e206664a4d2f765cf561df44d3d501d9568/runtime/src/prioritization_fee_cache.rs#L375-L385

This unfinalized map is only pruned when the receiving thread receives a bank finalized notification which is only triggered by an Optimistic Confirmation RPC notification:
https://github.com/anza-xyz/alpenglow/blob/5d7b7e206664a4d2f765cf561df44d3d501d9568/rpc/src/optimistically_confirmed_bank_tracker.rs#L199-L200
https://github.com/anza-xyz/alpenglow/blob/5d7b7e206664a4d2f765cf561df44d3d501d9568/runtime/src/prioritization_fee_cache.rs#L260-L262

In alpenglow we do not send OC notifications

AshwinSekar:
#### Problem
We no longer send the optimistically confirmed notification as Alpenglow has no notion of OC.
This breaks RPC and most importantly the prioritization fee cache, as it relies on the RPC notification in order to perform cleanup. See https://github.com/anza-xyz/alpenglow/issues/240.

#### Summary of Changes
Similar to inserting the new highest finalized slot as an OC slot in blockstore, also send the `BankNotification::OptimisticallyConfirmed`

In the future we can deprecate and look to remove the OC notifications/commitment level across the board.

Fixes #240

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -504,6 +504,10 @@ impl VotingLoop {
                 new_root, &e
             );
         }
+        // It is critical to send the OC notification in order to keep compatibility with
+        // the RPC API. Additionally the PrioritizationFeeCache relies on this notification
+        // in order to perform cleanup. In the future we will look to deprecate OC and remove
+        // these code paths.
         if let Some(config) = bank_notification_sender {
             config
                 .sender


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -504,6 +504,12 @@ impl VotingLoop {
                 new_root, &e
             );
         }
+        if let Some(config) = bank_notification_sender {
```
bw-solana, line 511: might be nice to add a comment here similar to what you have in the PR description

bw-solana (approved):

PR #239, opened 2025-07-04
voting_loop: save vote even if tx generation fails

AshwinSekar:
#### Problem
Unstaked nodes will never store their votes, this causes problems in the voted checks in the voting loop, which slows down catchup.

#### Summary of Changes
Always store the vote

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -829,6 +829,11 @@ impl VotingLoop {
         cert_pool: &mut CertificatePool<LegacyVoteCertificate>,
         context: &mut VotingContext,
     ) {
+        // Update and save the vote history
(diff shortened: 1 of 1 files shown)


qkniep:
Why would unstaked nodes run the voting loop at all?

AshwinSekar:
> Why would unstaked nodes run the voting loop at all?

Unstaked here can be ephemeral. In a failover setup the backup node will be running with a dummy identity key and operators will use the cli to enable the proper key which allows vote generation to succeed. In these cases we want the voting loop current slot, timers, and certificate pool to be up to date for fast failover.

Additionally it's cumbersome to maintain a separate pathway for memory cleanup - the voting loop is responsible for "rooting" most in-memory data structures in Agave based on the ingested finalization certificates and replayed banks.

PR #241, opened 2025-07-06
WIP, test(local-cluster): liveness after triple notar fallback

ksn6:
#### Problem
Let's get Alpenglow to have a node issue three notar fallbacks in an interated fashion, and then ensure that the cluster remains live.

We'll have five nodes with stakes:
- Node A (Leader): epsilon
- Node B: 20% - epsilon
- Node C: 40%
- Node D: 20%
- Node E: 20%

(1) Stage 1: Initially, the cluster is progressing normally, with Node A as the leader.

(2) Stage 2: A is Byzantine and entirely stops sending blocks to node C (by disabling turbine).
  - Node C now keeps issuing skips.
  - The cluster still progresses, since 60% of the stake is still active.

(3) Stage 3: A now additionally equivocates by:
  - (a) sending itself and B block b1
  - (b) sending D block b2
  - (c) sending E block b3

  Nodes A, B, D, and E issue two notar fallbacks each. This is because each node (a) observes node C's skip which has 40% stake and observes two other blocks besides their own received block with 20% stake each, satisfying the SafeToNotar condition twice.

  Node C, on the other hand, issues three notar fallbacks. This is because it observes votes for three such blocks.

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -45,8 +45,15 @@ impl SigVerifier for BLSSigVerifier {
             stats_updater.received += 1;
 
             let message = match packet.deserialize_slice(..) {
-                Ok(msg) => msg,
+                Ok(msg) => {
+                    println!("Success! :: {:?}", &packet);
+                    msg
+                }
                 Err(e) => {
+                    println!(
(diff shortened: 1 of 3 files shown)


qkniep:
It isn't necessary for liveness here, but what should also happen is that every (honest) node sends a SkipFallback vote as well (except for C, who already sent skip).

Issue #250, opened 2025-07-08
Refactor voting loop into a proper event driven loop


AshwinSekar:
#### Problem
Loop has evolved and is now quite ugly to look at and trace through.
We are additionally limited as we have to loop through every slot sequentially, there is no need to wait for a certificate on every slot like earlier versions of the paper.

#### Proposed Solution
Similar to the paper, lets refactor to an event driven approach

Inputs:
- Certificate pool thread ingests votes and sends ParentReady, BlockNotarized, SafeToNotar, SafeToSkip and Standstill
- Replay stage and broadcast stage sends Block. Lets move setting the block id in Bank to these stages as well.
- Blockstore sends FirstShred
- Main thread can handle firing timers for regular skip and crashed leader skip

PR #251, opened 2025-07-08
only send block to voting loop once block_id is available

AshwinSekar:
#### Problem
We send all replayed blocks to the voting loop. Since shredding is asynchronous, replayed leader blocks might not be eligible for voting yet as the block id is missing.

This adds complexity to the voting loop which has to check blockstore and potentially update the bank with the block_id. Additionally the voting loop doesn't handle this transient condition properly , see #249 for example.

#### Summary of Changes
Instead we:
1. For non leader blocks, in replay we set the block id and send the completed bank to the voting loop
2. For leader blocks, instead broadcast stage will set the block id and send the completed bank
3. The voting loop now will always have the block id for any incoming banks, the blockstore fetch can be removed.
4. `try_notar` can only fail transiently for certificate checks now, refactor the blocking in the initial loop to address this

Part of #250

### DIFF
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -377,8 +377,6 @@ impl VotingLoop {
 
                     // Check if replay has successfully completed
                     if let Some(bank) = pending_blocks.get(&current_slot) {
-                        debug_assert!(bank.is_frozen());
-                        debug_assert!(bank.block_id().is_some());
                         // Vote notarize
                         if Self::try_notar(
                             &my_pubkey,
--- a/core/src/alpenglow_consensus/voting_loop.rs
+++ b/core/src/alpenglow_consensus/voting_loop.rs
@@ -18,17 +18,14 @@ use {
         commitment_service::{
             AlpenglowCommitmentAggregationData, AlpenglowCommitmentType, CommitmentAggregationData,
         },
-        replay_stage::{
-            CompletedBlock, CompletedBlockReceiver, Finalizer, ReplayStage, MAX_VOTE_SIGNATURES,
-        },
+        replay_stage::{Finalizer, ReplayStage, MAX_VOTE_SIGNATURES},
         voting_service::VoteOp,
     },
     alpenglow_vote::vote::Vote,
     crossbeam_channel::{RecvTimeoutError, Sender},
-    solana_feature_set::FeatureSet,
     solana_gossip::cluster_info::ClusterInfo,
     solana_ledger::{
-        blockstore::Blockstore,
+        blockstore::{Blockstore, CompletedBlock, CompletedBlockReceiver},
         leader_schedule_cache::LeaderScheduleCache,
         leader_schedule_utils::{
             first_of_consecutive_leader_slots, last_of_consecutive_leader_slots, leader_slot_index,
@@ -46,7 +43,6 @@ use {
     },
     solana_sdk::{
         clock::{Slot, NUM_CONSECUTIVE_LEADER_SLOTS},
-        hash::Hash,
         pubkey::Pubkey,
         signature::{Keypair, Signature, Signer},
         timing::timestamp,
@@ -382,11 +378,11 @@ impl VotingLoop {
                     // Check if replay has successfully completed
                     if let Some(bank) = pending_blocks.get(&current_slot) {
                         debug_assert!(bank.is_frozen());
+                        debug_assert!(bank.block_id().is_some());
                         // Vote notarize
                         if Self::try_notar(
                             &my_pubkey,
                             bank.as_ref(),
-                            &blockstore,
                             &mut cert_pool,
                             &mut voting_context,
                         ) {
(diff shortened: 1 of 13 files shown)


core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -689,13 +657,10 @@ impl VotingLoop {
...
         let parent_bank_hash = bank.parent_hash();
-        let parent_block_id = Self::get_set_block_id(
-            my_pubkey,
-            bank.parent().expect("Cannot vote on genesis").as_ref(),
-            blockstore,
-        )
-        // To account for child of genesis and snapshots we allow default block id
-        .unwrap_or_default();
+        let parent_block_id = bank
+            .parent_block_id()
```
AshwinSekar, line 659: Sanity check, assume that we are leader for `bank.parent()`. Since we've completed replay and have the block id of `bank` we must have the block id of the parent:
- If we are the leader for `bank` shredding must have completed for `bank`. Since shredding is serial this means shredding for `bank.parent()` is also complete and the block id for the parent must have been set.
- If we are not the leader for `bank`, the leader must have viewed the last data shred for `bank.parent()` so shredding is complete as well

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -382,29 +378,44 @@ impl VotingLoop {
...
-                    // Ingest replayed blocks
-                    match completed_block_receiver
-                        .recv_timeout(timeout.saturating_sub(skip_timer.elapsed()))
-                    {
-                        Ok(CompletedBlock { slot, bank }) => {
-                            pending_blocks.insert(slot, bank);
-                        }
-                        Err(RecvTimeoutError::Timeout) => (),
-                        Err(RecvTimeoutError::Disconnected) => return,
+                    // Block on certificate ingest
```
AshwinSekar, line 406: Now the only reason that `try_notar` can fail transiently is because we have not ingested the appropriate certificate, so we add that here. (should've already been here I just forgot).

Will hopefully pull off the refactor later this week so that this is cleaner.

core/src/replay_stage.rs
```diff
@@ -883,6 +877,10 @@ impl ReplayStage {
 
                 let forks_root = bank_forks.read().unwrap().root();
                 let start_leader_time = if !is_alpenglow_migration_complete {
+                    // TODO(ashwin): This will be moved to the event coordinator once we figure out
+                    // migration
+                    for _ in completed_block_receiver.try_iter() {}
```
AshwinSekar, line 882: Plumbing the feature flag into broadcast was messy so I have it always send `CompletedBlock` messages and we throw them away here for now.

core/src/alpenglow_consensus/voting_loop.rs
```diff
@@ -728,31 +693,27 @@ impl VotingLoop {
...
-    /// because the block_id/bank hash is not yet populated.
     fn vote_notarize(
         my_pubkey: &Pubkey,
         bank: &Bank,
-        blockstore: &Blockstore,
         cert_pool: &mut CertificatePool<LegacyVoteCertificate>,
         voting_context: &mut VotingContext,
     ) -> bool {
         debug_assert!(bank.is_frozen());
+        debug_assert!(bank.block_id().is_some());
```
bw-solana, line 709: probably can remove this given the expect below

core/src/validator.rs
```diff
@@ -1539,6 +1539,7 @@ impl Validator {
             Arc::<RwLock<repair::repair_service::OutstandingShredRepairs>>::default();
         let cluster_slots =
             Arc::new(crate::cluster_slots_service::cluster_slots::ClusterSlots::default());
+        let (completed_block_sender, completed_block_receiver) = bounded(1000);
```
bw-solana, line 1544: would be nice to indicate why the `1000`. I'm assuming more than a few items in this channel indicates serious problems with the voting loop

ledger/src/blockstore.rs
```diff
@@ -192,6 +192,14 @@ impl<T> AsRef<T> for WorkingEntry<T> {
     }
 }
 
+pub struct CompletedBlock {
```
bw-solana, line 195: blockstore is already so cooked. Should find a better home before we migrate upstream

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -119,6 +119,20 @@ pub(super) fn get_chained_merkle_root_from_parent(
     })
 }
 
+/// Set the block id on the bank and send it for consideration in voting
+pub(super) fn set_block_id_and_send(
```
bw-solana, line 123: maybe we could make some `completed_block_sender.rs` and push the typedefs and some of these sender/receiver helpers in there.

Doesn't need to gate this PR, but I'll be more picky when we migrate

core/src/alpenglow_consensus/voting_loop.rs
bw-solana: we could ingest votes and completed blocks concurrently, right? Should we change this to a `select!` ?

ledger/src/blockstore.rs
AshwinSekar: agreed haha, as part of the refactor I'm thinking of creating a votor crate that contains all the constants, events, certificate pool that core can depend on

core/src/alpenglow_consensus/voting_loop.rs
AshwinSekar: yeah that's true, it's a bit of a mess trying to refactor this into a `select!` at the moment. As part of the refactor these will be handled in separate threads so that they can happen concurrently. Are you cool if we punt till then?

turbine/src/broadcast_stage/broadcast_utils.rs
AshwinSekar: will do once I create the new crate

AshwinSekar:
running on invalidator looks good:
Screenshot 2025-07-09 at 3 43 18 PM

bw-solana (approved):
LGTM.

I'd still prefer a `select!` w/ a timeout so that we can handle votes/certs & completed block messages simultaneously, but this is still better than current state

PR #252, opened 2025-07-08
certificate pool: fix conflicting vote check for notarize fallback

AshwinSekar:
#### Problem
Certificate pool doesn't allow a notarize for block b and then a notarize fallback for block b'

#### Summary of Changes
Update the conflicting vote check to compare block id / bank hash when dealing with 2 block based vote types

### DIFF
--- a/core/src/alpenglow_consensus.rs
+++ b/core/src/alpenglow_consensus.rs
@@ -90,6 +90,12 @@ pub enum VoteType {
     SkipFallback,
 }
 
+impl VoteType {
+    fn is_block_type(&self) -> bool {
+        matches!(self, Self::Notarize | Self::NotarizeFallback)
+    }
+}
+
 pub const fn conflicting_types(vote_type: VoteType) -> &'static [VoteType] {
     match vote_type {
         VoteType::Finalize => &[VoteType::NotarizeFallback, VoteType::Skip],
--- a/core/src/alpenglow_consensus.rs
+++ b/core/src/alpenglow_consensus.rs
@@ -91,7 +91,7 @@ pub enum VoteType {
 }
 
 impl VoteType {
-    fn is_block_type(&self) -> bool {
+    fn is_notarize_type(&self) -> bool {
         matches!(self, Self::Notarize | Self::NotarizeFallback)
     }
 }
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -267,7 +267,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
                 if pool.has_prev_vote(
                     validator_vote_key,
                     conflicting_type
-                        .is_block_type()
+                        .is_notarize_type()
                         .then_some(vote_key.as_ref())
                         .flatten(),
                 ) {
@@ -330,7 +330,7 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
             }
             _ => (None, None),
         };
-        let vote_key = (block_id.is_some()).then_some(VoteKey {
+        let vote_key = block_id.map(|_| VoteKey {
             block_id,
             bank_hash,
         });
--- a/core/src/alpenglow_consensus/certificate_pool.rs
+++ b/core/src/alpenglow_consensus/certificate_pool.rs
@@ -260,10 +260,17 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
         slot: Slot,
         vote_type: VoteType,
         validator_vote_key: &Pubkey,
+        vote_key: Option<VoteKey>,
     ) -> Option<VoteType> {
         for conflicting_type in conflicting_types(vote_type) {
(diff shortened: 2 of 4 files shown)


core/src/alpenglow_consensus/vote_pool.rs
```diff
@@ -115,8 +115,14 @@ impl<VC: VoteCertificate> VotePool<VC> {
         Ok(())
     }
 
-    pub fn has_prev_vote(&self, validator_key: &Pubkey) -> bool {
-        self.prev_votes.contains_key(validator_key)
+    pub(crate) fn has_prev_vote(&self, validator_key: &Pubkey, vote_key: Option<&VoteKey>) -> bool {
```
AshwinSekar, line 118: after we stabilize BLS we can refactor certificate pool to be slot based and hopefully clean up a lot of this. 


Also in favor of just removing the conflicting vote check and leaving it to slashing. Just respect the # of votes per type

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -323,8 +330,12 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
             }
             _ => (None, None),
         };
+        let vote_key = (block_id.is_some()).then_some(VoteKey {
+            block_id,
+            bank_hash,
+        });
```
wen-coding, line 336: nit:
let vote_key = block_id.map(|_| VoteKey {
    block_id,
    bank_hash,
});

core/src/alpenglow_consensus/vote_pool.rs
wen-coding: The conflicting vote check is added partly because I wanted to avoid double counting stake. If you send Notarize(b) and NotarizeFallback(b) I can't count your stake twice. Slashing doesn't happen real time, I don't think we can totally rely on slashing for everything.

AshwinSekar: yeah I think we should still have a check for double counting stake, something simple like https://github.com/qkniep/alpenglow/blob/main/src/consensus/pool/slot_state.rs#L395

Just thinking that we can't catch all of the slashing cases (esp inter window) so might be easier to leave it to the slashing client rather than stuffing it in here

wen-coding: That's fine, we can discuss that when we refactor certificate pool. But I definitely want to keep the double counting stake check.

core/src/alpenglow_consensus/certificate_pool.rs
```diff
@@ -260,10 +260,17 @@ impl<VC: VoteCertificate> CertificatePool<VC> {
...
     ) -> Option<VoteType> {
         for conflicting_type in conflicting_types(vote_type) {
             if let Some(pool) = self.vote_pools.get(&(slot, *conflicting_type)) {
-                if pool.has_prev_vote(validator_vote_key) {
+                if pool.has_prev_vote(
+                    validator_vote_key,
+                    conflicting_type
+                        .is_block_type()
+                        .then_some(vote_key.as_ref())
+                        .flatten(),
```
wen-coding, line 272: I believe only `Notarize` and `NotarizeFallback` have vote_key not equal to `(None, None)`, so it's also possible that you blindly pass down `vote_key` and then check whether `vote_key.bank_hash` is `None` inside `vote_pool.rs`, then you don't really need `.is_block_type()`, and `vote_key` can just be of type `VoteKey` instead of `Option<VoteKey>`.

core/src/alpenglow_consensus/vote_pool.rs
```diff
@@ -115,8 +115,14 @@ impl<VC: VoteCertificate> VotePool<VC> {
...
+        if let Some(vote_key) = vote_key {
+            self.prev_votes
+                .get(validator_key)
+                .is_some_and(|vs| vs.contains(vote_key))
+        } else {
+            self.prev_votes.contains_key(validator_key)
+        }
```
wen-coding, line 125: Can we also add an unittest?

core/src/alpenglow_consensus.rs
```diff
@@ -90,6 +90,12 @@ pub enum VoteType {
     SkipFallback,
 }
 
+impl VoteType {
+    fn is_block_type(&self) -> bool {
```
carllin: nit: maybe is_notarize_type?

core/src/alpenglow_consensus/vote_pool.rs
ksn6: Nit: should opt for match statements over if let Some(...).

core/src/alpenglow_consensus/vote_pool.rs
qkniep: Taking only `should_ignore_vote` but not `check_slashable_offence` from there is not enough to prevent double counting. You could still be counting a skip and a notar from the same validator, which is totally not safe!

PR #254, opened 2025-07-09
Upgrade to `solana-signature` v2.3.0 and `solana-bls-signatures`

samkim-crypto:
#### Problem
The official `solana-bls-signatures` crate is now published in crates.io.

#### Summary of Changes
The `solana-bls-signatures` crate depends on `solana-signatures` v2.3.0, so I first upgraded `solana-signatures` to v2.3.0 from v2.2.1. The main agave repo already uses `v2.3.0`, so it is safe to make the upgrade.

Then I replaced `solana-bls` to `solana-bls-signatures`.

Fixes #


core/src/alpenglow_consensus/vote_certificate.rs
```diff
@@ -141,10 +143,9 @@ impl VoteCertificate for CertificateMessage {
             self.bitmap.set(vote_message.rank as usize, true);
             // aggregate the signature
             // TODO(wen): put this into bls crate
```
wen-coding, line 145: nit: we should remove the TODO above.

wen-coding (approved):

core/src/alpenglow_consensus/vote_certificate.rs
samkim-crypto: Sorry I missed this, I'll leave a note to delete this on a next bls PR.

wen-coding: No worries, I'll clean up in https://github.com/anza-xyz/alpenglow/pull/214. Thanks for cleaning things up!

PR #256, opened 2025-07-09
Fix BLSKeypair in genesis config.

wen-coding:
A lot of BLSKeypair generation was added when we couldn't correctly do BLSKeypair::derive_from_signer. Fix them now to unblock https://github.com/anza-xyz/alpenglow/pull/214

### DIFF
--- a/genesis/src/lib.rs
+++ b/genesis/src/lib.rs
@@ -29,4 +29,5 @@ pub struct StakedValidatorAccountInfo {
     pub identity_account: String,
     pub vote_account: String,
     pub stake_account: String,
+    pub bls_pubkey: String,
 }
--- a/genesis/src/main.rs
+++ b/genesis/src/main.rs
@@ -8,7 +8,7 @@ use {
     itertools::Itertools,
     solana_account::{Account, AccountSharedData, ReadableAccount, WritableAccount},
     solana_accounts_db::hardened_unpack::MAX_GENESIS_ARCHIVE_UNPACKED_SIZE,
-    solana_bls::keypair::Keypair as BLSKeypair,
+    solana_bls::Pubkey as BLSPubkey,
     solana_clap_utils::{
         input_parsers::{
             cluster_type_of, pubkey_of, pubkeys_of, unix_timestamp_from_rfc3339_datetime,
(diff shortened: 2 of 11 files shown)


genesis/src/lib.rs
```diff
@@ -29,4 +29,5 @@ pub struct StakedValidatorAccountInfo {
     pub identity_account: String,
     pub vote_account: String,
     pub stake_account: String,
+    pub bls_pubkey: String,
```
AshwinSekar: I believe we want this to be an Option<String> to not interrupt TowerBFT usage.

genesis/src/main.rs
```diff
@@ -772,6 +798,7 @@ fn main() -> Result<(), Box<dyn error::Error>> {
     add_validator_accounts(
         &mut genesis_config,
         &mut bootstrap_validator_pubkeys.iter(),
```
AshwinSekar, line 803: kind of strange how `add_validator_accounts` is setup. Think it would be nicer to parse out these Pubkeys into a struct rather than relying on sequential ordering in groups of 3.

definitely not a problem for this PR, just pointing it out.

keygen/src/keygen.rs
```diff
@@ -388,6 +390,40 @@ fn app<'a>(num_threads: &'a str, crate_version: &'a str) -> Command<'a> {
                         .help("Overwrite the output file if it exists"),
                 )
         )
+        .subcommand(
+            Command::new("bls_pubkey")
+                .about("Display the BLS pubkey from a keypair file")
```
AshwinSekar: nit: might want to specify that this is an ED25519 keyfile not a BLS keypair file and it's showing the derived pubkey

AshwinSekar (commented):
Looks good just some small comments

genesis/src/main.rs
wen-coding: Yeah, agree this set up is a bit weird.

genesis/src/lib.rs
wen-coding: Changed.

keygen/src/keygen.rs
wen-coding: Changed.

PR #258, opened 2025-07-10
Upgrade `solana-keypair` to 2.2.3 and `solana-signature` to 2.3.0

samkim-crypto:
#### Problem
We want to update `solana-bls` to the official `solana-bls-signatures` that is published in crates.io. But `solana-bls-signatures` that was recently published depends on the most updated versions of `solana-keypair` and `solana-signature`.

#### Summary of Changes
We will eventually need to update the versions of `solana-keypair` and `solana-signature`, so I think it makes sense to upgrade the versions of these crates in preparation to switching over to the `solana-bls-signatures` crates.

There are basically just two breaking changes in these crates:
- `Keypair::from_bytes()` is deprecated in favor of `Keypair::try_from()`
- `Keypair::generate()` is deprecated in favor of `Keypair::new_from_array()`

Fixes #


samkim-crypto:
Replaced by https://github.com/anza-xyz/alpenglow/pull/254.

PR #261, opened 2025-07-10
refactor: move core::alpenglow_consensus -> votor crate

AshwinSekar:
In preparation for #250 lets move to a separate crate. Additionally `solana_core` was ignoring all arithmetic clippy warnings. I've reenabled them in `solana_votor` and switched offenders to `saturating` for now, we can audit and potentially add error logic for these operations later.

The voting loop will be dealt with in the next PR and block creation loop/certificate service will remain in core

This PR should not change any existing behavior


core/src/alpenglow_consensus.rs
```diff
@@ -1,193 +1,2 @@
-use {
```
AshwinSekar, line 1: moved all of this to `votor/src/lib.rs`

PR #263, opened 2025-07-11
Support reading bls_pubkey from json file.

wen-coding:
Support reading bls_pubkey from json file so we can later write the BLS pubkeys in https://github.com/anza-xyz/infra-invalidator/pull/161

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6957,6 +6957,7 @@ dependencies = [
  "chrono",
  "clap 2.33.3",
  "rpassword",
+ "serde_json",
  "solana-bls-signatures",
  "solana-clock",
  "solana-cluster-type",
--- a/clap-utils/Cargo.toml
+++ b/clap-utils/Cargo.toml
@@ -13,6 +13,7 @@ edition = { workspace = true }
 chrono = { workspace = true, features = ["default"] }
 clap = "2.33.0"
 rpassword = { workspace = true }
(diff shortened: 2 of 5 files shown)


bw-solana:
Taking a step back - the goal we're trying to achieve is adding all the accounts we need into genesis so we can run a bunch of staked validators running alpenglow in the invalidator cluster. The problem today is we don't add the BLS pubkeys. You added a change such that we can provide the BLS pubkeys for all bootstrap nodes in genesis, but now we're figuring out the best way to actually get a handle to those pubkeys. Am I right so far?

Assuming that's correct, my understanding is there are 2 possible ways to do this:
1. Using `solana-keygen` to derive the BLS pubkey from the identity key inline in the invalidator pipeline. This is what you did with https://github.com/anza-xyz/infra-invalidator/pull/161
2. Storing the BLS pubkey in the invalidator repo. And this PR is helping us do that by pulling the BLS pubkey from some identity file?

Can you help me better understand what this PR is doing? Which JSON file are we extracting BLS pubkey from? Identity kepair? Vote keypair?

wen-coding:
After chatting with Brennan, we decided to keep https://github.com/anza-xyz/infra-invalidator/pull/161 as is, I'll close this PR for now.

bw-solana:
thanks for putting up with me! 😄

PR #264, opened 2025-07-12
Statically type vote pools 

carllin:
#### Problem
Vote pools use Option<BankId>, Option<BankHash>  to distinguish between notarize/non-notarize pools, making all the API's hard to digest and error-prone if people pass in the wrong options.

#### Summary of Changes
Two different types of pools, `DuplicateBlockVotePool` that can take multiple votes per validator and `SimpleVotePool` that can't. 

Next steps could make a typed pool for each of the 5 different types of pools in `vote_pools: BTreeMap<PoolId, VotePoolType>,` so that each vote type has an explicitly typed pool of either `DuplicateBlockVotePool` or `SimpleVotePool`

Fixes #

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -4,7 +4,9 @@ use {
         certificate_pool::{
             parent_ready_tracker::ParentReadyTracker,
             vote_certificate::{CertificateError, VoteCertificate},
-            vote_pool::{VoteKey, VotePool},
+            vote_pool::{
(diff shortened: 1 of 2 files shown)


AshwinSekar:
while you're here can we remove `bank_hash` from vote pool entirely? We plan to refactor votes to only be on `block_id` so we'll have to do this eventually.

carllin:
Yup can do, but seems like a job for a follow up PR in case anything breaks

PR #265, opened 2025-07-12
Send verified votes from BLSSigverifier to repair service.

wen-coding:


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -96,24 +165,92 @@ impl BLSSigVerifier {
...
+            epoch_schedule: EpochSchedule::default(),
+            epoch_stakes_map: Arc::new(HashMap::new()),
+            root_epoch: Epoch::default(),
+            epoch_stakes_queried: Instant::now() - EPOCH_STAKES_QUERY_INTERVAL,
+        };
+        verifier.update_epoch_stakes_map();
+        verifier
+    }
+
+    // TODO(wen): We should maybe create a epoch stakes service so all these objects
```
AshwinSekar, line 214: For a future pr what do you think about creating something in votor/src that keeps track of epoch stakes and the bls rank to pubkey map / any other context we need?
Things that require this context could subscribe to this service and be notified everytime the epoch rolls over (voting loop can set the new epoch stakes when it roots a bank in the next epoch).

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -96,24 +165,92 @@ impl BLSSigVerifier {
...
+    // only needing epoch stakes don't need to worry about bank_forks and banks.
+    fn update_epoch_stakes_map(&mut self) {
+        if self.epoch_stakes_queried.elapsed() < EPOCH_STAKES_QUERY_INTERVAL {
```
AshwinSekar, line 217: A bit hesitant about using wall clock time here because we might have shorter epochs during testing.

I don't have a great answer here, maybe using the `RootBankCache` and basing calls to `update_epoch_stakes_map` via the slot # on incoming votes.
A malicious voter could force use to make frequent calls to `update_epoch_stakes_map`, but we could refuse them without having to grab the `bank_forks` lock.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -59,6 +94,7 @@ impl SigVerifier for BLSSigVerifier {
     ) -> Result<(), SigVerifyServiceError<Self::SendType>> {
         // TODO(wen): just a placeholder without any batching.
         let mut stats = BLSSigVerifierStats::new();
+        let mut verified_votes: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
```
AshwinSekar, line 94: Can be in a follow up PR, but technically we don't need all of the slots, only the latest per validator is relevant for repair weight.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: That was the plan. I think it can be even higher than votor/src, in its own directory. So many things depend on stake, and currently they all need to grab EpochStakes from a bank. We only really need this new EpochStakesService to read and track root bank so it can update EpochStakes. All readers can blindly grab EpochStakes from this structure without caring about bank or bank_forks at all.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Yeah, maybe we should really start working on EpochStakesService now, that should definitely happen before we move to Agave master branch.

For a stop gap, can we just poke bank_forks every minute for now?

core/src/sigverifier/bls_sigverifier.rs
wen-coding: It can be done here, but if we start to track latest per validator, then maybe we could also consider Quentin's suggestion: https://github.com/anza-xyz/alpenglow/issues/30#issuecomment-3072735683

The certificate pool is the more convenient place to track latest state per validator maybe?

What do you think?

core/src/sigverifier/bls_sigverifier.rs
AshwinSekar: yeah i'm fine deferring until we have EpochStakesService, can we decrease the query interval? I think in some of the local cluster tests we have minimum epoch length which means we could rollover before the query happens here and votes get dropped?

core/src/sigverifier/bls_sigverifier.rs
AshwinSekar: for eager repair let's continue using repair weight. The block based repair that i'm implementing will use the certificate pool like quentin suggested.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: That's a good point, let me change to just poke bank_forks every minute and see if I break any tests. If so, I can provide a test_only API to adjust this interval.

wen-coding: Changed, turns out all tests still pass. I checked the flaky and slow running tests, those were either unrelated (not Alpenglow) or slow before this change.

wen-coding:
> LGTM lets file an issue to come back and fix all these bank forks locks once the EpochStakesService is ready

Sure, https://github.com/anza-xyz/alpenglow/issues/270 filed.

PR #266, opened 2025-07-12
InfluxDB doesn't understand u64 yet.

wen-coding:
#### Problem
[2025-07-12T18:01:30.524548697Z WARN  solana_metrics::metrics] submit response unsuccessful: 400 Bad Request {"error":"partial write: unable to parse 'bls_sig_verifier_stats,host_id=tiv1zkpDdumabxfZisVjuQgDzGcVSVDTEaHJD6ueVuK sent=1u,sent_failed=0u,received=1u,received_malformed=0u 1752343289602304527': invalid number dropped=0"}

#### Summary of Changes
Change back to i64 for now, change to u64 when it is fixed.

### DIFF
--- a/programs/sbf/Cargo.lock
+++ b/programs/sbf/Cargo.lock
@@ -9386,7 +9386,7 @@ dependencies = [
  "serde_derive",
  "solana-accounts-db",
  "solana-bloom",
- "solana-bls",
+ "solana-bls-signatures",
  "solana-entry",
  "solana-ledger",
  "solana-logger",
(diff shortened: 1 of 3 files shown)


programs/sbf/Cargo.lock
```diff
@@ -9386,7 +9386,7 @@ dependencies = [
  "serde_derive",
  "solana-accounts-db",
  "solana-bloom",
- "solana-bls",
+ "solana-bls-signatures",
```
alexpyattaev, line 9389: Is it normal that we are suddenly pulling a new crate here when change has nothing to do with it?

alexpyattaev (approved):
LGTM thank you!

programs/sbf/Cargo.lock
wen-coding: This is leftover from a previous PR, we should have generated this as well.

alexpyattaev: Ok just looks super sus to have cargo.lock changes for no apparent reason =)

PR #245, opened 2025-07-07
Remove has_new_vote_been_rooted and voted_signatures

0xPraedico:
#### Problem

With Alpenglow votes don't actually land as transactions on chain. These checks are obsolete.

#### Summary of Changes

- Removed has_new_vote_been_rooted checks
- Removed voted_signatures tracking

Fixes #237

PR #267, opened 2025-07-13
smuggle voting_loop and block_creation loop out of solana_core::alpenglow_consensus

AshwinSekar:
This PR does not change any existing behavior
voting_loop.rs -> votor/src
block_creation_loop.rs -> core/src

### DIFF
--- a/core/src/commitment_service.rs
+++ b/core/src/commitment_service.rs
@@ -1,6 +1,6 @@
 use {
     crate::consensus::{tower_vote_state::TowerVoteState, Stake},
-    crossbeam_channel::{bounded, unbounded, Receiver, RecvTimeoutError, Sender},
+    crossbeam_channel::{bounded, select, unbounded, Receiver, RecvTimeoutError, Sender},
     solana_measure::measure::Measure,
     solana_metrics::datapoint_info,
     solana_rpc::rpc_subscriptions::RpcSubscriptions,
@@ -122,32 +122,27 @@ impl AggregateCommitmentService {
             }
 
             let mut aggregate_commitment_time = Measure::start("aggregate-commitment-ms");
-            let tower_result = receiver.recv_timeout(Duration::from_millis(500));
-            let ag_result = ag_receiver.recv_timeout(Duration::from_millis(500));
-
-            let commitment_slots = match (tower_result, ag_result) {
-                (Ok(aggregation_data), _) => {
-                    let data = receiver.try_iter().last().unwrap_or(aggregation_data);
+            let commitment_slots = select! {
+                recv(receiver) -> msg => {
+                    let data = msg?;
+                    let data = receiver.try_iter().last().unwrap_or(data);
                     let ancestors = data.bank.status_cache_ancestors();
                     if ancestors.is_empty() {
                         continue;
                     }
                     Self::update_commitment_cache(block_commitment_cache, data, ancestors)
                 }
-                (_, Ok(aggregation_data)) => {
-                    let data = ag_receiver.try_iter().last().unwrap_or(aggregation_data);
+                recv(ag_receiver) -> msg => {
+                    let data = msg?;
+                    let data = ag_receiver.try_iter().last().unwrap_or(data);
                     Self::alpenglow_update_commitment_cache(
                         block_commitment_cache,
                         data.commitment_type,
                         data.slot,
                     )
                 }
-                (Err(RecvTimeoutError::Disconnected), Err(RecvTimeoutError::Disconnected)) => {
-                    return Err(RecvTimeoutError::Disconnected)
-                }
-                _ => continue,
+                default(Duration::from_secs(1)) => continue
             };
-
             aggregate_commitment_time.stop();
 
             datapoint_info!(
--- a/core/src/commitment_service.rs
+++ b/core/src/commitment_service.rs
@@ -1,6 +1,6 @@
 use {
     crate::consensus::{tower_vote_state::TowerVoteState, Stake},
-    crossbeam_channel::{unbounded, Receiver, RecvTimeoutError, Sender},
+    crossbeam_channel::{bounded, unbounded, Receiver, RecvTimeoutError, Sender},
     solana_measure::measure::Measure,
     solana_metrics::datapoint_info,
     solana_rpc::rpc_subscriptions::RpcSubscriptions,
@@ -77,10 +77,11 @@ impl AggregateCommitmentService {
             Sender<TowerCommitmentAggregationData>,
             Receiver<TowerCommitmentAggregationData>,
         ) = unbounded();
+        // This channel should not grow unbounded, cap at 1000 messages for now
         let (ag_sender, ag_receiver): (
             Sender<AlpenglowCommitmentAggregationData>,
             Receiver<AlpenglowCommitmentAggregationData>,
-        ) = unbounded();
+        ) = bounded(1000);
 
         (
             sender,
@@ -100,7 +101,6 @@ impl AggregateCommitmentService {
                             &subscriptions,
                             &exit,
                         ) {
-                            info!("#ASH: disconnecting");
                             break;
                         }
                     })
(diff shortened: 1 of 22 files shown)


votor/src/root_utils.rs
```diff
@@ -0,0 +1,216 @@
...
+            .sender
+            .send(BankNotification::OptimisticallyConfirmed(new_root))
+            .unwrap();
+    }
+
+    Some(new_root)
+}
+
+/// Sets the new root, this should be kept one-to-one with
+/// ReplayStage::check_and_handle_new_root
```
AshwinSekar: kind of shitty but I'll remember to keep this 1:1 with replay stage until AG is active and we can delete the replay stage version

core/src/commitment_service.rs
```diff
@@ -120,29 +121,33 @@ impl AggregateCommitmentService {
...
-                        Self::update_commitment_cache(block_commitment_cache, data, ancestors)
-                    }
-                    CommitmentAggregationData::AlpenglowCommitmentAggregationData(data) => {
-                        Self::alpenglow_update_commitment_cache(
-                            block_commitment_cache,
-                            data.commitment_type,
-                            data.slot,
-                        )
+            let tower_result = receiver.recv_timeout(Duration::from_millis(500));
+            let ag_result = ag_receiver.recv_timeout(Duration::from_millis(500));
```
bw-solana: this looks scary. Reduced the timeout in half but will wait on receiving on both. I'm assuming only one of these will typically have messages coming across

votor/src/commitment.rs
```diff
@@ -0,0 +1,26 @@
...
+pub fn alpenglow_update_commitment_cache(
+    commitment_type: AlpenglowCommitmentType,
+    slot: Slot,
+    commitment_sender: &Sender<AlpenglowCommitmentAggregationData>,
+) {
+    if let Err(e) = commitment_sender.send(AlpenglowCommitmentAggregationData {
+        commitment_type,
+        slot,
+    }) {
+        trace!("commitment_sender failed: {:?}", e);
```
bw-solana: should this be higher log level? Failing to send here seems bad

core/src/replay_stage.rs
```diff
@@ -4402,63 +4380,43 @@ impl ReplayStage {
     pub fn handle_new_root(
```
bw-solana: we need a tree of all the `<module>_set/handle_root` calls 🤯

core/src/replay_stage.rs
```diff
@@ -4402,63 +4380,43 @@ impl ReplayStage {
...
         new_root: Slot,
         bank_forks: &RwLock<BankForks>,
```
bw-solana: It's so annoying that the RwLock is on the outside 😭

bw-solana (commented):
Left a few comments. Overall re-structuring seems like an improvement. Looks like local cluster testing is failing though and seems related to changes made

core/src/commitment_service.rs
carllin: maybe a select across channels here

```
select! {
            recv(receiver) -> msg => {
                match msg {
                    Ok(data) => {
                        // process TowerCommitmentAggregationData
                    }
                    Err(_) => break,
                }
            }
            recv(ag_receiver) -> msg => {
                match msg {
                    Ok(data) => {
                        // process AlpenglowCommitmentAggregationData
                    }
                    Err(_) => break,
                }
            }
            default(Duration::from_millis(500)) => {
```

votor/src/root_utils.rs
carllin: Isi possible to pass in a closure that executes some custom code after `set_bank_forks_root()` so that replay can pass in the custom handling of the `progress_map` and `tbft_structs`

Otherwise seems like a pain to maintain if somebody touches the replay version

core/src/voting_service.rs
```diff
@@ -122,17 +113,47 @@ impl VotingService {
...
-                        vote_history_storage.as_ref(),
-                        vote_op,
-                        connection_cache.clone(),
-                        additional_listeners.as_ref(),
-                        &mut staked_validators_cache,
-                    );
+                let mut vote_receiver_disconnected = false;
+                let mut bls_receiver_disconnected = false;
+                loop {
+                    if vote_receiver_disconnected && bls_receiver_disconnected {
```
carllin: might be cleaner to have this as a `while vote_receiver_open || bls_receiver_open {` loop

carllin: do we only want to disconnect if both fail? seems like we should exit if either replay or voting thread dies

core/src/commitment_service.rs
AshwinSekar: it's tricky because if one of the channels disconnects we'll end up in a case where we don't receive from the other channel (since the disconnected channel is always ready for `recv`). I'll look into reworking this.

core/src/voting_service.rs
AshwinSekar: it's setup like this because the migration code isn't ready, so we only instantiate voting loop if ff is active on genesis. This means the channel is immediately disconnected for tests that don't use the feature flag.

I agree this is not the correct approach, i'll rework to what you have suggested instead.

votor/src/root_utils.rs
AshwinSekar: yeah so much smarter, unified it here 10214b5f59c4ed88991fca5dd7274059d37b6ece

core/src/commitment_service.rs
AshwinSekar: fixed here 3f369131a74529e5fc5261d9249ed6d1fcf56e87
The alpenglow sender is hacked to not be dropped when the feature flag is off, when we deal with a migration I'll clean it up properly.

votor/src/commitment.rs
AshwinSekar: good call, I bounded the channel and added some logging here 679f84bcc65478e7b57dca22c0bc3f351368096d
If the channel fails to send for disconnection, it could just be normal shutdown. In the next PR I'll use this value to set the exit flag and quick exit the voting loop as well.

core/src/replay_stage.rs
AshwinSekar: I added a little diagram for the replay stuff now, but agree some of the inner structs have confusing patterns
https://github.com/anza-xyz/alpenglow/blob/5617f2b52c0edc798be3dffe81cec57ea87ea04e/votor/src/root_utils.rs#L102

core/src/voting_service.rs
AshwinSekar: done cbad0632b2c63daacba457ef0d6256259ae01d87

votor/src/root_utils.rs
```diff
@@ -0,0 +1,229 @@
...
+        highest_super_majority_root,
+    )?;
+
+    drop_bank_sender
+        .send(removed_banks)
+        .unwrap_or_else(|err| warn!("bank drop failed: {:?}", err));
+
+    // Dropping the bank_forks write lock and reacquiring as a read lock is
+    // safe because updates to bank_forks are only made by a single thread.
+    // TODO(ashwin): Once PR #254 lands move this back to ReplayStage
```
bw-solana: it landed :)

votor/src/voting_loop.rs
```diff
@@ -37,31 +43,67 @@ use {
...
+const MAX_VOTE_SIGNATURES: usize = 200;
+
+// Implement a destructor for Finalizer.
+impl Drop for Finalizer {
+    fn drop(&mut self) {
+        self.exit_sender.clone().store(true, Ordering::Relaxed);
+    }
+}
+
+pub enum BLSOp {
```
bw-solana, line 100: feels like this will need its own home eventually... like `votor::bls`, but it would be very lonely right now, so I'm okay leaving it here. If/when we start building this out more or have some helper functions, we can revisit

bw-solana (approved):
LGTM

votor/src/voting_loop.rs
AshwinSekar: hold on for the next PR, now that i've smuggled the voting loop I can rip apart the giant file haha :) 
Screenshot 2025-07-16 at 10 41 21 AM

votor/src/root_utils.rs
AshwinSekar: woops i'm dyslexic https://github.com/anza-xyz/alpenglow/pull/245

Issue #122, opened 2025-03-30
Add metrics to block creation loop


AshwinSekar:
For inspiration here's a couple of stages in Agave that are well instrumented:
https://github.com/anza-xyz/agave/blob/5c9878b6bba2d781a15e2ff068ce38885f3ff343/core/src/replay_stage.rs#L311
https://github.com/anza-xyz/agave/blob/5c9878b6bba2d781a15e2ff068ce38885f3ff343/core/src/banking_stage.rs#L97

fahimahmedx:
Would love to take this on!

AshwinSekar:
Lets hold off on the voting loop portion for now, i'm planning to refactor some of the code this week.

As for the block creation loop:
> time spent waiting for notification before block starts to get created, starting from [line 211](https://github.com/anza-xyz/alpenglow/blob/34ee0461f27f5abba3499b9c145a907dbf3f1b8e/core/src/alpenglow_consensus/block_creation_loop.rs#L211) to line 227.

This is the only one I would omit, shouldn't be useful 

> count on number of VoteNotRooted error

This will be removed in #245 

The biggest points of interest as you pointed out is how many times `ReplayIsBehind` and noting when we decide to skip production of the leader window because of this. Measuring how much time `start_leader_retry_replay` ends up waiting for replay would also be usefull

wen-coding:
Marking this done, Thanks @fahimahmedx

PR #269, opened 2025-07-15
block_creation_loop: add metrics

fahimahmedx:
### Problem
Solves #122 for the block creation loop by adding in metrics to track the metrics in [this comment](https://github.com/anza-xyz/alpenglow/issues/122#issuecomment-3047284188).

### Summary of Changes
In this PR, I add two types of metrics for the block creation loop. `BlockCreationLoopMetrics` which provides high-level performance monitoring for the entire block creation loop, and `SlotMetrics` which provides granular monitoring for individual slot production attempts inside of the block creation loop.

---

#### BlockCreationLoopMetrics

These metrics provides high-level performance monitoring for the entire block creation loop.

| Metric | Type | Description |
|--------|---------|----------|
| `loop_count` | Counter | Total number of leader windows processed |
| `window_production_elapsed` | Duration | Total time spent producing all the slots for the given window |
| `bank_filled_completion_elapsed_hist` | Histogram | Distribution of completion times for filled banks |
| `bank_timeout_completion_elapsed_hist` | Histogram | Distribution of completion times for timed-out banks |
| `bank_filled_completion_count` | Counter | Banks completed due to being filled |
| `bank_timeout_completion_count` | Counter | Banks completed due to timeout |

#### SlotMetrics

These metrics provides granular monitoring for individual slot production attempts inside of the block creation loop.

| Metric | Type | Description |
|--------|---------|----------|
| `slot` | Identifier | Current slot being monitored |
| `attempt_count` | Counter | Number of attempts to start this slot (`AlreadyHaveBank` error) |
| `replay_is_behind_count` | Counter | Times replay was behind parent slot (`ReplayIsBehind` error) |
| `startup_verification_incomplete_count` | Counter | Times startup verification was incomplete (`StartupVerificationIncomplete` error) |
| `already_have_bank_count` | Counter | Times bank already existed for slot |
| `slot_delay_hist` | Histogram | Distribution of the times elapsed between attempting to start producing a slot -> successfully starting production of the slot  |
| `replay_is_behind_cumulative_wait_elapsed` | Duration | Total time waiting for replay to catch up |
| `replay_is_behind_wait_elapsed_hist` | Histogram | Distribution of individual replay wait times |

#### todo list
Add metrics for the following:
- [x] count on the number of times a replay is behind (`ReplayIsBehind` error). `replay_is_behind_count`
- [x] count on number of `StartupVerificationIncomplete` error. `startup_verification_incomplete_count`
- [x] time it takes to create and insert a bank as `bank_completion_elapsed`
- [x] time it takes to create a window, starting from after the voting loop gets notified & node is prepared [here](https://github.com/anza-xyz/alpenglow/blob/331c3bd905ac4f9180f4a38868cacc848f55f25b/core/src/block_creation_loop.rs#L230-L287). `window_production_elapsed`
- [x] of the time it takes to create a window of slots, how long does it take to create each slot in this loop [here](https://github.com/anza-xyz/alpenglow/blob/331c3bd905ac4f9180f4a38868cacc848f55f25b/core/src/block_creation_loop.rs#L232-L287). `slot_production_elapsed`
- [x]  number of times we hit bank completion via timeout. `bank_timeout_completion_count`
- [x] number of times we hit bank completion via bank getting filled. `bank_filled_completion_count`
- [x] how much time `start_leader_retry_replay()` ends up waiting for replay (specifically, how long it took to wait on `wait_timeout_while`) as `replay_is_behind_wait_elapsed`

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -358,7 +358,7 @@ fn start_leader_retry_replay(
                 return Ok(());
             }
             Err(StartLeaderError::ReplayIsBehind(_)) => {
-                metrics.replay_is_behind_count += 1;
+                metrics.replay_is_behind_count.fetch_add(1, Ordering::Relaxed);
 
                 trace!(
                     "{my_pubkey}: Attempting to produce slot {slot}, however replay of the \
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -79,6 +79,8 @@ struct BlockCreationLoopMetrics {
     loop_count: AtomicUsize,
     replay_is_behind_count: AtomicUsize,
     delay_after_replay_is_behind_elapsed: AtomicU64,
+    record_receiver_timeout_count: AtomicUsize,
+    record_receiver_disconnected_count: AtomicUsize,
 }
 
 impl BlockCreationLoopMetrics {
@@ -91,11 +93,7 @@ impl BlockCreationLoopMetrics {
         if self.last_report.should_update(report_interval_ms) {
             datapoint_info!(
                 "block-creation-loop-metrics",
-                (
-                    "loop_count",
-                    self.loop_count.load(Ordering::Relaxed),
-                    i64
-                ),
+                ("loop_count", self.loop_count.load(Ordering::Relaxed), i64),
                 (
                     "replay_is_behind_count",
                     self.replay_is_behind_count.swap(0, Ordering::Relaxed),
@@ -104,6 +102,18 @@ impl BlockCreationLoopMetrics {
                 (
                     "delay_after_replay_is_behind_elapsed",
                     self.delay_after_replay_is_behind_elapsed
+                        .swap(0, Ordering::Relaxed),
+                    i64
+                ),
+                (
+                    "record_receiver_timeout_count",
+                    self.record_receiver_timeout_count
+                        .swap(0, Ordering::Relaxed),
+                    i64
+                ),
+                (
+                    "record_receiver_disconnected_count",
+                    self.record_receiver_disconnected_count
                         .swap(0, Ordering::Relaxed),
                     i64
                 ),
@@ -135,6 +145,7 @@ fn start_receive_and_record_loop(
     exit: Arc<AtomicBool>,
     poh_recorder: Arc<RwLock<PohRecorder>>,
     record_receiver: Receiver<Record>,
+    metrics: &mut BlockCreationLoopMetrics,
 ) {
     while !exit.load(Ordering::Relaxed) {
         // We need a timeout here to check the exit flag, chose 400ms
@@ -155,9 +166,16 @@ fn start_receive_and_record_loop(
             }
             Err(RecvTimeoutError::Disconnected) => {
                 info!("Record receiver disconnected");
+                metrics
+                    .record_receiver_disconnected_count
+                    .fetch_add(1, Ordering::Relaxed);
                 return;
             }
-            Err(RecvTimeoutError::Timeout) => (),
+            Err(RecvTimeoutError::Timeout) => {
+                metrics
+                    .record_receiver_timeout_count
+                    .fetch_add(1, Ordering::Relaxed);
+            }
         }
     }
 }
@@ -211,7 +229,7 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
     let exit_c = exit.clone();
     let p_rec = poh_recorder.clone();
     let receive_record_loop = thread::spawn(move || {
-        start_receive_and_record_loop(exit_c, p_rec, record_receiver);
+        start_receive_and_record_loop(exit_c, p_rec, record_receiver, &mut metrics);
     });
 
     while !exit.load(Ordering::Relaxed) {
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -436,7 +436,6 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
                 trace!("{my_pubkey}: finished leader window {start_slot}-{end_slot}");
                 break;
             }
-            let mut slot_production_start = Measure::start("slot_production");
 
             // Although `slot - 1`has been cleared from `poh_recorder`, it might not have finished processing in
             // `replay_stage`, which is why we use `start_leader_retry_replay`
@@ -446,12 +445,6 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
                 error!("{my_pubkey}: Unable to produce {slot}, skipping rest of leader window {slot} - {end_slot}: {e:?}");
                 break;
             }
-
-            // Record individual slot production time
-            slot_production_start.stop();
-            let _ = metrics
-                .slot_production_elapsed_hist
-                .increment(slot_production_start.as_us());
         }
         window_production_start.stop();
         metrics
@@ -492,9 +485,16 @@ fn start_leader_retry_replay(
 ) -> Result<(), StartLeaderError> {
     let my_pubkey = ctx.my_pubkey;
     let timeout = block_timeout(leader_slot_index(slot));
+    let mut slot_production_start = Measure::start("slot_production");
     while !timeout.saturating_sub(skip_timer.elapsed()).is_zero() {
         match maybe_start_leader(slot, parent_slot, ctx, metrics) {
             Ok(()) => {
+                // Record successful slot production time
+                slot_production_start.stop();
+                let _ = metrics
+                    .slot_production_elapsed_hist
+                    .increment(slot_production_start.as_us());
+
                 return Ok(());
             }
             Err(StartLeaderError::ReplayIsBehind(_)) => {
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -202,6 +202,11 @@ impl BlockCreationLoopMetrics {
                     i64
                 ),
             );
+
+            // .swap() resetted most of the metrics to 0, but not the histograms or non-atomic values. reset them.
+            self.slot_production_elapsed_hist.clear();
+            self.bank_completion_elapsed_hist.clear();
+            self.last_report = now;
         }
     }
 }
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -76,7 +76,7 @@ pub struct ReplayHighestFrozen {
 
 #[derive(Default)]
 struct BlockCreationLoopMetrics {
-    last_report: AtomicInterval,
+    last_report: u64,
     loop_count: u64,
     replay_is_behind_count: AtomicUsize,
     startup_verification_incomplete_count: AtomicUsize,
@@ -112,7 +112,10 @@ impl BlockCreationLoopMetrics {
             return;
         }
 
-        if self.last_report.should_update(report_interval_ms) {
+        let now = timestamp();
+        let elapsed_ms = now - self.last_report;
+
+        if elapsed_ms > report_interval_ms {
             datapoint_info!(
                 "block-creation-loop-metrics",
                 ("loop_count", self.loop_count, i64),
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -77,7 +77,7 @@ pub struct ReplayHighestFrozen {
 #[derive(Default)]
 struct BlockCreationLoopMetrics {
     last_report: AtomicInterval,
-    loop_count: AtomicUsize,
+    loop_count: u64,
     replay_is_behind_count: AtomicUsize,
     startup_verification_incomplete_count: AtomicUsize,
     already_have_bank_count: AtomicUsize,
@@ -92,7 +92,7 @@ struct BlockCreationLoopMetrics {
 
 impl BlockCreationLoopMetrics {
     fn is_empty(&self) -> bool {
-        0 == self.loop_count.load(Ordering::Relaxed) as u64
+        0 == self.loop_count
             + self.replay_is_behind_count.load(Ordering::Relaxed) as u64
             + self
                 .startup_verification_incomplete_count
@@ -115,7 +115,7 @@ impl BlockCreationLoopMetrics {
         if self.last_report.should_update(report_interval_ms) {
             datapoint_info!(
                 "block-creation-loop-metrics",
-                ("loop_count", self.loop_count.load(Ordering::Relaxed), i64),
+                ("loop_count", self.loop_count, i64),
                 (
                     "replay_is_behind_count",
                     self.replay_is_behind_count.swap(0, Ordering::Relaxed),
@@ -449,7 +449,7 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
         metrics
             .window_production_elapsed
             .fetch_add(window_production_start.as_us(), Ordering::Relaxed);
-        metrics.loop_count.fetch_add(1, Ordering::Relaxed);
+        metrics.loop_count += 1;
         metrics.report(1000);
     }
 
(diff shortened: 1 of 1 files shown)


core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -89,6 +90,37 @@ pub struct LeaderWindowNotifier {
     pub window_notification: Condvar,
 }
 
+#[derive(Default)]
+struct BlockCreationLoopTiming {
```
carllin: nit: no real timing being measured here, so maybe `BlockCreationLoopStats` or `BlockCreationLoopMetrics`

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -341,6 +377,7 @@ fn start_leader_retry_replay(
     parent_slot: Slot,
     skip_timer: Instant,
     ctx: &LeaderContext,
+    timing: &mut BlockCreationLoopTiming,
```
carllin: nit: rename `timing`

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -89,6 +90,37 @@ pub struct LeaderWindowNotifier {
...
+        self.maybe_submit();
+    }
+
+    fn maybe_submit(&mut self) {
+        let now = timestamp();
+        let elapsed_ms = now - self.last_submit;
+
+        if elapsed_ms > 1000 {
+            datapoint_info!(
+                "block-creation-loop-timing-stats",
```
carllin: nit: rename `timing`

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -89,6 +90,37 @@ pub struct LeaderWindowNotifier {
     pub window_notification: Condvar,
 }
 
+#[derive(Default)]
+struct BlockCreationLoopTiming {
+    last_submit: u64,
```
carllin: https://github.com/anza-xyz/agave/blob/5c9878b6bba2d781a15e2ff068ce38885f3ff343/core/src/banking_stage.rs#L98 I think `AtomicInterval` is the more canonical way to do this now

core/src/alpenglow_consensus/block_creation_loop.rs
```diff
@@ -350,6 +387,8 @@ fn start_leader_retry_replay(
                 return Ok(());
             }
             Err(StartLeaderError::ReplayIsBehind(_)) => {
+                timing.replay_is_behind_count += 1;
```
carllin: maybe a metric for how long it took to wait on `wait_timeout_while` as well

core/src/alpenglow_consensus/block_creation_loop.rs
fahimahmedx: Thanks for spotting this, will implement! Few questions, as I'd love to better learn & understand this:

1. Why are `AtomicUsize` values like `dropped_forward_packets_count` typecased into a `i64` instead of `u64` before being logged in the code below, given that `AtomicUsize` implements `Usize` under the hood which is unsigned?

https://github.com/anza-xyz/agave/blob/5c9878b6bba2d781a15e2ff068ce38885f3ff343/core/src/banking_stage.rs#L240-L245

2. In the same code above, why `AtomicUsize` instead of `AtomicU64`?

core/src/alpenglow_consensus/block_creation_loop.rs
fahimahmedx: Thanks for the suggestion, added as `consume_buffered_packets_elapsed`

fahimahmedx:
Metrics for the count of `RecvTimeoutError::Disconnected` error, and count of `RecvTimeoutError::Timeout` error were unadded from my PR, as it's on a separate thread (concurrency issues) and on its own loop that isn't as related to block creation.

If these two metrics are something you'd find useful, I think it'd be more appropriate to design it in a separate metrics struct specific for the `start_receive_and_record_loop()` function which can be done in a future PR. Would love your thoughts on this!

core/src/alpenglow_consensus/block_creation_loop.rs
AshwinSekar: The metrics server expects signed types, so we need to convert to `i64`

fahimahmedx:
Fixed the merge conflict! 

I believe a previous check failed. Are you able to rerun the CI jobs & share what the check failures were? @AshwinSekar

fahimahmedx:
Fixed the lints, ready for a CI @AshwinSekar! and correct, I'm unable to see the buildkite failures (below is what I see).

image

core/src/block_creation_loop.rs
```diff
@@ -73,6 +75,135 @@ pub struct ReplayHighestFrozen {
     pub freeze_notification: Condvar,
 }
 
+#[derive(Default)]
+struct BlockCreationLoopMetrics {
+    last_report: AtomicInterval,
+    loop_count: AtomicUsize,
```
AshwinSekar: Do these need to be atomics? These are only incremented by one thread right?

core/src/block_creation_loop.rs
```diff
@@ -73,6 +75,135 @@ pub struct ReplayHighestFrozen {
...
+                    "bank_completion_elapsed_min",
+                    self.bank_completion_elapsed_hist.minimum().unwrap_or(0),
+                    i64
+                ),
+                (
+                    "bank_completion_elapsed_max",
+                    self.bank_completion_elapsed_hist.maximum().unwrap_or(0),
+                    i64
+                ),
+            );
```
AshwinSekar, line 182: After we report we should also reset all the metrics, otherwise it'll become difficult to compare nodes that start up at different times.

AshwinSekar (commented):
Thanks for doing this! Just a couple more comments to get this over the line

core/src/block_creation_loop.rs
```diff
@@ -310,15 +483,20 @@ fn start_leader_retry_replay(
...
-        match maybe_start_leader(slot, parent_slot, ctx) {
+        match maybe_start_leader(slot, parent_slot, ctx, metrics) {
             Ok(()) => {
                 return Ok(());
             }
             Err(StartLeaderError::ReplayIsBehind(_)) => {
+                metrics
+                    .replay_is_behind_count
+                    .fetch_add(1, Ordering::Relaxed);
+
```
carllin: @AshwinSekar could this potentially be triggered a lot for the same  potential leader slot `Y` if you have a situation where:

1. The parent slot is `X`, bank is not frozen
2. There is `highest_frozen_slot > X` so we don't block on the wait condition here: https://github.com/anza-xyz/alpenglow/blob/ccf1550c5ce9fe399288610f5cdea33279ad4c05/core/src/block_creation_loop.rs#L519

core/src/block_creation_loop.rs
```diff
@@ -73,6 +75,135 @@ pub struct ReplayHighestFrozen {
...
 
+#[derive(Default)]
+struct BlockCreationLoopMetrics {
+    last_report: AtomicInterval,
+    loop_count: AtomicUsize,
+    replay_is_behind_count: AtomicUsize,
+    startup_verification_incomplete_count: AtomicUsize,
+    already_have_bank_count: AtomicUsize,
+    bank_timeout_completion_count: AtomicUsize,
+    bank_filled_completion_count: AtomicUsize,
```
carllin: After doing a lot of debugging over the years, I think it might be better to break up some of these giant metrics

For instance this set is all associated with `maybe_start_leader()`, and not necessarily loop timing, so it might be better to break them out into a separate `StartLeaderFailures` struct that can be reported at the same time as everything else, but is organizationally clearer

core/src/block_creation_loop.rs
```diff
@@ -73,6 +75,135 @@ pub struct ReplayHighestFrozen {
...
+
+    replay_is_behind_wait_elapsed: AtomicU64,
+    window_production_elapsed: AtomicU64,
+    slot_production_elapsed_hist: histogram::Histogram,
```
carllin: nit: `slot_delay_hist` to distinguish from the actual duration of the slot

core/src/block_creation_loop.rs
```diff
@@ -262,25 +408,52 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
...
+            {
                 error!("{my_pubkey}: Unable to produce {slot}, skipping rest of leader window {slot} - {end_slot}: {e:?}");
                 break;
             }
+
+            // Record individual slot production time
+            slot_production_start.stop();
+            let _ = metrics
+                .slot_production_elapsed_hist
+                .increment(slot_production_start.as_us());
```
carllin: There's also an earlier call to `start_leader_retry_replay` https://github.com/anza-xyz/alpenglow/blob/ccf1550c5ce9fe399288610f5cdea33279ad4c05/core/src/block_creation_loop.rs#L349-L358 that probably also needs these metrics, should we just move this metric inside of `start_leader_retry_replay`

core/src/block_creation_loop.rs
fahimahmedx: Good point, let me take a look to confirm.

Out of curiosity, what are the downsides to accidentally using Atomic variables in areas where there isn’t multiple threads?

For example, are atomic variables not as performant due to it needing an extra step of acquiring a lock?

core/src/block_creation_loop.rs
fahimahmedx: Acknowledged! Will fix

core/src/block_creation_loop.rs
AshwinSekar: Shouldn't be a noticeable performance hit, just adds confusion for the reader as they would expect the metrics to be shared by multiple threads.

fahimahmedx: Thanks for the feedback, implemented!

core/src/block_creation_loop.rs
fahimahmedx: Implemented for the remaining ones that weren't being reset already.

core/src/block_creation_loop.rs
fahimahmedx: implemented!

core/src/block_creation_loop.rs
fahimahmedx: Thanks for the feedback, implemented!

core/src/block_creation_loop.rs
```diff
@@ -339,11 +523,18 @@ fn start_leader_retry_replay(
                         |hfs| *hfs < parent_slot,
                     )
                     .unwrap();
+                wait_start.stop();
+                metrics
+                    .replay_is_behind_wait_elapsed
+                    .fetch_add(wait_start.as_us(), Ordering::Relaxed);
             }
             Err(e) => return Err(e),
```
fahimahmedx, line 531: @carllin Currently, I only record the slot production delay measurement `slot_delay_start` when the slot production was successful, as my thinking is that recording the failed slot productions (when we return the `Err`) may skew the `slot_delay_hist` histogram metric.

Is this something you think makes sense, or should I also be recording the measurement to `slot_delay_hist` when it fails?

core/src/block_creation_loop.rs
fahimahmedx: @carllin Currently, I only record the slot production delay measurement `slot_delay_start` when the slot production was successful [here](https://github.com/anza-xyz/alpenglow/blob/672266d3c119c796deded929b36451dff20663ac/core/src/block_creation_loop.rs#L495), as my thinking is that recording the failed slot productions (when we return the `Err`) may skew the `slot_delay_hist` histogram metric.

Is this something you think makes sense, or should I also be recording the measurement to `slot_delay_hist` when it fails [here](https://github.com/anza-xyz/alpenglow/blob/672266d3c119c796deded929b36451dff20663ac/core/src/block_creation_loop.rs#L531)?

carllin: I think that makes sense because this wait behind metric: https://github.com/anza-xyz/alpenglow/blob/672266d3c119c796deded929b36451dff20663ac/core/src/block_creation_loop.rs#L527-L529 should capture all the waits that eventually lead to a skip timeout https://github.com/anza-xyz/alpenglow/blob/672266d3c119c796deded929b36451dff20663ac/core/src/block_creation_loop.rs#L490 or an Ok(())

Just realized we have a loop count but that doesn't give us a count for number of attempts to start a leader block https://github.com/anza-xyz/alpenglow/blob/672266d3c119c796deded929b36451dff20663ac/core/src/block_creation_loop.rs#L480C4-L531. This will make it hard to figure out which slots were delayed.

For debugging, It might even be more useful to report for every slot we are the leader for, a separate set of metrics, and put the delay metrics in there. This is similar to the per slot replay metrics we track https://github.com/anza-xyz/alpenglow/blob/07866dc758021298ecf507bafff93e4fca121bda/ledger/src/blockstore_processor.rs#L1481-L1483

fahimahmedx: Implemented in "seperate slot metrics into its own struct" commit by splitting slot related metrics into its own `SlotMetrics` struct, and added metrics on number of attempts to start a leader block. 

Is this what you're looking for?

core/src/block_creation_loop.rs
fahimahmedx: Implemented via `SlotMetrics`

core/src/block_creation_loop.rs
carllin: yup looks good!

core/src/block_creation_loop.rs
fahimahmedx: Or would you like it to be specifically in its own `StartLeaderFailures` struct? @carllin

fahimahmedx:
@AshwinSekar @carllin Would love to follow up on if there's anything else needed for this PR to be merged!

core/src/block_creation_loop.rs
```diff
@@ -262,11 +451,21 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
...
 
+            // Assert that the bank has been cleared
             assert!(!poh_recorder.read().unwrap().has_bank());
 
+            // Record bank completion time
+            bank_completion_measure.stop();
+            let _ = metrics
+                .bank_completion_elapsed_hist
+                .increment(bank_completion_measure.as_us());
+
```
carllin: do we want to increment this even if we timed out here: https://github.com/anza-xyz/alpenglow/blob/0f5164aa6f8a8e8a36fce766d30aa921e65893b3/core/src/block_creation_loop.rs#L456, we could keep a separate metric for that timeout instead

carllin (dismissed):
just a nit

core/src/block_creation_loop.rs
carllin: I think this is fine as is!

core/src/block_creation_loop.rs
fahimahmedx: good catch, would be useful to be able to see how long each type of completion takes! 

Fixed by seperating the `bank_completion_elapsed_hist` into `bank_timeout_completion_count` and `bank_filled_completion_count`.

For next steps, are we waiting on @AshwinSekar's review?

carllin: All good, I think we can just resolve conflicts and merge!

fahimahmedx:
Rebased to fix the merge conflicts caused by #343

Note that the following command clippy (which I believe is used in the buildkite workflow) no longer works on my side after the `master` branch was upstreamed. Wonder if the buildkite workflow will still pass?

> cargo +nightly-2024-11-22 clippy --workspace --all-targets --features dummy-for-ci-check,frozen-abi -- --deny=warnings --deny=clippy::default_trait_access --deny=clippy::arithmetic_side_effects --deny=clippy::manual_let_else --deny=clippy::used_underscore_binding

Note that I don't have repo permissions to merge my PR after it's been approved. If this is approved, could you also merge it? @carllin

PR #273, opened 2025-07-16
certificate_pool: use prev vote instead of VoteHistory in safe_to_*

AshwinSekar:
#### Problem
As part of #250 certificate pool ingest will be in a separate thread that doesn't have access to VoteHistory

#### Summary of Changes
Instead of using VoteHistory, use our previous votes from the pool itself. Future PR will initialize cert pool from vote history on startup.

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -6,9 +6,7 @@ use {
             vote_certificate::{CertificateError, VoteCertificate},
             vote_pool::{VoteKey, VotePool},
         },
-        conflicting_types,
-        vote_history::VoteHistory,
-        vote_to_certificate_ids, CertificateId, Stake, VoteType,
+        conflicting_types, vote_to_certificate_ids, CertificateId, Stake, VoteType,
         MAX_ENTRIES_PER_PUBKEY_FOR_NOTARIZE_LITE, MAX_ENTRIES_PER_PUBKEY_FOR_OTHER_TYPES,
         MAX_SLOT_AGE, SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP,
         SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP, SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY,
@@ -487,11 +485,7 @@ impl CertificatePool {
     /// (ii) At least 20% of stake voted to notarize `b` and at least 60% of stake voted to either notarize `b` or skip `slot`
     /// and we have not already cast a notarize fallback for this `b`
     /// If all the above hold, return `Some(block_id, bank_hash)` for the `b`
-    pub fn safe_to_notar(&self, slot: Slot, vote_history: &VoteHistory) -> Vec<(Hash, Hash)> {
-        if vote_history.its_over(slot) {
-            return vec![];
-        }
-
+    pub fn safe_to_notar(&self, my_vote_pubkey: &Pubkey, slot: Slot) -> Vec<(Hash, Hash)> {
         let Some(epoch_stakes) = self
             .epoch_stakes_map
             .get(&self.epoch_schedule.get_epoch(slot))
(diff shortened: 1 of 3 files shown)


votor/src/certificate_pool/vote_pool.rs
```diff
@@ -120,6 +120,14 @@ impl VotePool {
         Ok(())
     }
 
+    // Get the previous notarization vote, only used for safe to notar to figure out previous notar vote
+    pub(crate) fn get_prev_notarize_vote(&self, validator_key: &Pubkey) -> Option<(Hash, Hash)> {
```
AshwinSekar, line 124: i trust you got a nice way to do this with your refactor 🙏

carllin:
> Future PR will initialize cert pool from vote history on startup

Maybe an issue so we don't forget?

votor/src/certificate_pool.rs
```diff
@@ -553,28 +550,21 @@ impl CertificatePool {
...
     /// - votedStake(s) is the cumulative stake of all nodes who voted notarize or skip on s
     /// - topNotarStake(s) the highest of cumulative notarize stake per block in s
-    pub fn safe_to_skip(&self, slot: Slot, vote_history: &VoteHistory) -> bool {
-        if vote_history.its_over(slot) {
-            return false;
-        }
-
-        if vote_history.voted_skip_fallback(slot) {
-            return false;
-        }
```
carllin, line 563: are these checks replicated somewhere else?

votor/src/certificate_pool.rs
```diff
@@ -487,11 +485,7 @@ impl CertificatePool {
     /// (ii) At least 20% of stake voted to notarize `b` and at least 60% of stake voted to either notarize `b` or skip `slot`
     /// and we have not already cast a notarize fallback for this `b`
     /// If all the above hold, return `Some(block_id, bank_hash)` for the `b`
-    pub fn safe_to_notar(&self, slot: Slot, vote_history: &VoteHistory) -> Vec<(Hash, Hash)> {
-        if vote_history.its_over(slot) {
-            return vec![];
-        }
-
```
carllin, line 494: is there an equivalent check to this in the new code?

votor/src/certificate_pool.rs
```diff
@@ -525,14 +521,15 @@ impl CertificatePool {
         {
             let b_block_id = block_id.unwrap();
             let b_bank_hash = bank_hash.unwrap();
-            if vote_history.voted_notar_fallback(slot, b_block_id, b_bank_hash) {
+
+            if !voted_skip
+                && my_prev_notarize_vote
+                    .is_none_or(|(prev_block_id, _)| prev_block_id == b_block_id)
```
carllin, line 527: This is equivalent to the above check?
```
if !vote_history.voted(slot) {
            return vec![];
        }
```

Edit: makes sense because in VoteHistory 
```
/// The slots which this node has cast either a notarization or skip vote
    voted: HashSet<Slot>,
```
only covers notarization and skip

votor/src/certificate_pool.rs
```diff
@@ -525,14 +521,15 @@ impl CertificatePool {
...
+            {
+                // We either have not voted for the slot or we voted notarize on this block.
+                // Not eligble for safe to notar
                 continue;
             }
-            if let Some((prev_block_id, prev_bank_hash)) = b_prime {
-                if prev_block_id == b_block_id && prev_bank_hash == b_bank_hash {
```
carllin, line 532: we no longer need to check the bank hash because we're relying only on block id now?

votor/src/certificate_pool.rs
AshwinSekar: ah good catch forgot to update this part in the voting loop. Should roughly match now

votor/src/certificate_pool.rs
AshwinSekar: yeah we're eventually going to remove all bank hash references here and in votes so i've been in ignoring bank hash in new code that I write and then we can remove the outdated stuff once things stabilize.

PR #276, opened 2025-07-16
Bls cert all to all

wen-coding:
Send and ingest BLS cert all-to-all.


votor/src/certificate_pool.rs
```diff
@@ -101,13 +103,16 @@ pub struct CertificatePool {
     root_epoch: Epoch,
     /// The certificate sender, if set, newly created certificates will be sent here
     certificate_sender: Option<Sender<(CertificateId, CertificateMessage)>>,
+    /// The voting sender, used to send certificates all-to-all
+    voting_sender: Option<Sender<BLSOp>>,
```
carllin: don't we already have a `certificate_sender` here that sends certificates to `CertificateService` to store in blockstore and broadcast all to all?

carllin: hmmm not convinced that `CertificatePool` should own these senders, pollutes the API

Seems better for CertificatePool to return in its API's when certs have been finished and the voting loop can take care of sending sending/storing these certs

wen-coding: As far as I know certificate_sender was created when we still wanted to do cert repair, and that's no longer the plan.

wen-coding: I'm fine with moving it there because voting loop already uses voting_sender. I didn't choose that plan because I don't want to copy all certificates from CertificatePool back to voting loop, let me see if I can somehow reduce the copy around.

wen-coding: Changed to send certificates from voting loop instead.

AshwinSekar: I think it's fine to remove `CertificateService` now that we no longer need certificate repair.
Storing in blockstore is nice for traceability, but it's not strictly necessary, i'm fine sending the certificate A2A directly in the certificate pool / voting loop. If we want to still store in blockstore we can add a thread in votor as well.

votor/src/certificate_pool.rs
```diff
@@ -315,17 +283,45 @@ impl CertificatePool {
...
     }
 
-    pub(crate) fn insert_certificate(&mut self, cert_id: CertificateId, cert: VoteCertificate) {
-        self.completed_certificates.insert(cert_id, cert);
+    pub(crate) fn insert_certificate(
+        &mut self,
+        cert_id: CertificateId,
+        cert: VoteCertificate,
+        send_certificates: bool,
+    ) -> Result<Arc<CertificateMessage>, AddVoteError> {
```
carllin: I think we can 
1. delete the `send_certificates` bool here
2. Pass in the actual `Arc<CertificateMessage>` instead of constructing it here
3. function no longer needs to return anything
 
Looks like this:

```
    pub(crate) fn insert_certificate(
        &mut self,
        cert_id: CertificateId,
        new_cert: Arc<CertificateMessage>,
    ) {
        self.completed_certificates.insert(cert_id, new_cert);
        match cert_id {
            CertificateId::NotarizeFallback(slot, block_id, bank_hash) => {
                self.parent_ready_tracker
                    .add_new_notar_fallback((slot, block_id, bank_hash));
                if self
                    .highest_notarized_fallback
                    .map_or(true, |(s, _, _)| s < slot)
                {
                    self.highest_notarized_fallback = Some((slot, block_id, bank_hash));
                }
            }
            CertificateId::Skip(slot) => self.parent_ready_tracker.add_new_skip(slot),
            CertificateId::Finalize(slot) | CertificateId::FinalizeFast(slot, _, _) => {
                if self.highest_finalized_slot.map_or(true, |s| s < slot) {
                    self.highest_finalized_slot = Some(slot);
                }
            }
            CertificateId::Notarize(_, _, _) => (),
        }
    }
```

votor/src/certificate_pool.rs
```diff
@@ -204,76 +205,43 @@ impl CertificatePool {
...
-            })
+                match vote_pool {
+                    VotePoolType::SimpleVotePool(pool) => pool.add_to_certificate(&mut vote_certificate),
+                    VotePoolType::DuplicateBlockVotePool(pool) => pool.add_to_certificate(voted_block_key.as_ref().expect("Duplicate block pool for {vote_type:?} expects a voted block key for certificate {cert_id:?}"), &mut vote_certificate),
+                }.map_err(AddVoteError::Certificate)?;
+            }
+            let new_cert = self.insert_certificate(cert_id, vote_certificate, true)?;
+            new_certificates.push(new_cert);
+        }
+        Ok(new_certificates)
```
carllin: After this: https://github.com/anza-xyz/alpenglow/pull/276#discussion_r2216998325 can be maybe cleaned up to:

```
 fn update_certificates(
        &mut self,
        vote: &Vote,
        voted_block_key: Option<VotedBlockKey>,
        total_stake: Stake,
    ) -> Vec<Arc<CertificateMessage>>, AddVoteError {
        let slot = vote.slot();
    
        vote_to_certificate_ids(vote)
            .iter()
            .filter_map(|id| {
                if self.completed_certificates.contains_key(id) {
                    None
                } else {
                    self.find_new_certificates(*id, slot, voted_block_key.as_ref(), total_stake)
                }
            })
            //  └────────────── Result<Arc<CertificateMessage>, AddVoteError>
            .collect()       // <Result<Vec<Arc<CertificateMessage>>, AddVoteError>>
    }
    
    fn find_new_certificates(
        &mut self,
        cert_id: CertificateId,
        slot: Slot,
        voted_block_key: Option<&VotedBlockKey>,
        total_stake: Stake,
    ) -> Option<Arc<CertificateMessage>>, AddVoteError {
        let (limit, vote_types) = certificate_limits_and_vote_types(&cert_id);
    
        let accumulated: Stake = vote_types
            .iter()
            .filter_map(|vt| {
                self.vote_pools.get(&(slot, *vt)).map(|pool| match pool {
                    VotePoolType::SimpleVotePool(p) => p.total_stake(),
                    VotePoolType::DuplicateBlockVotePool(p) => {
                        p.total_stake_by_voted_block_key(
                            voted_block_key.expect("duplicate‑block pool needs key"),
                        )
                    }
                })
            })
            .sum();
    
        // Skip if the certificate is not yet complete.
        if (accumulated as f64) / (total_stake as f64) < limit {
            return Ok(None);
        }
    
        let mut cert = VoteCertificate::new(cert_id);
        for vt in vote_types {
            if let Some(pool) = self.vote_pools.get(&(slot, *vt)) {
                match pool {
                    VotePoolType::SimpleVotePool(p) => p.add_to_certificate(&mut cert),
                    VotePoolType::DuplicateBlockVotePool(p) => p.add_to_certificate(
                        voted_block_key.expect("duplicate‑block pool needs key"),
                        &mut cert,
                    ),
                }
                .map_err(AddVoteError::Certificate)?;
            }
        }
    
        let new_cert = Arc::new(cert.certificate().clone());
        self.insert_certificate(cert_id, new_cert.clone()); 
        Some(new_cert)// already returns a Result<Arc<…>>
    }
```

votor/src/certificate_pool.rs
```diff
@@ -414,11 +427,46 @@ impl CertificatePool {
...
+            CertificateType::NotarizeFallback => CertificateId::NotarizeFallback(
+                certificate.slot,
+                certificate.block_id.unwrap(),
+                certificate.replayed_bank_hash.unwrap(),
+            ),
+            CertificateType::Skip => CertificateId::Skip(certificate.slot),
+        };
+        if self.completed_certificates.contains_key(&certificate_id) {
+            return Ok(vec![]);
+        }
```
carllin: can this be implemented as an into/from method from `CertificateType` to `CertificateId`

votor/src/certificate_pool.rs
```diff
@@ -414,11 +427,46 @@ impl CertificatePool {
...
+        let new_certificate = self.insert_certificate(
+            certificate_id,
+            VoteCertificate::from(certificate_message.clone()),
+            true,
+        )?;
+        Ok(vec![new_certificate])
+    }
+
```
carllin: can this be a test only method?

votor/src/certificate_pool.rs

votor/src/certificate_pool.rs
wen-coding: Hmm, in `update_certificates` the problem with
```
vote_to_certificate_ids(vote)
            .iter()
            .filter_map(|id| {
```
 is that I can't return early on an Error. Otherwise the way you suggested would work.

votor/src/certificate_pool.rs

votor/src/certificate_pool.rs

votor/src/voting_loop.rs
```diff
@@ -797,6 +799,7 @@ impl VotingLoop {
         vote_receiver: &VoteReceiver,
         cert_pool: &mut CertificatePool,
         commitment_sender: &Sender<AlpenglowCommitmentAggregationData>,
+        bls_sender: &Sender<BLSOp>,
```
AshwinSekar, line 802: aplogies for the merge conflicts, after you rebase #279 we can add the bls sender here:
https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/certificate_pool_service.rs#L37
And then send the certificates that are returned here:
https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/certificate_pool_service.rs#L160

votor/src/certificate_pool.rs
carllin: Ah, missed that

Is there a reason `pool.add_to_certificate()` has to return an error, seems like pools should gurantee that all signatures within the pool are valid

In vote_certificate::aggregate() i see these possible errors:

1.`SignatureProjective::try_from(*signature).map_err(|_| CertificateError::InvalidSignature)?` Seems like this should be verified before adding to the pool
2. `if bitmap.len() <= vote_message.rank as usize { return Err(CertificateError::ValidatorDoesNotExist);}` Also seems like this should be verified before adding to the pool

Issue #243, opened 2025-07-07
Fast crashed leader timeout


qkniep:
#### Problem
It seems we are currently not distinguishing between the case where we received no shreds and where we received at least one shred. In case of crashed leaders, we thus unnecessarily wait the extra `BLOCKTIME` amount of time they would have to produce their block.

#### Proposed Solution
In the case where we see not a single shred for the current slot, we can assume the leader is crashed and timeout already `BLOCKTIME` earlier than we otherwise would. This way we get rid of crashed leaders faster.

We have to be a bit careful here. Technically, we can only deduct the full amount of `BLOCKTIME` if we assume honest leaders to send out their first slice *immediately* upon entering their slot. Otherwise, we may want to account for this additional delay. One way to do this would be to define a minimum slice frequency of `BLOCKTIME / k` and deduct only `BLOCKTIME - BLOCKTIME / k`.

Alternatively, especially if we conservatively account for Turbine/Rotor with 2 * Delta, we can say this small delay is already accounted for within that waiting time.

Edit: Actually it would be enough to configure a `max_time_for_first_slice`, then we can deduct `BLOCKTIME - max_time_for_first_slice `.

AshwinSekar:
Changes involved:
- Add new events for `FirstShred` and `TimeoutCrashedLeader` here  https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/event.rs#L30
- Add a sender/receiver pair for `FirstShred`. The sender should notify here https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/turbine/src/retransmit_stage.rs#L661 when the first shred is received by retransmit stage
- Update the skip timer to additionally send the `TImeoutCrashedLeader` event https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/skip_timer.rs#L139
- Update the event loop to `try_skip_window` if we receive a `TimeoutCrashedLeader` for slot `s` but have not received a `FirstShred` event for slot `s` https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/event_handler.rs#L182
[...]

Issue #242, opened 2025-07-07
Fix standstill timeout


qkniep:
If I understand correctly, this corresponds to the standstill timeout we propose in the paper to be something like 10 seconds:

https://github.com/anza-xyz/alpenglow/blob/5429f250b243eb8c334abc90a89eaa94d57dfa71/core/src/alpenglow_consensus/voting_loop.rs#L422

Maybe it would make sense to increase this? Setting this to just 1 second seems very aggressive, considering that we expect this to never be necessary in practice. This is only for recovering from the theoretical problem of concentrated packet loss scenarios (e.g. 50% of nodes' vote packets across the entire network being dropped). Every time we trigger this unnecessarily, we waste bandwidth.

Even more importantly: This timer should only reset when seeing a new block finalized and not on every ParentReady!

AshwinSekar:
#279 Increases the timer to 10 seconds and makes it only reset when a new block is finalized.
The remaining work is to broadcast certificates as well here:
https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/certificate_pool_service.rs#L123

We need to add a function to the `certificate_pool` to send these certificates since the `highest_finalized_slot` after #276 lands

PR #279, opened 2025-07-17
WIP: event driven voting loop to handle all slots concurrently

AshwinSekar:
#### Problem
Sequential loop is hard to follow and slow.

todo list:
- Plug in set root
- Add skip timer
- Ingest own votes into cert pool
- remove voting_loop.rs
- cleanup & testing
- add detailed follow up issues for delegation

#### Summary of Changes

Fixes #


AshwinSekar:
Follow ups:
- #242 
- https://github.com/anza-xyz/alpenglow/issues/243
- #283
- #284
- #285
- #286
- #287
- #288
- #289
- #290
- #291
- #292
- #293

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -127,10 +127,12 @@ pub(super) fn set_block_id_and_send(
     block_id: Hash,
 ) -> Result<()> {
     bank.set_block_id(Some(block_id));
-    completed_block_sender.send(CompletedBlock {
-        slot: bank.slot(),
-        bank,
-    })?;
+    if bank.is_frozen() {
```
carllin, line 130: why is this new check necessary?

votor/src/certificate_pool.rs
```diff
@@ -252,26 +256,41 @@ impl CertificatePool {
                 }
 
                 match cert_id {
-                    CertificateId::Notarize(_, _, _) => (),
+                    CertificateId::Notarize(slot, block_id, bank_hash) => {
+                        events.push(VotorEvent::BlockNotarized((slot, block_id, bank_hash)));
+                        if self.is_finalized(slot) {
+                            events.push(VotorEvent::Finalized((slot, block_id, bank_hash)));
+                        }
```
carllin, line 263: do we need this check if we already have the finalized event below?

Seems like the events loop already tracks the finalized/notarized blocks so if there's any order-dependent activity, should be checked there

core/src/replay_stage.rs
```diff
@@ -3492,10 +3510,11 @@ impl ReplayStage {
...
                         }
                     }
-                } else {
-                    None
                 };
-                bank.set_block_id(block_id);
+
+                if bank.block_id().is_none() {
+                    bank.set_block_id(block_id);
+                }
```
carllin, line 3517: nice, fixes the leader block id issue 👍

votor/src/certificate_pool_service.rs
```diff
@@ -0,0 +1,267 @@
...
+                        trace!("{}: unable to push vote into pool {}", ctx.my_pubkey, e);
+                        continue;
+                    }
+                };
+            }
+        }
+        Ok(())
+    }
+
+    fn add_produce_block_event(
```
carllin, line 193: hehe what determines if something like this vs `add_message_and_maybe_update_commitment()` goes into `voting_utils.rs` vs stays here

votor/src/certificate_pool_service.rs
```diff
@@ -0,0 +1,267 @@
...
+    pub(crate) my_vote_pubkey: Pubkey,
+    pub(crate) blockstore: Arc<Blockstore>,
+    pub(crate) root_bank_cache: RootBankCache,
+    pub(crate) leader_schedule_cache: Arc<LeaderScheduleCache>,
+
+    // TODO: for now we ingest our own votes into the certificate pool
+    // just like regular votes. However do we need to convert
+    // Vote -> BLSMessage -> Vote?
+    // consider adding a separate pathway in cert_pool.add_transaction for ingesting own votes
+    pub(crate) own_vote_receiver: BLSVerifiedMessageReceiver,
```
carllin, line 51: can we instead just clone the same sender as `bls_receiver` and pass that sender to the voting loop, so that we don't need a separate receiver for our own votes?

votor/src/certificate_pool_service.rs
```diff
@@ -0,0 +1,267 @@
...
+                    Ok(Some(finalized_slot)) => {
+                        // Reset standstill timer
+                        debug_assert!(finalized_slot > highest_finalized_slot);
+                        highest_finalized_slot = finalized_slot;
+                        standstill_timer = Instant::now();
+                        // Set root
+                        let root_bank = ctx.root_bank_cache.root_bank();
+                        if root_bank.slot() > current_root {
+                            current_root = root_bank.slot();
+                            cert_pool.handle_new_root(root_bank);
```
carllin, line 169: This is to handle when the finalized slot is potentially ahead of our local roots?

Because this isn't dependent on highest_finalized_slot, i.e. the root is being updated in the event loop, should we just move this check to the beginning of the entire certificate pool service loop

votor/src/skip_timer.rs
```diff
@@ -0,0 +1,170 @@
...
+
+pub(crate) struct SkipTimerManager {
+    heap: BinaryHeap<SkipTimer>,
+    order: VecDeque<u64>,
+    max_timers: usize,
+    next_id: u64,
+}
+
+impl SkipTimerManager {
+    pub(crate) fn new(max_timers: usize) -> Self {
```
carllin, line 65: max_timers just in case you have an ungodly number of forks? i.e. notarization and skip certificates for the same slots over and over?

votor/src/skip_timer.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        let id = self.next_id;
+        self.next_id = self.next_id.wrapping_add(1);
+
+        // To account for restarting from a snapshot in the middle of a leader window
+        // or from genesis, we compute the exact length of this leader window:
+        let leader_window_offset = leader_slot_index(start_slot);
+        let remaining = NUM_CONSECUTIVE_LEADER_SLOTS
+            .checked_sub(leader_window_offset as u64)
```
carllin, line 89: nit: might be better to make remaining a utility function like `leader_slot_index`, otherwise we have to assume the internals of `leader_slot_index` use `NUM_CONSECUTIVE_LEADER_SLOTS`

votor/src/certificate_pool_service.rs
AshwinSekar: I left it at as a separate receiver in case we want to go down the path in the TODO and avoid wrapping and unwrapping the vote. If we don't want to do that i can unify the receiver, wdyt?

votor/src/certificate_pool_service.rs
AshwinSekar: yeah I could move it to the main loop but then it'll fire every iteration, I figured it would be better to at least have it behind a new finalization event so that we grab the read lock less frequently.

There's no problem if the cert pool root is slightly behind the bank forks root, just need to perform cleanup eventually.

votor/src/skip_timer.rs
AshwinSekar: I believe in a happy small cluster the rate of ParentReady events (finishing a leader window) could be 1600 ms + some network delay. However the timer takes 2400 ms to finish firing all 4 events.
Thus we make timers faster than we clear them, so we need to have a cap. 
more details and some ideas in  #298

votor/src/certificate_pool_service.rs
AshwinSekar: which side of the bed I woke up on that day :)
jk, i'm open to moving these functions around. Perhaps `add_message_and_maybe_update_commitment()` should live in this file instead?

votor/src/certificate_pool.rs
AshwinSekar: > Seems like the events loop already tracks the finalized/notarized blocks so if there's any order-dependent activity, should be checked there

The problem is that the slow finalization certificate only contains the slot #, while we need the full block to process in the event loop. However the fast finalization certificate contains all the information. To avoid sending 2 separate events for slow and fast finalization, I handled it here so that we always send a finalization event with the information:
- If you receive the slow finalization certificate first, then the notarization certificate, notify the event loop here.
- If you first receive the notarization certificate, then the slow finalization certificate, notify the event loop on line # 277
- If you receive a fast finalization certificate, then immediately notify the event loop

turbine/src/broadcast_stage/broadcast_utils.rs
AshwinSekar: very rarely, the leader block finishes shredding before replay!
in that case we should have the replay thread send `Block` instead of the broadcast thread.

votor/src/skip_timer.rs
AshwinSekar: good call, https://github.com/anza-xyz/alpenglow/issues/303

votor/src/certificate_pool.rs
carllin: oh because we don't have the hash/block id in the Final vote

Even if you get finalization before notarization, you still notify RPC of finalization which seems fine

votor/src/certificate_pool_service.rs
carllin: For now I think anything specific to this module that isn't also used by replay should probably just live here

votor/src/certificate_pool_service.rs
carllin: Could just send an enum I think that covers both?

votor/src/skip_timer.rs
carllin: > However the timer takes 2400 ms to finish firing all 4 events.

Hmmm the timer is just +800ms for the timeout, but shouldn't there be a 1:1 between the leader window and the timer firing if there's no forks?

AshwinSekar: you're correct yeah, there should only be a fixed amount of timers because of the 1:1 between ParentReady and timer creation. 
Forks shouldn't play a role, we only set one timer per leader window.

votor/src/certificate_pool_service.rs
wen-coding: Trying to address https://github.com/anza-xyz/alpenglow/pull/300/files#r2220502653 and now I'm confused.  The `new_finalized_slot` can totally be ahead of local root. We might have a SlowFinalization cert arriving before the corresponding Notarization cert, and we might not have this new finalized block at all, so it will be a while before we have the block, have it replayed, and perform actual `set_root`. Is it a problem if we just do `cert_pool.handle_new_root(root_bank);` to discard everything before `new_finalized_slot` ? I think this function used to be called only when we run `set_root` in `voting_loop`.

AshwinSekar: The root bank is from bank forks via the root bank cache, the finalization event is just a trigger to check the cache. So cert pool can't root before the event loop.

wen-coding: Then is the `new_finalized_slot` check really helping you here? Let's say we mark slot `s` finalized because we received some votes or certs, the cert pool will send out `VotorEvent::Finalized` and then we reach here and do the root bank cache check, this only succeeds if:
1. the block is in bank_forks and replay finished
2. the `VotorEvent::Finalized` has been handled and set_root has been done
If either of the above fails, we won't even check root bank cache until some day another slot got finalized.

It seems like this new_finalized_slot check might make `cert_pool.handle_new_root` happen slower.

AshwinSekar: the actual slot doesn't matter here, it's just the timing of when to check the root bank cache.
For example lets say the cert pool root is 1.
The event loop roots 2.
The cert pool see's that 3 is finalized, now the cert pool will root 2. 

If we don't gate it behind the finalization event, we'll end up checking the lock quite frequently. It's not a big deal if the cert pool root is slightly behind the event loop.

wen-coding: If you are worried about checking the lock too frequently then a simple interval might help you more. I think root bank cache is supposed to solve the problem where we grab bank_forks read lock too often. If we are worried about even querying root bank cache then it seems like it's not doing its job.

That said, I do think the correct solution is for certificate pool to get root bank notification from EpochStakesService @akhi3030 is working on.

AshwinSekar: i mean it's just wasteful to check the root bank more frequently, we know in steady state rooting occurs roughly every 400 ms, similar to finalization so it makes sense to check when we receive the finalization notification.
alternatively we could have the event loop notify the cert pool when it roots a block if you want some tighter synchronization.
The epoch stakes service will only notify when the epoch rolls over, which is too infrequent for cert pool cleanup.
Is there a use case you're thinking that requires we check the root more frequently?

wen-coding: It probably works in normal cases. But hidden assumptions like this should at least be specified in comments so in the future new members of the team don't think it's a bug and spend time debugging it. For example, the assumptions that:
1. Querying root bank cache is costly and should not be done frequently (I still don't understand the reasoning here)
2. We do expect that in some cases cert_pool will only be cleaned on next finalized slot so it can be a bit out of sync with local roots and that's fine

votor/src/certificate_pool_service.rs
AshwinSekar: done here https://github.com/anza-xyz/alpenglow/pull/309

votor/src/certificate_pool_service.rs
AshwinSekar: done here #308

votor/src/certificate_pool_service.rs
AshwinSekar: 1) I'm not saying it's overly costly, i'm just saying it's unnecessary. We only expect the root bank to change ~ 400ms there's no need to check for updates in a tight loop while ingesting bls messages every couple ms. 
I would argue it's more confusing to a reader, because such logic would indicate that we expect the root bank to be changing quite frequently, and it's imperative that we observe this (which is not true).

2) Even if we did check in the loop, the cert_pool root can still be behind the bank forks root. The only way to ensure full synchronicity would be to share the cert pool and update it behind a write lock when setting root, which is completely unnecessary for our purposes.

votor/src/certificate_pool_service.rs
```diff
@@ -0,0 +1,267 @@
...
+            // TODO: we need set identity here as well
+            Self::add_produce_block_event(
+                &mut highest_parent_ready,
+                &cert_pool,
+                &mut ctx,
+                &mut events,
+            );
+
+            if standstill_timer.elapsed() > STANDSTILL_TIMEOUT {
+                events.push(VotorEvent::Standstill(highest_finalized_slot));
```
wen-coding, line 123: Can we just use cert_pool.highest_finalized_slot() here?
I think this highest_finalized_slot is just used to:
1. check whether the returned `highest_finalized_slot` is larger than the local copy, which `certificate_pool` already checks before returning
2. used in Standstill event here, which we can get from `cert_pool.highest_finalized_slot()`.

AshwinSekar: great idea, I think that should work.

PR #281, opened 2025-07-18
refactor: move `BlsSigVerifierStats` into its own module and clean it up a bit

akhi3030:
Previously submission of stats was intermixed with the business logic and there were some minor issues.

This PR separates `BlsSigVerifier` and its stats into separate modules which makes things a bit cleaner.  This also allows us to eliminate some code duplication and restrict visibility.


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -90,17 +51,16 @@ impl SigVerifier for BLSSigVerifier {
         packet_batches: Vec<PacketBatch>,
     ) -> Result<(), SigVerifyServiceError<Self::SendType>> {
         // TODO(wen): just a placeholder without any batching.
-        let mut stats = BLSSigVerifierStats::new();
-        let mut verified_votes: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
+        let mut verified_votes = HashMap::new();
 
         for packet in packet_batches.iter().flatten() {
-            stats.received += 1;
+            self.stats.received += 1;
```
wen-coding: We are allocating a new stats every time because we want to do parallel processing later.

akhi3030: Thanks for the feeback @wen-coding.  I've made some improvements based on your feedback.  PTAL.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -1,6 +1,8 @@
 //! The BLS signature verifier.
 //! This is just a placeholder for now, until we have a real implementation.
 
+mod stats;
```
wen-coding, line 4: This module currently only contains stats for BLSSigverifier right? 
Does it make sense to move bls_sigverifier into bls_sigverifier/ directory as well?

akhi3030: Based on the rules of modules work in rust, if we move `bls_sigverifier.rs` into the `bls_sigverifier/` directory, it will have to be renamed to `bls_sigverifier/mod.rs`.  

The general trend is not to name files `mod.rs` as that makes it harder for some tools to function and depending on your IDE setup, can become confusing if you have too many open files that are all named `mod.rs`.

wen-coding (approved):

Issue #284, opened 2025-07-19
Propagate errors in CertificatePoolService


AshwinSekar:
Error handling in this file is kind of a mess, lets propagate errors properly and take the appropriate action of setting the exit flag and exiting, or logging and continuing depending on the error.

PR #300, opened 2025-07-20
BLS cert all to all.

wen-coding:
Broadcast BLS cert all-to-all.


votor/src/certificate_pool.rs
```diff
@@ -395,15 +376,37 @@ impl CertificatePool {
...
+        let new_certficates_to_send = match message {
+            BLSMessage::Vote(vote_message) => {
+                self.add_vote(my_vote_pubkey, vote_message, events)?
+            }
+            BLSMessage::Certificate(certificate_message) => {
+                self.add_certificate(certificate_message, events)?
+            }
+        };
+        // If we have a new highest finalized slot, return it
+        let new_finalized_slot = if self.highest_finalized_slot != current_highest_finalized_slot {
```
AshwinSekar: nit
```suggestion
        let new_finalized_slot = if self.highest_finalized_slot > current_highest_finalized_slot {
```

votor/src/certificate_pool_service.rs
```diff
@@ -157,19 +158,38 @@ impl CertificatePoolService {
...
-                        // Reset standstill timer
-                        debug_assert!(finalized_slot > highest_finalized_slot);
-                        highest_finalized_slot = finalized_slot;
-                        standstill_timer = Instant::now();
-                        // Set root
-                        let root_bank = ctx.root_bank_cache.root_bank();
-                        if root_bank.slot() > current_root {
-                            current_root = root_bank.slot();
-                            cert_pool.handle_new_root(root_bank);
+                    Ok((new_finalized_slot, new_certificates_to_send)) => {
```
AshwinSekar: could we move this entire match statement to a helper function that returns a result?
Then if the helper function returns an error corresponding to the bls_sender being disconnected, or AddVoteError::CertificateSendError we can shutdown, if it's a "unable to push vote into pool" type of error we can `continue`

Then in a future PR we can also bubble this error all the way up to the `new` fn for shutting down #284

AshwinSekar (commented):
Overall looks great! just a couple of comments

votor/src/certificate_pool_service.rs
wen-coding: This is an unbounded channel, so I believe it will only return SendError on channel disconnect.

Helper added.

votor/src/certificate_pool_service.rs
```diff
@@ -138,53 +218,33 @@ impl CertificatePoolService {
                 return Ok(());
             }
 
-            let bls_messages = select! {
+            let bls_messages: Vec<BLSMessage> = select! {
                 recv(ctx.own_vote_receiver) -> msg => {
-                    std::iter::once(msg?).chain(ctx.own_vote_receiver.try_iter())
+                    std::iter::once(msg?).chain(ctx.own_vote_receiver.try_iter()).collect()
```
AshwinSekar, line 227: do we need to collect here?

votor/src/certificate_pool_service.rs
```diff
@@ -76,6 +76,85 @@ impl CertificatePoolService {
...
+        // Send new certificates to peers
+        for certificate in new_certificates_to_send {
+            // bls_sender is an unbounded channel, so error only happens if the receiver is disconnected
+            if bls_sender
+                .send(BLSOp::PushCertificate {
+                    certificate: certificate.clone(),
+                })
+                .is_err()
+            {
+                return Err(AddVoteError::CertificateSenderError);
```
AshwinSekar:
```suggestion
                return Err(AddVoteError::VotingServiceSenderDisconnected);
```

votor/src/certificate_pool_service.rs
wen-coding: We do if we need to pass in `&mut ctx` later. We need `&mut ctx` instead of `&ctx` because we need to update `root_bank_cache`. I can theoretically pass in individual members, but then linter complains this helper function has too many arguments.

I believe this collect doesn't make a copy.

votor/src/certificate_pool_service.rs
```diff
@@ -75,6 +76,85 @@ impl CertificatePoolService {
         Self { t_ingest }
     }
 
+    fn maybe_update_root_and_send_new_certificates(
```
bw-solana, line 79: should this go in `root_utils.rs`?

votor/src/certificate_pool_service.rs
```diff
@@ -75,6 +76,85 @@ impl CertificatePoolService {
...
+            // Set root
+            let root_bank = root_bank_cache.root_bank();
+            if root_bank.slot() > *current_root {
+                *current_root = root_bank.slot();
+                cert_pool.handle_new_root(root_bank);
+            }
+        }
+        // Send new certificates to peers
+        for certificate in new_certificates_to_send {
+            // bls_sender is an unbounded channel, so error only happens if the receiver is disconnected
```
bw-solana: should we bound this?

core/src/voting_service.rs
```diff
@@ -397,9 +412,9 @@ mod tests {
         )
     }
 
-    #[allow(clippy::disallowed_methods)]
-    #[test]
-    fn test_send_bls_message() {
+    #[test_case(true)]
+    #[test_case(false)]
```
bw-solana: nit: might be nice to just define the BLS ops here rather than passing a bool and then building them out. Especially if we go beyond 2 enums

votor/src/certificate_pool/vote_certificate.rs
```diff
@@ -25,12 +25,8 @@ pub enum CertificateError {
     BlsError(#[from] BlsError),
     #[error("Invalid pubkey")]
     InvalidPubkey,
-    #[error("Invalid signature")]
-    InvalidSignature,
     #[error("Validator does not exist for given rank")]
```
bw-solana: can we print the index in the error message?

votor/src/certificate_pool_service.rs
AshwinSekar: I think the sending certificates part should probably stay here.
We could move this part to `root_utils.rs`. 
```
  // Reset standstill timer
            debug_assert!(new_finalized_slot > *highest_finalized_slot);
            *highest_finalized_slot = new_finalized_slot;
            *standstill_timer = Instant::now();
            // Set root
            let root_bank = root_bank_cache.root_bank();
            if root_bank.slot() > *current_root {
                *current_root = root_bank.slot();
                cert_pool.handle_new_root(root_bank);
            }
```

Maybe cleaner to create a struct with all the mutables from the loop to pass around instead of all the arguments as well:
```
&mut current_root,
&mut highest_finalized_slot,
&mut standstill_timer,
```

core/src/voting_service.rs
wen-coding: Done.

votor/src/certificate_pool/vote_certificate.rs
wen-coding: Done.

votor/src/certificate_pool_service.rs
wen-coding: Since this is verified and forwarded by our own certificate pool before getting here, and we should only send once per (slot, cert_type), I think it's okay if we don't bound the channel.

bw-solana: If we bound, we can pre-allocate a ring buffer

votor/src/certificate_pool_service.rs
wen-coding: In that case I think the standstill_timer maybe doesn't belong here either....

Since this is code refactoring, can this change maybe go into another PR instead of piggying back on this BLS cert all to all change?

bw-solana: I'm okay w/ doing it in a follow-up 👍

votor/src/certificate_pool_service.rs
wen-coding: Hmm okay, that's fair. Although I probably won't build certificate resend here, I'll just drop and log error, since in normal cases the buffer shouldn't overflow.

The BLSOps to send out should be biggest during standstill handling. Currently we are saying after 10 seconds standstill, send out all own votes and certificates, since 10/0.4 = 25, and we can have <= 5 votes and <= 5 certs per slot, I'll cap the channel at 512.

bw-solana: 512 sounds reasonable to me. Might be nice to log a drop metric as well (easier to query at scale and see if we have a bug)

core/src/tvu.rs
```diff
@@ -298,7 +298,10 @@ impl Tvu {
         let (cost_update_sender, cost_update_receiver) = unbounded();
         let (drop_bank_sender, drop_bank_receiver) = unbounded();
         let (voting_sender, voting_receiver) = unbounded();
-        let (bls_sender, bls_receiver) = unbounded();
+        // The BLS sender channel should be mostly used during standstill handling,
+        // there could be 10s/400ms = 25 slots, <=5 votes and <=5 certificates per slot,
+        // we cap the channel at 512 to give some headroom.
```
bw-solana, line 303: thanks for the comment!

bw-solana (approved):

votor/src/certificate_pool_service.rs
wen-coding: Sure. Err currently we have no stats in certificate_pool_service at all. Can I do that in a follow up PR to properly add stats for drops and everything else?

core/src/tvu.rs
qkniep: the 25 slots is not an upper bound though, only what we would expect if the network recovered in the meantime (if e.g. we have a network partition for >10s this does not hold)

wen-coding: But if the network is in a partition where everyone is in a partition with <60% stake, then we shouldn't have many votes and certs constructed during the partition? If one partition is > 60% stake, then others should be able to observe new finalized cert and continue during recovery.

PR #301, opened 2025-07-20
event_handler: unify the completed_block_sender and event_sender

### LINKED ISSUE #283: Unify completed_block_sender and event_sender
#### Problem
We maintain a separate sender for CompletedBlock events that we wrap here:
https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/event_handler.rs#L136

#### Proposed Solution
Just send it wrapped and use the event sender so we don't have to select! on this extra receiver.

AshwinSekar:
#### Problem
We maintain a separate sender for `Block` events. Instead just use the regular event sender.
Having the event sender available to tvu will also let us  send `FirstShred` events in the future.

#### Summary of Changes
Remove the `completed_block_sender/receiver` and plumb the `event_sender` instead.

Fixes #283


core/src/replay_stage.rs
```diff
@@ -3611,18 +3612,18 @@ impl ReplayStage {
...
+                // 1) Replay finishes before shredding, broadcast_stage will take care of
                 //      notifying votor
                 // 2) Shredding finishes before replay, we notify here
                 //
                 // For non leader banks (2) is always true, so notify here
                 if *is_alpenglow_migration_complete && bank.block_id().is_some() {
                     // Leader blocks will not have a block id, broadcast stage will
                     // take care of notifying the voting loop
-                    let _ = completed_block_sender.send(CompletedBlock {
+                    let _ = votor_event_sender.send(VotorEvent::Block(CompletedBlock {
```
AshwinSekar, line 3623: this and `broadcast_utils` is the actual change on the sender side, rest is the corresponding plumbing

carllin (approved):

PR #304, opened 2025-07-22
[votor] Replace `VoteCertificate` with either `VoteCertificateBuilder` or `CertificateMessage`

samkim-crypto:
#### Problem
The existing BLS vote certificate logic had an inefficiency related to costly type conversions during signature aggregation. The `VoteCertificate` type, a wrapper around `CertificateMessage`, stored its signature in the compact "affine" form. However, because signature aggregation is performed in "projective" form, each call to aggregate required the existing signature to be converted from affine to projective, incurring a non-trivial performance cost on a critical path.

#### Summary of Changes
To address this, I added a `VoteCertificateBuilder` type which holds the certificate signature in its projective form. This results in much cleaner and more performance aggregation function. Once all votes are aggregated, `VoteCertificateBuilder::build` can be called to produce the final `CertificateMessage`, converting the signature back to the compact affine form only once at the end.

With the new builder, I removed the redundant `VoteCertificate`, as its role is now cleanly handled by `VoteCertificateBuilder` for aggregation and `CertificateMessage` for the final product.

Additionally, I added both checked (`aggregate`) and unchecked (`aggregate_unchecked`) versions of the aggregation function. Since signatures are pre-verified during the sigverify phase, the consensus logic can use the more performant unchecked version.

Fixes #


votor/src/certificate_pool/vote_certificate.rs
```diff
@@ -33,73 +31,92 @@ pub enum CertificateError {
...
-            );
-            bitmap.set(vote_message.rank as usize, true);
-            // aggregate the signature
-            current_signature.aggregate_with([&vote_message.signature])?;
+            self.bitmap.set(vote_message.rank as usize, true);
         }
-        *signature = Signature::from(current_signature);
-        Ok(())
+        let signature_iter = messages.iter().map(|vote_message| &vote_message.signature);
+        Ok(self.signature.aggregate_with(signature_iter)?)
```
wen-coding, line 71: Sorry I forgot to ask, do you think this function should really go into CertificateMessage?

samkim-crypto: Sorry are you stating that the `aggregate` function should go into `CertificateMessage` instead of `VoteCertificate` (now `VoteCertificateBuilder`) or are you stating that maybe we should remove `aggregate` from votor entirely?

wen-coding: The former, now all we need from `VoteCertificate` is the `aggregate` function, I wonder whether we should just move that into `BLSMessage::CertificateMessage` in the vote program and remove `VoteCertificate` entirely. What do you think?

samkim-crypto: Oh yeah this is actually what I just addressed in this PR.

I removed `VoteCertificate` and replaced it with `CertificateMessage` in most places. For example see [here](https://github.com/anza-xyz/alpenglow/pull/304/files#diff-e64c2ddc2a7e91389eebee0b2aa3a234a69cc8286c9f7281a366a9ab8276b0fbR89).

But I did add `VoteCertificateBuilder` (maybe I can rename this to `CertificateMessageBuilder` instead) that is primarily used for signature aggregation. The issue is that `CertificateMessage` stores the signature in a more compact `Signature` form while signature aggregation can only be done in `SignatureProjective` form. So the previous aggregation function took the `Signature` type, converted it to `SignatureProjective`, then convert it back to `Signature` type after aggregation.

The `VoteCertificateBuilder` now just holds the `SignatureProjective` type from the start. Once the signatures are all aggregated, then it can use the [build](https://github.com/anza-xyz/alpenglow/pull/304/files#diff-dee427ddc9e4276709b33437fd08dc54e0d52ef6555e3c12308f662ebeef89c2R104) function to create `CertificateMessage`. It should be more efficient this way since it removes unnecessary costs related to type conversions.

I was planning on writing a detailed PR description, but I just wanted to check if the CI check passed first.

votor/src/certificate_pool.rs
```diff
@@ -555,11 +556,13 @@ impl CertificatePool {
     /// the size of the certificate
     #[cfg(test)]
     pub fn slot_notarized_fallback(&self, slot: Slot) -> Option<usize> {
+        // TODO: remove this line once `CertificateMessage` is moved into the repo`
+        use crate::certificate_pool::vote_certificate::get_vote_count;
```
wen-coding: I don't think we need get_vote_count any more, it was only used in tests and those tests have since been removed, I'm removing it in https://github.com/anza-xyz/alpenglow/pull/300

votor/src/certificate_pool.rs
```diff
@@ -555,11 +556,13 @@ impl CertificatePool {
...
         self.completed_certificates
             .iter()
             .find_map(|(cert_id, cert)| {
                 matches!(cert_id, CertificateId::NotarizeFallback(s,_,_) if *s == slot)
-                    .then_some(cert.vote_count())
+                    .then_some(get_vote_count(cert))
```
wen-coding: We don't need to return vote count here as I mentioned above, that test only checks if the certificate exists.

votor/src/certificate_pool/vote_certificate.rs
```diff
@@ -33,73 +31,92 @@ pub enum CertificateError {
     InvalidVoteType,
 }
 
-//TODO(wen): Maybe we can merge all the below functions into CertificateMessage.
+/// A builder for creating a `CertificateMessage` by efficiently aggregating BLS signatures.
 #[derive(Clone)]
-pub struct VoteCertificate(CertificateMessage);
+pub struct VoteCertificateBuilder {
```
wen-coding, line 32: Nice, let's change the file name as well.

votor/src/certificate_pool/vote_certificate.rs
```diff
@@ -33,73 +31,92 @@ pub enum CertificateError {
...
-            SignatureProjective::identity()
-        } else {
-            SignatureProjective::try_from(*signature)
-                .map_err(|_| CertificateError::InvalidSignature)?
-        };
-
-        // aggregate the votes
-        let bitmap = &mut self.0.bitmap;
+    /// Aggregates a slice of `VoteMessage`s into the builder.
+    #[allow(dead_code)]
```
wen-coding: If we will only use aggregate_unchecked in the future, we can remove this one.

votor/src/certificate_pool/vote_pool.rs
```diff
@@ -105,9 +102,8 @@ impl SimpleVotePool {
         true
     }
 
-    pub fn add_to_certificate(&self, output: &mut VoteCertificate) -> Result<(), CertificateError> {
-        output.aggregate(self.vote_entry.transactions.iter())?;
-        Ok(())
+    pub fn add_to_certificate(&self, output: &mut VoteCertificateBuilder) {
+        output.aggregate_unchecked(&self.vote_entry.transactions)
```
wen-coding: Instead of calling a function which panics inside, can we use `output.aggregate` which returns an error and panic here? A bit easier for constructing tests.

votor/src/certificate_pool.rs
samkim-crypto: Okay great! I removed this during the rebase.

votor/src/certificate_pool/vote_certificate.rs
samkim-crypto: Done!

votor/src/certificate_pool/vote_certificate.rs
samkim-crypto: Related to your comment below, I just removed the unchecked version and kept the safe one. A potential panic will now happen in `add_to_certificate`.

votor/src/certificate_pool/vote_pool.rs
samkim-crypto: Yep, done!

wen-coding (approved):

votor/src/certificate_pool/vote_pool.rs
```diff
@@ -102,8 +102,10 @@ impl SimpleVotePool {
         true
     }
 
-    pub fn add_to_certificate(&self, output: &mut VoteCertificate) {
-        output.aggregate(self.vote_entry.transactions.iter())
+    pub fn add_to_certificate(&self, output: &mut VoteCertificateBuilder) {
+        output
+            .aggregate(&self.vote_entry.transactions)
+            .expect("Incoming vote message signatures are assumed to be valid")
```
carllin, line 102: just double checking, we've made all the appropriate checks in the vote pool such that the two errors in `aggregate` can't happen right?

https://github.com/anza-xyz/alpenglow/blob/eba6a912468773b970921929a2e4eab5c888dfdc/votor/src/certificate_pool/vote_certificate_builder.rs#L64

and 

https://github.com/anza-xyz/alpenglow/blob/eba6a912468773b970921929a2e4eab5c888dfdc/votor/src/certificate_pool/vote_certificate_builder.rs#L71

wen-coding: Yes, they should not happen unless our check was screwed up, in that case we can panic.

samkim-crypto: Yeah both of these checks should be guaranteed by the sigverify stage.
- If the sigverify stage cannot find a valid public key to verify the signature, then it will discard the vote or cert packet
- The only way the `aggregate_with` function will fail is if the signature is invalid. The sigverify stage will obviously guarantee this as well

Issue #270, opened 2025-07-15
Create an EpochStakesService to allow users of EpochStakes avoid reading bank_forks


wen-coding:
#### Problem
We need EpochStakes in a lot of places, for example, in certificate_pool, and in BLSSigverifier.

Currently they all need to somehow get hold of root_bank to update the EpochStakes they use, this might cause too many data structures repeatedly querying bank_forks for root_bank unnecessarily.

#### Proposed Solution
Let's create EpochStakesService, this should be the only data structure reading bank_forks root_bank so it can update EpochStakes, all the other users should just be passed a reference of EpochStakesService and read EpochStakes blindly from there.

akhi3030:
I assume this issue is about addressing https://github.com/anza-xyz/alpenglow/blob/331c3bd905ac4f9180f4a38868cacc848f55f25b/core/src/sigverifier/bls_sigverifier.rs#L216 and https://github.com/anza-xyz/alpenglow/blob/331c3bd905ac4f9180f4a38868cacc848f55f25b/votor/src/certificate_pool.rs#L147.  

Looking at how the function is used in `bls_sigverifier.rs`, I am trying to understand what the exact problem is with the current code.  

> Currently they all need to somehow get hold of root_bank to update the EpochStakes they use, this might cause too many data structures repeatedly querying bank_forks for root_bank unnecessarily.

The actual update only happens once every 60seconds.  Is that too much querying for the various data structures mentioned above?  Or is there some other issue with the existing code?

wen-coding:
We don't like everyone to get hold of bank_forks, because we've been bitten by bank_forks deadlock before. Some benign code secretly grabbing bank_forks lock very deep, and we happily grabbed bank_forks lock before calling this benign code. Too many readers could also starve the writer.

Also, all these data structures don't care about other things in BankForks and Bank at all, all they need is EpochStakes to figure out who on a specific slot has how much stake. They don't really need that dependency.

We can't change the fact that `EpochStakes` is produced in `Bank` and you do need to get hold of root through `BankForks` to get your `EpochStakes` update, but we can decouple `EpochStakes` producer and consumer through this `EpochStakesService`, so the voting loop only needs to notify a few people when root changed. And EpochStakesService can get the root change much faster if needed.

Does that make sense?

wen-coding:
Just check who are using EpochStakes and how they get EpochStakes now.

wen-coding:
[...]

PR #306, opened 2025-07-22
feat: Introduce a EpochStakesService

akhi3030:
Part of https://github.com/anza-xyz/alpenglow/issues/270.

As discussed in the above issue, introduces a new `EpochStakesService` and uses it in `BLSSigVerifier`.

Additional components will be updated in subsequent PRs.

### DIFF
--- a/runtime/src/epoch_stakes_service.rs
+++ b/runtime/src/epoch_stakes_service.rs
@@ -1,0 +1,77 @@
+use {
+    crate::{
+        bank_forks::BankForks,
+        epoch_stakes::{BLSPubkeyToRankMap, EpochStakes},
+    },
+    parking_lot::RwLock as PlRwLock,
+    solana_sdk::{
+        clock::{Epoch, Slot},
+        epoch_schedule::EpochSchedule,
+    },
+    std::{
+        collections::HashMap,
+        sync::{Arc, RwLock as StdRwLock},
+        thread,
+        time::Duration,
+    },
+};
+
+const EPOCH_STAKES_QUERY_INTERVAL: Duration = Duration::from_secs(60);
+
+struct State {
+    stakes: HashMap<Epoch, EpochStakes>,
+    epoch_schedule: EpochSchedule,
+}
+
+/// Updates the epoch stakes state from the bank forks.
+///
+/// If new state was found, returns the corresponding root epoch and the state.
+fn update_state(bank_forks: &Arc<StdRwLock<BankForks>>, epoch: Epoch) -> Option<(Epoch, State)> {
+    let root_bank = bank_forks.read().unwrap().root_bank();
+    let root_epoch = root_bank.epoch();
+    (root_epoch > epoch).then(|| {
+        (
+            root_epoch,
+            State {
+                stakes: root_bank.epoch_stakes_map().clone(),
+                epoch_schedule: root_bank.epoch_schedule().clone(),
+            },
+        )
+    })
+}
+
+/// A service that regularly updates the epoch stakes state from the bank forks
+/// and exposes various methods to access the state.
+pub struct EpochStakesService {
+    state: Arc<PlRwLock<State>>,
+}
+
+impl EpochStakesService {
+    pub fn new(bank_forks: Arc<StdRwLock<BankForks>>) -> Self {
+        let (mut last_update_epoch, state) = update_state(&bank_forks, Epoch::default()).unwrap();
+        let state = Arc::new(PlRwLock::new(state));
+
+        {
+            let state = state.clone();
+            thread::spawn(move || loop {
+                thread::sleep(EPOCH_STAKES_QUERY_INTERVAL);
+                if let Some((new_update_epoch, update)) =
+                    update_state(&bank_forks, last_update_epoch)
+                {
+                    last_update_epoch = new_update_epoch;
+                    *state.write() = update;
+                }
+            });
+        }
+        Self { state }
+    }
+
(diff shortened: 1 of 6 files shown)


runtime/src/epoch_stakes_service.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/// A service that regularly updates the epoch stakes state from the bank forks
+/// and exposes various methods to access the state.
+pub struct EpochStakesService {
+    state: Arc<PlRwLock<State>>,
+}
+
+impl EpochStakesService {
+    pub fn new(bank_forks: Arc<StdRwLock<BankForks>>) -> Self {
```
AshwinSekar: nit: Prefer `RootBankCache` if the only access of `BankForks` is to get the `root_bank()`

runtime/src/epoch_stakes_service.rs
```diff
@@ -0,0 +1,80 @@
...
+    },
+    std::{
+        collections::HashMap,
+        sync::{Arc, RwLock as StdRwLock},
+        thread,
+        time::Duration,
+    },
+};
+
+const EPOCH_STAKES_QUERY_INTERVAL: Duration = Duration::from_secs(60);
```
AshwinSekar: Would prefer to derive this value based on `root_bank.get_slots_in_epoch`.
In tests we can modify the length of epochs, (minimum of 32 slots = 12.8 seconds), so we could end up serving stale state.

AshwinSekar: more unsolicited suggestions, we already detect when the `root_bank` moves to a new epoch in `bank_forks` here https://github.com/anza-xyz/alpenglow/blob/5b5c49fb47e2bf40bf66b8e5d5082ad37e3f506a/runtime/src/bank_forks.rs#L438

Similar to sending an AbsRequest, we could plumb a notification to `epoch_stakes_service` with the new `epoch_stakes`, then we don't have to worry about a query interval and can wait on the notification.

wen-coding: Agreed. If you are one of the very few people tracking root bank changes you can be more aggressive instead of sleeping, like plumb a notification when the EpochStakes changes in root.

runtime/src/epoch_stakes_service.rs
akhi3030: Nice, updated.

runtime/src/epoch_stakes_service.rs
akhi3030: I like this idea.  Just to make sure I understand correctly, the suggestion is that I create a new channel, I pass the sender to `do_set_root_return_metrics` and the receiver to `EpochStakesService`.  Then whenever a new epoch is detected, `do_set_root_return_metrics` will send a message to `EpochStakesService`?

AshwinSekar: An alternative to passing the sender in to the root fn and having to update the callgraph, you could register a crossbeam receiver as a subscriber for "root bank to new epoch" notifications and notify them. 

If you check out `BankForks::register_dumped_slot_subscriber` you can see an example of how we do it for dumped slots (it uses a mutex, but we could just pass in a crossbeam receiver instead).

This gives us some flexibility in the future if we want to add more listeners on rooting in a new epoch.

akhi3030: introduced the channel now.  PTAL.

runtime/src/epoch_stakes_service.rs
```diff
@@ -0,0 +1,72 @@
...
+        let mut prev_epoch = Epoch::default();
+        let state = Arc::new(PlRwLock::new(State {
+            stakes: HashMap::new(),
+            epoch_schedule: EpochSchedule::default(),
+        }));
+
+        {
+            let state = state.clone();
+            thread::spawn(move || loop {
+                let () = new_epoch_receiver.recv().unwrap();
```
AshwinSekar: if the receiver disconnects we should exit the loop, otherwise we won't be able to shut down the validator cleanly

runtime/src/bank_forks.rs
```diff
@@ -329,6 +332,24 @@ impl BankForks {
...
+    /// Register a new subscriber interested in hearing about new epochs.
+    pub fn register_new_epoch_subscriber(&mut self, tx: Sender<()>) {
+        self.new_epoch_subscribers.push(tx);
+    }
+
+    /// Call to notify subscribers of new epochs.
+    fn notify_new_epoch_subscribers(&mut self) {
+        let mut channels_to_drop = vec![];
+        for (ind, tx) in self.new_epoch_subscribers.iter().enumerate() {
+            if let Err(SendError(())) = tx.send(()) {
```
AshwinSekar: I think we can just send `root_bank : Arc<Bank>` the the subscriber so they don't have to perform an additional lookup.

runtime/src/epoch_stakes_service.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+/// A service that regularly updates the epoch stakes state from the bank forks
+/// and exposes various methods to access the state.
+pub struct EpochStakesService {
+    state: Arc<PlRwLock<State>>,
+}
+
+impl EpochStakesService {
+    pub fn new(new_bank_receiver: Receiver<Arc<Bank>>) -> Self {
```
AshwinSekar: lets take in the current root bank here to initialize with, just so we can serve data until the first epoch rolls over.

akhi3030: good idea.  Now added.

AshwinSekar (approved):

PR #307, opened 2025-07-22
Add CertificatepoolServiceStats.

wen-coding:


votor/src/certificate_pool_service.rs
```diff
@@ -201,10 +224,12 @@ impl CertificatePoolService {
                 &cert_pool,
                 &mut ctx,
                 &mut events,
+                &mut stats,
             );
 
             if standstill_timer.elapsed() > STANDSTILL_TIMEOUT {
                 events.push(VotorEvent::Standstill(highest_finalized_slot));
+                CertificatePoolServiceStats::incr_u16(&mut stats.standstill_reset);
```
AshwinSekar: We don't need a separate metric for `standstill_reset` since it resets everytime we finalize a block which the `new_finalized_slot` metric reports.
Can call this just `standstill` and only update it here.

votor/src/certificate_pool_service.rs
```diff
@@ -276,6 +304,7 @@ impl CertificatePoolService {
             return;
         }
         *highest_parent_ready = new_highest_parent_ready;
+        CertificatePoolServiceStats::incr_u16(&mut stats.new_parent_ready);
```
AshwinSekar: we should increment this on line 294, right now it will only notify on new highest parent ready.

votor/src/certificate_pool_service.rs
```diff
@@ -304,6 +333,7 @@ impl CertificatePoolService {
                     skipping production of {start_slot}-{end_slot}",
                     ctx.my_pubkey,
                 );
+                CertificatePoolServiceStats::incr_u16(&mut stats.new_parent_ready_missed_window);
```
AshwinSekar:
```suggestion
                CertificatePoolServiceStats::incr_u16(&mut stats.parent_ready_missed_window);
```

votor/src/certificate_pool_service/stats.rs
```diff
@@ -0,0 +1,88 @@
...
+pub(crate) struct CertificatePoolServiceStats {
+    pub(crate) add_message_failed: u32,
+    pub(crate) certificates_sent: u16,
+    pub(crate) certificates_dropped: u16,
+    pub(crate) new_finalized_slot: u16,
+    pub(crate) new_parent_ready: u16,
+    pub(crate) new_parent_ready_missed_window: u16,
+    pub(crate) new_root: u16,
+    pub(crate) received_votes: u32,
+    pub(crate) received_certificates: u32,
```
AshwinSekar, line 18: we should also add some stats for everytime we insert a new certificate (to cover both receiving and constructing) for each type by plugging this into `certificate_pool.rs`.

votor/src/certificate_pool_service.rs
wen-coding: Hmm, I'm confused. I think on this line you pointed out we also reset standstill timer without a new finalized slot, this is purely based on timer expiration?

AshwinSekar: you reset the standstill timer if either:
1) We see a new finalized slot
2) We reach 10 seconds and notify the event loop

(1) is not interesting and is covered by the `new_finalized_slot` metric. (2) is very interesting and should be put in a metric either here or in the event loop.

(1) + (2) together in one metric doesn't tell us anything. Does that make sense?

votor/src/certificate_pool_service/stats.rs
wen-coding: That's a good point, but I think this datapoint_info! line already has quite some data. We should probably give certificate_pool its own stats and datapoint_info! lines, so we can count cert construction and cert construction errors as well. That should be another PR though.

votor/src/certificate_pool_service.rs
wen-coding: Sure, in that case I think I shouldn't increment standstill in case 1) at all, changed.

votor/src/certificate_pool_service.rs
```diff
@@ -320,7 +348,8 @@ impl CertificatePoolService {
                     parent_block,
                     // TODO: we can just remove this
                     skip_timer: Instant::now(),
-                }))
+                }));
+                CertificatePoolServiceStats::incr_u16(&mut stats.parent_ready);
```
AshwinSekar:
```suggestion
                CertificatePoolServiceStats::incr_u16(&mut stats.parent_ready_produce_window);
```

AshwinSekar (commented):
Looks good thanks! Just one more rename

votor/src/certificate_pool_service/stats.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub(crate) add_message_failed: u32,
+    pub(crate) certificates_sent: u16,
+    pub(crate) certificates_dropped: u16,
+    pub(crate) new_finalized_slot: u16,
+    pub(crate) new_root: u16,
+    pub(crate) parent_ready: u16,
+    pub(crate) parent_ready_missed_window: u16,
+    pub(crate) received_votes: u32,
+    pub(crate) received_certificates: u32,
+    pub(crate) standstill_expired: bool,
```
AshwinSekar: lets just call this standstill, expired makes it sound like we have completed standstill

wen-coding: Done.

PR #310, opened 2025-07-23
wip

akhi3030:
#### Summary of Changes

Fixes #


runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+    // Ensures that the initial and updated state in the service is as expected.
+    #[test]
+    fn verify_initial_and_updated_state() {
+        let n_key_pairs = 10;
+        let bank = get_bank(n_key_pairs);
+        let epoch = bank.epoch();
+        let slot = bank.slot();
+        let (tx, rx) = unbounded();
+        let service = Arc::new(EpochStakesService::new(bank.clone(), epoch, rx));
+        let map = service.get_key_to_rank_map(slot).unwrap();
```
bw-solana, line 109: nit: `s/map/key_to_rank_map`

runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+            1_000_000_000,
+            &validator_keypairs,
+            stakes_vec,
+        );
+        Arc::new(Bank::new_for_tests(&genesis.genesis_config))
+    }
+
+    // Ensures that the initial and updated state in the service is as expected.
+    #[test]
+    fn verify_initial_and_updated_state() {
```
bw-solana, line 102: nit: maybe `future_epoch` instead of `updated`?

runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+        // Get a new bank sufficiently far in the future to ensure that the epoch has changed.
+        let new_bank = Arc::new(Bank::warp_from_parent(
+            bank,
+            &Pubkey::default(),
+            DEFAULT_SLOTS_PER_EPOCH * 2,
+            solana_accounts_db::accounts_db::CalcAccountsHashDataSource::IndexForTests,
+        ));
+
+        tx.send(new_bank.clone()).unwrap();
+        sleep(Duration::from_millis(10));
```
bw-solana, line 121: 10ms seems like an eternity, but I've seen weird flakiness with these types of timeouts when we run a bunch of these cargo tests in parallel (like in CI).

Maybe we can clone the receiver and spin until empty (with some timeout check and test fail condition)?

runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+        thread::sleep,
+    };
+
+    fn get_bank(n_key_pairs: usize) -> Arc<Bank> {
+        let validator_keypairs = (0..n_key_pairs)
+            .map(|_| ValidatorVoteKeypairs::new_rand())
+            .collect::<Vec<_>>();
+        let stakes_vec = (0..validator_keypairs.len())
+            .map(|i| 1_000 - i as u64)
+            .collect::<Vec<_>>();
```
ksn6, line 91:
```suggestion
        let stakes_vec = (0..validator_keypairs.len())
            .map(|i| validator_keypairs.len() - i as u64)
            .collect::<Vec<_>>();
```

runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+    fn verify_initial_and_updated_state() {
+        let n_key_pairs = 10;
+        let bank = get_bank(n_key_pairs);
+        let epoch = bank.epoch();
+        let slot = bank.slot();
+        let (tx, rx) = unbounded();
+        let service = Arc::new(EpochStakesService::new(bank.clone(), epoch, rx));
+        let map = service.get_key_to_rank_map(slot).unwrap();
+        assert_eq!(map.len(), n_key_pairs);
+
```
ksn6, line 111: Probably worth doing `assert!(service.get_key_to_rank_map(your_new_slot).is_none());` here.

runtime/src/epoch_stakes_service.rs
```diff
@@ -66,3 +66,62 @@ impl EpochStakesService {
...
+            DEFAULT_SLOTS_PER_EPOCH * 2,
+            solana_accounts_db::accounts_db::CalcAccountsHashDataSource::IndexForTests,
+        ));
+
+        tx.send(new_bank.clone()).unwrap();
+        sleep(Duration::from_millis(10));
+
+        let slot = new_bank.slot();
+        let map = service.get_key_to_rank_map(slot).unwrap();
+        assert_eq!(map.len(), n_key_pairs);
```
ksn6, line 125: Could we make this test stronger by asserting the value of the contents rather than just the length?

PR #311, opened 2025-07-23
Reject certificates older than root and add certificate pool stats

wen-coding:
Add CertificatePoolStats as promised in https://github.com/anza-xyz/alpenglow/pull/307/files#r2223736794

During the process, also realized we are not currently rejecting certificates older than root, which is a bug.

Fixed bug and added tests.

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -394,8 +394,6 @@ impl CertificatePool {
         message: &BLSMessage,
         events: &mut Vec<VotorEvent>,
     ) -> Result<(Option<Slot>, Vec<Arc<CertificateMessage>>), AddVoteError> {
-        // We add stats reporting here because we should almost always have a message.
-        self.stats.maybe_report();
         let current_highest_finalized_slot = self.highest_finalized_slot;
         let new_certficates_to_send = match message {
             BLSMessage::Vote(vote_message) => {
@@ -777,6 +775,10 @@ impl CertificatePool {
     /// not distinguished by the pubkey.
     pub fn update_pubkey(&mut self, new_pubkey: Pubkey) {
         self.parent_ready_tracker.update_pubkey(new_pubkey);
+    }
+
+    pub fn report_stats(&mut self) {
+        self.stats.report();
     }
 }
 
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -777,8 +777,8 @@ impl CertificatePool {
         self.parent_ready_tracker.update_pubkey(new_pubkey);
     }
 
-    pub fn report_stats(&mut self) {
-        self.stats.report();
+    pub fn maybe_report(&mut self) {
+        self.stats.maybe_report();
     }
 }
 
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -432,14 +432,14 @@ impl CertificatePool {
             "Validator stake is zero for pubkey: {validator_vote_key}"
         );
 
-        CertificatePoolStats::incr_u32_array(&mut self.stats.incoming, true);
+        self.stats.incoming_votes = self.stats.incoming_votes.saturating_add(1);
         if slot < self.root {
-            CertificatePoolStats::incr_u32_array(&mut self.stats.out_of_range, true);
+            self.stats.out_of_range_votes = self.stats.out_of_range_votes.saturating_add(1);
             return Err(AddVoteError::UnrootedSlot);
         }
         // We only allow votes
         if slot > self.root.saturating_add(MAX_SLOT_AGE) {
-            CertificatePoolStats::incr_u32_array(&mut self.stats.out_of_range, true);
+            self.stats.out_of_range_votes = self.stats.out_of_range_votes.saturating_add(1);
             return Err(AddVoteError::SlotInFuture);
         }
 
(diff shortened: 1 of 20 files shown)


votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    pub fn maybe_report(&mut self) {
+        if self.last_reported.elapsed() >= STATS_REPORT_INTERVAL {
+            self.report();
+            self.reset();
+        }
+    }
+
+    fn reset(&mut self) {
```
bw-solana: could just set to default here

votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+            ("out_of_range_votes", self.out_of_range[0] as i64, i64),
+            (
+                "out_of_range_certificates",
+                self.out_of_range[1] as i64,
+                i64
+            ),
+        );
+
+        datapoint_info!(
+            "certificate_pool_ingested_votes_by_type",
```
bw-solana: could probably drop the `_by_type` postfix for these metrics just to reduce verbosity

votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub fn incr_u32(value: &mut u32) {
+        *value = value.saturating_add(1);
+    }
+
+    pub fn incr_ingested_vote_type(&mut self, vote_type: VoteType) {
+        let index = vote_type as usize;
+
+        if self.ingested_votes_by_type.len() <= index {
+            self.ingested_votes_by_type
+                .resize(index.saturating_add(1), 0);
```
bw-solana: don't love this.. maybe we could use something like `strum` to help derive number of enums and set the Vec size properly up front?

votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+const STATS_REPORT_INTERVAL: Duration = Duration::from_secs(10);
+
+#[derive(Debug)]
+pub(crate) struct CertificatePoolStats {
+    pub(crate) conflicting_votes: u32,
+    pub(crate) event_safe_to_skip: u32,
+    pub(crate) event_safe_to_notarize: u32,
+
+    pub(crate) exits: [u32; 2],
+    pub(crate) incoming: [u32; 2],
```
bw-solana: probably easier to just have `incoming_votes` and `incoming_certs` fields

votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+            exits: [0, 0],
+            incoming: [0, 0],
+            out_of_range: [0, 0],
+            new_certs_by_type: [Vec::new(), Vec::new()],
+            ingested_votes_by_type: Vec::new(),
+            last_reported: Instant::now(),
+        }
+    }
+
+    pub fn incr_u32_array(value: &mut [u32; 2], is_vote: bool) {
```
bw-solana: I'm assuming we could get rid of this if we get rid of the arrays

votor/src/certificate_pool/stats.rs
```diff
@@ -0,0 +1,249 @@
...
+        if is_vote {
+            value[0] = value[0].saturating_add(1);
+        } else {
+            value[1] = value[1].saturating_add(1);
+        }
+    }
+
+    pub fn incr_u32(value: &mut u32) {
```
bw-solana: Do we really need this? I think we can just call `self.value = self.value.saturating_add(1);` directly

votor/src/certificate_pool/stats.rs
wen-coding: Done.

votor/src/certificate_pool/stats.rs
wen-coding: Done.

votor/src/certificate_pool/stats.rs
wen-coding: Actually, since the type of Vote/Cert in Alpenglow will not change that fast, I can use the `last_constant + 1`. If we ever change this, we should see panic in tests because I was using `array[index]`. Does that work?

votor/src/certificate_pool/stats.rs
wen-coding: Done.

votor/src/certificate_pool/stats.rs
wen-coding: Removed.

votor/src/certificate_pool/stats.rs
wen-coding: I was just being lazy, removed.

votor/src/certificate_pool/stats.rs
bw-solana: sounds okay to me

votor/src/certificate_pool.rs
```diff
@@ -387,6 +394,8 @@ impl CertificatePool {
         message: &BLSMessage,
         events: &mut Vec<VotorEvent>,
     ) -> Result<(Option<Slot>, Vec<Arc<CertificateMessage>>), AddVoteError> {
+        // We add stats reporting here because we should almost always have a message.
+        self.stats.maybe_report();
```
bw-solana: I'm thinking this might fit better higher up the stack (like in `certificate_pool_ingest` main loop). If we move it up there, we could even make this a sub-component of the `CertificatePoolServiceStats` reporting

wen-coding: Hmm I don't think the two stats should be mixed together, they each maintain the status of the corresponding data structure.

But I can move the trigger there. Done.

votor/src/certificate_pool_service.rs
```diff
@@ -275,7 +275,10 @@ impl CertificatePoolService {
                     return Ok(());
                 }
             }
-            stats.maybe_report();
+            if stats.maybe_report() {
+                // Stats were reported, report the certificate pool stats
+                cert_pool.report_stats();
```
bw-solana: if we're going to decouple the service and pool stats, might as well fully decouple them and just call something like this in succession:
```
stats.maybe_report();
cert_pool.maybe_report();
```

wen-coding: Done.

bw-solana (approved):

Issue #294, opened 2025-07-19
Move vote and BLS types to a module under `votor/src`


AshwinSekar:
We want to retire the vote program repo. Move all useful types back to votor.

wen-coding:
Those are now under votor-messages.

PR #315, opened 2025-07-25
refactor: remove unused `alpenglow-vote` items

ksn6:
#### Problem
Remove unused items under `vote/src/alpenglow`.

Fixes https://github.com/anza-xyz/alpenglow/issues/294.

### DIFF
--- a/votor/src/vote_history.rs
+++ b/votor/src/vote_history.rs
@@ -44,7 +44,7 @@ impl VoteHistoryVersions {
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiExample),
-    frozen_abi(digest = "4weZDM6uKNGzT7SgQMqFrqXCuZ5uG9WJg2ZHRCJBTdZu")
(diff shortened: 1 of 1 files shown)


vote/src/alpenglow/accounting.rs
```diff
@@ -109,158 +97,3 @@ impl EpochCredit {
         self.prev_credits = PodU64::from(prev_credits);
     }
 }
-
-/// Authorize the given pubkey to withdraw or sign votes. This may be called multiple times,
```
wen-coding, line 113: Nice. But can we get rid of credits as well?

wen-coding (approved):

vote/src/alpenglow/accounting.rs
ksn6: I've removed dependencies on `EpochCredits` in `alpenglow`, although I'd rather keep the `EpochCredits` field for now. Removing this field would mean that I'd have to also refactor / remove the field in `alpenglow-vote` to ensure that sizes all match.

wen-coding: That's fine, we can always change it later.

PR #316, opened 2025-07-25
refactor: remove alpenglow from gossip

ksn6:
#### Problem
We need to remove Alpenglow transaction logic from gossip.

#### Summary of Changes
Remove Alpenglow transaction logic from gossip.

### DIFF
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -4186,7 +4186,7 @@ fn run_duplicate_shreds_broadcast_leader(vote_on_duplicate: bool) {
                     );
                     gossip_vote_index += 1;
                     gossip_vote_index %= MAX_VOTES;
-                    cluster_info.push_vote_at_index(vote_tx, gossip_vote_index, false);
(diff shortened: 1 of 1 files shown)


core/src/cluster_info_vote_listener.rs
```diff
@@ -301,8 +301,9 @@ impl ClusterInfoVoteListener {
                 !packet_batch[0].meta().discard()
             })
             .filter_map(|(tx, packet_batch)| {
-                let (vote_account_key, vote, ..) = vote_parser::parse_vote_transaction(&tx)
-                    .or_else(|| vote_parser::parse_alpenglow_vote_transaction(&tx))?;
+                // let (vote_account_key, vote, ..) = vote_parser::parse_vote_transaction(&tx)
+                //     .or_else(|| vote_parser::parse_alpenglow_vote_transaction(&tx))?;
```
wen-coding: We can just remove line 304~305?

ksn6: yep pushed this in `cleanup` above

wen-coding (approved):

PR #319, opened 2025-07-26
Handle commitment sender disconnect and unify exit code in CertificatePoolService.

wen-coding:


votor/src/event_handler.rs
```diff
@@ -401,7 +401,7 @@ impl EventHandler {
             false,
             voting_context,
         ));
-        alpenglow_update_commitment_cache(
+        let _ = alpenglow_update_commitment_cache(
```
carllin, line 404: whats the purpose of adding an error if we just ignore it here?

wen-coding: This function is called from both `EventHandler `and `CertificatePoolService`. This PR only handles the error on the `CertificatePoolService` side, the `EventHandler` error handling can be done in a future PR maybe? I suppose `EventHandler` should also exit on disconnect.

bw-solana (approved):
LGTM. Would be good to hook up to `EventHandler` in the future

carllin (approved):

PR #320, opened 2025-07-26
refactor: minor cleanups

akhi3030:
- Use `let Some` bindings instead of checking for `Some` and then `unwrap`ing.
- Remove a function that performs a trivial computation and associated tests.


bw-solana (approved):
LGTM

PR #322, opened 2025-07-26
wip

akhi3030:
#### Summary of Changes

Fixes #


core/src/validator.rs
```diff
@@ -1329,10 +1330,17 @@ impl Validator {
             Some(stats_reporter_sender.clone()),
             exit.clone(),
         );
+        let root_bank_cache = {
+            let (tx, rx) = unbounded();
+            bank_forks.write().unwrap().register_new_bank_subscriber(tx);
```
wen-coding, line 1335: I do like the push model much much more than the pull model. At least, you will not suddenly acquire a bank_forks read lock several call levels deep.

Hmm, it seems to me `notify_new_bank_subscriber` is unfortunately a wrong name, it actually notifies about root bank whenever the `epoch_stakes` in the new root bank changes. That's enough for `EpochStakes`, but we do need `root_slot` as well in certificate pool and BLSSigverifier at least, otherwise we risk wasting resources on a lot of old votes and certs.

If we are guaranteed to only have one root bank cache globally, can we be a bit more aggressive and gets notified of every new root bank in `root_utils::set_root`? @AshwinSekar @bw-solana what do you think?

bw-solana: Agree - need to update on every new root bank. Per epoch is going to cause major problems.

I'm also curious if all of the consumers can handle eventually consistent (w/ respect to `bank_forks`) root. Spot checking, it seems like most of these are for cleaning up data as the root advances, which should be tolerant to slightly slower updates, but we should analyze each use.

I think it would also be good to quantify the root bank access pattern. My intuition is that we are extremely read heavy on the root bank and thus this scheme to reduce lock contention would be beneficial, but it would be nice to quantify under load (and stash this empirical data in a comment if we pursue this idea).

Also, if we pursue this idea, I would love to have some type that wraps the `Arc`/`RwLock`. Then we could hide `read`/`unwrap` and such in the APIs and it would be a lot easier to add logging later if needed for debugging. We wish we did this with `bank_forks` before it spiraled out of control

PR #326, opened 2025-07-28
Remove MAX_SLOT_AGE check.

wen-coding:
Remove MAX_SLOT_AGE check.

It was added because we were afraid someone will send a lot of future votes to arbitrarily increase certificate pool memory usage, but it's in the way of single validator rejoining cluster.

We will only receive votes/certs up to two Epochs ahead, because any future slot without EpochStakes will be automatically dropped.

With 80% non-byzantine nodes, the amount of damage can be done by sending future votes is limited, and the attackers should be easily filterable.

Remove this check for now.

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -7,6 +7,7 @@ use {
             vote_certificate_builder::{CertificateError, VoteCertificateBuilder},
             vote_pool::{DuplicateBlockVotePool, SimpleVotePool, VotePool, VotePoolType},
         },
+        commitment::AlpenglowCommitmentError,
         conflicting_types,
         event::VotorEvent,
         vote_to_certificate_ids, Block, CertificateId, Stake, VoteType,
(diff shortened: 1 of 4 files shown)


AshwinSekar (approved):
LGTM, @carllin for sanity check since I believe you suggested the MAX_SLOT_AGE change initially.

This range check doesn't really buy us anything, as attackers could spam votes in range anyway. I think it's best to defer to the network layer to rate limit or disconnect from attackers with unusual traffic patterns.

carllin (approved):

wen-coding:
> the old limit was 512, the 2 epoch limit is 432k to 864k w/ mainnet config. Would it be safer to just increase the limit to something really large like 20k?
> 
> Agree that network packet throttling should help mitigate the damage as well, but I'm assuming we'll allow a rate significantly more than what would be expected during typical operation, which opens the door to abuse.
> 
> Of course.. I think it would still be really hard to OOM validators with this attack, so not sure there's really much of an incentive... so I'm fine either way

Hmm, 20k is possible for a borked validator restarting (some validator operator sleeping when the validator is dead). With 80% non-byzantine validators, I think the amount of damage future votes can do is limited. Let's just remove this check for now, should be very easy to add back if it ever becomes a real attack.

PR #327, opened 2025-07-29
block_creation: add blockstore check for leader blocks

### LINKED ISSUE #290: Add blockstore check when checking if it is our turn to start our leader window
Totally missed this earlier 😅 , when checking the leader schedule we use the wrong function which doesn't additionally check if we already have a shred present in blockstore.
https://github.com/anza-xyz/alpenglow/blob/fbb25b3014fe9744cf79db9c6c82dc747e14ab6f/votor/src/certificate_pool_service.rs#L216-L220.

Either update this to use the blockstore variant, or add checking downstream.

AshwinSekar:
#### Problem
We do not check blockstore before instruction the block creation loop to produce window. Although broadcast will save us from sending duplicate shreds, we should catch it before we get there.

#### Summary of Changes

Fixes #290


votor/src/certificate_pool_service.rs
```diff
@@ -358,6 +357,14 @@ impl CertificatePoolService {
         let start_slot = *highest_parent_ready;
         let end_slot = last_of_consecutive_leader_slots(start_slot);
 
+        if (start_slot..=end_slot).any(|s| ctx.blockstore.has_existing_shreds_for_slot(s)) {
```
AshwinSekar, line 360: So in Poh/TowerBFT land we cache our next leader slot:
https://github.com/anza-xyz/alpenglow/blob/e84f02b3adcdc535a28d76db7f21c10d689a7a30/core/src/replay_stage.rs#L3127

This will check blockstore to see if the leader slot already has shreds in blockstore, and if so skip it:
https://github.com/anza-xyz/alpenglow/blob/e84f02b3adcdc535a28d76db7f21c10d689a7a30/ledger/src/leader_schedule_cache.rs#L143-L145


However we again check blockstore in poh recorder to account for the wild poh reset case we saw some years back https://github.com/solana-labs/solana/pull/32679:
https://github.com/anza-xyz/alpenglow/blob/e84f02b3adcdc535a28d76db7f21c10d689a7a30/poh/src/poh_recorder.rs#L879-L888

This PR maintains existing behavior by checking all 4 leader slots for shreds, however I don't believe this is necessary. In steady state we will never attempt to produce a block inside a leader window without producing the first slot of the leader window. 

The only corner cases are something like WFSM restart, or if a leader restarts in the middle of it's leader window.
In both cases this block can't get finalized anyway, because tryNotar doesn't allow you to vote for a block intra window without voting on its parent. @carllin lmk if you see any problems with this reasoning, if not I think we can just check for shreds in the first leader slot instead.

carllin: > however I don't believe this is necessary. In steady state we will never attempt to produce a block inside a leader window without producing the first slot of the leader window.

Agreed this is true, but also the corner case is still annoying because it could cause duplicate blocks on the network, which  would get ignored but triggers an annoying sequence of handler events. 

I don't think its that much extra work to check all 4 slots anyways, and gives operators logs to figure out why their slot got skipped

carllin (approved):

PR #329, opened 2025-07-29
Turn off all to all during partition test

wen-coding:
Add alpenglow_port_override, this can arbitrarily override alpenglow port for validator B in validator A during tests.

We can do arbitrary interesting things in the future:
1. Turn off all to all during partition test, just give random address as blackholes and all votes/certs will be dropped on the floor
2. Eavesdropping on any validator pair (we can replace additional_listeners probably)
3. Arbitrarily intercept communication from A to B and replace it with votes/certs we want


core/src/staked_validators_cache.rs
```diff
@@ -45,6 +45,12 @@ pub struct StakedValidatorsCache {
 
     /// Whether to include the running validator's socket address in cache entries
     include_self: bool,
+
+    /// Optional override for Alpenglow port, used for testing purposes
+    alpenglow_port_override: Option<Arc<RwLock<HashMap<Pubkey, SocketAddr>>>>,
+
+    /// size of alpenglow_port_override, used to determine if we need to refresh the cache
+    alpenglow_port_override_size: usize,
```
bw-solana: these probably deserve their own type. `AlpenglowPortOverride` or something. Then we can just have this new type here and hide some of the implementation details

core/src/staked_validators_cache.rs
```diff
@@ -132,14 +147,23 @@ impl StakedValidatorsCache {
...
+
             if let Some(alpenglow_socket) = node.alpenglow_socket {
-                alpenglow_sockets.push(alpenglow_socket);
+                let socket = alpenglow_port_override
+                    .as_ref()
+                    .and_then(|m| m.get(&node.pubkey))
+                    .cloned()
+                    .unwrap_or(alpenglow_socket);
+
+                alpenglow_sockets.push(socket);
```
bw-solana: might be nice to have a mode like
```
if Some(alpenglow_port_override) {
    alpenglow override API that checks each socket for override and such
} else {
    quick iterate, check, and push
}
```

core/src/staked_validators_cache.rs
```diff
@@ -170,6 +194,20 @@ impl StakedValidatorsCache {
...
+            let alpenglow_port_override_size = alpenglow_port_override.read().unwrap().len();
+            if alpenglow_port_override_size != self.alpenglow_port_override_size {
+                self.alpenglow_port_override_size = alpenglow_port_override_size;
+                trace!(
+                        "refreshing cache entry for epoch {} due to alpenglow port override size change",
+                        self.cur_epoch(slot)
+                    );
+                self.refresh_cache_entry(self.cur_epoch(slot), cluster_info, access_time);
+            }
+        }
```
bw-solana: this seems a little dicey. What if we add 1 alpenglow override and remove another?

Maybe we should either:

1. Force update the staked validators cache when modifying alpenglow overrides
2. Set a flag (probably atomic bool within our new Alpenglow overrides struct) that indicates we need to update cache entries

core/src/voting_service.rs
```diff
@@ -99,8 +107,17 @@ impl VotingService {
...
     ) -> Self {
+        let (additional_listeners, alpenglow_port_override) =
+            if let Some(test_override) = test_override {
+                (
+                    Some(test_override.additional_listeners),
+                    Some(test_override.alpenglow_port_override),
+                )
+            } else {
+                (None, None)
+            };
```
bw-solana: this might be a little cleaner:
```
let (additional_listeners, alpenglow_port_override) = test_override
    .map(|t| (Some(t.additional_listeners), Some(t.alpenglow_port_override)))
    .unwrap_or((None, None));
```

local-cluster/src/integration_tests.rs
```diff
@@ -504,12 +516,25 @@ pub fn run_cluster_partition<C>(
     info!("PARTITION_TEST start partition");
     on_partition_start(&mut cluster, &mut context);
     turbine_disabled.store(true, Ordering::Relaxed);
+    // Make all to all votes/certs not able to reach each other by overriding the
+    // alpenglow port override to SocketAddr which no one is listening on.
+    let blackhole_addr = solana_net_utils::bind_to_localhost()
```
bw-solana: might be good to add a real blockhole sentinel `SocketAddr` here

core/src/staked_validators_cache.rs
bw-solana: I'm not sure if we would want `Option<AlpenglowPortOverride>` or hide the `Option` within.. either can probably work.

This probably just impacts how we setup all the tests and such as far as if we pass `None` or `AlpenglowPortOverride::default()` which is effectively set to `None`. We could also have `AlpenglowPortOverride::new()` For setting up empty Option/Arc/Lock/Hashmap

core/src/staked_validators_cache.rs
wen-coding: I'll add a timestamp into `AlpenglowPortOverride`. Since we only have one writer, this should be enough.

It's hard to force update staked validator cache from local-cluster tests.

core/src/staked_validators_cache.rs
wen-coding: Changed to Option<AlpenglowPortOverride> here. It's a bit obvious we don't have an override at all in non-test cases.

core/src/staked_validators_cache.rs
wen-coding: Done.

core/src/voting_service.rs
wen-coding: Done.

local-cluster/src/integration_tests.rs
wen-coding: Done.

carllin:
Think these hacks are fine for now, but it would be nice to design a better partitioning mechanism in local cluster so we don't have to do this for every single component @ksn6. For instance, I think we mentioned looking to partition off the TPU or gossip at one point as well

core/src/staked_validators_cache.rs
```diff
@@ -45,6 +46,12 @@ pub struct StakedValidatorsCache {
 
     /// Whether to include the running validator's socket address in cache entries
     include_self: bool,
+
+    /// Optional override for Alpenglow port, used for testing purposes
+    alpenglow_port_override: Option<AlpenglowPortOverride>,
+
+    /// timestamp of the last alpenglow port override
+    alpenglow_port_override_timestamp: Instant,
```
bw-solana: you can tuck this in the `AlpenglowPortOverride` type now

core/src/voting_service.rs
```diff
@@ -89,6 +91,49 @@ pub struct VotingService {
     thread_hdl: JoinHandle<()>,
 }
 
+#[derive(Clone)]
+pub struct AlpenglowPortOverride {
+    override_map_with_timetamp: Arc<RwLock<(HashMap<Pubkey, SocketAddr>, Instant)>>,
```
bw-solana: what does the timestamp buy us that a bool doesn't?
Might be easier to reason about with a separe `update_pending` `AtomicBool`

core/src/staked_validators_cache.rs
```diff
@@ -170,6 +187,19 @@ impl StakedValidatorsCache {
         cluster_info: &ClusterInfo,
         access_time: Instant,
     ) -> (&[SocketAddr], bool) {
+        // Check if self.alpenglow_port_override has a different size.
```
bw-solana: comment needs update

core/src/staked_validators_cache.rs
```diff
@@ -170,6 +187,19 @@ impl StakedValidatorsCache {
...
+        // Immediately refresh the cache if it does.
+        if let Some(alpenglow_port_override) = &self.alpenglow_port_override {
+            if alpenglow_port_override.has_new_override(self.alpenglow_port_override_timestamp) {
+                self.alpenglow_port_override_timestamp = alpenglow_port_override.timestamp();
+                trace!(
+                        "refreshing cache entry for epoch {} due to alpenglow port override size change",
```
bw-solana: same here

local-cluster/src/integration_tests.rs
```diff
@@ -504,12 +517,25 @@ pub fn run_cluster_partition<C>(
...
+            .map(|node| (*node.pubkey(), blackhole_addr)),
+    );
+    alpenglow_port_override.update_override(new_override);
     sleep(partition_duration);
 
     on_before_partition_resolved(&mut cluster, &mut context);
     info!("PARTITION_TEST remove partition");
     turbine_disabled.store(false, Ordering::Relaxed);
+    // Restore the alpenglow port override to the default, so that the nodes can communicate again.
+    alpenglow_port_override.update_override(HashMap::new());
```
bw-solana: might be nice to have a `clear` method or something that can call `alpenglow_port_override.update_override(HashMap::new());` internally. Just in case we ever wanted to change the type

bw-solana (commented):
Looking pretty good. Left a few more suggestions

core/src/voting_service.rs
wen-coding: If we have multiple readers then each one need to cache their own timestamp to know when to pull value.

It works under one writer/one reader case, but:
1. The reader needs to reset update_pending to false by grabbing a writer lock, this is a bit messy
2. Putting update_pending into a different lock is actually more error prone, because putting them in two variables means that you are making assumptions on writer speed:
```
override_map: Arc<RwLock<HashMap<Pubkey, SocketAddr>>>,
update_pending: Arc<AtomicBool>,

```
This means if the writer does:
1. update `override_map`
2. update `update_pending` to `true`
And the reader does
1. read `update_pending`
2. read `override_map` if value in 1 is `true`
3. set `update_pending` to `false`
Then if the writer sends map1 and then map2, it can happen in this sequence
1. writer stores `map1` to `override_map`
2. writer stores `true` to `update_pending`
3. reader reads `update_pending`
4. reader reads `map1` from `override_map`
5. writer stores `map2` to `override_map`
6. writer stores `true` to `update_pending`
7. reader sets `update_pending` to `false`

Then map2 is never read by reader. Generally I believe it's recommended practice to put things related into the same lock.

Did I correctly guess what you had in mind?

core/src/staked_validators_cache.rs
```diff
@@ -761,6 +798,7 @@ mod tests {
             Duration::from_secs(5),
             5,
             false,
+            None,
```
ksn6, line 811: Could we add in a test using the new fields introduced?

core/src/voting_service.rs
ksn6: Could we add in a comment that explains this data structure?

bw-solana: ahhh, I didn't realize we might have multiple instances of `StakedValidatorsCache` that all need to make their independent updates. Thanks for clarifying. In this case, I'm cool with the timestamp approach here.

Also agree that having the map and timestamp under the same lock simplifies things. It might be a little cleaner to structure things something like this:
```
pub type AlpenglowPortOverride = Arc<RwLock<AlpenglowPortOverrideInner>>;

struct AlpenglowPortOverrideInner {
    override_map: HashMap<Pubkey, SocketAddr>,
    timestamp: Instant,
}
```

carllin:
Could be per component, but would be nice for the underlying implementation to just work on the specific ports that need to be sectioned off. No need to dig into the validator everytime to plumb these custom hacks

core/src/voting_service.rs
wen-coding: Nah I think we would only have one `StakedValidatorsCache`. But the bad case I mentioned above will also happen for one reader/one writer, as long as the writer writes fast enough (or the reader reads slow enough).

core/src/staked_validators_cache.rs
wen-coding: I need to cache a timestamp locally so I can compare with the one in RwLock to see whether I need to refresh. Any suggestions here?

core/src/voting_service.rs
bw-solana: If we're pretty sure that we only need one `StakedValidatorsCache`, then I would probably do something like this:

1. struct as described here: https://github.com/anza-xyz/alpenglow/pull/329/files#r2243900625 except use `pending_update: bool` instead of `timestamp: Instant`
2. tests can grab write lock and insert overrides. They set the `pending_update` to `true` when they're done
3. reader grabs read lock and checks `pending_update`. If it's `true`, it calls `refresh_cache_entry`
4. `refresh_cache_entry` grabs the write lock, updates all the entries, sets `pending_update` to `false`. Note this would require restructuring `refresh_cache_entry` a bit to grab the lock before the node loop and just hold it (don't grab read locks for each individual node)

bw-solana: That said, I'm okay keeping a `last_modified` timestamp in `AlpenglowPortOverrideInner` and `last_updated` timestamp in `StakedValidatorsCache`, similar to what you have here

core/src/staked_validators_cache.rs
bw-solana: I'm okay with what you have here.

If we want to tuck it inside `AlpenglowPortOverride`, then we don't need two timestamps (see other comment below)

core/src/voting_service.rs
wen-coding: That's definitely doable, but it's more invasive to refresh_cache_entry, and it's just a bit weird a reader of the data structure will grab write lock, at this point we should probably switch to channel instead....

I'll stick with timestamp for now until we feel it's too ugly.

core/src/staked_validators_cache.rs
wen-coding: Done.

core/src/staked_validators_cache.rs
wen-coding: Done.

local-cluster/src/integration_tests.rs
wen-coding: Done.

core/src/staked_validators_cache.rs
wen-coding: Done.

core/src/voting_service.rs
wen-coding: Done.

core/src/staked_validators_cache.rs
```diff
@@ -132,14 +144,19 @@ impl StakedValidatorsCache {
 
         let mut validator_sockets = Vec::new();
         let mut alpenglow_sockets = Vec::new();
-
         for node in nodes {
             validator_sockets.push(node.tpu_socket);
+
             if let Some(alpenglow_socket) = node.alpenglow_socket {
-                alpenglow_sockets.push(alpenglow_socket);
+                let socket = if let Some(alpenglow_port_override) = &self.alpenglow_port_override {
```
bw-solana: Sorry, I did a bad job at explaining my previous desire. I was actually thinking this check could live on the outside of the for loop (so we just call it once) and setup a closure so that it would look something like:
```rust
let push_alpenglow_socket = if let Some(alpenglow_port_override) = &self.alpenglow_port_override {
    let alpenglow_port_override_rl = alpenglow_port_override.read().unwrap();
    |pubkey, socket| {alpenglow_sockets.push(alpenglow_port_override_rl.get_overridden_socket(pubkey, socket));}
} else {
    |_, socket| {alpenglow_sockets.push(alpenglow_socket);}
}

for node in nodes {
    validator_sockets.push(node.tpu_socket);
    if let Some(alpenglow_socket) = node.alpenglow_socket {
        push_alpenglow_socket(&node.pubkey, &alpenglow_socket);
    }
}
```
^I'm sure the above code is broken, but hopefully gets the idea across. I think what you have is more readable, but we make the same check for every node and also acquire the read lock repeatedly. Given this is just for testing and number of nodes is likely to be small, maybe it's fine.

bw-solana (approved):
I'm okay w/ what you have.

We could probably streamline things a bit in `refresh_cache_entry` to avoid redundant checks and grabbing the read lock for every node (while not polluting the function w/ test code too much), but I've given you enough grief for one PR 🤣

core/src/staked_validators_cache.rs
wen-coding: I see. If we only want one read lock every time, we have several options:
1. make AlpenglowPortOverrideInner visible to users, so we can actually do what you described above
2. Just copy out the whole HashMap locally, so you can do local lookup
3. Send locallist of Vec<(PubKey, SocketAddr)> to AlpenglowPortOverride and get back a Vec<SocketAddr>

Given this is only used in tests and the maps are all tiny, the performance should all be similar. What do you feel?

bw-solana: Yeah, I agree that performance shouldn't be a problem with small N. I don't have a strong preference

wen-coding: Sounds good, I submitted a change to copy out the whole HashMap for now.

PR #332, opened 2025-07-31
improve ci

yihau:
#### Problem

our ci is very close to the timeout.

#### Summary of Changes

- split it into 3
- exclude solana-cargo-build-sbf since we won't touch the logic in Alpenglow. (any errors will still be caught when submitting a PR back to Agave)


wen-coding (approved):

yihau:
I guess the red `certificate_pool::tests::test_get_certs_for_standstill` isn't related to this PR 🙈

PR #333, opened 2025-07-31
Refactor safe_to_notar and safe_to_skip into SlotStakeCounters.

wen-coding:
We used to check safe_to_notar and safe_to_skip on every vote, there are two problems:
1. Too many unnecessary checks, we don't need to check them until at least I voted or we got new Notarize/Skip votes
2. We repeatedly send out the same event upstream, these events should only be sent once

Added SlotStakeCounters to update internal state on only corresponding votes.

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -487,10 +487,12 @@ impl CertificatePool {
                 return Ok(vec![]);
             }
             Some(entry_stake) => {
-                let fallback_vote_counters = self.slot_stake_counters_map.entry(slot).or_default();
+                let fallback_vote_counters = self
+                    .slot_stake_counters_map
+                    .entry(slot)
+                    .or_insert_with(|| SlotStakeCounters::new(total_stake));
                 fallback_vote_counters.add_vote(
                     vote,
-                    total_stake,
                     entry_stake,
                     my_vote_pubkey == &validator_vote_key,
                     events,
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -87,6 +87,105 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
     }
 }
 
+#[derive(Debug, Default)]
+struct FallbackVoteCounters {
+    my_first_vote: Option<Vote>,
+    total_stake: Stake,
+    skip_total: Stake,
+    notarize_total: Stake,
+    notarize_entry_total: BTreeMap<Hash, Stake>,
+    top_notarized_stake: Stake,
+    safe_to_notar_sent: Vec<Hash>,
+    safe_to_skip_sent: bool,
+}
+
+impl FallbackVoteCounters {
+    pub fn add_vote(
+        &mut self,
+        vote: &Vote,
+        total_stake: Stake,
+        entry_stake: Stake,
+        is_my_own_vote: bool,
+        events: &mut Vec<VotorEvent>,
+    ) -> (usize, bool) {
+        match vote {
+            Vote::Skip(_) => self.skip_total = entry_stake,
+            Vote::Notarize(vote) => {
+                self.notarize_entry_total
+                    .insert(*vote.block_id(), entry_stake);
+                self.notarize_total = self.notarize_entry_total.values().sum();
+                self.top_notarized_stake = self.top_notarized_stake.max(entry_stake);
+            }
+            _ => return (0, false), // Not interested in other vote types
+        }
+        if self.my_first_vote.is_none() && is_my_own_vote {
+            self.my_first_vote = Some(*vote);
+            self.total_stake = total_stake;
+        }
+        if self.my_first_vote.is_some() {
+            let slot = vote.slot();
+            let old_safe_to_notar_sent = self.safe_to_notar_sent.len();
+            let mut new_safe_to_skip = false;
+            // Check safe to notar
+            for (block_id, stake) in &self.notarize_entry_total {
+                if !self.safe_to_notar_sent.contains(block_id)
+                    && self.is_safe_to_notar(block_id, stake)
+                {
+                    events.push(VotorEvent::SafeToNotar((slot, *block_id)));
+                    self.safe_to_notar_sent.push(*block_id);
+                }
+            }
+            // Check safe to skip
+            if !self.safe_to_skip_sent && self.is_safe_to_skip() {
+                events.push(VotorEvent::SafeToSkip(slot));
+                new_safe_to_skip = true;
+                self.safe_to_skip_sent = true;
+            }
+            (
+                self.safe_to_notar_sent
+                    .len()
+                    .saturating_sub(old_safe_to_notar_sent),
+                new_safe_to_skip,
+            )
+        } else {
+            (0, false) // No new safe to notar or skip
+        }
+    }
+
+    fn is_safe_to_notar(&self, block_id: &Hash, stake: &Stake) -> bool {
+        // White paper v1.1 page 22: The event is only issued if the node voted in slot s already,
+        // but not to notarize b. Moreover:
+        // notar(b) >= 40% or (skip(s) + notar(b) >= 60% and notar(b) >= 20%)
+        if let Some(Vote::Notarize(my_vote)) = self.my_first_vote.as_ref() {
+            if my_vote.block_id() == block_id {
+                return false; // I voted for the same block, no need to send NotarizeFallback
+            }
+        }
+        let skip_ratio = self.skip_total as f64 / self.total_stake as f64;
+        let notarized_ratio = *stake as f64 / self.total_stake as f64;
+        // Check if the block fits condition (i) 40% of stake holders voted notarize
+        notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY
+            // Check if the block fits condition (ii) 20% notarized, and 60% notarized or skip
+            || (notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP
+                && notarized_ratio + skip_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP)
+    }
+
+    fn is_safe_to_skip(&self) -> bool {
+        // White paper v1.1 page 22: The event is only issued if the node voted in slot s already,
+        // but not to skip s. Moreover:
+        // skip(s) + Sum of all notarize - (max in notarize(b)) >= 40%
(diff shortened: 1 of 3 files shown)


votor/src/certificate_pool.rs
```diff
@@ -87,6 +87,105 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
     }
 }
 
+#[derive(Debug, Default)]
+struct FallbackVoteCounters {
```
AshwinSekar: lets move this to a new file in `votor/src/certificate_pool/`

votor/src/certificate_pool.rs
```diff
@@ -87,6 +87,105 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
...
+                self.notarize_total = self.notarize_entry_total.values().sum();
+                self.top_notarized_stake = self.top_notarized_stake.max(entry_stake);
+            }
+            _ => return (0, false), // Not interested in other vote types
+        }
+        if self.my_first_vote.is_none() && is_my_own_vote {
+            self.my_first_vote = Some(*vote);
+            self.total_stake = total_stake;
+        }
+        if self.my_first_vote.is_some() {
```
AshwinSekar: we can make this even more granular, e.g. only check `is_safe_to_skip` if `my_first_vote` was a notarize. Then we could also remove the check in the function.

I don't think we need to keep track of `my_first_vote: Vote` fully, could have an enum with `my_vote : Notarize | Skip | None`

votor/src/certificate_pool.rs
AshwinSekar: Also thinking we name it something more generic like `SlotStakeCounters`, so in the future we could just aggregate certificates using `notarize_entry_total` (and add a `finalize_total`, notarize fallback, skip fallback etc.) and then we can remove the per type vote pools.

votor/src/certificate_pool.rs
```diff
@@ -87,6 +87,105 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
...
+        entry_stake: Stake,
+        is_my_own_vote: bool,
+        events: &mut Vec<VotorEvent>,
+    ) -> (usize, bool) {
+        match vote {
+            Vote::Skip(_) => self.skip_total = entry_stake,
+            Vote::Notarize(vote) => {
+                self.notarize_entry_total
+                    .insert(*vote.block_id(), entry_stake);
+                self.notarize_total = self.notarize_entry_total.values().sum();
```
AshwinSekar: Can we just += entry_stake?

votor/src/certificate_pool.rs
```diff
@@ -87,6 +87,105 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
...
+
+impl FallbackVoteCounters {
+    pub fn add_vote(
+        &mut self,
+        vote: &Vote,
+        total_stake: Stake,
+        entry_stake: Stake,
+        is_my_own_vote: bool,
+        events: &mut Vec<VotorEvent>,
+    ) -> (usize, bool) {
```
AshwinSekar: would rather just pass in the mut stats struct here, so in the future we can increment more stats without polluting the return type.

votor/src/certificate_pool.rs

We can definitely move more totals into here, but hmm, not sure we can totally remove vote pools, we need to at least check for duplicate votes and conflicting votes, we can discuss when we try to do that.

votor/src/certificate_pool.rs
wen-coding: Err, we do need to check the block_id in Notarize in safe_to_notar. I don't think saving Vote instead of VoteType saves us that much any way.

I can move check on my_first_vote out, but safe_to_notar also need to check my_first_vote, so it's nice to keep the check here, and we can keep comments in the paper in one location. I'm leaning a bit towards checking inside the function, because I can put the comments in the function, but I can be convinced.

votor/src/certificate_pool.rs
wen-coding: We should make it += (entry_stake - old_entry_stake). The entry_stake here is really all stake voting the same (slot, block_id).

Sum seems straightforward, performance shouldn't matter much because we should only have a few entries. But I can change it to += (entry_stake - old_entry_stake) if you feel strongly about it.

votor/src/certificate_pool.rs

votor/src/certificate_pool/slot_stake_counters.rs
```diff
@@ -0,0 +1,301 @@
...
+                self.notarize_entry_total
+                    .insert(*vote.block_id(), entry_stake);
+                self.notarize_total = self.notarize_entry_total.values().sum();
+                self.top_notarized_stake = self.top_notarized_stake.max(entry_stake);
+            }
+            _ => return, // Not interested in other vote types
+        }
+        if self.my_first_vote.is_none() && is_my_own_vote {
+            self.my_first_vote = Some(*vote);
+            self.total_stake = total_stake;
```
carllin: can total stake be passed to this object on construction so we don't have to pass it every time

votor/src/certificate_pool/slot_stake_counters.rs
```diff
@@ -0,0 +1,301 @@
...
+            }
+        }
+    }
+
+    fn is_safe_to_notar(&self, block_id: &Hash, stake: &Stake) -> bool {
+        // White paper v1.1 page 22: The event is only issued if the node voted in slot s already,
+        // but not to notarize b. Moreover:
+        // notar(b) >= 40% or (skip(s) + notar(b) >= 60% and notar(b) >= 20%)
+        if let Some(Vote::Notarize(my_vote)) = self.my_first_vote.as_ref() {
+            if my_vote.block_id() == block_id {
```
carllin, line 85: Do we also want to avoid revoting on second and third blocks we have already voted on

votor/src/certificate_pool/slot_stake_counters.rs
```diff
@@ -0,0 +1,301 @@
...
+
+    fn is_safe_to_skip(&self) -> bool {
+        // White paper v1.1 page 22: The event is only issued if the node voted in slot s already,
+        // but not to skip s. Moreover:
+        // skip(s) + Sum of all notarize - (max in notarize(b)) >= 40%
+        if let Some(Vote::Notarize(_)) = self.my_first_vote.as_ref() {
+            error!(
+                "{} {} {}",
+                self.skip_total, self.notarize_total, self.top_notarized_stake
+            );
```
carllin: might be interesting to log the voted block id here as well

votor/src/certificate_pool/slot_stake_counters.rs
```diff
@@ -0,0 +1,301 @@
...
+    ) {
+        match vote {
+            Vote::Skip(_) => self.skip_total = entry_stake,
+            Vote::Notarize(vote) => {
+                self.notarize_entry_total
+                    .insert(*vote.block_id(), entry_stake);
+                self.notarize_total = self.notarize_entry_total.values().sum();
+                self.top_notarized_stake = self.top_notarized_stake.max(entry_stake);
+            }
+            _ => return, // Not interested in other vote types
```
carllin: I think we can separate this into two separate functions 

`add_skip` and `add_notarize` that call a common method which encapsulates the logic below so that the argument list isn't as big

votor/src/certificate_pool.rs
```diff
@@ -183,7 +186,7 @@ impl CertificatePool {
         transaction: &VoteMessage,
         validator_vote_key: &Pubkey,
         validator_stake: Stake,
-    ) -> bool {
+    ) -> Option<Stake> {
```
carllin, line 189: I think grouping the return value into an `entry_stake` can be a bit confusing in some cases, i.e. for skip votes it's the total_stake which is a bit confusing, whereas for the notarization votes its the entry stake

Maybe here we can return an 

```
enum NewVote {
      Skip(Stake),
      Notarize(Slot, BlockId, Stake),
}
```

And then match on this result to either call `add_notarize` or `add_skip` once we break up ` add_vote` https://github.com/anza-xyz/alpenglow/blob/69a256e12d885287cf3dbeef9dea64d250bad7d6/votor/src/certificate_pool/slot_stake_counters.rs#L25

votor/src/certificate_pool/slot_stake_counters.rs
wen-coding: Only Notarize vote counts here, there can only be one Notarize. (NotarizeFallback doesn't count here.)

votor/src/certificate_pool/slot_stake_counters.rs
wen-coding: Oops, this was for debugging, changed to trace!

votor/src/certificate_pool/slot_stake_counters.rs
wen-coding: Sure, done.

votor/src/certificate_pool.rs
wen-coding: Changed, but I don't like this change, the code is getting more complicated, see if I implemented what you imagined?

votor/src/certificate_pool/slot_stake_counters.rs
carllin: ah yeah, makes sense

votor/src/certificate_pool.rs
carllin: @wen-coding hmm yeah, seems like we can't avoid passing the `vote` either, feel free to revert back to the entry_stake and the separate vote, sorry for the randomizaiton

wen-coding: No problem. If in the future when we move more stuff into SlotStakeCounters we can always refactor this into an enum again as needed.

votor/src/certificate_pool/slot_stake_counters.rs
```diff
@@ -0,0 +1,303 @@
...
+        }
+        if self.my_first_vote.is_none() && is_my_own_vote {
+            self.my_first_vote = Some(*vote);
+        }
+        if self.my_first_vote.is_some() {
```
bw-solana: nit: could reduce nesting by early returning
```
if self.my_first_vote.is_none() {
    // Don't push votor events if I haven't voted yet
    return;
}
```

votor/src/certificate_pool.rs
bw-solana: sum feels less error prone, but I believe it will still cause us to build an iterator each time (even though there is 1 element). Probably not a big deal but 🤷‍♂️

votor/src/certificate_pool/slot_stake_counters.rs
wen-coding: Done.

votor/src/certificate_pool.rs

carllin (approved):

PR #343, opened 2025-08-02
Integrate upstream

bw-solana:
#### Summary of Changes

Fixes #


votor/src/voting_loop.rs
```diff
@@ -0,0 +1,865 @@
+//! The Alpenglow voting loop, handles all three types of votes as well as
```
AshwinSekar: this file should be deleted

Issue #286, opened 2025-07-19
Add metrics to the CertificatePoolService, SkipTimerService, and EventHandler


AshwinSekar:

AshwinSekar:
CertificatePoolService metrics are added in #307

PR #344, opened 2025-08-04
wip

akhi3030:
#### Summary of Changes

Fixes #


votor/src/timer_manager.rs
```diff
@@ -0,0 +1,48 @@
...
+pub(crate) struct TimerManager {
+    timers: Arc<RwLock<Timers>>,
+}
+
+impl TimerManager {
+    pub(crate) fn new(event_sender: Sender<VotorEvent>) -> Self {
+        let timers = Arc::new(RwLock::new(Timers::new(event_sender)));
+        {
+            let timers = Arc::clone(&timers);
+            thread::spawn(move || loop {
```
AshwinSekar: we'll want to pass the exit flag in so that this thread can shutdown properly

votor/src/event_handler.rs
```diff
@@ -206,13 +206,19 @@ impl EventHandler {
             // Received a parent ready notification for `slot`
             VotorEvent::ParentReady { slot, parent_block } => {
                 info!("{my_pubkey}: Parent ready {slot} {parent_block:?}");
+                assert_eq!(slot, first_of_consecutive_leader_slots(slot));
```
AshwinSekar: if we perform a full cluster restart from a block in the middle of a leader window we'll need the initial ParentReady to kick things off which will not be the first of the window.

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,156 @@
...
+            Self::WaitDeltaTimeout { window: _, timeout } => Some(*timeout),
+            Self::WaitDeltaBlock { window: _, timeout } => Some(*timeout),
+            Self::Done => None,
+        }
+    }
+}
+
+/// Maintains all active timer states for windows of slots.
+pub(super) struct Timers {
+    /// Timers are indexed by slots.  It is possible to have multiple active
```
AshwinSekar: I don't see the reason we need to have multiple active timers for a given slot, the first one firing should take care of any skip votes that need to be sent.

akhi3030: agreed and fixed.

turbine/src/retransmit_stage.rs
```diff
@@ -664,6 +671,9 @@ impl RetransmitStats {
                             rpc_subscriptions,
                             slot_status_notifier,
                         );
+                        votor_event_sender
```
AshwinSekar: I think it would make sense to move this into notify_subscribers to keep in line with the RPC / geyser FirstShred notifications

votor/src/event_handler.rs
```diff
@@ -203,16 +208,28 @@ impl EventHandler {
                 Self::try_final(my_pubkey, block, vctx, &mut votes);
             }
 
+            VotorEvent::FirstShred(slot) => {
```
AshwinSekar, line 211: Lets add a log here similar to the other events

votor/src/timer_manager.rs
```diff
@@ -0,0 +1,63 @@
...
+            event_sender,
+        )));
+        let handle = {
+            let timers = Arc::clone(&timers);
+            thread::spawn(move || {
+                while !exit.load(Ordering::Relaxed) {
+                    let duration = match timers.write().progress() {
+                        None => {
+                            // No active timers, sleep for some arbitrary amount.
+                            Duration::from_millis(100)
```
qkniep, line 46: I guess it's important that this number is smaller than the minimum time any newly added timer would take to expire, which it is, but maybe the comment should mention that

votor/src/event.rs
```diff
@@ -35,10 +35,16 @@ pub enum VotorEvent {
...
+    FirstShred(Slot),
+
     /// The pool has marked the given block as a ready parent for `slot`
     ParentReady { slot: Slot, parent_block: Block },
 
-    /// The skip timer has fired for the given slot
+    /// Timeout to inspect whether the leader has crashed.
+    TimeoutCrashedLeader(Slot),
+
+    /// Timeout to inspect whether the slot should be skipped.
```
qkniep: this timeout also triggers skipping the entire remaining leader window, not just a specific slot

votor/src/event.rs
```diff
@@ -35,10 +35,16 @@ pub enum VotorEvent {
...
     BlockNotarized(Block),
 
+    /// Received the first shred for the slot.
+    FirstShred(Slot),
+
     /// The pool has marked the given block as a ready parent for `slot`
     ParentReady { slot: Slot, parent_block: Block },
 
-    /// The skip timer has fired for the given slot
+    /// Timeout to inspect whether the leader has crashed.
```
qkniep: maybe make it clearer that this timeout is about *early detection* that is only possible for honest-but-crashed leaders

turbine/src/retransmit_stage.rs
```diff
@@ -664,6 +671,9 @@ impl RetransmitStats {
...
+                            .send(VotorEvent::FirstShred(slot))
```
qkniep: right now this is sent for every shred, right? this is by far the largest spam of `VotorEvent`s we have anywhere right now, so actually sending once might be good

qkniep: also maybe this should be in `retransmit()` or `retransmit_shred()` directly

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,287 @@
...
+                }
+
+                let mut events = Vec::with_capacity(2);
+                let slot = window.pop().unwrap();
+                events.push(VotorEvent::Timeout(slot));
+                if window.is_empty() {
+                    *self = Self::Done;
+                    return events;
+                }
+                events.push(VotorEvent::TimeoutCrashedLeader(*window.last().unwrap()));
```
qkniep: there is no reason to ever send both, or to send `TimeoutCrashedLeader` at any later point at all, since `Timeout` is the "stronger" one of the two

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,287 @@
...
+        assert!(!window.is_empty());
+        window.reverse();
+        let timeout = Instant::now().checked_add(delta_timeout).unwrap();
+        (Self::WaitDeltaTimeout { window, timeout }, timeout)
+    }
+
+    /// Call to make progress on the state machine.
+    ///
+    /// Returns a potentially empty list of events that should be sent.
+    fn progress(&mut self, delta_block: Duration) -> Vec<VotorEvent> {
```
qkniep: then (see below) this can also simply return `Option<VotorEvent>`

votor/src/timer_manager.rs
```diff
@@ -0,0 +1,63 @@
+//! Controls the queueing and firing of skip timer events for use
+//! in the event loop.
+//! TODO: Make this mockable in event_handler for tests
```
qkniep: nit: this should probably be a standard TODO comment:
```suggestion
// TODO: Make this mockable in event_handler for tests
```

votor/src/timer_manager.rs
```diff
@@ -0,0 +1,63 @@
...
+        },
+        thread::{self, JoinHandle},
+        time::{Duration, Instant},
+    },
+    timers::Timers,
+};
+
+/// A manager of timer states.  Uses a background thread to trigger next ready
+/// timers and send events.
+pub(crate) struct TimerManager {
```
qkniep, line 25: I didn't mind the name containing `Skip`, since both timeouts are for triggering skip votes, but I don't mind this name either...

votor/src/timer_manager/timers.rs
akhi3030: I was going based on https://github.com/qkniep/alpenglow/blob/7f099d5aa6a5044b3b7e9901b431eed3250ea11d/src/consensus/votor.rs#L240 where it seems that we send crashed leader for every slot in the window.  I can make the change here.  We should also update the other repo then?

akhi3030: I created https://github.com/qkniep/alpenglow/pull/70.  To make sure I understood the changes suggested here correctly.  Once https://github.com/qkniep/alpenglow/pull/70 is approved, I can make the appropriate changes here.

turbine/src/retransmit_stage.rs
AshwinSekar: This is only sent when we receive the first shred in the slot

votor/src/timer_manager/timers.rs
akhi3030: Should be fixed now.

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+        assert!(!window.is_empty());
+        window.reverse();
+        let timeout = Instant::now().checked_add(delta_timeout).unwrap();
+        (Self::WaitDeltaTimeout { window, timeout }, timeout)
+    }
+
+    /// Call to make progress on the state machine.
+    ///
+    /// Returns a potentially empty list of events that should be sent.
+    fn progress(&mut self, delta_block: Duration) -> Option<VotorEvent> {
```
akhi3030: Feedback from @bw-solana: we could use `ENTRY_COALESCE_DURATION` to actually give the leader enough time to put the first bytes of data on the network.  I'll inspect this further.

votor/src/timer_manager.rs
akhi3030: Thanks, I somehow found the renaming a bit more logical.

votor/src/timer_manager/timers.rs
qkniep: You mean as the additional time after `DELTA_TIMEOUT` before timing out for crashed leaders? Note, that this is/becomes a protocol parameter. We might want to keep `ENTRY_COALESCE_DURATION` client-specific. If I understand correctly what `ENTRY_COALESCE_DURATION` does, the time we set at the protocol level becomes an effective upper bound on what clients can choose for this parameter.

turbine/src/retransmit_stage.rs
akhi3030: I also think it might make sense to move this into `retransmit()` or `retransmit_shred()`.  However, AFAICT that seems to require a larger change especially if we do not want to rely on the stats struct to figure out when we have the first shred.  

Ultimately, this event should come from the blockstore maybe?  I'll keep it in `notify_subscribers` as suggested in the other comment for now.

votor/src/timer_manager/timers.rs
akhi3030: OK, I will not add this for now.  We can revisit later.

turbine/src/retransmit_stage.rs
AshwinSekar: It can come from blockstore, but it would delay the event as the blockstore thread does a lot more work to process shreds. Maybe not a huge deal depending on how tight the timing is.

turbine/src/retransmit_stage.rs
```diff
@@ -768,6 +777,9 @@ fn notify_subscribers(
             .unwrap()
             .notify_first_shred_received(slot);
     }
+    votor_event_sender
+        .send(VotorEvent::FirstShred(slot))
+        .unwrap();
```
AshwinSekar, line 782: This can be addressed in a follow up, but we should have proper handling for all event sending points. If this sender is disconnected we should panic as written here, but additionally if the queue is full we should at least output some error logs / metrics, would hate to block here and not know.

votor/src/event_handler.rs
```diff
@@ -203,16 +208,29 @@ impl EventHandler {
...
+                    timer_manager.write().set_timeouts(slot);
                 }
             }
 
+            VotorEvent::TimeoutCrashedLeader(slot) => {
+                info!("{my_pubkey}: TimeoutCrashedLeader {slot}");
+                if vctx.vote_history.voted(slot) || received_shred.contains(&slot) {
+                    return Ok(votes);
+                }
+                Self::try_skip_window(my_pubkey, slot, vctx, &mut votes);
```
AshwinSekar, line 231: A note for the future, as part of https://github.com/anza-xyz/alpenglow/issues/286 we should add some metrics for how often we skip_window based on crashed leader / timeouts

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+            delta_block,
+            timers: HashMap::new(),
+            heap: BinaryHeap::new(),
+            event_sender,
+        }
+    }
+
+    /// Call to set timeouts for a new window of slots.
+    pub(super) fn set_timeouts(&mut self, slot: Slot) {
+        assert_eq!(self.heap.len(), self.timers.len());
```
AshwinSekar, line 120: In a follow up we should also add metrics around the timers. Think it would be good to have visibility on the size of these structs.

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+        let mut ret_timeout = None;
+        let now = Instant::now();
+        loop {
+            assert_eq!(self.heap.len(), self.timers.len());
+            match self.heap.pop() {
+                None => break,
+                Some(Reverse((next_fire, slot))) => {
+                    if now < next_fire {
+                        ret_timeout = Some(match ret_timeout {
```
AshwinSekar: Nit could be cleaner:
```
ret_timeout = Some(ret_timeout.map_or(next_fire, |r| std::cmp::min(r, next_fire)));
```

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl TimerState {
+    /// Creates a new instance of the state machine.
+    ///
+    /// Also returns the next time the timer should fire.
+    fn new(slot: Slot, delta_timeout: Duration) -> (Self, Instant) {
+        let mut window = (slot..=last_of_consecutive_leader_slots(slot)).collect::<Vec<_>>();
+        assert!(!window.is_empty());
+        window.reverse();
```
AshwinSekar: Would it make more sense to use a VecDeque and pop_front so we don't have to store the window backwards?

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+                ret
+            }
+            Self::Done => None,
+        }
+    }
+
+    /// When would this state machine next be able to make progress.
+    fn next_fire(&self) -> Option<Instant> {
+        match self {
+            Self::WaitDeltaTimeout { window: _, timeout } => Some(*timeout),
```
AshwinSekar: nit you can combine the arms:
`Self::WaitDeltaTimeout { timeout, .. } | Self::WaitDeltaBlock { timeout, .. } => Some(*timeout),`

votor/src/timer_manager/timers.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    #[test]
+    fn timer_state_machine() {
+        let slot = 0;
+        let (mut timer_state, next_fire) = TimerState::new(slot, Duration::from_micros(1));
+
+        let next_fire = next_fire.duration_since(Instant::now());
+        let delta_block = Duration::from_micros(1);
+
+        thread::sleep(next_fire);
```
AshwinSekar: What do you think about having timer_state.progress and timers.progress take in an Instant that we pass in from the TimerManager. Then we can avoid having to sleep in test code using something like mock_instant. 

Don't feel too strongly about this since we're dealing with micro seconds here haha, just a thought if we want to do more complex orderings in future tests

votor/src/timer_manager/timers.rs
bw-solana: yes, `ENTRY_COALESCE_DURATION` is just client specific. I believe FD will actually wait forever (well.. an entire slot) before shredding entries.

As long as we think through the case of a completely idle cluster and make sure we don't skip those well behaved leaders, I'm okay with whatever we choose.

In the truly idle case, I believe an FD node will essentially wait 400ms (maybe 380ms right now?), pad out 32 data shreds worth of data, erasure code it, and then send it out

turbine/src/retransmit_stage.rs
akhi3030: agreed... we need to think hard about how to handle situations where the queues are full.  It suggests that we are not keeping up and we need to apply back pressure.  Not an easy problem.

votor/src/event_handler.rs
akhi3030: agreed.

votor/src/timer_manager/timers.rs
akhi3030: agreed

votor/src/timer_manager/timers.rs
akhi3030: thanks, fixed.

votor/src/timer_manager/timers.rs
akhi3030: good idea... it makes the code less reliant on the "outside".  I'll make the change.

votor/src/timer_manager/timers.rs
akhi3030: I don't like either of the solutions particularly too much.  As a reader if you think that is cleaner, I'll make the change.

AshwinSekar: I think it's more intuitive to read if we pop front, the window will be in the order we send timeout events

votor/src/timer_manager/timers.rs
qkniep: If we want to continue allowing the current FD behavior, we have to completely remove the early timeout for crashed leaders again.

Otherwise, I would suggest we do introduce a global max. coalesce time for the *first* slice only, that all clients need to adopt. This should be much smaller than the block time to be effective, so lower than current `ENTRY_COALESCE_DURATION`, e.g. 40ms.

For current parameters, the early crashed leader timeout could reduce the idle time caused by each crashed leader by: `DELTA_BLOCK / (DELTA_BLOCK + 3*DELTA = 400 / (400 + 3*250) = approx. 35%` (with 31% already saved when choosing 40ms as first slice coalesce time).

AshwinSekar (approved):
Implementation LGTM, approving pending resolution of the coalesce discussion

Issue #293, opened 2025-07-19
Add optimistic block creation


AshwinSekar:
`block_creation_loop.rs` doesn't start a leader window until it receives a `ParentReady`.

Instead we could also consume `VotorEvent::Block` here, and start our leader window early, potentially abandoning it if we do not receive the appropriate certificate.

In order to abandon properly we need to send the signal to broadcast stage so that it calls `finish_prev_slot`

AshwinSekar:
> Send a special slice as slice k+1 indicating that the parent changed to (new_parent_slot, new_parent_hash).

Good catch, we'll want to update `finish_prev_slot` to handle this special shred, and for replay to throw away current execution state for the slot if it encounters this special shred.

What should the leader do if it reaches the max block size before it receives the `ParentReady`? I'm assuming it should wait after producing some # of slices?

AshwinSekar:
Both of these suggestions make sense, we should have some sort of reset indicator at the Entry level.
For the first pass we use this to indicate that the block is empty (ignore all current state changes treat it as an empty bank).
Afterwards we can implement logic to handle actually restarting the bank, updating the relevant structures and rerouting the transactions back in.

qkniep:
I hope it's clear that this is not a long-term solution, because with this it definitely does not hold that "This comes at no downside to the leader, regardless of what the previous leader does." The previous leader can always cause an honest leader to lose their first block just by delaying their last block enough for it to not be notarized.

What is the parent info in the shreds currently used for? What can it even be used for? It only has the slot number (actually not even that, just the offset) and not the hash, right?

The simplest way to do this, seems to be to only have the parent info (incl. hash) once at the beginning of a block (and not use the parent slot in shreds at all).

bw-solana:
@ksn6 - Milestones look good to me. We'll also need to figure out if/how we reset parts of the cost model (e.g. CUs used and other cost dimensions) and which milestone that falls in

PR #345, opened 2025-08-04
feat(entry): update Entry with parent ready info for optimistic block packing

ksn6:
#### Problem
We'd like to introduce parent ready information to `Entry` for optimistic block production.

#### Summary of Changes
Add an `Option<(Slot, Hash)>` field to `Entry` and propagate changes throughout `alpenglow`.

See https://github.com/anza-xyz/alpenglow/issues/293 for details + discussion.

### DIFF
--- a/entry/src/entry.rs
+++ b/entry/src/entry.rs
@@ -141,6 +141,8 @@ pub struct Entry {
 
     /// Store additional information about the entry (e.g., updated ParentReady information for
     /// Alpenglow optimistic block production).
+    ///
+    /// TODO(karthik): serde(default) doesn't work here - why not?
     #[serde(skip_serializing_if = "Option::is_none")]
     pub special: Option<SpecialEntry>,
 }


entry/src/entry.rs
```diff
@@ -130,6 +131,19 @@ pub struct Entry {
...
+    /// is initially None, as we optimistically assume that a block that we receive but haven't yet
+    /// received a certificate for, will eventually be finalized. Most of the time, we expect this
+    /// to be the case. Occasionally, the disseminated block will be skipped - in this case, "parent
+    /// ready" in the Alpenglow algorithm will not be generated for the block that was just
+    /// disseminated, but rather to a previous block, at which point this field will be set to
+    /// Some((slot, hash)) where (slot, hash) refers to said previous block.
+    ///
+    /// TODO(karthik): should we instead turn Entry into an Enum with two variants: one for the
+    /// current Entry type and another for (Slot, Hash)?
+    pub updated_parent_block: Option<(Slot, Hash)>,
```
AshwinSekar: we'll want to throw in a  #[serde(default)] for compatibility.
I'd suggest making this an enum field, SpecialEntry, as we will have other special entries apart from the parent switching one (footer and maybe header) in the future

ksn6: Done

qkniep:
Where do we put the parent slot + hash in the first entry of the block? Which under optimistic block production would usually simply be the block first seen by the block producer. Shouldn't these two cases be encoded in the same place?

PR #346, opened 2025-08-05
Put discarded packets in stats separately.

wen-coding:

`Failed to deserialize BLS message: the size limit has been reached`


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -44,6 +44,12 @@ impl SigVerifier for BLSSigVerifier {
         for packet in packet_batches.iter().flatten() {
             stats_updater.received += 1;
 
+            if packet.meta().discard() {
```
ksn6, line 57: We have a test that verifies the accuracy of the stats capturer, though don't have a test that relies on this logic being here (i.e., the `continue` in particular). Could we either add in such a test or a TODO to write such a test in the future?

wen-coding: Hmm, I'm using assert_eq!(verifier.stats.sent, 0), also checking that receiver is empty now, is that enough?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -44,6 +44,12 @@ impl SigVerifier for BLSSigVerifier {
...
+                // This can happen if a cert is forwarded by multiple sources.
```
bw-solana, line 58: is there any other reason we would mark this packet for discard?

wen-coding: We could discard for any other verification error as well. In the future we could also mark discard for BLS verification failures, when that happens we would adjust the stats accordingly. @samkim-crypto

Issue #289, opened 2025-07-19
Plug in Set-Identity from the admin RPC and add handling in CertificatePoolService as well


AshwinSekar:
`VotorEvent::SetIdentity` is not being sent, update the admin RPC to send this event.

In `CertificatePoolService` after we receive a new `ParentReady` we check to see if we should notify the block creation loop to start our window. We do this by using the leader schedule and our pubkey. At this point we should see if our pubkey has been changed, and if so update it and any other relevant certificate pool / parent ready tracker info.

Issue #274, opened 2025-07-16
On startup, initialize the certificate pool with our own votes


AshwinSekar:
#### Problem
See #273 we will rely on the certificate pool to have our own votes.
This is now important for safe_to checks.
This has always been important for threshold checks.

#### Proposed Solution
Make sure we add our own votes on startup into the certificate pool, also probably a good idea to do this when we set-identity as well, for weird cases where the backup node has not received the votes from the original node.

In any case this might not be completely necessary as during standstill we will refresh our votes and then they will be inserted into the certificate pool, but it doesn't hurt

PR #348, opened 2025-08-05
log: add some additional logging in cert pool / local-cluster

AshwinSekar:
Adds some trace logs on certificate insert and refresh, and updates phase 1 logging for the local cluster test

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -89,6 +89,7 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
 
 #[derive(Default)]
 pub struct CertificatePool {
+    my_pubkey: Pubkey,
     // Vote pools to do bean counting for votes.
     vote_pools: BTreeMap<PoolId, VotePoolType>,
     /// Completed certificates
@@ -131,6 +132,7 @@ impl CertificatePool {
         let parent_ready_tracker = ParentReadyTracker::new(my_pubkey, root_block);
 
         let mut pool = Self {
+            my_pubkey,
             vote_pools: BTreeMap::new(),
             completed_certificates: BTreeMap::new(),
             highest_notarized_fallback: None,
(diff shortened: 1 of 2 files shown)


votor/src/certificate_pool.rs
```diff
@@ -696,8 +700,9 @@ impl CertificatePool {
     pub fn get_certs_for_standstill(&self) -> Vec<Arc<CertificateMessage>> {
         self.completed_certificates
             .iter()
-            .filter_map(|(_, cert)| {
+            .filter_map(|(cert_id, cert)| {
                 if Some(cert.certificate.slot()) >= self.highest_finalized_with_notarize {
+                    trace!("{}: Refreshing certificate {:?}", self.my_pubkey, cert_id);
```
wen-coding, line 705: This is for standstill, we can just say sending certificate for standstill?

votor/src/certificate_pool.rs
```diff
@@ -686,6 +689,7 @@ impl CertificatePool {
     /// This avoids the need to recreate the entire certificate pool since it's
     /// not distinguished by the pubkey.
     pub fn update_pubkey(&mut self, new_pubkey: Pubkey) {
+        self.my_pubkey = new_pubkey;
```
wen-coding, line 692: Not relevant for this PR: err, this is our set-id implementation currently?

Guess we need to update safe_to_notar and safe_to_skip when pubkey changes to filter self-votes from the pool?

AshwinSekar: yep, currently we don't handle set identity at all for the certificate pool, but we should plug it in at some point #289

votor/src/certificate_pool.rs
AshwinSekar: It just copies the same log as voting to rerfresh

wen-coding: I don't have strong preference here, just in case we have another cert refresh path in the future, then somehow put standstill in log will make the logic source easier to debug.

votor/src/certificate_pool.rs
wen-coding: When set-id is done, would vote history be updated to pull all the votes done by the same pubkey but by another instance from the cert pool?

AshwinSekar: Yeah, currently vote_history -> cert_pool is not implemented #274 
We should dump our own votes in when we startup and when we set identity

wen-coding (approved):

PR #349, opened 2025-08-05
Only return Notarize/Finalize/FinalizeFast cert for highest finalized slot in standstill, do not return votes for this slot

wen-coding:


votor/src/certificate_pool.rs
```diff
@@ -697,10 +698,15 @@ impl CertificatePool {
         self.completed_certificates
             .iter()
             .filter_map(|(_, cert)| {
-                if Some(cert.certificate.slot()) >= self.highest_finalized_with_notarize {
-                    Some(cert.clone())
-                } else {
-                    None
+                match Some(cert.certificate.slot()).cmp(&self.highest_finalized_with_notarize) {
+                    Ordering::Greater => Some(cert.clone()),
+                    Ordering::Equal => match cert.certificate.certificate_type() {
```
AshwinSekar: nit: Not a huge deal but if we have a Notarze, Finalize, & Fast finalize for the slot we should only send the Fast Finalize, this code will send all 3.

AshwinSekar (commented):
Do you mind also updating the >= to > in vote_history votes_cast_since?

votor/src/certificate_pool.rs
AshwinSekar: Maybe it would be easier if `highest_finalized_with_notarize` also stored the certificate and then we only filtered by > here?

wen-coding: The only special case is FastFinalize, with that we don't need Notarize and Finalize, so added a boolean to indicate the presence of FastFinalize.

votor/src/certificate_pool.rs
```diff
@@ -698,15 +700,35 @@ impl CertificatePool {
...
 
     pub fn get_certs_for_standstill(&self) -> Vec<Arc<CertificateMessage>> {
+        let (highest_finalized_with_notarize_slot, has_fast_finalize) = self
+            .highest_finalized_with_notarize
+            .unwrap_or((self.root, false));
         self.completed_certificates
             .iter()
             .filter_map(|(cert_id, cert)| {
-                if Some(cert.certificate.slot()) >= self.highest_finalized_with_notarize {
+                let cert_to_send = match cert
```
AshwinSekar: I think it would be cleaner if we avoided the nesting here with a single match statement, also seems easier to use `cert_id` instead of `cert` indirection
```rust
match (cert_id.slot().cmp(&highest_finalized_with_notarize_slot), cert_id.certificate_type, has_fast_finalize) => {
  (Ordering::Greater, _, _) 
  | (Ordering::Equal, Finalize | Notarize, false),
  | (Ordering::Equal, FastFinalize, true) => Some(cert.clone()),
  (Ordering::Equal, FastFinalize, false) => panic!("Programmer Error"),
  _ => None
}
```
Something like this?

I wouldn't panic here, we can totally get FastFinalize cert after highest_finalized_with_notarize_slot is set.

AshwinSekar: hmm but in that case we update the bool right?
```
 if self
                    .highest_finalized_with_notarize
                    .map_or(true, |(s, _)| s <= slot)
                {
                    self.highest_finalized_with_notarize = Some((slot, true));
                }
```
since you use a "<=" here?

votor/src/certificate_pool.rs
```diff
@@ -698,15 +700,32 @@ impl CertificatePool {
...
     pub fn get_certs_for_standstill(&self) -> Vec<Arc<CertificateMessage>> {
+        let (highest_finalized_with_notarize_slot, has_fast_finalize) = self
+            .highest_finalized_with_notarize
+            .unwrap_or((self.root, false));
         self.completed_certificates
             .iter()
             .filter_map(|(cert_id, cert)| {
-                if Some(cert.certificate.slot()) >= self.highest_finalized_with_notarize {
+                let cert_to_send = match (
+                    cert.certificate
```
AshwinSekar: nit: we can just do `cert_id.slot()`

votor/src/certificate_pool.rs
wen-coding: It only works when this whole thing is still single threaded. Sure.

votor/src/certificate_pool.rs

PR #350, opened 2025-08-05
refactor: rip out the epoch stakes service.

akhi3030:
We have decided to just use the root bank cache everywhere for now.


bw-solana (approved):
LGTM

PR #351, opened 2025-08-06
Handle discarded packets separately in stats

wen-coding:

Failed to deserialize BLS message: the size limit has been reached


bw-solana:
Closed due to #343 stomping master branch. Replaced by #352

PR #352, opened 2025-08-06
Put discarded packets in stats separately.

wen-coding:

Failed to deserialize BLS message: the size limit has been reached

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -414,8 +414,14 @@ mod tests {
         let packet_batches = vec![PinnedPacketBatch::new(packets).into()];
         assert!(verifier.send_packets(packet_batches).is_ok());
         assert_eq!(verifier.stats.sent, 0);
+        assert_eq!(verifier.stats.sent_failed, 0);
+        assert_eq!(verifier.stats.verified_votes_sent, 0);
+        assert_eq!(verifier.stats.verified_votes_sent_failed, 0);
         assert_eq!(verifier.stats.received, 1);
         assert_eq!(verifier.stats.received_discarded, 1);
+        assert_eq!(verifier.stats.received_malformed, 0);
+        assert_eq!(verifier.stats.received_no_epoch_stakes, 0);
+        assert_eq!(verifier.stats.received_votes, 0);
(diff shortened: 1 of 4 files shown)


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -54,6 +54,11 @@ impl SigVerifier for BLSSigVerifier {
         for packet in packet_batches.iter().flatten() {
             stats_updater.received += 1;
 
+            if packet.meta().discard() {
+                stats_updater.received_discarded += 1;
```
bw-solana, line 58: do we understand all the reasons a packet will be discarded before this? It would be really nice to understand if the packet was deduped vs. malformed or other sanitization reasons.

Don't necessarily need to tackle in this PR, but if we at least know the broad categories that feed into this, it would be nice to include in a comment in `StatsUpdater` struct def

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -389,4 +394,28 @@ mod tests {
...
+        packet
+            .populate_packet(None, &message)
+            .expect("Failed to populate packet");
+        packet.meta_mut().set_discard(true);
+        let packets = vec![packet];
+        let packet_batches = vec![PinnedPacketBatch::new(packets).into()];
+        assert!(verifier.send_packets(packet_batches).is_ok());
+        assert_eq!(verifier.stats.sent, 0);
+        assert_eq!(verifier.stats.received, 1);
+        assert_eq!(verifier.stats.received_discarded, 1);
```
bw-solana, line 421: should we verify everything in stats? (besides maybe `last_stats_logged`)

core/src/sigverifier/bls_sigverifier.rs
wen-coding: I believe the current places which set discard have their own stats, we are not adding new ones here, so I don't think that stat belongs here.

When we add new ones due to BLS rate limiting or verification failure we will add additional stats.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

PR #355, opened 2025-08-06
blockstore: add alternative shred and block versions columns

AshwinSekar:
#### Problem
As part of alpenglow, we need to be able to repair and store up to 3 alternate versions of a block separately
from the block that is received via turbine. We need to enable blockstore to ingest and store these blocks.

#### Summary of Changes
After discussion with Steve and Carl we have taken the approach that I believe will allow us to enable this with the smallest surface area:
- Add alternate versions of the columns relevant to shred ingest (SlotMeta, ErasureMeta, Index, ShredData, MerkleRootMeta)
- Add a BlockVersions column which keeps track of available complete versions in order to serve repair requests and facilitate replay / intrawindow safe to notar checks
- Future PR will modify the do_insert_shreds call path to accept a location to store these repaired shreds in the appropriate column
- Future PR will allow for replaying banks from an alternate column, and update the appropriate tracking in Bank/BankForks
- In the case where we have a block that we have to fetch via block id repair and it ends up getting rooted, we will move the relevant data from these alternate columns back into the original columns. This avoids having to update the full blockstore API,  RPC and any downstream tooling (e.g. ledger-tool). This situation should be rare during steady state, so we are fine eating the copy cost.

In V2 we can consider migrating from the original columns completely (Turbine blocks could be stored with Hash::default) and updating downstream dependencies

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -257,6 +257,7 @@ pub struct Blockstore {
     blocktime_cf: LedgerColumn<cf::Blocktime>,
     code_shred_cf: LedgerColumn<cf::ShredCode>,
     data_shred_cf: LedgerColumn<cf::ShredData>,
+    block_versions_cf: LedgerColumn<cf::BlockVersions>,
     dead_slots_cf: LedgerColumn<cf::DeadSlots>,
     duplicate_slots_cf: LedgerColumn<cf::DuplicateSlots>,
     erasure_meta_cf: LedgerColumn<cf::ErasureMeta>,
@@ -272,6 +273,11 @@ pub struct Blockstore {
     transaction_memos_cf: LedgerColumn<cf::TransactionMemos>,
     transaction_status_cf: LedgerColumn<cf::TransactionStatus>,
     transaction_status_index_cf: LedgerColumn<cf::TransactionStatusIndex>,
+    alt_meta_cf: LedgerColumn<cf::AlternateSlotMeta>,
+    alt_erasure_meta_cf: LedgerColumn<cf::AlternateErasureMeta>,
+    alt_index_cf: LedgerColumn<cf::AlternateIndex>,
+    alt_data_shred_cf: LedgerColumn<cf::AlternateShredData>,
+    alt_merkle_root_meta_cf: LedgerColumn<cf::AlternateMerkleRootMeta>,
 
     highest_primary_index_slot: RwLock<Option<Slot>>,
     max_root: AtomicU64,
@@ -405,6 +411,7 @@ impl Blockstore {
         let blocktime_cf = db.column();
         let code_shred_cf = db.column();
         let data_shred_cf = db.column();
+        let block_versions_cf = db.column();
         let dead_slots_cf = db.column();
         let duplicate_slots_cf = db.column();
         let erasure_meta_cf = db.column();
@@ -420,6 +427,11 @@ impl Blockstore {
         let transaction_memos_cf = db.column();
         let transaction_status_cf = db.column();
         let transaction_status_index_cf = db.column();
+        let alt_meta_cf = db.column();
+        let alt_erasure_meta_cf = db.column();
+        let alt_index_cf = db.column();
+        let alt_data_shred_cf = db.column();
+        let alt_merkle_root_meta_cf = db.column();
 
         // Get max root or 0 if it doesn't exist
         let max_root = roots_cf
@@ -440,6 +452,7 @@ impl Blockstore {
             blocktime_cf,
             code_shred_cf,
             data_shred_cf,
+            block_versions_cf,
             dead_slots_cf,
             duplicate_slots_cf,
             erasure_meta_cf,
@@ -455,6 +468,12 @@ impl Blockstore {
             transaction_memos_cf,
             transaction_status_cf,
             transaction_status_index_cf,
+            alt_meta_cf,
+            alt_erasure_meta_cf,
+            alt_index_cf,
+            alt_data_shred_cf,
+            alt_merkle_root_meta_cf,
+
             highest_primary_index_slot: RwLock::<Option<Slot>>::default(),
             new_shreds_signals: Mutex::default(),
             completed_slots_senders: Mutex::default(),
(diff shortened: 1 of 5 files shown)


ledger/src/blockstore/column.rs
```diff
@@ -677,25 +778,45 @@ impl TypedColumn for columns::Index {
...
-        // https://github.com/anza-xyz/agave/issues/3570
-        let index: bincode::Result<blockstore_meta::Index> = config.deserialize(data);
-        match index {
-            Ok(index) => Ok(index),
-            Err(_) => {
-                let index: blockstore_meta::IndexFallback = config.deserialize(data)?;
-                Ok(index.into())
-            }
-        }
+        blockstore_meta::deserialize_index(data)
```
AshwinSekar, line 781: @steviez not sure what the latest is on the slot meta and index column migrations, i've replicated the migration logic for the alternate columns, but if you prefer I can write the new columns in the V2 format.

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,51 @@ pub struct DuplicateSlotProof {
...
+pub enum BlockStatus {
+    /// The block is being ingested, of the shreds received there are no conflicts
+    Incomplete,
+
+    /// The block `block_id` has been fully ingested, has consistent shreds and is ready for replay
+    Complete { block_id: Hash },
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct BlockVersions {
```
AshwinSekar, line 428: @carllin I think storing this column is best rather than trying to rebuild it during startup and storing it in memory. 
Allows us to easily lookup from the appropriate column when serving block_id repair https://github.com/anza-xyz/alpenglow/pull/356/files#diff-683c825caa839726df6cf4a62f9fbc160e96d424fb26f400d9a790e464c3a61bR42

ledger/src/blockstore/column.rs
```diff
@@ -133,6 +163,26 @@ pub mod columns {
...
+    ///
+    /// Similar to [`ShredData`], however this column is only populated for alternate
+    /// versions fetched via block_id based repair.
+    ///
+    /// * index type: `(slot: u64, shred_index: u64, block_id: Hash)`
+    /// * value type: [`Vec<u8>`]
+    pub struct AlternateShredData;
+
+    #[derive(Debug)]
+    /// The block verisons
```
bw-solana: versions*

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,51 @@ pub struct DuplicateSlotProof {
...
+    pub turbine_version: BlockStatus,
+    pub repaired_versions: [(Hash, BlockStatus); 3],
```
bw-solana, line 435: are repaired versions mutually exclusive from the turbine version? Whatever the answer, it seems like a good thing to document here

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,54 @@ pub struct DuplicateSlotProof {
...
+    /// We are unable to recover the block with the shreds ingested
+    Conflicting,
+
+    /// The block `block_id` has been fully ingested, has consistent shreds and is ready for replay
+    Complete { block_id: Hash },
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct BlockVersions {
+    pub turbine_version: BlockStatus,
```
bw-solana, line 430: are we guaranteed to only have 1 turbine version? Is this because we will filter out conflicting shreds the don't match the first version received? Is this determined via CMS?

Edit: Reading more, it seems like it could be 1 version or it could be Conflicting state and be a mish-mash of versions

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,54 @@ pub struct DuplicateSlotProof {
     pub shred2: shred::Payload,
 }
 
+#[derive(Deserialize, Serialize)]
+pub enum BlockStatus {
+    /// The block is being ingested, of the shreds received there are no conflicts
+    Incomplete,
+
+    /// We are unable to recover the block with the shreds ingested
+    Conflicting,
```
bw-solana, line 421: Is the expectation that this only applies to turbine because repair will request a specific Block ID? Or can this state be hit for repaired slots transiently and trigger some shred dump/re-repair?

ledger/src/blockstore/column.rs
bw-solana: FYI @cpubot is the one that wrote the V2 columns + upgrade/downgrade handling

ledger/src/blockstore.rs
```diff
@@ -2363,6 +2441,29 @@ impl Blockstore {
             .collect()
     }
 
+    pub fn get_data_shred_from_location(
```
bw-solana, line 2444: How are you envisioning these two APIs (`get_data_shred_from_location` & `get_block_location`) being used?

I'm wondering if it makes sense to make `get_block_location` available as pub or if this should be kept as a blockstore implementation detail. For example, we could just take in `block_id` into `get_data_shred_from_location` and call into `get_block_location` under the hood.

But I don't know if this extra information about block location might actually be useful for repair/replay/etc.

ledger/src/blockstore_meta.rs
AshwinSekar: Exactly, the Turbine version is just 1 shred per index on whatever we received through Turbine. In most cases this will be a Consistent version, however if there is malicious activity we could end up getting a conflicting / mish mash of versions, at which point we might need to repair a version.

ledger/src/blockstore_meta.rs
AshwinSekar: If we get unlucky and request a repairs from malicious nodes, even a repaired version could end up with conflicting shreds. A follow up PR implements a resolver that will detect these conflicts and dump & re-repair when it happens.

ledger/src/blockstore.rs
AshwinSekar: I haven't fully implemented all the users of this api, but at least for repair there is some nuance here:
- We only want to serve block_id repair if we actually have the complete block, which is checked via `get_block_location`
- In TowerBFT repair, we serve the repair request regardless (only out of the Turbine column)

I'll keep it as pub for now and we can reevaluate once the implementation is complete

bw-solana: Sounds good. Can re-evaluate and shuffle as needed once we have the consumers hooked up

PR #356, opened 2025-08-06
repair: add new block_id based repair variants

AshwinSekar:
#### Problem
We need to be able to fetch alternate versions of a block and store them in the appropriate blockstore columns

#### Summary of Changes
Create a block id based version of orphan, highest shred, and shred, and plug in the responder.

Unlike TowerBFT the responder only fulfills these requests if it has the *complete* block for the shred(s) that are being requested.

### DIFF
--- a/core/src/repair/malicious_repair_handler.rs
+++ b/core/src/repair/malicious_repair_handler.rs
@@ -1,6 +1,7 @@
 use {
     super::{repair_handler::RepairHandler, repair_response::repair_response_packet_from_bytes},
     solana_clock::Slot,
+    solana_hash::Hash,
     solana_ledger::{
         blockstore::Blockstore,
         shred::{Nonce, SIZE_OF_DATA_SHRED_HEADERS},
@@ -36,13 +37,23 @@ impl RepairHandler for MaliciousRepairHandler {
         &self,
         slot: Slot,
         shred_index: u64,
+        block_id: Option<Hash>,
         dest: &SocketAddr,
         nonce: Nonce,
     ) -> Option<Packet> {
-        let mut shred = self
-            .blockstore
-            .get_data_shred(slot, shred_index)
-            .expect("Blockstore could not get data shred")?;
+        let mut shred = match block_id {
+            None => self.blockstore.get_data_shred(slot, shred_index),
+            Some(block_id) => {
+                let location = self
+                    .blockstore()
+                    .get_block_location(slot, block_id)
+                    .expect("Unable to fetch block location from blockstore")?;
+                self.blockstore()
+                    .get_data_shred_from_location(slot, shred_index, location)
+            }
+        }
+        .expect("Blockstore could not get data shred")?;
+
         if self
             .config
             .bad_shred_slot_frequency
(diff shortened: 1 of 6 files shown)


ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,51 @@ pub struct DuplicateSlotProof {
     pub shred2: shred::Payload,
 }
 
+#[derive(Deserialize, Serialize)]
+pub enum BlockStatus {
+    /// The block is being ingested, of the shreds received there are no conflicts
+    Incomplete,
+
+    /// The block `block_id` has been fully ingested, has consistent shreds and is ready for replay
+    Complete { block_id: Hash },
```
carllin: Do we need an in between status for incomplete but in conflict, or is that automatically dumped?

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,51 @@ pub struct DuplicateSlotProof {
...
+}
+
+#[derive(Deserialize, Serialize)]
+pub struct BlockVersions {
+    pub turbine_version: BlockStatus,
+    pub repaired_versions: [(Hash, BlockStatus); 3],
```
carllin: Tuples can be hard to track, would it be cleaner to augment the BlockStatus::Incomplete to also track the first hash it gets?

Can then implement a `BlockStatus::hash() -> Hash` function for convenience

ledger/src/blockstore_meta.rs
```diff
@@ -365,6 +412,51 @@ pub struct DuplicateSlotProof {
...
+}
+
+/// Which column an associated block currently resides
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum BlockLocation {
+    Turbine,
+    Repair { block_id: Hash },
```
carllin: nit: For clarity for column "location", should we rename to `Primary { block_id: Hash }` and `Duplicate { block_id: Hash }`

carllin (commented):
looks good, just nits

ledger/src/blockstore_meta.rs
qkniep: I don't think that works, by definition we can only know the block hash once it is complete

ledger/src/blockstore_meta.rs
qkniep: the blockstore should not care about this at some point consensus decides, at this point we could call a `cleanup()` function on blockstore to delete any other blocks in that slot

ledger/src/blockstore_meta.rs
qkniep: I think, as it is currently, it is clearer what it means, `Turbine` is just the block we receive via block dissemination (that's also why we don't know the `block_id` yet), whereas `Primary` implies that it is the correct block for that slot as decided by consensus, as soon as we know that, we would no longer have to store the `Duplicate`s

ledger/src/blockstore_meta.rs
AshwinSekar: yeah that block_id is fetched from the merkle root of the last FEC set, so we can't know it until we've received enough of the block.

carllin: Hmm, so what is stored in the first Hash of this tuple before we receive enough of the block, just Hash::default?

AshwinSekar: It's whatever is tagged onto the repair request. I might have been misunderstanding your initial question, were you suggesting that Incomplete also uses the hash from the repair request? Because then the Turbine column would have to use Hash::default or some different sentinel

PR #357, opened 2025-08-06
repair: populate a nonce <> BlockLocation lookup for use in shred insert

AshwinSekar:
#### Problem
Shreds received from block_id based requests need to be inserted into the appropriate blockstore columns, however ingest doesn't know anything about the shred response.

#### Summary of Changes
Similar to the OutstandingRequests cache, maintain a separate BlockLocationLookup.
After sigverify, lookup the location and tag it along the shred payload to window service.

### DIFF
--- a/core/src/repair/ancestor_hashes_service.rs
+++ b/core/src/repair/ancestor_hashes_service.rs
@@ -938,6 +938,7 @@ mod test {
         solana_runtime::bank_forks::BankForks,
         solana_signer::Signer,
         solana_streamer::socket::SocketAddrSpace,
+        solana_turbine::block_location_lookup::BlockLocationLookup,
         std::collections::HashMap,
         trees::tr,
     };
@@ -1403,8 +1404,7 @@ mod test {
                 repair_validators: None,
                 repair_whitelist,
                 wen_restart_repair_slots: None,
-                block_location_lookup:
-                    solana_turbine::block_location_lookup::BlockLocationLookup::new_arc(),
+                block_location_lookup: BlockLocationLookup::new_arc(),
             };
 
             let (ancestor_hashes_replay_update_sender, ancestor_hashes_replay_update_receiver) =
--- a/core/src/repair/ancestor_hashes_service.rs
+++ b/core/src/repair/ancestor_hashes_service.rs
@@ -930,15 +930,14 @@ mod test {
         solana_hash::Hash,
         solana_keypair::Keypair,
         solana_ledger::{
-            blockstore::make_many_slot_entries, get_tmp_ledger_path,
-            get_tmp_ledger_path_auto_delete, shred::Nonce,
+            block_location_lookup::BlockLocationLookup, blockstore::make_many_slot_entries,
+            get_tmp_ledger_path, get_tmp_ledger_path_auto_delete, shred::Nonce,
         },
         solana_net_utils::sockets::bind_to_localhost_unique,
         solana_perf::packet::Packet,
         solana_runtime::bank_forks::BankForks,
         solana_signer::Signer,
         solana_streamer::socket::SocketAddrSpace,
-        solana_turbine::block_location_lookup::BlockLocationLookup,
         std::collections::HashMap,
         trees::tr,
     };
--- a/core/src/repair/ancestor_hashes_service.rs
+++ b/core/src/repair/ancestor_hashes_service.rs
@@ -1403,6 +1403,8 @@ mod test {
                 repair_validators: None,
                 repair_whitelist,
                 wen_restart_repair_slots: None,
+                block_location_lookup:
+                    solana_turbine::block_location_lookup::BlockLocationLookup::new_arc(),
             };
 
             let (ancestor_hashes_replay_update_sender, ancestor_hashes_replay_update_receiver) =
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -2,8 +2,8 @@
 use {
     crate::repair::standard_repair_handler::StandardRepairHandler,
     crossbeam_channel::unbounded,
+    solana_ledger::block_location_lookup::BlockLocationLookup,
     solana_ledger::{blockstore::Blockstore, get_tmp_ledger_path_auto_delete},
-    solana_turbine::block_location_lookup::BlockLocationLookup,
 };
 use {
     crate::{
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -128,18 +128,14 @@ impl ShredRepairType {
         }
     }
 
-    pub fn block_location(&self) -> BlockLocation {
+    pub fn block_id(&self) -> Option<Hash> {
         match self {
-            // AG repair ingests into an alternate column
             ShredRepairType::OrphanForBlockId(_, bid)
             | ShredRepairType::HighestShredForBlockId(_, _, bid)
-            | ShredRepairType::ShredForBlockId(_, _, bid) => {
-                BlockLocation::Alternate { block_id: *bid }
-            }
-            // TowerBFT repair ingests into the original column populated by turbine
+            | ShredRepairType::ShredForBlockId(_, _, bid) => Some(*bid),
             ShredRepairType::Orphan(_)
             | ShredRepairType::HighestShred(_, _)
-            | ShredRepairType::Shred(_, _) => BlockLocation::Original,
+            | ShredRepairType::Shred(_, _) => None,
         }
     }
 }
@@ -1257,7 +1253,11 @@ impl ServeRepair {
         };
         let peer = repair_peers.sample(&mut rand::thread_rng());
         let nonce = outstanding_requests.add_request(repair_request, timestamp());
-        let location = repair_request.block_location();
+        let location = repair_request
+            .block_id()
+            .map_or(BlockLocation::Original, |block_id| {
+                BlockLocation::Alternate { block_id }
+            });
         repair_info
             .block_location_lookup
             .add_location(nonce, location);
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -1,7 +1,9 @@
 #[cfg(test)]
 use {
     crate::repair::standard_repair_handler::StandardRepairHandler,
+    crossbeam_channel::unbounded,
     solana_ledger::{blockstore::Blockstore, get_tmp_ledger_path_auto_delete},
+    solana_turbine::block_location_lookup::BlockLocationLookup,
 };
 use {
     crate::{
@@ -10,7 +12,7 @@ use {
             duplicate_repair_status::get_ancestor_hash_repair_sample_size,
             quic_endpoint::RemoteRequest,
             repair_handler::RepairHandler,
-            repair_service::{OutstandingShredRepairs, RepairStats, REPAIR_MS},
+            repair_service::{OutstandingShredRepairs, RepairInfo, RepairStats, REPAIR_MS},
             request_response::RequestResponse,
             result::{Error, RepairVerifyError, Result},
         },
@@ -33,7 +35,10 @@ use {
     },
     solana_hash::{Hash, HASH_BYTES},
     solana_keypair::{signable::Signable, Keypair},
-    solana_ledger::shred::{self, Nonce, ShredFetchStats, SIZE_OF_NONCE},
+    solana_ledger::{
+        blockstore_meta::BlockLocation,
+        shred::{self, Nonce, ShredFetchStats, SIZE_OF_NONCE},
+    },
     solana_packet::PACKET_DATA_SIZE,
     solana_perf::{
         data_budget::DataBudget,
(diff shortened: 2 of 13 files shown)


core/src/window_service.rs
```diff
@@ -191,7 +191,13 @@ fn run_check_duplicate(
 #[allow(clippy::too_many_arguments)]
 fn run_insert<F>(
```
AshwinSekar, line 193: next PR will actually insert it into the tagged column

core/src/repair/ancestor_hashes_service.rs
```diff
@@ -1403,6 +1403,8 @@ mod test {
                 repair_validators: None,
                 repair_whitelist,
                 wen_restart_repair_slots: None,
+                block_location_lookup:
+                    solana_turbine::block_location_lookup::BlockLocationLookup::new_arc(),
```
bw-solana: might be prettier to add `use solana_turbine::block_location_lookup::BlockLocationLookup`

turbine/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![allow(clippy::arithmetic_side_effects)]
 
 mod addr_cache;
+pub mod block_location_lookup;
```
bw-solana: feels a little funky putting this in turbine.. but I'm not sure where the right home is.. maybe blockstore?

turbine/src/sigverify_shreds.rs
```diff
@@ -148,8 +152,9 @@ fn run_shred_sigverify<const K: usize>(
     deduper: &Deduper<K, [u8]>,
     shred_fetch_receiver: &Receiver<PacketBatch>,
     retransmit_sender: &EvictingSender<Vec<shred::Payload>>,
-    verified_sender: &Sender<Vec<(shred::Payload, /*is_repaired:*/ bool)>>,
+    verified_sender: &Sender<Vec<(shred::Payload, /*is_repaired:*/ bool, BlockLocation)>>,
```
bw-solana, line 155: do we still need `is_repaired`? Could this be inferred from `BlockLocation`?

turbine/src/sigverify_shreds.rs
```diff
@@ -243,14 +248,19 @@ fn run_shred_sigverify<const K: usize>(
...
         })
-        .partition_map(|(shred, repair)| {
-            if repair {
+        .partition_map(|(shred, nonce)| {
+            if let Some(nonce) = nonce {
+                // Only block id repair needs to insert shreds in a special column,
+                // Regular repair will insert in the Turbine column
+                let location = block_location_lookup
+                    .get_location(nonce)
+                    .unwrap_or(BlockLocation::Turbine);
```
bw-solana: Seems like this should be fine (and hopefully LRU sized such that it doesn't happen often)... although it's a little strange that we could have shreds in the `repairs` vec w/ `BlockLocation::Turbine`

turbine/src/sigverify_shreds.rs
```diff
@@ -270,10 +280,10 @@ fn run_shred_sigverify<const K: usize>(
...
     let shreds = shreds
         .into_iter()
-        .map(|shred| (shred, /*is_repaired:*/ false));
-    let repairs = repairs
-        .into_iter()
-        .map(|shred| (shred, /*is_repaired:*/ true));
+        .map(|shred| (shred, /*is_repaired:*/ false, BlockLocation::Turbine));
+    let repairs = repairs.into_iter().map(|(shred, location)| {
+        (shred, /*is_repaired:*/ true, location)
+    });
```
bw-solana: could we just add the `true` for `is_repaired` above?

core/src/window_service.rs
```diff
@@ -191,7 +191,13 @@ fn run_check_duplicate(
...
     thread_pool: &ThreadPool,
-    verified_receiver: &Receiver<Vec<(shred::Payload, /*is_repaired:*/ bool)>>,
+    verified_receiver: &Receiver<
+        Vec<(
+            shred::Payload,
+            /*is_repaired:*/ bool,
+            solana_ledger::blockstore_meta::BlockLocation,
```
bw-solana: `use solana_ledger::blockstore_meta::BlockLocation` might be cleaner

core/src/repair/serve_repair.rs
```diff
@@ -122,6 +127,19 @@ impl ShredRepairType {
...
+    pub fn location_to_insert_response(&self) -> BlockLocation {
+        match self {
+            ShredRepairType::OrphanForBlockId(_, bid)
+            | ShredRepairType::HighestShredForBlockId(_, _, bid)
+            | ShredRepairType::ShredForBlockId(_, _, bid) => {
+                BlockLocation::Repair { block_id: *bid }
+            }
+            ShredRepairType::Orphan(_)
+            | ShredRepairType::HighestShred(_, _)
+            | ShredRepairType::Shred(_, _) => BlockLocation::Turbine,
```
bw-solana: Conceptually this makes sense because we don't know block ID when we request these shreds. But maybe we need to rename `BlockLocation::Turbine`... maybe something like `BlockLocation::Original` or `BlockLocation::BlockIdUnknown`

Doesn't need to happen now

turbine/src/sigverify_shreds.rs
AshwinSekar: `is_repaired` is used for wen_restart. Unfortunately BlockLocation isn't good enough, since TowerBFT repair will still insert the shred into the Turbine column which is indistinguishable from turbine shred ingest.

turbine/src/sigverify_shreds.rs
AshwinSekar: Yeah maybe the naming isn't great, but TowerBFT repair still inserts into the Turbine column. In a future where we remove TowerBFT repair entirely, then all repaired shreds will have BlockLocation::Repair

core/src/repair/serve_repair.rs
AshwinSekar: yeah i'm open to suggestions:
- Turbine ingest -> BlockLocation::Turbine
- TowerBFT repair -> BlockLocation::Turbine
- AG repair -> BlockLocation::Repair

If we remove TowerBFT repair this would be perfect, but maybe for now we should rename it to something like Original or First for clarity

core/src/repair/ancestor_hashes_service.rs
AshwinSekar: ah yeah my editor did something weird, fixed here https://github.com/anza-xyz/alpenglow/pull/357/commits/2099e781e43982f09343031ed4c19e4b9f04ad4b

turbine/src/lib.rs
AshwinSekar: agreed, moved to ledger 4535550ae4a683d699a0379fddcea6bd794ef9a1

turbine/src/sigverify_shreds.rs
AshwinSekar: good catch, done 44926c8cbfbab31cbc81361f3e872b0f07ade6ce

turbine/src/sigverify_shreds.rs
bw-solana: Is it truly just for TowerBFT repair? Seems like in some extreme case we could saturate the LRU cache

bw-solana (dismissed):
LGTM

turbine/src/sigverify_shreds.rs
AshwinSekar: oh I see your point, yeah in this case it's actually bad to insert this into the wrong column. 
Let me rework this to throw away shreds and alert us if the cache is saturated

turbine/src/sigverify_shreds.rs
```diff
@@ -243,14 +248,35 @@ fn run_shred_sigverify<const K: usize>(
...
                 // not retranmitted.
-                Either::Right(shred::Payload::from(shred))
+                Either::Right((
+                    shred::Payload::from(shred),
+                    /* is_repaired */ true,
+                    location,
+                ))
             } else {
                 // Share the payload between the retransmit-stage and the
                 // window-service.
```
bw-solana: would be a great candidate to pull into a sub-function

bw-solana (dismissed):

core/src/repair/serve_repair.rs
```diff
@@ -122,6 +127,19 @@ impl ShredRepairType {
             | ShredRepairType::ShredForBlockId(slot, _, _) => *slot,
         }
     }
+
+    pub fn location_to_insert_response(&self) -> BlockLocation {
+        match self {
+            ShredRepairType::OrphanForBlockId(_, bid)
+            | ShredRepairType::HighestShredForBlockId(_, _, bid)
+            | ShredRepairType::ShredForBlockId(_, _, bid) => {
+                BlockLocation::Repair { block_id: *bid }
```
carllin: maybe making a `fn block_id() -> Option<BlockId>` method on `ShredRepairType` and casing on that to determine `BlockLocation` can make it more explicit this location is based on blockid

core/src/repair/serve_repair.rs
```diff
@@ -122,6 +127,19 @@ impl ShredRepairType {
             | ShredRepairType::ShredForBlockId(slot, _, _) => *slot,
         }
     }
+
+    pub fn location_to_insert_response(&self) -> BlockLocation {
```
carllin: nit: block_location() or storage_location() might be clearer

core/src/repair/serve_repair.rs
carllin: Original makes sense to me, better than my suggestion of Primary

ledger/src/block_location_lookup.rs
```diff
@@ -0,0 +1,36 @@
...
+//!
+//! We don't actively cleanup this cache, its size is managed according to Lru policy
+
+use {
+    crate::{blockstore_meta::BlockLocation, shred::Nonce},
+    lru::LruCache,
+    std::sync::{Arc, RwLock},
+};
+
+pub struct BlockLocationLookup {
```
carllin, line 16: what was the reason we couldn't augment the existing `OutstandingRequests` cache with the location?

core/src/repair/serve_repair.rs
```diff
@@ -122,6 +127,19 @@ impl ShredRepairType {
...
+        match self {
+            ShredRepairType::OrphanForBlockId(_, bid)
+            | ShredRepairType::HighestShredForBlockId(_, _, bid)
+            | ShredRepairType::ShredForBlockId(_, _, bid) => {
```
carllin: was wondering if it would ever be beneficial to have repair by fec set merkle granularity, but this seems better/simpler for now

ledger/src/block_location_lookup.rs
AshwinSekar: outstanding requests verification is performed before sigverify and we pop the request as soon as we've received the appropriate amount of responses. So this would require holding onto the request for longer and only popping it off once we've additionally sigverified the response.
Seemed cleaner to split this out into a different lookup.

core/src/repair/serve_repair.rs
AshwinSekar: I did consider also having one for FEC set, but I don't think it makes a difference. In the cases where we only need to re-repair a few shreds from an FEC set (and thus known the merkle root), the responder can still give us the wrong shred so we'll need to compare the merkle root during ingest.

core/src/repair/serve_repair.rs
AshwinSekar: I like having this directly spit out a BlockLocation instead of having the indirection

core/src/repair/serve_repair.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/357/commits/21b106957e4933682896d4b20b066d2f68ab2f3e

turbine/src/sigverify_shreds.rs
AshwinSekar: done [aa9ea9c](https://github.com/anza-xyz/alpenglow/pull/357/commits/aa9ea9c21820098376c365009d7400e81fe4a6c5)

core/src/repair/serve_repair.rs
AshwinSekar: done [3beee61](https://github.com/anza-xyz/alpenglow/pull/357/commits/3beee61357251c0fa02ace6b8a37d395101ef969)

core/src/repair/serve_repair.rs
carllin: Was thinking something like 
```
self.block_id()
                .map(|bid| BlockLocation::Repair { block_id: *bid })
                .or_else(BlockLocation::Original);
```

core/src/repair/serve_repair.rs
carllin: yeah better to get all the wrong things at once and then retry i think

core/src/repair/serve_repair.rs
AshwinSekar: done e1cd2db0e1bcb4f98164aa3681c29267f586fa88

PR #360, opened 2025-08-06
refactor: do not store epoch related state in `CertificatePool`

akhi3030:
Instead of caching epoch related state in the `CertificatePool`, pass it into the pool.  This removes dependencies between banks and other related structs like the RootBankCache, and BankForks, etc.

### DIFF
--- a/votor/src/certificate_pool.rs
+++ b/votor/src/certificate_pool.rs
@@ -800,10 +800,10 @@ mod tests {
             .map(|_| ValidatorVoteKeypairs::new_rand())
             .collect::<Vec<_>>();
         let bank_forks = create_bank_forks(&validator_keypairs);
-        let bank = bank_forks.read().unwrap().root_bank();
+        let root_bank = bank_forks.read().unwrap().root_bank();
         (
             validator_keypairs,
-            CertificatePool::new_from_root_bank(Pubkey::new_unique(), &bank, None),
+            CertificatePool::new_from_root_bank(Pubkey::new_unique(), &root_bank, None),
             bank_forks,
         )
     }


votor/src/certificate_pool.rs
```diff
@@ -668,20 +636,18 @@ impl CertificatePool {
     /// Cleanup any old slots from the certificate pool
     pub fn handle_new_root(&mut self, bank: Arc<Bank>) {
```
akhi3030: who is calling this and does it need to pass in the bank.

qkniep: this should not be needed at all: pool itself determines finalization, finalization itself is enough to trigger cleanup within pool

akhi3030: Indeed.  We covered this in the morning.  I plan on adjusting the code here in steps.  First, I just want to stop tracking epoch stakes and related structs in this struct.

AshwinSekar: the only use for bank here was to support epoch stakes. If we do not have epoch stakes then we can immediately pass the finalized slot here in the caller (and avoid having to take the lock in the caller).

Just a side note, RootBankCache was removed in master, it'd be a good idea to cherry-pick and update to the new SharableRootBank at some point. https://github.com/anza-xyz/agave/pull/7300

akhi3030: Indeed... I plan on making these changes in steps so as to keep the PR small and self contained.

votor/src/certificate_pool.rs
```diff
@@ -713,17 +678,14 @@ impl CertificatePool {
 
 pub fn load_from_blockstore(
     my_pubkey: &Pubkey,
-    root_bank: &Bank,
+    bank: &Bank,
```
wen-coding: We are still passing in root_bank right? Why rename it bank?

votor/src/certificate_pool.rs
```diff
@@ -361,12 +353,14 @@ impl CertificatePool {
                 }
                 if self.highest_finalized_slot.is_none_or(|s| s < slot) {
                     self.highest_finalized_slot = Some(slot);
+                    self.cleanup();
```
wen-coding: If we are going to implement Notarize+Skip certs in slot s+8, this is too soon. You need to make sure when the leader starts producing slot s+8 the certs for s are still around.

I also want to see how this cleanup will be impacted by local ParentReady tracker.

votor/src/certificate_pool_service.rs
```diff
@@ -296,15 +294,22 @@ impl CertificatePoolService {
     ///
     /// If a new finalization slot was recognized, returns the slot
     fn add_message_and_maybe_update_commitment(
+        bank: &Bank,
```
wen-coding: this is root_bank right? We should name it so.

votor/src/certificate_pool_service.rs
```diff
@@ -96,11 +95,10 @@ impl CertificatePoolService {
...
             CertificatePoolServiceStats::incr_u16(&mut stats.new_finalized_slot);
             // Set root
-            let root_bank = root_bank_cache.root_bank();
-            if root_bank.slot() > *current_root {
+            let bank = root_bank_cache.root_bank();
+            if &bank.slot() > current_root {
                 CertificatePoolServiceStats::incr_u16(&mut stats.new_root);
-                *current_root = root_bank.slot();
-                cert_pool.handle_new_root(root_bank);
+                *current_root = bank.slot();
```
wen-coding: What do we use `current_root` for if we don't need `handle_new_root` any more?

votor/src/certificate_pool_service.rs
```diff
@@ -296,15 +294,22 @@ impl CertificatePoolService {
...
         cert_pool: &mut CertificatePool,
         votor_events: &mut Vec<VotorEvent>,
         commitment_sender: &Sender<AlpenglowCommitmentAggregationData>,
     ) -> Result<(Option<Slot>, Vec<Arc<CertificateMessage>>), AddVoteError> {
-        let (new_finalized_slot, new_certificates_to_send) =
-            cert_pool.add_message(my_vote_pubkey, message, votor_events)?;
+        let (new_finalized_slot, new_certificates_to_send) = cert_pool.add_message(
+            bank.epoch_schedule(),
+            bank.epoch_stakes_map(),
+            bank.slot(),
```
wen-coding: Just curious, why is passing three members of a Bank better than passing bank itself?

akhi3030: because then the pool depends upon the bank for information that only changes once an epoch.  We need to clean things up so that we are not unnecessarily using banks for information that should be stored in an object like EpochInfo.

votor/src/certificate_pool.rs
```diff
@@ -1985,7 +1985,11 @@ mod tests {
             )
             .is_ok());
         let cert_4 = CertificateMessage {
-            certificate: Certificate::new(CertificateType::Finalize, 4, None),
```
wen-coding: This is to test the case where we have Finalize but not accompanied Notarize certificate, I believe in that case you should not clean up this aggressively. (White paper page 43, "or a finalization certificate for s *together* with a notarization certificate for b in the same slots")

votor/src/certificate_pool.rs
akhi3030: I am not sure I understand this comment.  CC: @qkniep and @kobi-lizard.  Maybe you can help add some light.

votor/src/certificate_pool_service.rs
akhi3030: now removed.

votor/src/certificate_pool.rs
wen-coding: Also, finalization cert can arrive before corresponding notarization cert, and for standstill we need both(white paper page 43), I think you should actually cleanup when `highest_finalized_slot_with_notarize` changes.

akhi3030: I'll take a closer look at the paper again tomorrow.

AshwinSekar: I believe wen's point about s+8 is that the leader should pack the notarize and skip aggregates for slot s - 8 in slot s for rewards. 
To implement this, when we send over the `VotorEvent::ProduceWindow` we want to attach these aggregates so that the block creation loop can put them in the block. 

This already isn't possible in the current implementation, since `handle_new_root` will cleanup the `vote_pools` too early 😅  - imagine we root slot 10 and then are leader for 12 - 15, we won't have the aggregates for 4 - 7 available!

Since we're refactoring this code, it would be nice to implement cleanup in such a way that enables us to implement rewards down the line.

wen-coding: Yeah, we are already cleaning up a bit too aggressively, this is even more aggressive.

akhi3030: for now I have reverted the change so we are only cleaning up on new root slots.  We can revisit in a separate PR.

votor/src/certificate_pool.rs
akhi3030: reverted.

votor/src/certificate_pool.rs
```diff
@@ -699,9 +676,8 @@ impl CertificatePool {
     }
 
     pub fn get_certs_for_standstill(&self) -> Vec<Arc<CertificateMessage>> {
-        let (highest_finalized_with_notarize_slot, has_fast_finalize) = self
-            .highest_finalized_with_notarize
-            .unwrap_or((self.root, false));
+        let (highest_finalized_with_notarize_slot, has_fast_finalize) =
+            self.highest_finalized_with_notarize.unwrap_or((0, true));
```
bw-solana, line 680: can you explain why we default to `true` for `has_fast_finalize` here?

votor/src/certificate_pool.rs
```diff
@@ -816,28 +790,36 @@ mod tests {
...
+        CertificatePool,
+        Arc<RwLock<BankForks>>,
+    ) {
         // Create 10 node validatorvotekeypairs vec
         let validator_keypairs = (0..10)
             .map(|_| ValidatorVoteKeypairs::new_rand())
             .collect::<Vec<_>>();
         let bank_forks = create_bank_forks(&validator_keypairs);
-        let root_bank = bank_forks.read().unwrap().root_bank();
+        let bank = bank_forks.read().unwrap().root_bank();
```
bw-solana: `root_bank` here seems more descriptive

votor/src/certificate_pool.rs
```diff
@@ -87,7 +87,30 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
...
+    let Some((vote_key, _)) = epoch_stakes
+        .bls_pubkey_to_rank_map()
+        .get_pubkey(rank as usize)
+    else {
+        return Err(AddVoteError::InvalidRank(rank));
+    };
+    let stake = epoch_stakes.vote_account_stake(vote_key);
+    if stake == 0 {
+        // Since we have a valid rank, this should never happen, there is no rank for zero stake.
+        panic!("Validator stake is zero for pubkey: {vote_key}");
```
bw-solana, line 109: not your fault and doesn't need to be part of this PR, but it would probably be better to return an error here

votor/src/certificate_pool.rs
```diff
@@ -668,22 +648,19 @@ impl CertificatePool {
     }
 
     /// Cleanup any old slots from the certificate pool
-    pub fn handle_new_root(&mut self, bank: Arc<Bank>) {
-        let new_root = bank.slot();
-        self.root = new_root;
+    pub fn cleanup(&mut self, root_slot: Slot) {
```
bw-solana: cleanup fills a little vague. Maybe `cleanup_old_state`, `prune_old_state`, `cleanup_state_behind_root`?

votor/src/certificate_pool_service.rs
```diff
@@ -296,15 +286,22 @@ impl CertificatePoolService {
...
         cert_pool: &mut CertificatePool,
         votor_events: &mut Vec<VotorEvent>,
         commitment_sender: &Sender<AlpenglowCommitmentAggregationData>,
     ) -> Result<(Option<Slot>, Vec<Arc<CertificateMessage>>), AddVoteError> {
-        let (new_finalized_slot, new_certificates_to_send) =
-            cert_pool.add_message(my_vote_pubkey, message, votor_events)?;
+        let (new_finalized_slot, new_certificates_to_send) = cert_pool.add_message(
+            root_bank.epoch_schedule(),
+            root_bank.epoch_stakes_map(),
+            root_bank.slot(),
```
bw-solana, line 300: I don't necessarily have a problem with it, but any particular reason you chose to pass these 3 elements instead of passing the whole bank (and referencing later)?

akhi3030: The basic idea to get the certificate pool ready for https://github.com/anza-xyz/alpenglow/issues/369.  The pool should only need specific information that can only changes once an epoch.  We are still passing in the root_slot which I want to improve in a future PR as per https://github.com/anza-xyz/alpenglow/issues/368.

votor/src/certificate_pool.rs
akhi3030: Agreed, I'll create an issue which could be addressed in a future PR.

wen-coding: It was an error, but this shouldn't happen so we changed to a panic.
https://github.com/anza-xyz/alpenglow/pull/214#discussion_r2164493858

akhi3030: Created https://github.com/anza-xyz/alpenglow/issues/382.

votor/src/certificate_pool.rs
akhi3030: https://github.com/anza-xyz/alpenglow/pull/360/commits/ab3f209a71fff5b8a8882af5e46bc4260318cd70

votor/src/certificate_pool.rs
wen-coding: Yeah, this seems wrong, can't assume that.

votor/src/certificate_pool.rs
```diff
@@ -862,6 +847,7 @@ mod tests {
 
     fn add_skip_vote_range(
         pool: &mut CertificatePool,
+        bank: &Bank,
```
wen-coding, line 850: this is also root_bank right?

votor/src/certificate_pool.rs
akhi3030: I am naively assuming that if slot=0 is genesis then it will implicitly be notarized and finalized.  I can change it to false if that makes more sense.

votor/src/certificate_pool_service/stats.rs
```diff
@@ -11,7 +11,6 @@ pub(crate) struct CertificatePoolServiceStats {
     pub(crate) certificates_sent: u16,
     pub(crate) certificates_dropped: u16,
     pub(crate) new_finalized_slot: u16,
-    pub(crate) new_root: u16,
```
wen-coding, line 14: Maybe we should add a count for cleanup we've done....

votor/src/certificate_pool.rs
wen-coding: has_fast_finalize means a FastFinalize cert exists, so we don't send out either Fast or Notarize certs.

votor/src/certificate_pool.rs
akhi3030: https://github.com/anza-xyz/alpenglow/pull/360/commits/6733b2f1187647e63a8bf01cbf8bc1aced30b21f

votor/src/certificate_pool_service/stats.rs
akhi3030: do you mean the number of times we have called `cleanup`?

wen-coding: Yup, so if suddenly we don't accept any incoming votes/certs we can check whether a cleanup happened.

votor/src/certificate_pool.rs
akhi3030: changed to `false`.

votor/src/certificate_pool.rs
```diff
@@ -87,7 +87,30 @@ impl From<AlpenglowCommitmentError> for AddVoteError {
     }
 }
 
-#[derive(Default)]
+fn get_key_and_stakes(
```
qkniep, line 90: this also probably shouldn't be in `certificate_pool`

votor/src/certificate_pool.rs
```diff
@@ -467,18 +446,18 @@ impl CertificatePool {
         });
         let vote_type = VoteType::get_type(vote);
         if let Some(conflicting_type) =
-            self.has_conflicting_vote(slot, vote_type, &validator_vote_key, &block_id)
+            self.has_conflicting_vote(vote_slot, vote_type, &validator_vote_key, &block_id)
```
qkniep, line 449: this should not to be part of this PR, but why do we turn the `rank` into the vote key and then use the vote key to index validators? I think we should only use the rank within all of Pool

votor/src/certificate_pool.rs
```diff
@@ -511,12 +490,13 @@ impl CertificatePool {
 
     fn add_certificate(
         &mut self,
+        root_slot: Slot,
```
qkniep, line 493: this should not care about the root at all, but I guess that will be part of #368

qkniep (approved):

akhi3030:
This PR was accidentally merged.  I'll make a new one with some of the changes requested above.

PR #363, opened 2025-08-06
Send SetIdentity from admin_rpc_service.

wen-coding:


validator/src/admin_rpc_service.rs
```diff
@@ -1417,6 +1438,14 @@ mod tests {
             actual_validator_id,
             expected_validator_id.pubkey().to_string()
         );
+        let event = votor_event_receiver
+            .recv()
+            .expect("Failed to receive SetIdentity event");
+        if let VotorEvent::SetIdentity = event {
```
AshwinSekar, line 1444: nit: can use
`assert_matches!(event, VotorEvent::SetIdentity)`
instead of the if else

AshwinSekar (approved):
Lets also remove the TODO here https://github.com/anza-xyz/alpenglow/blob/a7a2a15396c727de38b77797b586ba8c2444d66e/votor/src/event_handler.rs#L322

As a follow up I think for #274 in `handle_set_identity` after we restore the new vote history we can send out all our votes via the own vote sender so they get inserted into the certificate pool if for whatever reason they weren't initially there.

validator/src/admin_rpc_service.rs
wen-coding: Err, tried that, got:
error[E0658]: use of unstable library feature `assert_matches`
    --> validator/src/admin_rpc_service.rs:1444:9
     |
1444 |         assert_matches!(event, VotorEvent::SetIdentity);
     |         ^^^^^^^^^^^^^^
     |
     = note: see issue #82775 <https://github.com/rust-lang/rust/issues/82775> for more information

So maybe we stick with the ugly if for now?

AshwinSekar:
> Will the operators forget to copy over the vote history file?

This is well documented in the failover process that you have to copy over the file else you risk submitting slasable votes. In TowerBFT we have the cli argument `--require_tower` which will fail to set-identity if the tower file is missing (tower & vote history files are named by identity pubkey). Could be a good idea to create an equivalent for vote history.

PR #364, opened 2025-08-07
feat(entry): introducing `BlockComponent` and `BlockMarker`

ksn6:
#### Problem
We need a unified, extensible data structure to represent the components of a Solana block, covering both normal transaction entries and special markers.

The structure must support current needs - such as optimistic block production in Alpenglow and block footer markers from (SIMD 307)[https://github.com/solana-foundation/solana-improvement-documents/pull/307] - while remaining flexible for new marker types in the future.

#### Summary of Changes
We introduce a new enum, `BlockComponent`, which represents one logical element within a block:

```
pub enum BlockComponent {
    Entries(Vec<Entry>),
    BlockMarker(VersionedBlockMarker),
}
```

- `Entries` - contains a vector of `Entry`-s (the most common case)

`BlockMarker` - a special marker payload that can store different kinds of markers, versioned for future extensibility

```
Entries Variant
┌─────────────────────────────────────────┐
│ Entry Count                  (8 bytes)  │
├─────────────────────────────────────────┤
│ bincode Entry 0           (variable)    │
├─────────────────────────────────────────┤
│ bincode Entry 1           (variable)    │
├─────────────────────────────────────────┤
│ ...                                     │
├─────────────────────────────────────────┤
│ bincode Entry N-1         (variable)    │
└─────────────────────────────────────────┘

BlockMarker Variant
┌─────────────────────────────────────────┐
│ Entry Count = 0              (8 bytes)  │
├─────────────────────────────────────────┤
│ Marker Version               (2 bytes)  │
├─────────────────────────────────────────┤
│ Marker Data Length, in bytes (2 bytes)  │
├─────────────────────────────────────────┤
│ Marker Data               (variable)    │
└─────────────────────────────────────────┘

BlockMarkerV0 Layout
┌─────────────────────────────────────────┐
│ Variant ID                   (1 byte)   │
├─────────────────────────────────────────┤
│ Variant Data              (variable)    │
└─────────────────────────────────────────┘

BlockMarkerV1 Layout
┌─────────────────────────────────────────┐
│ Variant ID                   (1 byte)   │
├─────────────────────────────────────────┤
│ Variant Data              (variable)    │
└─────────────────────────────────────────┘

BlockFooterV0 Layout
┌─────────────────────────────────────────┐
│ Producer Time Nanos          (8 bytes)  │
├─────────────────────────────────────────┤
│ User Agent Length            (1 byte)   │
├─────────────────────────────────────────┤
│ User Agent Bytes          (0-255 bytes) │
└─────────────────────────────────────────┘

ParentReadyUpdateV0 Layout
┌─────────────────────────────────────────┐
│ Parent Slot                  (8 bytes)  │
├─────────────────────────────────────────┤
│ Parent Block ID             (32 bytes)  │
└─────────────────────────────────────────┘
```

#### Migration and Compatibility
- Initially, most `BlockComponents` will be `Entries`.

- Marker variants such as `ParentReadyUpdate` and `BlockFooter` will be used where special marker data is needed.

- Versioned markers ensure forward compatibility when adding new marker types.

#### Resources
- Alpenglow optimistic block production design: https://github.com/anza-xyz/alpenglow/issues/293

- SIMD 307: Block footers proposal: https://github.com/solana-foundation/solana-improvement-documents/pull/307#issuecomment-3161916201

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -458,10 +458,13 @@ impl BlockComponent {
         let entries = bincode::deserialize::<Vec<Entry>>(data)
             .map_err(|e| BlockComponentError::DeserializationFailed(e.to_string()))?;
         let cursor = bincode::serialized_size(&entries)
-            .map_err(|e| BlockComponentError::SerializationFailed(e.to_string()))? as usize;
+            .map_err(|e| BlockComponentError::SerializationFailed(e.to_string()))?
+            as usize;
 
         // Handle remaining data
-        let remaining_bytes = data.get(cursor..).ok_or(BlockComponentError::CursorOutOfBounds)?;
+        let remaining_bytes = data
+            .get(cursor..)
+            .ok_or(BlockComponentError::CursorOutOfBounds)?;
 
         match (entries.is_empty(), remaining_bytes.is_empty()) {
             (true, true) => Ok(Self::Entries(Vec::new())),
@@ -611,12 +614,12 @@ impl<'de> Deserialize<'de> for VersionedBlockMarker {
 // ============================================================================
 
 /// Writes a variant ID and byte length prefix, then appends the data.
-/// 
+///
 /// # Format
 /// - Byte 0: Variant ID (u8)
 /// - Bytes 1-2: Data length in little-endian (u16)
 /// - Bytes 3+: Variant data
-/// 
+///
 /// Returns the complete serialized bytes.
 fn write_variant_with_length(variant_id: u8, data: &[u8]) -> Result<Vec<u8>, BlockComponentError> {
     let num_bytes: u16 = data
@@ -633,12 +636,12 @@ fn write_variant_with_length(variant_id: u8, data: &[u8]) -> Result<Vec<u8>, Blo
 }
 
 /// Reads a variant ID and byte length prefix from data.
-/// 
+///
 /// # Format Expected
 /// - Byte 0: Variant ID (u8)
 /// - Bytes 1-2: Data length in little-endian (u16)
 /// - Bytes 3+: Variant data (exactly `length` bytes)
-/// 
+///
 /// Returns (variant_id, payload_data) or an error.
 fn read_variant_with_length(data: &[u8]) -> Result<(u8, &[u8]), BlockComponentError> {
     // Get variant ID
@@ -1856,18 +1859,24 @@ mod tests {
         // First test that MAX_ENTRIES itself fails
         let result = BlockComponent::validate_entries_length(BlockComponent::MAX_ENTRIES);
         assert!(result.is_err());
-        assert_eq!(result.unwrap_err(), BlockComponentError::TooManyEntries {
-            count: BlockComponent::MAX_ENTRIES,
-            max: BlockComponent::MAX_ENTRIES,
-        });
+        assert_eq!(
+            result.unwrap_err(),
+            BlockComponentError::TooManyEntries {
+                count: BlockComponent::MAX_ENTRIES,
+                max: BlockComponent::MAX_ENTRIES,
+            }
+        );
 
         // Test that MAX_ENTRIES + 1 also fails
         let result = BlockComponent::validate_entries_length(BlockComponent::MAX_ENTRIES + 1);
         assert!(result.is_err());
-        assert_eq!(result.unwrap_err(), BlockComponentError::TooManyEntries {
-            count: BlockComponent::MAX_ENTRIES + 1,
-            max: BlockComponent::MAX_ENTRIES,
-        });
+        assert_eq!(
+            result.unwrap_err(),
+            BlockComponentError::TooManyEntries {
+                count: BlockComponent::MAX_ENTRIES + 1,
+                max: BlockComponent::MAX_ENTRIES,
+            }
+        );
 
         // Test that MAX_ENTRIES - 1 succeeds
         let result = BlockComponent::validate_entries_length(BlockComponent::MAX_ENTRIES - 1);
@@ -1925,10 +1934,13 @@ mod tests {
         // So we'll test the validation logic directly
         let result = BlockComponent::validate_entries_length(BlockComponent::MAX_ENTRIES);
         assert!(result.is_err());
-        assert_eq!(result.unwrap_err(), BlockComponentError::TooManyEntries {
-            count: BlockComponent::MAX_ENTRIES,
-            max: BlockComponent::MAX_ENTRIES,
-        });
+        assert_eq!(
+            result.unwrap_err(),
+            BlockComponentError::TooManyEntries {
+                count: BlockComponent::MAX_ENTRIES,
+                max: BlockComponent::MAX_ENTRIES,
+            }
+        );
     }
 
     #[test]


entry/src/entry_batch.rs
```diff
@@ -0,0 +1,555 @@
...
+/// # Serialization Format
+/// The binary format is:
+/// ```
+/// [8 bytes: entries_len as u64 little-endian]
+/// [variable: serialized entries using bincode, concatenated]
+/// [variable: optional VersionedSpecialEntry if present]
+/// ```
+pub struct EntryBatch {
+    pub entries: Vec<Entry>,
+    pub special: Option<VersionedSpecialEntry>,
```
AshwinSekar: Special entries can't have any entries present `Vec<Entry>` right?
What do you think about making this an enum instead:
- If Vec<Entry>::len() > 0, deserialize as Vec<Entry>
- Else deserialize as a VersionedSpecialEntry

ksn6: Nice, done.

entry/src/entry_batch.rs
```diff
@@ -0,0 +1,1268 @@
...
+/// - Entries can be present with or without special data
+/// - Special data can be present with or without entries
+/// - Only restriction: entries cannot be both present and special data present simultaneously
+/// - Empty entries with no special data is allowed
+///
+/// # Serialization Format
+/// The binary format is:
+/// [8 bytes: entries_len as u64 little-endian]
+/// [variable: serialized entries using bincode, concatenated]
+/// [variable: optional VersionedSpecialEntry if present]
```
carllin: nit: Image from the SIMD might be clearer

entry/src/entry_batch.rs
```diff
@@ -0,0 +1,1268 @@
...
+/// A versioned wrapper around SpecialEntry for backward compatibility.
+///
+/// # Serialization Format
+/// [2 bytes: version as u16 little-endian]
+/// [variable: serialized SpecialEntry]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct VersionedSpecialEntry {
+    pub version: u16,
+    pub inner: SpecialEntry,
+}
```
carllin: Why are certain structs versioned with a version field and others are enums?

carllin: Can we include a description at the top of the file of how we imagine this versioning will all work together?

entry/src/entry_batch.rs
```diff
@@ -0,0 +1,1268 @@
...
+/// - Only restriction: entries cannot be both present and special data present simultaneously
+/// - Empty entries with no special data is allowed
+///
+/// # Serialization Format
+/// The binary format is:
+/// [8 bytes: entries_len as u64 little-endian]
+/// [variable: serialized entries using bincode, concatenated]
+/// [variable: optional VersionedSpecialEntry if present]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EntryBatch {
```
AshwinSekar: Instead of the version nesting, what do you think about having a VersionedEntryBatch where V0 has `Vec<Entry> | ParentReadyUpdate`. If we want to add a new version of ParentReadyUpdate we can add a new version of VersionedEntryBatch with the additional variant ParentReadyUpdateV2.

entry/src/entry_batch.rs
ksn6: Done

entry/src/entry_batch.rs
ksn6: Nice - this cuts down on a lot of code.

entry/src/entry_batch.rs
ksn6: I considered this, though, the current code ends up being a bit more ergonomic and less repetitive. I'd rather users not have to re-specify all structures just to upgrade a single component.

entry/src/block_component.rs
```diff
@@ -0,0 +1,2165 @@
...
+/// │ Variant ID                   (1 byte)   │
+/// ├─────────────────────────────────────────┤
+/// │ Variant Data              (variable)    │
+/// └─────────────────────────────────────────┘
+/// ```
+///
+/// ### BlockMarkerV1 Layout
+/// ```text
+/// ┌─────────────────────────────────────────┐
+/// │ Variant ID                   (1 byte)   │
```
AshwinSekar, line 58: SIMD also specifies Variant Length. We'll want to agree and update either the SIMD or the code accordingly

entry/src/block_component.rs
```diff
@@ -0,0 +1,2165 @@
...
+/// │ Footer Data               (variable)    │
+/// └─────────────────────────────────────────┘
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum VersionedBlockFooter {
+    V0(BlockFooterV0),
+    Current(BlockFooterV0),
+}
+
+/// Version 0 block footer containing production metadata.
```
AshwinSekar: SIMD also specifies 1 for the version here

entry/src/block_component.rs
```diff
@@ -0,0 +1,2165 @@
...
+            entries.push(entry);
+            cursor += entry_size;
+        }
+
+        // Handle remaining data
+        match data.get(cursor..) {
+            Some(remaining) if !remaining.is_empty() => {
+                // Remaining data should be a marker
+                let marker = VersionedBlockMarker::from_bytes(remaining)?;
+                if entries.is_empty() {
```
AshwinSekar: think it would be better to case on whether entry_count == 0 and deserialize accordingly. Cant' have both in the same entry batch.

entry/src/block_component.rs
```diff
@@ -0,0 +1,2165 @@
...
+    pub fn to_bytes(&self) -> Result<Vec<u8>, bincode::Error> {
+        let mut buffer = Vec::new();
+
+        match self {
+            Self::Entries(entries) => {
+                Self::validate_entries_length(entries.len())
+                    .map_err(|e| bincode::Error::new(bincode::ErrorKind::Custom(e)))?;
+
+                // Write entry count
+                buffer.extend_from_slice(&(entries.len() as u64).to_le_bytes());
```
AshwinSekar: Do we need to do this manually? Doesn't bincode automatically serialize a Vec<> with the length and each entry?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2165 @@
...
+///
+/// # Serialization Format
+/// ```text
+/// ┌─────────────────────────────────────────┐
+/// │ Version                      (2 bytes)  │
+/// ├─────────────────────────────────────────┤
+/// │ Marker Data               (variable)    │
+/// └─────────────────────────────────────────┘
+/// ```
+#[derive(Debug, Clone, PartialEq, Eq)]
```
AshwinSekar, line 208: apart from BlockComponent with the actual nuance in serialize/deserialize, can we just `#[derive(Serialize, Deserialize)]` on the remaining enum/structs? looks like the format is the same as stock serde.

entry/src/block_component.rs
ksn6: Nice catch - pinged the SIMD thread; seems a bit redundant to have `length` here given that it's easily determinable from `block_user_agent_len`and the rest of the fields.

entry/src/block_component.rs
ksn6: Done.

Variants also start at 1 for some reason - pinged the SIMD thread about this.

entry/src/block_component.rs
ksn6: Done

entry/src/block_component.rs
ksn6: Done

entry/src/block_component.rs
ksn6: Sadly, we can't easily - bincode encodes enum variants using 4 bytes and Vec<...> lengths using 8 bytes, and there doesn't appear to be an easy way of customizing this. Please let me know if you find a way to do so.

These are the two issues preventing a very elegant #[Serialize, Deserialize] for every struct here. E.g., see ParentReadyUpdateV1, where we're able to do this.

entry/src/block_component.rs
carllin: saves 2 bytes eh 💰

ksn6: #gottacatchemall

entry/src/block_component.rs
AshwinSekar: There is the variable int encoding configuration flag, which will allow us to do this although FD would also have to do this for all fields, as i believe it's not possible to specify this flag for just a single part of the serialization.

Got it, can we also derive for BlockFooterV1 then? Mainly want to avoid as much as the boiler plate as possible / maintaining custom serialization and introducing subtle errors

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+/// │ User Agent Bytes          (0-255 bytes) │
+/// └─────────────────────────────────────────┘
+/// ```
+///
+/// ### ParentReadyUpdateV1 Layout
+/// ```text
+/// ┌─────────────────────────────────────────┐
+/// │ Parent Slot                  (8 bytes)  │
+/// ├─────────────────────────────────────────┤
+/// │ Parent Block ID             (32 bytes)  │
```
wen-coding, line 82: Hmm, I know this is as described in the paper, but what happens if someone sends ParentReadyUpdate(s, block_id_1) and ParentReadyUpdate(s, block_id_2) and some people get the first while others get the second?
I'm mostly wondering how we guarantee "only switch once" and everyone switches to the same block_id

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BlockComponent {
+    Entries(Vec<Entry>),
+    BlockMarker(VersionedBlockMarker),
+}
+
+/// A versioned block marker supporting multiple format versions.
+///
+/// Provides backward compatibility through versioned variants. During deserialization,
+/// older versions are upgraded to the `Current` variant for forward compatibility.
```
wen-coding, line 198: What do you mean, how does the upgrade happen?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+/// ┌─────────────────────────────────────────┐
+/// │ Producer Time Nanos          (8 bytes)  │
+/// ├─────────────────────────────────────────┤
+/// │ User Agent Length            (1 byte)   │
+/// ├─────────────────────────────────────────┤
+/// │ User Agent Bytes          (0-255 bytes) │
+/// └─────────────────────────────────────────┘
+/// ```
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct BlockFooterV1 {
```
wen-coding, line 294: Totally okay for now, but in the future maybe we should put all new versions into new files?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+    pub block_user_agent: Vec<u8>,
+}
+
+// ============================================================================
+// Parent Ready Update Types
+// ============================================================================
+
+/// Versioned parent ready update for optimistic block packing in Alpenglow.
+///
+/// Used to signal when a parent block is ready for building upon. Always deserializes
```
wen-coding: nit: We will only send this when switching to another parent right?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+        } else {
+            Ok(())
+        }
+    }
+
+    /// Serializes to bytes.
+    ///
+    /// # Errors
+    /// Returns a bincode error if serialization fails or if validation fails.
+    pub fn to_bytes(&self) -> Result<Vec<u8>, bincode::Error> {
```
wen-coding: nit: instead of every `pub fn` returning a different error, do you think we should define `BlockComponentError`?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+        match self {
+            Self::Entries(entries) => {
+                Self::validate_entries_length(entries.len())
+                    .map_err(|e| bincode::Error::new(bincode::ErrorKind::Custom(e)))?;
+
+                buffer = bincode::serialize(entries)?;
+            }
+            Self::BlockMarker(marker) => {
+                // Write zero entry count
+                buffer.extend_from_slice(&0u64.to_le_bytes());
```
wen-coding, line 429: If it's 0u64, does endian matter?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+
+    /// Deserializes from bytes with validation.
+    ///
+    /// # Errors
+    /// Returns a bincode error if deserialization fails or data is invalid.
+    ///
+    /// TODO(karthik): fuzz test this function.
+    /// TODO(karthik): should we error out if there are remaining bytes after deserialization?
+    pub fn from_bytes(data: &[u8]) -> Result<Self, bincode::Error> {
+        const ENTRY_COUNT_SIZE: usize = 8;
```
wen-coding, line 447: nit: would be nice that this is a constant we share with encoding...

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+            ))
+        })?;
+
+        match (entries.is_empty(), remaining_bytes.is_empty()) {
+            (true, true) => Ok(Self::Entries(Vec::new())),
+            (true, false) => {
+                let marker = VersionedBlockMarker::from_bytes(remaining_bytes)?;
+                Ok(Self::BlockMarker(marker))
+            }
+            (false, true) => Ok(Self::Entries(entries)),
```
wen-coding, line 477: Can you put a comment there why this would happen?

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+        let mut buffer = Vec::with_capacity(2 + marker_bytes.len());
+        buffer.extend_from_slice(&self.version().to_le_bytes());
+        buffer.extend_from_slice(&marker_bytes);
+
+        Ok(buffer)
+    }
+
+    /// Deserializes from bytes, creating appropriate variant based on version.
+    fn from_bytes(data: &[u8]) -> Result<Self, bincode::Error> {
+        const VERSION_SIZE: usize = 2;
```
wen-coding, line 558: nit: would be nice to share this VERSION_SIZE with to_bytes.

entry/src/block_component.rs
```diff
@@ -0,0 +1,2149 @@
...
+// BlockMarkerV1 Implementation
+// ============================================================================
+
+impl BlockMarkerV1 {
+    /// Serializes to bytes with variant prefix.
+    fn to_bytes(&self) -> Result<Vec<u8>, bincode::Error> {
+        let Self::BlockFooter(footer) = self;
+        let footer_bytes = footer.to_bytes()?;
+
+        let mut buffer = Vec::with_capacity(1 + footer_bytes.len());
```
wen-coding, line 536: nit: maybe we should put these constants describing length of a field somewhere near each other?

entry/src/block_component.rs
qkniep: `ParentReadyUpdate` appears as part of a block, we reach consensus on the block, so either everyone sees it or no one. If there is more than one of these in a given block, we should consider the block invalid.

carllin: @wen-coding  I think you treat it like any other duplicate block, stash it in one of the alternative 3 versions until one gets confirmed

entry/src/block_component.rs
ksn6: Variants now start at 0 per the SIMD. All set here.

entry/src/block_component.rs
ksn6: Can't derive for `BlockFooterV1` either - it's the same issue as before:

```
#[derive(Clone, PartialEq, Eq, Debug)]
pub struct BlockFooterV1 {
    pub block_producer_time_nanos: u64,
    pub block_user_agent: Vec<u8>,
}
```

We plan on using a single byte to denote `block_user_agent.len()`, while a Vec would use 8 bytes.

Having `block_user_agent_len` as a `u8` locally doesn't work either, since `block_user_agent` needs to store some sort of variable length data structure, and bincode will end up writing out the length of that data structure.

FWIW - I agree with your concern re. avoiding as much boiler plate as possible to avoid subtle bugs. Let's aim to land this for now + start getting optimistic block packing implementation going. I'll add in a TODO that auto-serializes with proc-macros correctly.

entry/src/block_component.rs
ksn6: Agreed with @qkniep - I believe that this constitutes slashable behavior.

entry/src/block_component.rs
ksn6: This behavior isn't implemented yet, since Current and V1 are the only "active" versions present. The idea behind this comment is:

- Say we're running a validator that has V5 of some struct Foo. Concretely, Current refers to V5.
- We receive a message that is V2 of Foo from a validator running an older version.
- When deserializing, we will (a) parse the Foo V2, turn this into a Foo V3, then a Foo V4, and then a Foo V5 (i.e. the current).

entry/src/block_component.rs
ksn6: Sure - will do this in a separate PR.

entry/src/block_component.rs
ksn6: Yep, updated.

entry/src/block_component.rs
ksn6: What would the alternative be?

entry/src/block_component.rs
ksn6: We never directly use this constant in encoding.

entry/src/block_component.rs
ksn6: Not sure why this would happen, although it's probably a good idea to support the "everything is empty" case. Added a comment to clarify.

entry/src/block_component.rs
ksn6: We never directly use this constant in encoding.

entry/src/block_component.rs
wen-coding: This is fine.

entry/src/block_component.rs
ksn6: Done.

entry/src/block_component.rs
ksn6: There's a good point made in the SIMD thread; we'll want to keep the length field in. Context: https://github.com/solana-foundation/solana-improvement-documents/pull/307#discussion_r2267154201

PR #381, opened 2025-08-07
BLS: use base-3 encoding in solana-signer-store.

wen-coding:
We didn't have real base-3 encoding before in BLS. Now that solana-signer-store is ready, we can use the base-2 or base-3 encoding there.

Also cutting off ending zeroes in input bitmap, this can make BLS bitmap much smaller.


votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -1,46 +1,62 @@
...
 }
 
 /// A builder for creating a `CertificateMessage` by efficiently aggregating BLS signatures.
 #[derive(Clone)]
 pub struct VoteCertificateBuilder {
     certificate: Certificate,
     signature: SignatureProjective,
-    bitmap: BitVec<u8, Lsb0>,
+    input_bitmap_1: BitVec<u8, Lsb0>,
+    input_bitmap_2: BitVec<u8, Lsb0>,
```
carllin, line 45: can we add a comment on what the difference is between these two bitmaps

votor/src/certificate_pool.rs
```diff
@@ -246,15 +246,16 @@ impl CertificatePool {
...
                 if let Some(vote_pool) = self.vote_pools.get(&(slot, *vote_type)) {
-                match vote_pool {
-                    VotePoolType::SimpleVotePool(pool) => pool.add_to_certificate(&mut vote_certificate_builder),
-                    VotePoolType::DuplicateBlockVotePool(pool) => pool.add_to_certificate(block_id.as_ref().expect("Duplicate block pool for {vote_type:?} expects a block id for certificate {cert_id:?}"), &mut vote_certificate_builder),
-                };
-            }
+                    match vote_pool {
+                        VotePoolType::SimpleVotePool(pool) => pool.add_to_certificate(is_first_vote_type, &mut vote_certificate_builder),
+                        VotePoolType::DuplicateBlockVotePool(pool) => pool.add_to_certificate(block_id.as_ref().expect("Duplicate block pool for {vote_type:?} expects a block id for certificate {cert_id:?}"), is_first_vote_type, &mut vote_certificate_builder),
+                    };
```
carllin: Passing along this `is_first_vote_type` seems a bit awkward and error prone, because it relies on the ordering returned by `certificate_limits_and_vote_types`

1. Because `VoteCertificateBuilder` already tracks the type of certificate it is building, `VoteCertificateBuilder::aggregate()` should be able to deduce if passed a `VoteType` whether to add to bitmap1 or bitmap2
2. The vote pool `add_to_certificate()` method should be able to just pass its internal `VoteType` to the `vote_certificate_builder`

votor/src/certificate_pool/vote_certificate_builder.rs
wen-coding: Done.

votor/src/certificate_pool.rs
wen-coding: Yeah I originally wrote it that way, then felt I leaked certificate_limits_and_vote_types and VoteType into VoteCertificateBuilder unnecessarily. But I don't have strong opinions about it, I can change it back so you see if you like that better.
Actually, now that I think about it, maybe we should just add the VoteType for bitmap1 and/or bitmap2 to the certificate itself, adds 2 bytes but it's much less error prone, what do you think?

wen-coding: Err, VoteType is a local data structure.... But I think we can add it to vote-message

wen-coding: Okay, I reverted to using VoteType now.

votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -1,46 +1,71 @@
...
-    bitmap: BitVec<u8, Lsb0>,
+    // For some certificates we need two bitmaps, for example, NotarizeFallback
+    // certificates have Notarize and NotarizeFallback votes, so we need two bitmaps
+    // to represent them. The order of the VoteType is defined in certificate_limits_and_vote_types.
+    // We normally put fallback votes in the second bitmap.
+    // The order of the VoteType is important, if you change it, you might interpret
+    // the bitmap incorrectly.
+    // Some certificates (like Finalize) only need one bitmap, then the second bitmap
+    // will be empty.
+    input_bitmap_1: BitVec<u8, Lsb0>,
```
samkim-crypto, line 44: Do you think we can use specific names for the fields? Maybe we can use `primary_signer_bitmap` and `fallback_signer_bitmap` or something like these, perhaps?

votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -50,53 +75,266 @@ impl VoteCertificateBuilder {
...
+        if new_length > MAXIMUM_VALIDATORS {
+            error!(
+                "Bitmap length exceeds maximum allowed: {} should be caught during aggregation",
+                MAXIMUM_VALIDATORS
+            );
+            return Err(CertificateError::ValidatorDoesNotExist(new_length as u16));
+        }
+
+        let mut bitmap1 = self.input_bitmap_1.clone();
+        let mut bitmap2 = self.input_bitmap_2.clone();
```
samkim-crypto: Are these clones necessary? I think we can just do `let mut input_bitmap_1 = self.input_bitmap_1`. The function `.last_one()` above does already borrow the bitmap, so we do need to move it to the top of the function to satisfy the borrow checker.

```
    let mut input_bitmap_1 = self.input_bitmap_1;
    let mut input_bitmap_2 = self.input_bitmap_2;

    let last_one_1 = input_bitmap_1 // use local variable
        .last_one()
        .map_or(0, |i| i.saturating_add(1));
    let last_one_2 = input_bitmap_2 // use local variable
        .last_one()
        .map_or(0, |i| i.saturating_add(1));
    let new_length = last_one_1.max(last_one_2);
    if new_length > MAXIMUM_VALIDATORS {
        error!(
            "Bitmap length exceeds maximum allowed: {} should be caught during aggregation",
            MAXIMUM_VALIDATORS
        );
        return Err(CertificateError::ValidatorDoesNotExist(new_length as u16));
    }

    input_bitmap_1.resize(new_length, false);
    input_bitmap_2.resize(new_length, false);
```

carllin (approved):

samkim-crypto (commented):
Looks very clean! I have two minor comments, but looks good to me!

votor/src/certificate_pool/vote_certificate_builder.rs
wen-coding: I thought about that, but in the future we might use this struct to aggregate other bitmaps. For example, the [Alpenglow SIMD]([url](https://github.com/solana-foundation/solana-improvement-documents/pull/326)) says we should put aggregates of Notarize and Skip for slot `s` in the block for `s+8`, we could use the same base-3 encoding for that. So I'll probably keep these names for now.

votor/src/certificate_pool/vote_certificate_builder.rs
wen-coding: Good point, changed.

samkim-crypto (approved):

PR #384, opened 2025-08-08
refactor: additional feedback for #360

akhi3030:
#360 was accidentally merged and not all review comments were addressed.  This PR addresses the remaining comments.


qkniep (commented):
LGTM

Just a sanity check for me: Whether we set `has_fast_finalize` for genesis to `true` or `false` is completely arbitrary because we have to treat genesis separately anyways, right?

wen-coding (approved):

bw-solana:
yeah, we special case this for 'parent ready' as well

PR #385, opened 2025-08-08
Read new identity from cluster_info in certificate_pool

wen-coding:


votor/src/certificate_pool_service.rs
```diff
@@ -206,10 +212,19 @@ impl CertificatePoolService {
 
         // Ingest votes into certificate pool and notify voting loop of new events
         while !ctx.exit.load(Ordering::Relaxed) {
-            // TODO: we need set identity here as well
+            // Update the current pubkey if it has changed
+            {
+                let new_pubkey_guard = ctx.my_pubkey.read().unwrap();
```
AshwinSekar: if we're going to grab a lock i'd prefer to just pass cluster_info here instead of adding a new one. Otherwise we can have another receiver for set identity events to check in the select below

wen-coding: Adding an event receiver is probably over kill since this event is pretty rare. I considered passing cluster_info, the problem is that we are going to have a race between event loop resetting vote_history and certificate pool setting identity.

But I don't feel strongly about it, if you are sure the race will not do any harm then I can change to cluster_info.

AshwinSekar: It should be no difference as this implementation can race anyway unless you hold the lock for the entire ingestion process.

If you change pubkey from staked to unstaked:
- you might still emit events based on the staked nodes votes, however you are unable to sign with the unstaked key

If you change pubkey from unstaked to staked:
- The unstaked node could not have signed any votes in the first place, so technically safeto* cannot trigger in the first place

Changing between two staked nodes is not supported (1 node_pubkey per vote account).

So I think just checking cluster_info is fine.

wen-coding: That's fair, changed.

votor/src/certificate_pool_service.rs
```diff
@@ -170,26 +172,31 @@ impl CertificatePoolService {
...
+            channel_name
+        );
         ctx.exit.store(true, Ordering::Relaxed);
         Err(())
     }
 
     // Main loop for the certificate pool service, it only exits when any channel is disconnected
     fn certificate_pool_ingest_loop(mut ctx: CertificatePoolContext) -> Result<(), ()> {
         let mut events = vec![];
+        let mut my_current_pubkey = ctx.cluster_info.id();
```
AshwinSekar: nit: rename to my_pubkey to be inline with the rest

votor/src/certificate_pool_service.rs
```diff
@@ -206,10 +213,17 @@ impl CertificatePoolService {
 
         // Ingest votes into certificate pool and notify voting loop of new events
         while !ctx.exit.load(Ordering::Relaxed) {
-            // TODO: we need set identity here as well
+            // Update the current pubkey if it has changed
+            let new_pubkey = ctx.cluster_info.id();
+            if my_current_pubkey != new_pubkey {
```
AshwinSekar: A log here would be good, otherwise LGTM

AshwinSekar (approved):

votor/src/certificate_pool_service.rs
wen-coding: Done.

votor/src/certificate_pool_service.rs
wen-coding: Done.

AshwinSekar (approved):

PR #386, opened 2025-08-09
Bubble up errors in event handler.

wen-coding:


AshwinSekar (approved):
Thanks for cleaning this up,
just as a broader sanity check is there any reason that `insert_vote_and_create_bls_message` can fail for some sort of transient error? or any error from this function indicates that we something is seriously wrong and we should shutdown.

AshwinSekar (changes requested):
Actually sorry approved too early, if i'm tracing the code correctly this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?

Since we're here do you mind fixing it? Also doesn't have to be in this PR but would be nice to have an error.rs to move all the various Error types to

wen-coding:
> Actually sorry approved too early, if i'm tracing the code correctly this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?
> 
> Since we're here do you mind fixing it? Also doesn't have to be in this PR but would be nice to have an error.rs to move all the various Error types to

That's a good point. I assume any error during generate_vote_tx is transient? I've changed them to an error log.

AshwinSekar:
> > Actually sorry approved too early, if i'm tracing the code correctly this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?
> > Since we're here do you mind fixing it? Also doesn't have to be in this PR but would be nice to have an error.rs to move all the various Error types to
> 
> That's a good point. I assume any error during generate_vote_tx is transient? I've changed them to an error log.

So `NonVoting` and `HotSpare` are normal for RPC / hot spare setups `info!` is fine here. 
`Failed` is due to startup scenarios (verification or wait for slot), or situations where the vote account is not setup / other misconfigurations. I would use `warn! `here, or if we really wanted to be precise split this into two enums and warn for the startup related ones and error for the misconfigurations.

Not sure where NoRankFound should fit in, I assume this can also happen if the validator does not meet the top 2k cap? Or even then would we expect the rank to exist?

wen-coding:
NoRankFound can happen if you are scheduled to vote for next Epich, so in
the current Epoch you can’t vote, so I’ll also mark it transient. Can’t ask
users to set up voting at the last minute of current Epoch.

wen-coding:
> > > Actually sorry approved too early, if i'm tracing the code correctly this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?
> > > Since we're here do you mind fixing it? Also doesn't have to be in this PR but would be nice to have an error.rs to move all the various Error types to
> > 
> > 
> > That's a good point. I assume any error during generate_vote_tx is transient? I've changed them to an error log.
> 
> So `NonVoting` and `HotSpare` are normal for RPC / hot spare setups `info!` is fine here. `Failed` is due to startup scenarios (verification or wait for slot), or situations where the vote account is not setup / other misconfigurations. I would use `warn! `here, or if we really wanted to be precise split this into two enums and warn for the startup related ones and error for the misconfigurations.
> 
> Not sure where NoRankFound should fit in, I assume this can also happen if the validator does not meet the top 2k cap? Or even then would we expect the rank to exist?

Okay, I split GenerationError into GenerationInfo and GenerationWarning respectively.

votor/src/voting_utils.rs
```diff
@@ -30,8 +30,10 @@ pub enum GenerateVoteTxResult {
     // hot spare validator, not eligble for refresh
     // until set identity is invoked
     HotSpare,
+    // Wait to vote slot is not reached
+    WaitToVoteSlot,
     // failed generation, eligible for refresh
-    Failed,
+    Failed(String),
```
bw-solana: Should we call this `FailedToGenerate`? Because it looks like we call out some specific failures in the enum but then lump some into this `Failed` enum and then differentiate based on the `String`

votor/src/voting_utils.rs
```diff
@@ -63,11 +65,20 @@ pub enum BLSOp {
 
 #[derive(Debug, Error)]
 pub enum VoteError {
-    #[error("Unable to generate bls vote message")]
-    GenerationError(Box<GenerateVoteTxResult>),
+    #[error("Unable to generate bls vote message, transient error: {0:?}")]
+    GenerationInfo(Box<GenerateVoteTxResult>),
+
+    #[error("Unable to generate bls vote message, configuration error: {0:?}")]
+    GenerationWarning(Box<GenerateVoteTxResult>),
```
bw-solana: I think coupling these errors to the log level isn't the best.

I would propose naming `TransientGeneration` and `InvalidConfig`

votor/src/event_handler.rs
```diff
@@ -388,10 +414,10 @@ impl EventHandler {
         parent_block @ (parent_slot, parent_block_id): Block,
```
bw-solana, line 388: we should touch up the function comment here to indicate what the returned `Result` and `bool` mean

core/src/replay_stage.rs
```diff
@@ -2770,21 +2770,26 @@ impl ReplayStage {
...
+            return GenerateVoteTxResult::Failed("startup verification incomplete".into());
         }
 
         if authorized_voter_keypairs.is_empty() {
             return GenerateVoteTxResult::NonVoting;
         }
         if let Some(slot) = wait_to_vote_slot {
             if bank.slot() < slot {
-                return GenerateVoteTxResult::Failed;
+                return GenerateVoteTxResult::Failed(format!(
```
bw-solana: should this be `GenerateVoteTxResult::WaitToVoteSlot`?

votor/src/voting_utils.rs
bw-solana: Related - would it make more sense to call out all the appropriate categories in the enum? Or possibly a sub-enum? And then we could attach the error strings to the enum, similar to what was done for `VoteError`

wen-coding: It was an include-all item, I've split it into different enum values.

votor/src/voting_utils.rs
wen-coding: I assume you mean TransientError, changed.

votor/src/event_handler.rs
wen-coding: Done.

core/src/replay_stage.rs
wen-coding: Oops, good catch, changed.

votor/src/voting_utils.rs
bw-solana: yeah, `TransientError` works

votor/src/event_handler.rs
```diff
@@ -152,13 +152,39 @@ impl EventHandler {
...
             for vote in votes {
-                vctx.bls_sender.send(vote?).map_err(|_| SendError(()))?;
+                vctx.bls_sender.send(vote).map_err(|_| SendError(()))?;
             }
         }
 
         Ok(())
     }
 
+    fn generate_vote_message_and_push_to_votes(
```
AshwinSekar: nit: can we move this to voting_utils ?

AshwinSekar (approved):
overall LGTM

votor/src/event_handler.rs
wen-coding: Done.

votor/src/voting_utils.rs
```diff
@@ -262,3 +275,28 @@ pub(crate) fn insert_vote_and_create_bls_message(
         saved_vote_history: SavedVoteHistoryVersions::from(saved_vote_history),
     })
 }
+
+pub fn generate_vote_message_and_push_to_votes(
+    votes: &mut Vec<BLSOp>,
```
carllin: this makes the function cumbersome, can we keep the original call  to `votes.push()` from the caller and have this function return a `Result<BLSOp, VoteError>`

carllin:
>  this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?

What is the danger here, is this fixing an existing issue?

votor/src/voting_utils.rs
wen-coding: Currently this function intercepts transient error and doesn't propagate them upstream. If you take that away then there's not much point adding this function, however we need to replicate the "match InvalidConfig and TransientError" code in all 6 places.

wen-coding:
> > this still causes hot spares to shutdown right? Since any generate_vote_tx value that's not BLSMessage would result in a vote error?
> 
> What is the danger here, is this fixing an existing issue?

This is not fixing an existing issue, just change the event_handler.rs to follow normal Result<..., Error> handling, and distinguish between InvalidConfig and TransientError explicitly.

We want hot spares to not vote, do we want them to shut down? Then they are no longer "hot" spares.
The only behavior change is:
- startup verification incomplete will become an info log, not error log and not exiting (at Alpenglow side)
I don't think there is any behavior change on TowerBFT side.

votor/src/voting_utils.rs
carllin: I think it would be more readable to return `Result<Option<BLSOp>, VoteError> {` and have the caller run something like

```
if let Some(bls_op) = generate_vote_message(vote, is_refresh, vctx)? {
    votes.push(bls_op);
}
```

carllin:
> and not exiting (at Alpenglow side)

Ah because `WaitForStartupVerification` now maps to a `VoteError::TransientError` which is ignored and only logged?

This seems like fixing an existing bug, could we update the PR description to capture these

votor/src/voting_utils.rs
wen-coding: Sure, done.

wen-coding:
> > and not exiting (at Alpenglow side)
> 
> Ah because `WaitForStartupVerification` now maps to a `VoteError::TransientError` which is ignored and only logged?
> 
> This seems like fixing an existing bug, could we update the PR description to capture these

Actually, I think we might currently exit on any transient error on Alpenglow side:
https://github.com/anza-xyz/alpenglow/blob/be34e2f7690f25c3d70c753f6c92ca174257ef2b/votor/src/voting_utils.rs#L236
This will return GenerationError when the generation result is not a BLSMessage

Then the Errors are pushed into votes:
https://github.com/anza-xyz/alpenglow/blob/be34e2f7690f25c3d70c753f6c92ca174257ef2b/votor/src/event_handler.rs#L256

Then we exit on any error inside votes:
https://github.com/anza-xyz/alpenglow/blob/be34e2f7690f25c3d70c753f6c92ca174257ef2b/votor/src/event_handler.rs#L155

And the other problem is we will not exit as soon as there is a non-transient error.

Updating PR description.

votor/src/voting_utils.rs
```diff
@@ -262,3 +275,26 @@ pub(crate) fn insert_vote_and_create_bls_message(
         saved_vote_history: SavedVoteHistoryVersions::from(saved_vote_history),
     })
 }
+
+pub fn generate_vote_message_and_push_to_votes(
```
carllin: nit: function name can be generate_vote_message()

wen-coding: Good catch, done.

votor/src/voting_utils.rs
```diff
@@ -233,8 +249,14 @@ pub(crate) fn insert_vote_and_create_bls_message(
...
         GenerateVoteTxResult::BLSMessage(bls_message) => bls_message,
+        e @ GenerateVoteTxResult::NoAuthorizedVoter(_, _)
+        | e @ GenerateVoteTxResult::NoVoteState(_)
+        | e @ GenerateVoteTxResult::VoteAccountNotFound(_)
+        | e @ GenerateVoteTxResult::Tx(_) => {
+            return Err(VoteError::InvalidConfig(Box::new(e)));
+        }
         e => {
-            return Err(VoteError::GenerationError(Box::new(e)));
+            return Err(VoteError::TransientError(Box::new(e)));
```
carllin: so is essentially everything either an InvalidConfig or TransientError? so generate_vote_message_and_push_to_votes() will essentially never hit this case: https://github.com/anza-xyz/alpenglow/blob/7f052ba50e14d99cefa3e46d9c3ef89fab974f4f/votor/src/voting_utils.rs#L297 

Can we remove the Result<> return type on `generate_vote_message_and_push_to_votes` then?

votor/src/voting_utils.rs
```diff
@@ -233,8 +249,14 @@ pub(crate) fn insert_vote_and_create_bls_message(
     let bank = context.root_bank_cache.root_bank();
     let bls_message = match generate_vote_tx(&vote, &bank, context) {
         GenerateVoteTxResult::BLSMessage(bls_message) => bls_message,
+        e @ GenerateVoteTxResult::NoAuthorizedVoter(_, _)
+        | e @ GenerateVoteTxResult::NoVoteState(_)
+        | e @ GenerateVoteTxResult::VoteAccountNotFound(_)
+        | e @ GenerateVoteTxResult::Tx(_) => {
+            return Err(VoteError::InvalidConfig(Box::new(e)));
+        }
```
carllin: For maintainability/avoiding shooting ourselves in the foot, it would be nice to force anybody who adds a new `GenerateVoteTxResult` to bucket it into the right category

Maybe add a `is_invalid_config()`  and `is_transient_error` functions on `GenerateVoteTxResult` that goes through every case?

votor/src/voting_utils.rs
wen-coding: Only for GenerateVoteTxResult errors. If sender closes or vote history save fails it will return Error.

```
    context
        .own_vote_sender
        .send(bls_message.clone())
        .map_err(|_| SendError(()))?;

    // TODO: for refresh votes use a different BLSOp so we don't have to rewrite the same vote history to file
    let saved_vote_history =
        SavedVoteHistory::new(&context.vote_history, &context.identity_keypair)?;

```

votor/src/voting_utils.rs
wen-coding: Done.

PR #387, opened 2025-08-09
nit: refactor: minor improvements to `new_vote_pool` 

akhi3030:
Removes some code duplication and specify enum variants instead of using `_` so that the compiler helps us in figuring out all the places that need changes if we add or remove variants in future.


qkniep:
Makes sense to me.

I also dislike the name `DuplicateBlockVotePool` and "duplicate block" itself for that matter... Maybe we could rename it to `BlockHashVotePool` or to `BlockVotePool` (and `SimpleVotePool` to `SlotVotePool`)?

akhi3030:
there are subtle issues with this PR.  Closing.

PR #389, opened 2025-08-11
Preempt block production if there is ParentReady for slots after the current block production window.

wen-coding:


carllin:
Is this behavior documented in the latest version of the paper?

core/src/block_creation_loop.rs
```diff
@@ -512,6 +511,19 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
...
+            let highest_parent_ready_slot = leader_window_notifier
+                .highest_parent_ready
+                .read()
+                .unwrap()
+                .0;
+            if highest_parent_ready_slot > end_slot {
+                trace!(
+                    "{my_pubkey}: Skipping production of {slot} because highest parent ready slot is {highest_parent_ready_slot} > end slot {end_slot}"
+                );
+                break;
```
carllin: might be good to add an extra case `skipped_window_behind_parent_ready` to the `BlockCreationLoopMetrics`

wen-coding: Done.

core/src/block_creation_loop.rs
```diff
@@ -512,6 +519,20 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
                 break;
             }
 
+            // Check if the entire window is skipped.
+            let highest_parent_ready_slot = leader_window_notifier
```
AshwinSekar: I would move this into `start_leader_retry_replay` so that we can also check this when producing the first block of the leader window.

Also I don't think we need to be this precise, but technically we can give up as soon as we get a skip certificate >= start_slot. Probably more headache than it's worth to additionally track that though.

wen-coding: Done.

Yeah I thought about that, but when skipping validators normally skip rest of the window, so the difference is probably small. Let's start with this relatively conservative check and see if it introduces any bugs, then upgrade to more precise checks if necessary.

carllin (approved):

PR #390, opened 2025-08-12
wip

akhi3030:
#### Summary of Changes

Fixes #


votor/src/commitment.rs
```diff
@@ -1,42 +1,33 @@
...
+pub enum AlpenglowUpdateCommitmentCacheError {
     ChannelDisconnected,
+    ChannelFull,
 }
 
-pub enum AlpenglowCommitmentType {
+pub enum AlpenglowCommitment {
     /// Our node has voted notarize for the slot
-    Notarize,
+    Notarize(Slot),
```
qkniep, line 14: Shouldn't we pass the block hash here?

qkniep: Also, still think this name is highly confusing, sounds like the block is notarized (which would be a commitment status with much more weight). But if we keep it as is, could we rename it to something like `Voted` or `VotedNotar`?

wen-coding: Do we have a conclusion here? About to upstream this file to Agave repo soon.

PR #393, opened 2025-08-12
refactor: update the slab dependency

akhi3030:
`cargo audit` found a security vulnerability with this crate.  This PR updates the dependency


bw-solana:
this is totally fine, but just FYI, we typically fix something like this upstream (https://github.com/anza-xyz/agave/pull/7467) and then cherry-pick the commit to the forks

akhi3030:
> this is totally fine, but just FYI, we typically fix something like this upstream ([anza-xyz/agave#7467](https://github.com/anza-xyz/agave/pull/7467)) and then cherry-pick the commit to the forks

Thanks, I'll keep that in mind for next time.

PR #394, opened 2025-08-13
Add safe_to_notar_parent_tracker so we check whether parent block has NotarizeFallback cert.

wen-coding:


votor/src/event_handler.rs
```diff
@@ -177,6 +177,12 @@ impl EventHandler {
                 debug_assert!(bank.is_frozen());
                 let (block, parent_block) = Self::get_block_parent_block(&bank);
                 info!("{my_pubkey}: Block {block:?} parent {parent_block:?}");
+                ctx.block_parent_sender
```
AshwinSekar, line 180: When we fetch a block in order to check safeToNotar we won't necessarily give up on our replayed bank immediately (remember we can only replay one version of a bank at a time, and we only send the Block event after replay).

Instead I think it would be cleaner if we took the following approach:
If the thresholds indicate SafeToNotar for an intrawindow block then:
1) Check the block versions column to see if we've ingested the block https://github.com/anza-xyz/alpenglow/blob/48e6421fd63db15a839c0634345b847bb0830c81/ledger/src/blockstore.rs#L607
2) Use this to determine the location, and check for the parent block id / parent slot using:
parent block id: https://github.com/anza-xyz/alpenglow/blob/48e6421fd63db15a839c0634345b847bb0830c81/ledger/src/blockstore.rs#L595
parent slot: https://github.com/anza-xyz/alpenglow/blob/48e6421fd63db15a839c0634345b847bb0830c81/ledger/src/blockstore.rs#L587
3) If the (parent_slot, parent_block_id) has the notarize-fallback cert then emit the event
4) If (1) or (2) fail because we have not ingested the block, stash it in a "pending_safe_to_notar" set
5) Once we have ag_repair_service, send a notification that we need to fetch this block + its ancestors
6) Once we have ag repair, we will notify certificate pool service when we add a new version of a block. We can then check pending_safe_to_notar and send out the SafeToNotar event accordingly

This way we don't have to build out a separate tracker here in certificate pool, since blockstore/repair will keep track of these things anyway.

wen-coding: Good point about replay, which is why this is a draft.
I thought about doing it in event_handler, but I don't think we track a lot of states in there currently, most of the state keeping currently happens in certificate pool.
But I don't have strong feelings about this, letting certificate pool know about blockstore and repair is arguably worse.

PR #395, opened 2025-08-13
Switch to QUIC on all-to-all traffic.

wen-coding:


core/src/sigverifier/bls_sigverifier/stats.rs
```diff
@@ -57,6 +57,7 @@ impl BLSSigVerifierStats {
         if time_since_last_log < STATS_INTERVAL_DURATION {
             return;
         }
+        trace!("BLS Sig Verifier Stats: {:?}", self);
         datapoint_info!(
```
carllin: could make this a datapoint_trace

carllin (approved):

local-cluster/src/integration_tests.rs
```diff
@@ -71,6 +71,7 @@ pub const AG_DEBUG_LOG_FILTER: &str = "error,solana_core::replay_stage=info,\
         solana_core::block_creation_loop=trace,\
         solana_votor=trace,\
         solana_votor::vote_history_storage=info,\
+        solana_core::sigverifier::bls_sigverifier=trace,\
```
lijunwangs: why is this necessary, cannot we set it via ENV variable RUST_LOG?

core/src/validator.rs
```diff
@@ -517,8 +519,9 @@ impl ValidatorTpuConfig {
             ..Default::default()
         };
 
-        // vote and tpu_fwd share the same characteristics -- disallow non-staked connections:
+        // vote/tpu_fwd/alpenglow share the same characteristics -- disallow non-staked connections:
         let vote_quic_server_config = tpu_fwd_quic_server_config.clone();
+        let alpenglow_quic_server_config = tpu_fwd_quic_server_config.clone();
```
lijunwangs, line 524: This is probably okay, but I would like us to review all configurations and make sure it makes sense for alpenglow. Do they really match the current TPU forward configs other than disallowing non-staked connections?

local-cluster/src/integration_tests.rs
wen-coding: This is mostly for debugging local-cluster tests.

core/src/validator.rs
wen-coding: That's fair. This one is only for tests, so I think disallowing non-staked connections is enough. I'll change the production config and we can discuss there.

core/src/sigverifier/bls_sigverifier/stats.rs
wen-coding: This should be an info! log, I'll remove the log changes from this PR.

local-cluster/src/integration_tests.rs
wen-coding: Reverted this change.

validator/src/commands/run/execute.rs
```diff
@@ -1069,6 +1072,18 @@ pub fn execute(
...
+    let alpenglow_quic_server_config = QuicServerParams {
+        max_connections_per_peer: 1,
+        max_staked_connections: tpu_max_alpenglow_connections.try_into().unwrap(),
+        max_unstaked_connections: 0,
+        max_streams_per_ms,
+        max_connections_per_ipaddr_per_min: tpu_max_connections_per_ipaddr_per_minute,
+        coalesce: tpu_coalesce,
+        num_threads: tpu_alpenglow_receive_threads,
+        ..Default::default()
+    };
```
wen-coding, line 1085: @lijunwangs Added Alpenglow specific quic config here, does it make sense?

core/src/tpu.rs
```diff
@@ -239,6 +241,23 @@ impl Tpu {
         )
         .unwrap();
 
+        // Streamer for Alpenglow
+        let SpawnServerResult {
+            endpoints: _,
+            thread: alpenglow_quic_t,
+            key_updater: alpenglow_stream_key_updater,
+        } = spawn_server_multi(
+            "solQuicAlpenglow",
```
lijunwangs: The thread name only allows 15 characters in linux. we need to shorten more, how about "solQuicAlpglw"?

wen-coding: Done.

lijunwangs (approved):

PR #397, opened 2025-08-14
Add basic metrics to event_handler.

wen-coding:
Added basic event counting metrics for now.

### DIFF
--- a/votor/src/event_handler.rs
+++ b/votor/src/event_handler.rs
@@ -5,6 +5,7 @@ use {
     crate::{
         commitment::{alpenglow_update_commitment_cache, AlpenglowCommitmentType},
         event::{CompletedBlock, VotorEvent, VotorEventReceiver},
+        event_handler::stats::EventHandlerStats,
         root_utils::{self, RootContext},
         timer_manager::TimerManager,
         vote_history::{VoteHistory, VoteHistoryError},
@@ -33,6 +34,8 @@ use {
     },
     thiserror::Error,
 };
+
+mod stats;
 
 /// Banks that have completed replay, but are yet to be voted on
 /// in the form of (block, parent block)
@@ -74,6 +77,14 @@ pub(crate) struct EventHandler {
     t_event_handler: JoinHandle<()>,
 }
 
+struct LocalContext {
+    pub(crate) my_pubkey: Pubkey,
+    pub(crate) pending_blocks: PendingBlocks,
+    pub(crate) finalized_blocks: BTreeSet<Block>,
+    pub(crate) received_shred: BTreeSet<Slot>,
+    pub(crate) stats: EventHandlerStats,
+}
+
 impl EventHandler {
     pub(crate) fn new(ctx: EventHandlerContext) -> Self {
         let exit = ctx.exit.clone();
@@ -100,22 +111,25 @@ impl EventHandler {
             voting_context: mut vctx,
             root_context: rctx,
         } = context;
-        let mut my_pubkey = vctx.identity_keypair.pubkey();
-        let mut pending_blocks = PendingBlocks::default();
-        let mut finalized_blocks = BTreeSet::default();
-        let mut received_shred = BTreeSet::default();
+        let mut local_context = LocalContext {
+            my_pubkey: ctx.cluster_info.keypair().pubkey(),
+            pending_blocks: PendingBlocks::default(),
+            finalized_blocks: BTreeSet::default(),
+            received_shred: BTreeSet::default(),
+            stats: EventHandlerStats::new(),
+        };
 
         // Wait until migration has completed
-        info!("{my_pubkey}: Event loop initialized");
+        info!("{}: Event loop initialized", local_context.my_pubkey);
         Votor::wait_for_migration_or_exit(&exit, &start);
-        info!("{my_pubkey}: Event loop starting");
+        info!("{}: Event loop starting", local_context.my_pubkey);
 
         if exit.load(Ordering::Relaxed) {
             return Ok(());
         }
 
         // Check for set identity
-        if let Err(e) = Self::handle_set_identity(&mut my_pubkey, &ctx, &mut vctx) {
+        if let Err(e) = Self::handle_set_identity(&mut local_context.my_pubkey, &ctx, &mut vctx) {
             error!(
                 "Unable to load new vote history when attempting to change identity from {} \
                  to {} on voting loop startup, Exiting: {}",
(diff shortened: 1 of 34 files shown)


votor/src/event_handler/stats.rs
```diff
@@ -0,0 +1,363 @@
...
+            VotorEvent::Timeout(_) => 5,
+            VotorEvent::SafeToNotar(_) => 6,
+            VotorEvent::SafeToSkip(_) => 7,
+            VotorEvent::ProduceWindow(_) => 8,
+            VotorEvent::Finalized(_) => 9,
+            VotorEvent::Standstill(_) => 10,
+            VotorEvent::SetIdentity => 11,
+            // If you add an entry, change num_event_types below.
+        }
+    }
```
carllin: is it simpler to just cast VotorEvent to an index instead

votor/src/event_handler/stats.rs
```diff
@@ -0,0 +1,363 @@
...
+    pub(crate) leader_window_replaced: u16,
+    pub(crate) set_root: u16,
+    pub(crate) timeout_set: u16,
+
+    /// All of the following fields are in microseconds
+    pub(crate) receive_event_time: u32,
+    pub(crate) send_vote_time: u32,
+
+    pub(crate) received_events_count_and_timing: Vec<(u16, u32)>,
+    pub(crate) sent_votes: Vec<u16>,
```
carllin: Might be easier to understand/maintain with just `HashMap` types mapping from the eventtypes/vote types to the counts, easier to manage and performance not really a concern here

votor/src/event_handler/stats.rs
```diff
@@ -0,0 +1,363 @@
...
+            VotorEvent::Finalized(_) => 9,
+            VotorEvent::Standstill(_) => 10,
+            VotorEvent::SetIdentity => 11,
+            // If you add an entry, change num_event_types below.
+        }
+    }
+
+    pub fn new() -> Self {
+        let num_vote_types = (VoteType::SkipFallback as usize).saturating_add(1);
+        let num_event_types = 12; // Update this if you add more events.
```
carllin: seems brittle/error prone

votor/src/event_handler.rs
```diff
@@ -127,50 +142,72 @@ impl EventHandler {
...
+        local_context: &mut LocalContext,
     ) -> Result<Vec<BLSOp>, EventLoopError> {
         let mut votes = vec![];
+        let LocalContext {
+            ref mut my_pubkey,
+            ref mut pending_blocks,
+            ref mut finalized_blocks,
+            ref mut received_shred,
+            ref mut stats,
+        .. // The `..` is a good practice to ignore other fields you don't need
```
carllin: think this is common enough to not need a comment

votor/src/event_handler.rs
```diff
@@ -437,6 +480,7 @@ impl EventHandler {
             slot,
             &voting_context.commitment_sender,
         )?;
+        stats.commitment_updates = stats.commitment_updates.saturating_add(1);
```
carllin: This doesn't seem that useful given
1. the commitment update is just a channel send
2. we're already reporting the number of each of type of vote 
3. have timings covering each individual event handler

votor/src/event_handler/stats.rs
wen-coding: How do you do that on an enum like this?

votor/src/event_handler/stats.rs
wen-coding: Done

votor/src/event_handler/stats.rs
wen-coding: Removed.

votor/src/event_handler.rs
wen-coding: Removed.

votor/src/event_handler.rs
wen-coding: Removed.

votor/src/event_handler.rs
```diff
@@ -74,6 +78,14 @@ pub(crate) struct EventHandler {
     t_event_handler: JoinHandle<()>,
 }
 
+struct LocalContext {
```
AshwinSekar, line 81: nice, thanks for cleaning this part up

votor/src/event_handler/stats.rs
```diff
@@ -0,0 +1,337 @@
...
+            ),
+            ("set_root", self.set_root as i64, i64),
+            ("timeout_set", self.timeout_set as i64, i64),
+        );
+        datapoint_info!(
+            "event_handler_received_events_count_and_timing",
+            (
+                "block",
+                self.received_events_count_and_timing
+                    .get(&0)
```
AshwinSekar: I think manually using indices here is prone to error and creates a lot of boilerplate.

What do you think about making a separate enum for stats type and map `VotorEvent` => `StatsEvent`. `received_events_count_and_timing` can then be an map of `StatsEvent` to `(u16, u32)` and then we can use a loop here to report all the counts and timings. `StatsEvent` can implement `Display` for the actual metrics name.

wen-coding: Done.

wen-coding:
> Also doesn't have to be in this PR, but it would be really interesting to track how long it takes a slot to go from `FirstShred` -> `voted` -> `Finalized` (and specifically if it was a slow or fast finalization).

Done.

votor/src/event_handler/stats.rs
```diff
@@ -1,15 +1,41 @@
...
 const STATS_REPORT_INTERVAL: Duration = Duration::from_secs(10);
 
+#[derive(Debug, Clone)]
+struct SlotTracking {
+    start: Instant,
+    first_shred: Option<Instant>,
+    parent_ready: Option<Instant>,
+    vote_notarize: Option<Instant>,
+    vote_skip: Option<Instant>,
+    finalized: Option<Instant>,
```
carllin: nit: remove `is_fast_finalization` and replacing with `Option<(Instant, bool)>` to keep consistent with the Event format

votor/src/event_handler/stats.rs
```diff
@@ -142,7 +208,63 @@ impl EventHandlerStats {
                 ("count", *count as i64, i64)
             );
         }
+        // Only report if the slot is lower than highest_parent_ready
```
carllin: we might not have voted or replayed the block yet if we're going based on parent ready. 

Should we go based off when we've rooted instead?

votor/src/event_handler/stats.rs
wen-coding: Done.

votor/src/event_handler/stats.rs
wen-coding: Done.

PR #398, opened 2025-08-15
blockstore: ingest shreds into appropriate column based on BlockLocation

AshwinSekar:
#### Problem
We don't do anything with the tagged `BlockLocation` we receive from sigverify.

#### Summary of Changes
Use the `BlockLocation` to insert shreds into the appropriate columns in blockstore.
This updates the full insert shreds call graph with some notable differences:
- FEC recovery is not performed on alternate columns (this can be done after we enable coding shred repair)
- SlotMeta chaining and completed data sets tracking is not performed on alternate columns (this will be tracked elsewhere, as we do not automatically `generate_new_bank_forks` from the alternate column.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2039,8 +2039,8 @@ impl Blockstore {
             shred_source,
         )?;
         if matches!(location, BlockLocation::Original) {
-            // This tracking is not needed for alternate shred columns, as we intentionally switch banks in replay
-            // based on certificate conditions
+            // We don't currently notify RPC when we complete data sets in alternate columns. This can be extended in the future
+            // if necessary.
             newly_completed_data_sets.extend(completed_data_sets);
         }
         merkle_root_metas


ledger/src/blockstore.rs
```diff
@@ -683,29 +724,53 @@ impl Blockstore {
...
-                IteratorDirection::Reverse,
-            ))?
-            // `find` here, to skip the first element in case the erasure meta for fec_set_index is already present
-            .find(|((_, candidate_fec_set_index), _)| {
-                *candidate_fec_set_index != u64::from(fec_set_index)
-            })
-            // Do not consider sets from the previous slot
-            .filter(|((candidate_slot, _), _)| *candidate_slot == slot)
-        else {
+        let Some((candidate_fec_set_index, candidate_erasure_meta)) = (match location {
```
AshwinSekar, line 728: can't wait to just -32 here instead 😭

carllin: lmao because fec sets don't always contain 32 shreds atm?

ledger/src/blockstore.rs
```diff
@@ -1037,8 +1126,14 @@ impl Blockstore {
                     };
                 }
                 ShredType::Code => {
+                    // Coding shred repair is not implemented so coding shreds
+                    // should always be inserted into the Original column
+                    // We still pass location to support a future when we do repair
+                    // coding shreds
+                    assert_matches!(location, BlockLocation::Original);
```
carllin, line 1123: So if you get a coding shred from a different version than the original, it'll just get thrown out during the merkle checks?

ledger/src/blockstore.rs
```diff
@@ -1160,13 +1264,16 @@ impl Blockstore {
...
-        for (erasure_set, working_erasure_meta) in shred_insertion_tracker.erasure_metas.iter() {
+        for ((location, erasure_set), working_erasure_meta) in
+            shred_insertion_tracker.erasure_metas.iter()
+        {
             if !working_erasure_meta.should_write() {
                 // Not a new erasure meta
                 continue;
             }
             let (slot, _) = erasure_set.store_key();
             if self.has_duplicate_shreds_in_slot(slot) {
```
carllin, line 1263: can we eventually get rid of the duplicate column after pushing through the alternate columns? would be a healthy cleanup

ledger/src/blockstore.rs
```diff
@@ -1160,13 +1264,16 @@ impl Blockstore {
...
+                // TODO(ashwin): continue checking to notify the shred resolver
```
carllin, line 1264: whats the shred resolver?

ledger/src/blockstore.rs
```diff
@@ -1416,9 +1534,8 @@ impl Blockstore {
         })
     }
 
-    // Attempts to recover and retransmit recovered shreds (also identifying
-    // and handling duplicate shreds). Broadcast stage should instead call
-    // Blockstore::insert_shreds when inserting own shreds during leader slots.
+    /// Simlar to `insert_shreds_at_location_handle_duplicate`  but always inserts
+    /// shreds in the original column specified by `BlockLocation::Original`
     pub fn insert_shreds_handle_duplicate<'a, F>(
```
carllin, line 1527: Is this only used in a test here? 
https://github.com/anza-xyz/alpenglow/blob/0ef8f57639659490f43dcc0e6a4dea462128fb56/core/src/window_service.rs#L629-L642

Might be able to just tag those shreds in the caller with `BlockLocation::Original` and delete this entire function

ledger/src/blockstore.rs
```diff
@@ -1867,18 +2046,26 @@ impl Blockstore {
...
             index_meta.data_mut(),
             &shred,
+            location,
             write_batch,
             shred_source,
         )?;
-        newly_completed_data_sets.extend(completed_data_sets);
+        if matches!(location, BlockLocation::Original) {
+            // This tracking is not needed for alternate shred columns, as we intentionally switch banks in replay
+            // based on certificate conditions
```
carllin: hmmm what does this comment mean? I thought `newly_completed_data_sets` is only used for notifying listeners in the `CompletedDataSetsService`, the alternate slot metas still store the `completed_data_indexes` right?

ledger/src/blockstore.rs
```diff
@@ -1923,24 +2110,29 @@ impl Blockstore {
...
-                .map(Cow::Owned),
-            (_, ShredType::Code) => self
-                .get_coding_shred(slot, u64::from(index))
+                .get_data_shred_from_location(slot, u64::from(index), location)
                 .unwrap()
                 .map(shred::Payload::from)
                 .map(Cow::Owned),
+            (_, ShredType::Code) => {
+                // Coding shred repair is not implemented so we can only ask for coding shreds from the
+                // original column
```
carllin, line 2117: this function can be called even from outside the repair path right? 

Maybe comment should be "no erasure shreds are ever inserted outside of the BlockLocaiton::Original"

ledger/src/blockstore.rs
```diff
@@ -4803,7 +5063,7 @@ impl Blockstore {
     /// Also see [`find_slot_meta_in_cached_state`] and [`find_slot_meta_in_db_else_create`].
     fn find_slot_meta_else_create<'a>(
```
carllin, line 5052: maybe can rename to `find_original_slot_meta_else_create`

and  other `BlockLocation::Original` functions like 

`find_slot_meta_in_db_else_create` -> `find_original_slot_meta_in_db_else_create`
`find_slot_meta_in_cached_state` -> `find_original_slot_meta_in_cached_state`
(might be missing a few here)

ledger/src/blockstore.rs
```diff
@@ -664,16 +704,17 @@ impl Blockstore {
...
         let (slot, fec_set_index) = erasure_set.store_key();
 
         // Check the previous entry from the in memory map to see if it is the consecutive
         // set to `erasure set`
         let candidate_erasure_entry = erasure_metas
             .range((
-                Bound::Included(ErasureSetId::new(slot, 0)),
-                Bound::Excluded(erasure_set),
+                Bound::Included((location, ErasureSetId::new(slot, 0))),
+                Bound::Excluded((location, erasure_set)),
```
bw-solana, line 712: This means we will only walk back within the current location, right? This code hurts my brain

ledger/src/blockstore.rs
```diff
@@ -683,29 +724,53 @@ impl Blockstore {
             .map(|(erasure_set, erasure_meta)| {
                 (*erasure_set, Cow::Borrowed(erasure_meta.as_ref()))
             });
-        if candidate_erasure_set_and_meta.is_some() {
-            return Ok(candidate_erasure_set_and_meta);
+        if let Some(((loc, es), em)) = candidate_erasure_set_and_meta {
```
bw-solana, line 722: doesn't have to be this PR, but I'm thinking we should add some metrics here to understand how often we get a hit in memory

ledger/src/blockstore.rs
```diff
@@ -683,29 +724,53 @@ impl Blockstore {
...
+                    *candidate_fec_set_index != fec_set_index
+                })
+                // Do not consider sets from the previous slot
+                .and_then(|((candidate_slot, candidate_fec_set_index, bid), candidate_erasure_meta)| {
+                    debug_assert!(bid == block_id);
+                    (candidate_slot == slot).then_some((
+                        candidate_fec_set_index,
+                        candidate_erasure_meta
+                    ))
+                }),
```
bw-solana, line 746: a lot of almost duplicate code in here. And we don't even need the block ID beyond an assert check 😭 

I'm wondering if there is a way to structure this that dedups most of the logic

ledger/src/blockstore.rs
AshwinSekar: yeah essentially we don't repair coding shreds, so the only way we could get coding shreds is through turbine (or TowerBFT repair, prior to alpenglow). In this case it'll be subject to the current duplicate block checks, and could be thrown out if the merkle root does not match

ledger/src/blockstore.rs
AshwinSekar: yeah once the Alpenglow migration is done a lot of that tower bft related duplicate code can be removed. I still think having the duplicate column for holding slashing proofs is fine, but it can be opt-in.

ledger/src/blockstore.rs
AshwinSekar: future PR, will essentially be the replacement of repair_service to actually send out repair and re repair requests.

ledger/src/blockstore.rs
AshwinSekar: I kept this because Blockstore is a public API so we wanted to keep function signatures same for downstream. Steve mentioned he might wrap some things under `agave-unstable-api` so that we can cleanup some pub functions in the future

ledger/src/blockstore.rs
AshwinSekar: ah yeah, I was thinking that sending this tracking info to CDSS would be confusing for RPC and to actually notify when we switch the replayed bank to the alternate version.

But now that I think about it, i don't think we need this complexity, I can just notify here even while the shreds are inserted into the alternate column.

ledger/src/blockstore.rs
AshwinSekar: yeah exactly, it's why I ordered the map with the location as the first tuple, so that we don't return an erasure meta from a different BlockLocation

ledger/src/blockstore.rs
AshwinSekar: yeah I agree, it's annoying because the keys of the columns are different, and in the original column `fec_set_index` was erroneously stored as a `u64` instead of an `u32`. 
When I had added the `MerkleRootMeta` column in the past I was asked to use `u32` for `fec_set_index`, which is why I kept it as a `u32` here as well. 

I think I can rework this to first convert into a `u32` and strip off the `block_id` for better code reuse.

ledger/src/blockstore.rs
AshwinSekar: for now I don't want to send alternate column stuff to CDSS, more work to propagate block id here. we can explore it in the future, updated the comment here 6483d01d985ad26e3fdee7cbbbfebf665893476c

ledger/src/blockstore.rs
AshwinSekar: reworked this so that the alternate column just uses - 32 8b3e7e623442cf463f0df96ccd835b82e47181f8
once SIMD-0317 is activated we can update the rest of the code to -32 as well

bw-solana (approved):
LGTM

AshwinSekar:
image
invalidator run looks good, merging

PR #399, opened 2025-08-15
Add metrics to timer_manager.

wen-coding:


votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,67 @@
...
+        inner.max_heap_size = inner.max_heap_size.max(size as u16);
+    }
+
+    pub fn maybe_report(&self) {
+        let mut inner = self.inner.lock().unwrap();
+        if inner.last_report.elapsed() < STATS_REPORT_INTERVAL {
+            return;
+        }
+        datapoint_info!(
+            "timer_manager_stats",
```
bw-solana: I think we should name this `votor_timer_manager`

votor/src/timer_manager/timers.rs
```diff
@@ -125,6 +130,7 @@ impl Timers {
             self.heap.push(Reverse((next_fire, slot)));
             timer
         });
+        self.stats.set_timeout_with_heap_size(self.heap.len());
```
bw-solana: should this be tucked inside the `or_insert_with` closure?

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    pub fn set_timeout_with_heap_size(&self, size: usize) {
+        let mut inner = self.inner.lock().unwrap();
+        inner.set_timeouts = inner.set_timeouts.saturating_add(1);
+        inner.max_heap_size = inner.max_heap_size.max(size as u16);
+    }
+
+    pub fn set_heap_size(&self, size: usize) {
```
bw-solana: maybe call this `record_max_heap_size`?

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,67 @@
...
+        sync::Mutex,
+        time::{Duration, Instant},
+    },
+};
+
+const STATS_REPORT_INTERVAL: Duration = Duration::from_secs(10);
+
+struct TimerManagerStatsInner {
+    max_heap_size: u16,
+    set_timeouts: u16,
```
bw-solana: would be good to add comments to make it clear what these are recording

bw-solana: maybe call `set_timeouts` `set_timeout_count` ?

votor/src/timer_manager/stats.rs
wen-coding: Done.

votor/src/timer_manager/timers.rs
wen-coding: Either works, if you want to count only the effective set_timeouts then you do inside `or_insert_with`, changed.

votor/src/timer_manager/stats.rs
wen-coding: Not really sure it is the max, changed to `record_heap_size`.

votor/src/timer_manager/stats.rs
wen-coding: Done.

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    },
+};
+
+const STATS_REPORT_INTERVAL: Duration = Duration::from_secs(10);
+
+struct TimerManagerStatsInner {
+    /// The maximum heap size of the timers since the last report
+    max_heap_size: u16,
+    /// The number of times `set_timeout` was called with a new timer.
+    set_timeout_count: u16,
```
akhi3030: why are these `u16`?

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// The last time the stats were reported
+    last_report: Instant,
+}
+
+impl TimerManagerStatsInner {
+    pub fn new() -> Self {
```
akhi3030, line 22:
```suggestion
    fn new() -> Self {
```

I believe this is only called in this module.

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+        datapoint_info!(
+            "votor_timer_manager",
+            ("max_heap_size", inner.max_heap_size as i64, i64),
+            ("set_timeout_count", inner.set_timeout_count as i64, i64),
+        );
+        *inner = TimerManagerStatsInner::new();
+    }
+
+    #[cfg(test)]
+    pub fn get_numbers_for_tests(&self) -> (u16, u16) {
```
akhi3030: if we return `TimerManagerStatsInner` here, the compiler will help us in updating this function when add more stats.

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+        Self {
+            max_heap_size: 0,
+            set_timeout_count: 0,
+            last_report: Instant::now(),
+        }
+    }
+}
+
+pub struct TimerManagerStats {
+    inner: Mutex<TimerManagerStatsInner>,
```
akhi3030: AFAICT this module does not need to store stuff in a Mutex so I would recommend that the consumer of the struct should be storing it inside a lock as needed and not this module.

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,70 @@
...
+        inner.set_timeout_count = inner.set_timeout_count.saturating_add(1);
+        inner.max_heap_size = inner.max_heap_size.max(size as u16);
+    }
+
+    pub fn record_heap_size(&self, size: usize) {
+        let mut inner = self.inner.lock().unwrap();
+        inner.max_heap_size = inner.max_heap_size.max(size as u16);
+    }
+
+    pub fn maybe_report(&self) {
```
akhi3030: If this function is called twice without any recordings happening, then we will stats of 0 which can be very confusing when we are trying to debug stuff and all of a sudden we are seeing 0 values.

votor/src/timer_manager/timers.rs
bw-solana: Maybe we should record both to understand how often we’re trying to set multiple timers for a slot

votor/src/timer_manager/stats.rs
wen-coding: Because we report every 10 seconds, that's 25 slots, u8 is probably a bit dangerous, let's say we do have a lot of events and timeouts somehow, but u16 should be more than enough for these event counts.

akhi3030: I understand that reasoning but I am worried if there are some strange bugs and the code is getting triggered a ton of times, we will miss that.  If there is no good reason to arbitrarily limit the number bytes available, I would feel more comfortable with u64.

Also, further down, the code is using `as` to convert to u16 which is dangerous as it will wrap around so we will see strange behaviour which will only lead to confusion.

votor/src/timer_manager/stats.rs
wen-coding: What do you suggest instead? We should skip reporting when numbers are all zero?

votor/src/timer_manager/timers.rs
wen-coding: Done.

votor/src/timer_manager/stats.rs
wen-coding: Changed

votor/src/timer_manager/stats.rs
wen-coding: Done.

votor/src/timer_manager/stats.rs
wen-coding: Since this is stats module where the numbers don't need to be protected by a lock to be in sync with values outside the module, I believe it's generally recommended to put the Mutex inside the module instead, so
1. the locks are close to the members it protects
2. we can change synchronization inside stats without touching caller
3. we can generally just claim this struct thread safe and call it however we like

That said, this module is so simple and it will only ever be owned by this single caller, so I don't think it makes a big difference either way. I can move the Mutex to caller if you feel strongly about it.

votor/src/timer_manager/stats.rs
akhi3030: I am curious why we reset the metrics after reporting.  I have never seen metrics being reported like that.

- For heap: we just track the current heap size, no need to reset.
- for `set_timeout_count`: we keep monotonically increasing it and never reset it.  Under normal conditions, we will never see it overflow, if we ever see it overflow in our collection, we know something went wrong.  In fact, we should hopefully notice it increasing at a high rate before it overflows and we would intervene then.

votor/src/timer_manager/stats.rs
akhi3030: If the stats is not concerned about multthreading, putting it inside a lock randomly confuses the reader.  When I look at the module, I suspect that it is doing multithreading.  That adds cognitive overhead.

> we can generally just claim this struct thread safe and call it however we like

By this arguement, we have to put every struct inside a lock.  

I would very much appreciate it if we did not have the mutex here.

votor/src/timer_manager/stats.rs
wen-coding: Well, we don't save these stats numbers in disk, so when the validator restarts the numbers will suddenly become 0.

For heap, we do want to track max heap size within the reporting interval. You don't care the max size was 1000 in the last 10 seconds, as long as it comes back to normal now.

akhi3030: > Well, we don't save these stats numbers in disk, so when the validator restarts the numbers will suddenly become 0.

I don't understand, why is this an argument for resetting every 10 seconds?  We cannot help that validators would reset metrics when they restart (which hopefully is not happening every 10 seconds), so why do want to do that to our metrics?

> 
> For heap, we do want to track max heap size within the reporting interval. You don't care the max size was 1000 in the last 10 seconds, as long as it comes back to normal now.

Sorry, I don't understand.  I am proposing that we track the current heap size.  Then we will know what the heap was like at the end of every 10 second interval.  If it spiked earlier in the interval and then came back down, we will not know that and as you say, we don't care about that.

votor/src/timer_manager/stats.rs
wen-coding: Changed.

votor/src/timer_manager/stats.rs
wen-coding: > > Well, we don't save these stats numbers in disk, so when the validator restarts the numbers will suddenly become 0.
> 
> I don't understand, why is this an argument for resetting every 10 seconds? We cannot help that validators would reset metrics when they restart (which hopefully is not happening every 10 seconds), so why do want to do that to our metrics?

You were assuming the number will always monotonically increase over time. If there is a validator restart in the middle, you will see the number suddenly jump to 0, would that mess up with your metrics evaluation?

I think normally what you would evaluate here is: does my average rate of timeout setting suddenly spike up at specific interval.

So you would expect numbers like (every 10 seconds):

20, 20, 21, 19, ...., 0, 10000, ....

You can do similar math if you don't reset,
10020, 10040, 10061, 10080, ....
But if there is a restart in the middle, suddenly your numbers are messed up
10020, 10040, 1005 (a restart and abnormal number, you have to special case negative number in your calculation), 24, 15 (another restart)

Of course, depending on how you do metrics calculation, this is fixable.

> 
> > For heap, we do want to track max heap size within the reporting interval. You don't care the max size was 1000 in the last 10 seconds, as long as it comes back to normal now.
> 
> Sorry, I don't understand. I am proposing that we track the current heap size. Then we will know what the heap was like at the end of every 10 second interval. If it spiked earlier in the interval and then came back down, we will not know that and as you say, we don't care about that.

Sorry, what do you mean by "not reset"?

If my current max_heap_size is 100, and I've done reporting, what should I do now?

Are you saying we should set it to the next value reported? Then how do I keep the max over 10 seconds?

akhi3030: I don't think it is worth debating this point.  I think we can achieve the desired goals with what we have here and what I was proposing.  We will just have to have different type of complexity in the offchain metrics collection.

AshwinSekar: I think it would be good to give the product a try, influxQL doesn't support OVER so it's quite difficult to try to compare metrics across validators with different uptime if you don't reset.
Here's a [link ](https://metrics.solana.com:8888/sources/16/chronograf/data-explorer?query=SELECT%20median%28%22cumulative_runtime_us%22%29%20AS%20%22median_cumulative_runtime_us%22%20FROM%20%22testnet-invalidator%22.%22autogen%22.%22accounts_background_service%22%20WHERE%20time%20%3E%20%3AdashboardTime%3A%20AND%20time%20%3C%20%3AupperDashboardTime%3A%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29 )to our metrics dashboard to play around.

I think a workflow like you suggested is possible using Flux, but it's unintuitive to have this singular metric be cumulative while the rest of the timing/count metrics in the agave codebase are ephemeral

AshwinSekar: Regardless the suggestion is out of scope for this PR. If you find a way to better report & monitor, or generally unfuck our metrics 🙏 😅  it would be best approached as a project that updates all agave metrics, and derived dashboards + the comms to ask external consumers to update to the new approach. 

For now it's best to follow the convention set across the codebase.

votor/src/timer_manager/timers.rs
```diff
@@ -98,20 +99,24 @@ pub(super) struct Timers {
     heap: BinaryHeap<Reverse<(Instant, Slot)>>,
     /// Channel to send events on.
     event_sender: Sender<VotorEvent>,
+    /// Stats for the timer manager.
+    stats: Arc<Mutex<TimerManagerStats>>,
```
AshwinSekar: Would it make sense to split this instead:
- set_timeouts fills/returns metrics about timer insertion for reporting in EventHandler
- progress fills/returns metrics about heap size / timers popped for reporting in SkipTimerService

that way we could avoid having 2 threads updating the same stats and using a mutex?

wen-coding: We could do that, but then we depend on the knowledge that `set_timeout()` is always going to be called from one thread and `progress()` is always going to be called from one other thread. What makes it messier is that set_timeout and progress can both touch heap so we are going to have heap reported from both places?

I don't have strong feelings about it, but it just looks a bit weird because these two metrics are seen in the same object.

AshwinSekar: Oh I meant heap size is only reported from progress, is it necessary to report it from set_timeouts as well?

wen-coding: set_timeout pushes to heap, so it's actually the more important place. We could only report heap from there, but then we don't get update when heap is popped in progress.

votor/src/timer_manager/stats.rs
akhi3030: Thanks for the link Ashwin.  Agreed.

votor/src/timer_manager/timers.rs
carllin: hmm do we need to record the max heap size in `fn progress()`? the size can only stay the same or decrease there.

carllin: Is the mutex necessary here? `Timers` is already guarded by an RwLock in  the `timer_manager` and events handler

votor/src/timer_manager/stats.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct TimerManagerStats {
+    /// The maximum heap size of the timers since the last report
+    max_heap_size: u64,
+    /// The number of times `set_timeout` was called.
+    set_timeout_count: u64,
+    /// The number of times `set_timeout` was called, there was no
+    /// existing timer, so this operation succeeded.
+    set_timeout_succeed_count: u64,
```
carllin, line 16: The actual timeouts in the event handler seem much more interesting: https://github.com/anza-xyz/alpenglow/blob/767ff3e3710d68f0ec78d7d7ab78d1a1e5f362c5/votor/src/event_handler.rs#L282-L291

wen-coding: We do have event count for Timeout in event_handler already.

votor/src/timer_manager/timers.rs
wen-coding: Under normal cases it's fine, if we don't have set_timeout() in 10 seconds then this number isn't accurate any more, but I guess we don't care that much about this case.

wen-coding: `maybe_report()` happens outside `timers.write()`.

Since we don't have `progress()` touch stats any more, let me just do `maybe_report` from `set_timeout` so we can get rid of this mutex. It's a bit ugly, but at least we don't need to argue where to put mutex for this stats any more.

carllin (approved):

Issue #347, opened 2025-08-05
Update Standstill logic to only send votes certs > highest finalized slot


AshwinSekar:
We currently send any extra certificates and votes for the highest finalized slot, we should only send for slots greater than this (along with the FastFinalize or Finalize + Notarize cert for the finalized slot)

PR #400, opened 2025-08-15
ParentReady should be able to use NotarizeFallback/Notarize/FinalizeFast.

wen-coding:
NotarizeFallback/Notarize/FinalizeFast all tell us that (slot, hash) is notarized, so they should all count in ParentReadyTracker.

This is especially a problem in standstill handling after https://github.com/anza-xyz/alpenglow/issues/347, since we don't send out NotarizeFallback certs for highest finalized slot any more. (So now our local-cluster tests hangs a lot.)

### DIFF
--- a/votor/src/certificate_pool/parent_ready_tracker.rs
+++ b/votor/src/certificate_pool/parent_ready_tracker.rs
@@ -84,8 +84,8 @@ impl ParentReadyTracker {
         }
     }
 
-    /// Adds a new notar-fallback certificate
-    pub fn add_new_notar_fallback(
+    /// Adds a new notarize fallback certificate, we can use Notarize/NotarizeFallback/FastFinalize
+    pub fn add_new_notar_fallback_or_equivalent(
         &mut self,
         block @ (slot, _): Block,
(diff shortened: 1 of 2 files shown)


votor/src/certificate_pool/parent_ready_tracker.rs
```diff
@@ -84,8 +84,8 @@ impl ParentReadyTracker {
         }
     }
 
-    /// Adds a new notar-fallback certificate
-    pub fn add_new_notar_fallback(
+    /// Adds a new notarize fallback certificate, we can use Notarize/NotarizeFallback/FastFinalize
+    pub fn add_new_notar_fallback_or_equivalent(
```
AshwinSekar: nit: I would call this `add_new_notar_fallback_or_stronger`

AshwinSekar (approved):
thanks for catching this!

votor/src/certificate_pool/parent_ready_tracker.rs
wen-coding: Done.

PR #404, opened 2025-08-19
Add local trigger for ParentReady.

wen-coding:


wen-coding:
> a PR description would be nice :)

Done.

votor/src/event_handler.rs
```diff
@@ -387,6 +435,61 @@ impl EventHandler {
         Ok(votes)
     }
 
+    /// Under normal cases we should have a parent ready for first slot of every window.
+    /// But it could be we joined when the later slots of the window are finalized, then
+    /// we never saw the parent ready for the first slot and haven't voted for first slot
```
carllin, line 441: how is the problem that we haven't voted for the first slot resolved?

votor/src/event_handler.rs
```diff
@@ -387,6 +435,61 @@ impl EventHandler {
...
+    /// so we can't keep processing rest of the window. This is especially a problem for
+    /// cluster standstill.
```
carllin, line 443: Could you lay out the exact details/example of when this is a problem, because not every cluster standstill needs this fix

votor/src/event_handler.rs
```diff
@@ -387,6 +435,61 @@ impl EventHandler {
...
+        local_context: &mut LocalContext,
+    ) -> Option<(Slot, Block)> {
+        let (slot, block_id) = block;
+        let first_slot_of_window = first_of_consecutive_leader_slots(slot);
+        if first_slot_of_window == slot || first_slot_of_window == 0 {
+            // No need to trigger parent ready for the first slot of the window
+            return None;
+        }
+        if vctx.vote_history.highest_parent_ready_slot() >= Some(first_slot_of_window)
+            || !local_context.finalized_blocks.contains(&block)
```
carllin: If you root a later block and clear the finalized block from the the `finalized_blocks` set, is it guaranteed `add_missing_parent_ready` will also run on some later block in the window?

votor/src/event_handler.rs
```diff
@@ -387,6 +435,61 @@ impl EventHandler {
...
+        {
+            return None;
+        }
+        // If the block is missing, we can't trigger parent ready
+        let bank_forks_r = ctx.bank_forks.read().unwrap();
+        let bank = bank_forks_r.get(slot)?;
+        let first_slot_of_window_bank = bank_forks_r.get(first_slot_of_window)?;
```
carllin: what happens if a later root in the window has been set already and this `first_slot_of_window` has already been purged from bank forks?

votor/src/event_handler.rs
```diff
@@ -387,6 +435,61 @@ impl EventHandler {
...
+        drop(bank_forks_r);
+        if !bank.is_frozen() {
+            // We haven't finished replay for the block, so we can't trigger parent ready
+            return None;
+        }
+        if bank.block_id() != Some(block_id) {
```
AshwinSekar, line 482: Pointing this out as a sanity check, I believe it is true that a block within the leader window being finalized implies that the first slot in the window is also finalized (due to the consecutive check in tryNotar).
Thus it's impossible for this block_id to match, but first_slot_of_window_bank to not be an ancestor of bank.

votor/src/event_handler.rs
wen-coding: Once ParentReady is triggered for the first slot, we will vote Notarize for all previous blocks in the window.

votor/src/event_handler.rs
wen-coding: Added.

votor/src/event_handler.rs
wen-coding: We currently have a check that we only root a block if we have voted for the block:
https://github.com/anza-xyz/alpenglow/blob/2b71a1195d027188d0ca9c803f8356aa9bc58690/votor/src/event_handler.rs#L620
So, if we have rooted a later block, that means we don't need this ParentReady any more, we can keep voting for other blocks in the window.
We only need to add missing parent ready if we haven't rooted any block in the window.

votor/src/event_handler.rs
wen-coding: If somehow any block in the window is set to root, we will get a `None` here, and `add_missing_parent_ready` will return a None, in which case we will do nothing.

votor/src/event_handler.rs
wen-coding: Yeah, I think the only case block_id match here but we fail to get first_slot_of_window_bank later is we somehow rooted in the middle. Which means we somehow started voting in this window so we don't need this extra ParentReady.

Hmm, maybe the only exception is when we do a cluster restart from a slot in the middle of the window. In that case do we set up ParentReady and timeouts correctly now?

wen-coding:
> Echo-ing carl's comments I believe it's important to check the failure cases on state being cleaned up too early.
> 
> I think we need to ensure that we don't root bank forks until we're sure that we don't need to mark a local parent ready, probably easiest to enforce that we've emitted a ParentReady for the window, or simply change the order of the two checks.
> 
> The other place I was looking at was `should_ignore`, but I think that should also be fine if we change the order.

Sorry I'm confused, which order should I change here?

votor/src/event_handler.rs
AshwinSekar: yes cluster restart should be fine, we kick off a local parent ready when the node restarts.

my concern was could the `first_slot_of_window_bank` in our `bank_forks` be for a different version than the ancestor of `block_id`. I don't think this is possible though, just wanted to point it out.

AshwinSekar:
> > Echo-ing carl's comments I believe it's important to check the failure cases on state being cleaned up too early.
> > I think we need to ensure that we don't root bank forks until we're sure that we don't need to mark a local parent ready, probably easiest to enforce that we've emitted a ParentReady for the window, or simply change the order of the two checks.
> > The other place I was looking at was `should_ignore`, but I think that should also be fine if we change the order.
> 
> Sorry I'm confused, which order should I change here?

I was mistaken I think this is fine since as you pointed out we only cleanup after we have voted on the slot. Since the first leader slot requires a parent ready to vote, and the slots in the window require us to vote on the previous slot, i don't think it's possible that we need to get into a local trigger scenario but have rooted something in the window.

votor/src/event_handler.rs
carllin: Won't you get stuck again in this scenario if we do nothing, i.e. `C` can't vote? https://github.com/anza-xyz/alpenglow/blob/51f8095c22f2bf2ea410d27ad2949b128025c719/votor/src/event_handler.rs#L447-L449

votor/src/event_handler.rs
```diff
@@ -440,6 +440,20 @@ impl EventHandler {
...
+    ///    A 40%
+    ///    B 40%
+    ///    C 30%
+    /// A and B finalize block together up to slot 9, now A exited and C joined.
+    /// C sees block 9 as finalized, but it never had parent ready triggered for slot 8.
+    /// C can't vote for any slot in the window because there is no parent ready for slot 8.
+    /// While B is stuck because it is waiting for >60% of the votes to finalize slot 9.
+    /// The cluster will get stuck.
+    /// After we add the following function, C will see that block 9 is finalized yet
+    /// it never had parent ready for slot 8, so it will find the block id of 8 from
```
carllin: What if `C` started from a snapshot of `9` and has thus rooted `9`, and doesn't have `8` in its blockstore?

wen-coding: See Ashwin's explanation above: "yes cluster restart should be fine, we kick off a local parent ready when the node restarts."

wen-coding: https://github.com/anza-xyz/alpenglow/blob/fb7a69770bf6362b0ce372f98360ae3fbfd5ae69/votor/src/certificate_pool_service.rs#L210

votor/src/event_handler.rs
wen-coding: Hmm, so https://github.com/anza-xyz/alpenglow/blob/fb7a69770bf6362b0ce372f98360ae3fbfd5ae69/votor/src/certificate_pool_service.rs#L210
says we can trigger ParentReady from any slot (not just first slot of the window)

Should we trigger a ParentReady from finalized slot as well?

wen-coding: And, I don't think set_timeouts correctly handles setting timeouts from middle of a window?

votor/src/event_handler.rs
carllin: ah makes sense thanks

votor/src/event_handler.rs
carllin: ping

wen-coding: There are two cases where middle of the window can be set to root:
1. A ParentReady somehow triggered for first slot of the window, in which case C will be unstuck
2. Somehow we restarted on a slot in the middle of the window, in which case we should get a ParentReady on the slot we restart from.

That said, in case 2 we trigger a ParentReady on the slot we restart from, I'm now thinking I should do the same in this PR, just trigger a ParentReady on block 9 instead of block 8. Let me try that and see if local-cluster tests still pass.

wen-coding: Changed to trigger ParentReady on the finalized block, to be consistent with the case where we restart from the middle of a window.

Tests still pass.

votor/src/event_handler.rs
wen-coding: @AshwinSekar I changed ParentReady to trigger on the finalized block instead of the first slot of the window, to be consistent with restarting on a slot in the middle, what do you think? Tests look happy.

votor/src/event_handler.rs
carllin: Sanity check, does the ParentReady in the event handler have to be processed before we can root something? 

Can the parentready signal sit in the queue and be processed behind replay + root. I guess this is ok as long as the parentready signal eventually arrvies

wen-coding: We don't check ParentReady directly, currently `check_rootable_blocks` is called when a new `Finalize` cert comes in or when a block completes replay. When that happens all finalized are checked for potential of becoming the new root: https://github.com/anza-xyz/alpenglow/blob/ac1b0d7c95aaf4cf9845f0fb388915ca516da29a/votor/src/event_handler.rs#L619
```
                (slot > old_root
                    && vctx.vote_history.voted(slot)
                    && bank.is_frozen()
                    && bank.block_id().is_some_and(|bid| bid == block_id))

```
We check that:
1. slot is higher than old root
2. we voted for this slot (this is controlled by ParentReady)
3. the bank is frozen
4. the block_id for the bank in bank_forks is the same as that in finalized certs

carllin: So in this scenario here: https://github.com/anza-xyz/alpenglow/blob/51f8095c22f2bf2ea410d27ad2949b128025c719/votor/src/event_handler.rs#L447-L449

How is the bug triggered? what is the case in a node could see root 9 without triggering ParentReady on 8, I thought we always send a ParentReady on restart: https://github.com/anza-xyz/alpenglow/pull/404#discussion_r2287065278

wen-coding: The block 9 wasn't the root for C. C has root somewhere around block 0 I think, because it didn't join until A exited. It saw all the finalization certs A and B created, but never had a chance to vote on any block. Because C doesn't have ParentReady for block 8, it will never vote for block 8 and block 9, then the check "we voted for this slot" isn't true and we can't root block 8 and 9.

Does that make sense?

carllin: ah because we only trigger ParentReady on restart for the current root bank: https://github.com/anza-xyz/alpenglow/blob/fb7a69770bf6362b0ce372f98360ae3fbfd5ae69/votor/src/certificate_pool_service.rs#L207-L213

Can we solve this by triggering ParentReady for all the leaves in bank forks

wen-coding: We can, but that seems to be going the long way, and might not always work.
Think about what happened to a naive validator which joined with root at block 0.
Now the cluster is at block 10001, it needs to do ParentReady and repair all the blocks between 0 and 10001, which may or may not be available on its neighbors.

Instead, I believe the researchers designed the fast way:
- I joined the cluster and find out everyone else rooted block 10001
- I repaired and played block 10001, triggered a local ParentReady so I can Notarize block 10001, mark 10001 my new root, and trigger votes for 10002, 10003
- Now I can join the cluster and vote happily with everyone else

This is especially useful during cluster standstill, sync in the fastest way possible and move on.

votor/src/event_handler.rs
AshwinSekar: sorry I missed this, I believe this should be fine. The key is that we'll be able to vote skip and notarize-fallback if needed.

votor/src/event_handler.rs
```diff
@@ -387,6 +435,73 @@ impl EventHandler {
...
+        let Some(parent_block_id) = parent_bank.block_id() else {
+            // Maybe this bank is set to root after we drop bank_forks.
+            error!(
+                "{}: Unable to find block id for parent bank {parent_slot} to trigger parent ready",
+                local_context.my_pubkey
+            );
+            return None;
+        };
+        info!(
+            "{}: Triggering parent ready for missing first slot of window {slot} \
```
AshwinSekar: nit no longer first slot of window, maybe better to just say local trigger.

votor/src/event_handler.rs
```diff
@@ -387,6 +435,73 @@ impl EventHandler {
...
+            return None;
+        }
+        if vctx.vote_history.highest_parent_ready_slot() >= Some(first_slot_of_window)
+            || !local_context.finalized_blocks.contains(&block)
+        {
+            return None;
+        }
+        // If the block is missing, we can't trigger parent ready
+        let bank_forks_r = ctx.bank_forks.read().unwrap();
+        let bank = bank_forks_r.get(slot)?;
```
AshwinSekar: nit: get already clones the bank so you can just do 
`let bank = ctx.bank_forks.read().unwrap().get(slot)?;`

votor/src/event_handler.rs
wen-coding: Changed.

votor/src/event_handler.rs
wen-coding: Done.

votor/src/event_handler.rs
carllin: Yeah i guess it doesn't solve the problem if the new root is made off a non leaf node in bank forks, makes sense

votor/src/event_handler.rs
```diff
@@ -387,6 +435,70 @@ impl EventHandler {
...
+        vctx: &mut VotingContext,
+        local_context: &mut LocalContext,
+    ) -> Option<(Slot, Block)> {
+        let (slot, block_id) = block;
+        let first_slot_of_window = first_of_consecutive_leader_slots(slot);
+        if first_slot_of_window == slot || first_slot_of_window == 0 {
+            // No need to trigger parent ready for the first slot of the window
+            return None;
+        }
+        if vctx.vote_history.highest_parent_ready_slot() >= Some(first_slot_of_window)
```
carllin, line 471: This is safe because if a slot in this window got finalized, there cannot exist a skip certificate for this window, so there's no other fork that can have a higher parent ready right

votor/src/event_handler.rs
```diff
@@ -387,6 +435,70 @@ impl EventHandler {
...
+    /// While B is stuck because it is waiting for >60% of the votes to finalize slot 9.
+    /// The cluster will get stuck.
+    /// After we add the following function, C will see that block 9 is finalized yet
+    /// it never had parent ready for slot 9, so it will trigger parent ready for slot 9,
+    /// this means C will immediately vote Notarize for  slot 9, then vote Notarize for
+    /// all later slots. So B and C together can keep finalizing the blocks and unstuck the
+    /// cluster. If we get a finalization cert for later slots of the window and we have the
+    /// block replayed, trace back to the first slot of the window and emit parent ready.
+    fn add_missing_parent_ready(
+        block: Block,
```
carllin: nit: rename to finalized_block to emphasize this must be finalzied

votor/src/event_handler.rs
```diff
@@ -238,6 +269,19 @@ impl EventHandler {
                     received_shred,
                     stats,
                 )?;
+                if let Some((slot, block)) =
```
carllin: nit: `(ready_slot, parent_block)`

carllin (approved):
Looks good, thanks for thinking through the nuances

votor/src/event_handler.rs
wen-coding: It could also be that the cluster finalized this slot and the slots after them, but you are replaying blocks super slowly so you just got to this slot.

In any case, I think if we have ParentReady for a future slot in a future window, we shouldn't care about the current slot window any more, best strategy is fast forward to that future window.

votor/src/event_handler.rs
wen-coding: Done.

votor/src/event_handler.rs
wen-coding: Done.

PR #407, opened 2025-08-20
bug: handle failures to send in retransmit_stage

akhi3030:
#### Problem
If the receiving end of the channel is dropped, sending on it fails which currently results in a panic.  

#### Summary of Changes
Instead of panicking when that happens, exit the infinite loop more gracefully.

### DIFF
--- a/turbine/src/retransmit_stage.rs
+++ b/turbine/src/retransmit_stage.rs
@@ -253,7 +253,7 @@ fn retransmit(
             shred_buf.push(shreds);
         }
         Err(TryRecvError::Disconnected) => {
-            warn!("retransmint_receiver became disconnected.");
+            warn!("retransmit_receiver became disconnected.");
             return Err(());
         }
 


turbine/src/retransmit_stage.rs
```diff
@@ -244,15 +244,19 @@ fn retransmit(
...
     // Try to receive shreds from the channel without blocking. If the channel
     // is empty precompute turbine trees speculatively. If no cache updates are
     // made then block on the channel until some shreds are received.
     match retransmit_receiver.try_recv() {
         Ok(shreds) => {
             shred_buf.push(shreds);
         }
-        Err(TryRecvError::Disconnected) => return Err(RecvError),
+        Err(TryRecvError::Disconnected) => {
+            warn!("retransmint_receiver became disconnected.");
```
AshwinSekar: nit: 
```suggestion
            warn!("retransmit_receiver became disconnected.");
```

AshwinSekar (approved):

PR #410, opened 2025-08-21
feat: catchup to `agave`, until 2025-08-07 + bank-related modifications

ksn6:
NOTE: in addition to catching up, we make the following modifications:

(1) To accommodate https://github.com/anza-xyz/agave/commit/b82d3bbba3d5ab170bd65053b3b2420f44e9e9eb: 
- Eliminate `root_bank_cache.rs` and use `SharableBank` instead
- Gotta correct this spelling error in Agave later.

(2) To accommodate https://github.com/anza-xyz/agave/commit/dc0f51a3c743f875a165f8f9099f2beee4634b7d: 
- Eliminate `LeaderBankNotifier` in `block_creation_loop.rs`
- Write a "bank cleared notifier" condvar that notifies the loop once the bank has been cleared (or a timeout happens).

These modifications are important to make now rather than later; otherwise, we cannot cherry-pick all commits that get us on track to Agave. It's dangerous to selectively cherry-pick commits; we don't want to risk cherry-picking a later commit that indirectly depends on something we skipped, or else weird bugs may occur.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -371,7 +371,6 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
 
     // get latest identity pubkey during startup
     let mut my_pubkey = cluster_info.id();
-    // let leader_bank_notifier = poh_recorder.read().unwrap().new_leader_bank_notifier();
 
     let mut ctx = LeaderContext {
         my_pubkey,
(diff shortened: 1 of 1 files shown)


core/src/block_creation_loop.rs
```diff
@@ -469,55 +469,67 @@ pub fn start_loop(config: BlockCreationLoopConfig) {
...
             );
 
+            // Get the notifier at the start
+            let bank_cleared_notifier = poh_recorder.read().unwrap().bank_cleared_notifier.clone();
+
             // Start measuring bank completion time
             let mut bank_completion_measure = Measure::start("bank_completion");
 
-            leader_bank_notifier.wait_for_completed(remaining_slot_time);
+            let (lock, cvar) = &*bank_cleared_notifier;
```
AshwinSekar: for now lets just put a sleep rather than having the extra notifier here. I need to rework some stuff here, but currently we'll never get into case 2 (bank cleared by banking stage), so there's no reason to wait on a clear notification:
- sleep until remaining slot time
- keep remaining code as is to clear the bank.

ksn6: Done

wen-coding (approved):

Issue #365, opened 2025-08-07
Rename `BLSMessage` to `ConsensusMessage` or `AlpenglowMessage` or something appropriate


akhi3030:
BLS happens to be an implementation detail.  These are core consensus messages and do not have to contain the prefix BLS in them.

wen-coding:
If we are renaming CertificatePool to ConsensusPool, we can rename it to ConsensusMessage to be consistent.

But please don't touch any BLS in BLS verification or BLS aggregation. We do want to be explicit these are using BLS.

PR #413, opened 2025-08-21
refactor: rename `BLSMessage` to `ConsensusMessage`

akhi3030:
#### Problem
BLS happens to be implementation detail.  It should not be captured in the name of the message.

#### Summary of Changes

Renames `BLSMessage` to `ConsensusMessage`

Fixes #

Closes https://github.com/anza-xyz/alpenglow/issues/365.

### DIFF
--- a/votor/src/certificate_pool_service.rs
+++ b/votor/src/certificate_pool_service.rs
@@ -25,9 +25,7 @@ use {
         leader_schedule_utils::last_of_consecutive_leader_slots,
     },
     solana_pubkey::Pubkey,
-    solana_runtime::{
-        bank::Bank, bank_forks::SharableBank, vote_sender_types::BLSVerifiedMessageReceiver,
-    },
+    solana_runtime::{bank::Bank, bank_forks::SharableBank},
     solana_votor_messages::consensus_message::{CertificateMessage, ConsensusMessage},
     stats::CertificatePoolServiceStats,
     std::{


runtime/src/vote_sender_types.rs
```diff
@@ -1,11 +1,7 @@
...
     crossbeam_channel::{Receiver, Sender},
     solana_vote::vote_parser::ParsedVote,
-    solana_votor_messages::bls_message::BLSMessage,
 };
 
 pub type ReplayVoteSender = Sender<ParsedVote>;
 pub type ReplayVoteReceiver = Receiver<ParsedVote>;
-
-pub type BLSVerifiedMessageSender = Sender<BLSMessage>;
-pub type BLSVerifiedMessageReceiver = Receiver<BLSMessage>;
```
wen-coding, line 11: I think the code would look cleaner if you continue using these instead of using Sender/Receiver<ConsensusMessage> everywhere, just rename them.

akhi3030: I feel quite opposite to be honest.  The alias is longer to type than the actual type.  And The alias doesn't tell us what the actual type is so as a reader I have to keep looking up with it is.  

I have been reading the code a lot lately and found that the alias was not helping at all.

wen-coding: Sure, I don't have strong feelings about it.

core/src/tpu.rs
```diff
@@ -347,7 +348,7 @@ impl Tpu {
             let verifier = BLSSigVerifier::new(
                 root_bank_cache,
                 verified_vote_sender.clone(),
-                bls_verified_message_sender,
+                consensus_message_sender,
```
wen-coding: The incoming messages are also of type ConsensusMessage, I do prefer to keep verified in the name here.

votor/src/voting_utils.rs
```diff
@@ -75,7 +75,7 @@ impl GenerateVoteTxResult {
             | Self::WaitForStartupVerification
             | Self::WaitToVoteSlot(_)
             | Self::NoRankFound => false,
-            Self::Tx(_) | Self::BLSMessage(_) => false,
+            Self::Tx(_) | Self::Consensus(_) => false,
```
wen-coding: Why is this Consensus instead of ConsensusMessage?

core/src/tpu.rs
akhi3030: Updated in https://github.com/anza-xyz/alpenglow/pull/413/commits/5fcf524228dbf4d119b412073808ef0fe7632e36.

votor/src/voting_utils.rs
akhi3030: I thought it was a bit verbose.  Updated in https://github.com/anza-xyz/alpenglow/pull/413/commits/5fcf524228dbf4d119b412073808ef0fe7632e36.

wen-coding: LGTM, can approve after conflicts are resolved.

akhi3030: Merged the latest master into the branch and resolved the conflicts.

wen-coding (approved):

PR #416, opened 2025-08-23
Limit vote accounts to 2000 and remove those without BLS pubkey in Alpenglow epoch

wen-coding:
Per Alpenglow SIMD, we should limit the number of vote accounts to 2000.

We should also remove vote accounts without BLS pubkey because the BLS verification will never succeed for them.


feature-set/src/lib.rs
```diff
@@ -1366,6 +1371,7 @@ pub static FEATURE_NAMES: LazyLock<AHashMap<Pubkey, &'static str>> = LazyLock::n
         (raise_block_limits_to_100m::id(), "SIMD-0286: Raise block limit to 100M"),
         (raise_account_cu_limit::id(), "SIMD-0306: Raise account CU limit to 40% max"),
         (raise_cpi_nesting_limit_to_8::id(), "SIMD-0296: Raise CPI nesting limit from 4 to 8"),
+        (limit_validators_for_alpenglow::id(), "SIMD-0306: Limit validators for Alpenglow"),
```
AshwinSekar: is this supposed to be SIMD-0326?

feature-set/src/lib.rs
```diff
@@ -1125,6 +1126,10 @@ pub mod raise_cpi_nesting_limit_to_8 {
     solana_pubkey::declare_id!("6TkHkRmP7JZy1fdM6fg5uXn76wChQBWGokHBJzrLB3mj");
 }
 
+pub mod limit_validators_for_alpenglow {
```
AshwinSekar, line 1125: what's the purpose of having a separate feature flag here? to turn this on before the alpenglow migration?

runtime/src/bank.rs
```diff
@@ -212,6 +212,7 @@ pub(crate) mod tests;
 pub const SECONDS_PER_YEAR: f64 = 365.25 * 24.0 * 60.0 * 60.0;
 
 pub const MAX_LEADER_SCHEDULE_STAKES: Epoch = 5;
+const MAX_ALPENGLOW_VOTE_ACCOUNTS: usize = 2000;
```
AshwinSekar: think it would be better if this was a part of `vote_account.rs` instead

svm-feature-set/src/lib.rs
```diff
@@ -38,6 +38,7 @@ pub struct SVMFeatureSet {
     pub disable_zk_elgamal_proof_program: bool,
     pub reenable_zk_elgamal_proof_program: bool,
     pub raise_cpi_nesting_limit_to_8: bool,
+    pub limit_validators_for_alpenglow: bool,
```
AshwinSekar: hmm this isn't a svm feature right? since it only affects runtime

runtime/src/bank.rs
```diff
@@ -1155,7 +1156,7 @@ impl Bank {
         // genesis needs stakes for all epochs up to the epoch implied by
         //  slot = 0 and genesis configuration
         {
-            let stakes = bank.stakes_cache.stakes().clone();
+            let stakes = bank.get_stakes_for_epoch_stakes();
```
AshwinSekar: epoch stakes is also used for other purposes, e.g. the turbine shuffle. I don't think it's correct to remove the nodes that are less than the top 2000 from this map. Instead could we limit the bls pubkey map?

feature-set/src/lib.rs
wen-coding: Yes, I think it's a bit risky to turn on at the same time as Alpenglow migration, I want to turn it on one or two epochs before Alpenglow migration. So we can give a ping to people who are too lazy to update bls pubkey.

svm-feature-set/src/lib.rs
wen-coding: Good point, I was looking for RuntimeFeatureSet but didn't find it, then I saw:
`pub fn runtime_features(&self) -> SVMFeatureSet {`
And I assume this is it. Do we have a separate FeatureSet just for runtime?

feature-set/src/lib.rs
wen-coding: Oops. changed.

runtime/src/bank.rs
wen-coding: Since I'm touching Stakes which in turn calls VoteAccounts, I'm putting it in Stakes.

runtime/src/bank.rs
wen-coding: Limiting the bls pubkey map is not enough, since total_stake needs to change as well. Also currently the reward calculation and leader schedule calculation calls epoch_stakes.stakes() directly.

It's a bit weird to not count some people outside the 2000 limit as "not staked" and not giving them vote credits yet use them in Turbine shuffle. I think it's probably best to keep everything in sync.

runtime/src/bank.rs
wen-coding: Hmm, hold on, I need to turn it off in most of the Non-alpenglow tests, the bls pubkey check is causing trouble.

runtime/src/bank.rs
AshwinSekar: can we use total stake from the bls pubkey map? bit hesitant about changing these stake values on bank here.

Also for v1 we are not planning to change Turbine, so we should still keep propagating shreds to everyone until we have rotor/shadow turbine

wen-coding: Hmm, I thought Turbine now does forward to zero-staked nodes? We are not changing that in v1 launch.

I think it's just very confusing to limit voting to one stake distribution and rest of the code to another (e.g.
rewards calculation would be off because the percentage of stake distribution is different if we have 2000 validators,
leader schedule calculation may now put "non-staked" validators into leader schedule)

runtime/src/bank.rs
```diff
@@ -5646,6 +5646,18 @@ impl Bank {
     pub fn get_collector_fee_details(&self) -> CollectorFeeDetails {
         self.collector_fee_details.read().unwrap().clone()
     }
+
+    pub fn get_stakes_for_epoch_stakes(&self) -> Stakes<StakeAccount<Delegation>> {
```
carllin: nit: can we name this something clearer, like `get_alpenglow_epoch_stakes`, or `get_top_epoch_stakes`

feature-set/src/lib.rs
carllin: does this require a governance vote

vote/src/vote_account.rs
```diff
@@ -75,6 +75,39 @@ pub struct VoteAccounts {
     >,
 }
 
+impl VoteAccounts {
+    pub fn clone_and_filter_for_alpenglow(&self, maximum_accounts: usize) -> VoteAccounts {
+        let mut pubkeys_and_stakes_to_sort: Vec<(Pubkey, u64)> = self
+            .vote_accounts
+            .iter()
+            .filter_map(|(pubkey, (stake, vote_account))| {
+                if vote_account.bls_pubkey().is_some() && *stake != 0u64 {
```
carllin, line 95: is this sorting algorithm the same as the one used in the BLS rank map here? https://github.com/anza-xyz/alpenglow/blob/f114f25cb8822ffaefeb1cb575f29c721472db8b/runtime/src/epoch_stakes.rs#L39-L46. 

Would it be useful as a utility function that sorts any vector that has a stake and a pubkey so that it's easier to maintain the consistency between these two modules?

vote/src/vote_account.rs
```diff
@@ -75,6 +75,39 @@ pub struct VoteAccounts {
...
+            .filter_map(|(pubkey, (stake, vote_account))| {
+                if pubkeys_to_keep.contains_key(pubkey) {
+                    Some((*pubkey, (*stake, vote_account.clone())))
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+}
```
carllin:
```
pub fn clone_and_filter_for_alpenglow(&self, maximum_accounts: usize) -> VoteAccounts {
        if maximum_accounts == 0 {
            return VoteAccounts::default();
        }

        // Collect refs: (&Pubkey, stake, &VoteAccount), filter for BLS + nonzero stake.
        let mut entries: Vec<(&Pubkey, u64, &VoteAccount)> = self
            .vote_accounts
            .iter()
            .filter_map(|(pk, (stake, va))| {
                (va.bls_pubkey().is_some() && *stake != 0).then_some((pk, *stake, va))
            })
            .collect();
        
        // If there are more than we want, partition so the top-N by (stake desc, pubkey desc) are first.
        if entries.len() > maximum_accounts {
            let key = |e: &(&Pubkey, u64, &VoteAccount)| (Reverse(e.1), Reverse(*e.0));
            // Puts the element that would be at index `maximum_accounts` in sorted order into place,
            // with all "better" elements before it (order within the prefix is unspecified).
            let _ = entries.select_nth_unstable_by_key(maximum_accounts, key);
            entries.truncate(maximum_accounts);
        }

        // Clone only what we keep.
        entries
            .into_iter()
            .map(|(pk, stake, va)| (*pk, (stake, va.clone())))
            .collect()
```
Cool little cursor assisted version with fewer clones and iteration, and a Quick select, linear run time to find the `nth` largest element using `select_nth_unstable_by_key` instad of sorting the entire set

runtime/src/bank.rs
carllin: Turbine gets zero-staked nodes from gossip's tvu_peers right? 

1. https://github.com/anza-xyz/alpenglow/blob/f114f25cb8822ffaefeb1cb575f29c721472db8b/turbine/src/cluster_nodes.rs#L305
2. https://github.com/anza-xyz/alpenglow/blob/f114f25cb8822ffaefeb1cb575f29c721472db8b/turbine/src/cluster_nodes.rs#L348

so unstaked nodes I think should be unaffected.

feature-set/src/lib.rs
wen-coding: I don't think this one needs a governance vote, because it's straightforward and covered in SIMD 326 already. The other part (take 1.6 SOL out of vote account at beginning of Epoch) maybe, because we do want to make sure the validators aren't charged twice. I'm thinking charging around the same time we calculate EpochStakes, maybe add a bit somewhere to mark the fee has been charged for some specific Epoch, could use some suggestions there.

We can write the two into one SIMD.

svm-feature-set/src/lib.rs
wen-coding: Removed.

vote/src/vote_account.rs
wen-coding: Err, that was in another function, so it seems a bit overkill passing in a two-line function. But I don't feel strongly about it.

runtime/src/bank.rs
wen-coding: Actually plan to use this for pre-Alpenglow as well, pre-Alpenglow it just does a simple clone. Renamed to `copy_and_filter_from_stakes_cache`, does this sound better?

vote/src/vote_account.rs

I think we should disallow maximum_accounts to be 0, so I added a `panic!`.

runtime/src/bank.rs
wen-coding: Tests are now happy.

runtime/src/bank.rs
carllin: I think `filter` is a bit vague as to what the filter is, or implies a filter function is being passed.

a name that conveys it gets the "top" would be nice, maybe  something along the lines of `get_top_epoch_stakes`

wen-coding: Sure, done.

vote/src/vote_account.rs
carllin: a utility function might be nice just to remind ourselves that if the sort changes, we don't have to figure out where all the downstream effects are

wen-coding: Changed, how does this look?

wen-coding:
Err, I was implementing VAT collection and I realized this doesn't quite work with VAT collection. We have to make sure we only charge 1.6 SOL to whoever is selected here, so selection and charging 1.6 SOL probably need to happen together. I'll rethink the design and work on a SIMD next.

Issue #408, opened 2025-08-20
Some epoch level metrics for alpenglow


akhi3030:
@rogerANZA, @qkniep, and I brainstormed on what kind of metrics we would like to see for Alpenglow.

https://github.com/qkniep/alpenglow/pull/103 captures the intent at a high level.  This issue is to track the work to implement them in agave / Alpenglow fork.

Additional context discussed with @bw-solana:

- The amount of data we are proposing to collect doesn't seem to be a lot and the overhead should be fine.
- We probably do not want to report all the metrics from all the nodes in a single go at the epoch boundary.  These metrics are probably not very time sensitive so we should stagger the reporting by sending a bit of data at a time and pausing for random timeouts in between so that not all nodes are reporting at the same time.
- We have to think about how to report histograms.  We could report the actual buckets if that is not a lot of data.  Alternatively, we could either compute the standard stats (avg, std dev, 90, max, etc.) from them or instead of collecting histograms, we could track the actual stats.

akhi3030:
We were discussing that we would want to capture some stats on repair and bls signature mismatches.

PR #418, opened 2025-08-25
feat: implements tracking of vote and leader metrics

akhi3030:
Part of https://github.com/anza-xyz/alpenglow/issues/408.

- Implements tracking of vote and leader metrics.
- uses the timer manager to track when slots started to be able to compute durations when events are received.
- does not actually report data when a new epoch starts, that can be implemented in a subsequent PR.

### DIFF
--- a/votor/src/consensus_metrics.rs
+++ b/votor/src/consensus_metrics.rs
@@ -88,8 +88,8 @@ pub enum RecordBlockHashError {
     SlotNotFound,
 }
 
-/// Tracks various Alpenglow related metrics.
-pub struct AgMetrics {
+/// Tracks various Consensus related metrics.
+pub struct ConsensusMetrics {
     /// Used to track this node's view of how the other nodes on the network are voting.
     node_metrics: BTreeMap<Pubkey, NodeVoteMetrics>,
     /// Used to track when this node received blocks from different leaders in the network.
@@ -103,7 +103,7 @@ pub struct AgMetrics {
     current_epoch: Epoch,
 }
 
-impl AgMetrics {
+impl ConsensusMetrics {
     pub fn new(epoch: Epoch) -> Self {
         Self {
             node_metrics: BTreeMap::default(),
--- a/votor/src/event_handler.rs
+++ b/votor/src/event_handler.rs
@@ -245,7 +245,7 @@ impl EventHandler {
             VotorEvent::Block(CompletedBlock { slot, bank }) => {
                 debug_assert!(bank.is_frozen());
                 {
-                    let mut metrics_guard = vctx.ag_metrics.write();
+                    let mut metrics_guard = vctx.consensus_metrics.write();
                     match metrics_guard.record_block_hash_seen(*bank.collector_id(), slot) {
                         Ok(()) => (),
                         Err(err) => {
--- a/votor/src/lib.rs
+++ b/votor/src/lib.rs
@@ -4,8 +4,8 @@ use {
     std::time::Duration,
 };
 
-pub mod alpenglow_metrics;
 pub mod commitment;
+pub mod consensus_metrics;
 pub mod consensus_pool;
 mod consensus_pool_service;
 pub mod event;
--- a/votor/src/timer_manager.rs
+++ b/votor/src/timer_manager.rs
@@ -6,7 +6,7 @@ mod stats;
 mod timers;
 
 use {
-    crate::{alpenglow_metrics::AgMetrics, event::VotorEvent, DELTA_BLOCK, DELTA_TIMEOUT},
+    crate::{consensus_metrics::ConsensusMetrics, event::VotorEvent, DELTA_BLOCK, DELTA_TIMEOUT},
     crossbeam_channel::Sender,
     parking_lot::RwLock as PlRwLock,
     solana_clock::Slot,
@@ -32,13 +32,13 @@ impl TimerManager {
     pub(crate) fn new(
         event_sender: Sender<VotorEvent>,
         exit: Arc<AtomicBool>,
-        ag_metrics: Arc<PlRwLock<AgMetrics>>,
+        consensus_metrics: Arc<PlRwLock<ConsensusMetrics>>,
     ) -> Self {
         let timers = Arc::new(PlRwLock::new(Timers::new(
             DELTA_TIMEOUT,
             DELTA_BLOCK,
             event_sender,
-            ag_metrics,
+            consensus_metrics,
         )));
         let handle = {
             let timers = Arc::clone(&timers);
@@ -78,8 +78,8 @@ mod tests {
     fn test_timer_manager() {
         let (event_sender, event_receiver) = unbounded();
         let exit = Arc::new(AtomicBool::new(false));
-        let ag_metrics = Arc::new(PlRwLock::new(AgMetrics::new(0)));
-        let timer_manager = TimerManager::new(event_sender, exit.clone(), ag_metrics);
+        let consensus_metrics = Arc::new(PlRwLock::new(ConsensusMetrics::new(0)));
+        let timer_manager = TimerManager::new(event_sender, exit.clone(), consensus_metrics);
         let slot = 52;
         let start = Instant::now();
         timer_manager.set_timeouts(slot);
--- a/votor/src/timer_manager/timers.rs
+++ b/votor/src/timer_manager/timers.rs
@@ -1,6 +1,7 @@
 use {
     crate::{
-        alpenglow_metrics::AgMetrics, event::VotorEvent, timer_manager::stats::TimerManagerStats,
+        consensus_metrics::ConsensusMetrics, event::VotorEvent,
+        timer_manager::stats::TimerManagerStats,
     },
     crossbeam_channel::Sender,
     parking_lot::RwLock as PlRwLock,
@@ -53,7 +54,7 @@ impl TimerState {
         &mut self,
         delta_block: Duration,
         now: Instant,
-        ag_metrics: &PlRwLock<AgMetrics>,
+        consensus_metrics: &PlRwLock<ConsensusMetrics>,
     ) -> Option<VotorEvent> {
         match self {
             Self::WaitDeltaTimeout { window, timeout } => {
@@ -62,7 +63,7 @@ impl TimerState {
                     return None;
                 }
                 let slot = *window.front().unwrap();
-                ag_metrics.write().record_start_of_slot(slot);
+                consensus_metrics.write().record_start_of_slot(slot);
                 let timeout = now.checked_add(delta_block).unwrap();
                 *self = Self::WaitDeltaBlock {
                     window: window.to_owned(),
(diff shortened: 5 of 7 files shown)


core/src/replay_stage.rs
```diff
@@ -3618,6 +3623,8 @@ impl ReplayStage {
                 //
                 // For non leader banks (2) is always true, so notify here
                 if *is_alpenglow_migration_complete && bank.block_id().is_some() {
+                    // XXX: this is the leader producing the block?
```
bw-solana: The block associated with `bank` may or may not have been produced by this node. We can check by comparing `bank.collector_id` against our `pubkey`

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -88,6 +91,9 @@ impl SigVerifier for BLSSigVerifier {
                         stats_updater.received_malformed += 1;
                         continue;
                     };
+                    self.ag_metrics
+                        .write()
+                        .record_vote(*pubkey, vote, Duration::MAX);
```
bw-solana: do we also want to record skips?

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -201,6 +201,8 @@ pub(super) fn set_block_id_and_send(
 ) -> Result<()> {
     bank.set_block_id(Some(block_id));
     if bank.is_frozen() {
+        // XXX: this is where the block was received over the network?
```
bw-solana: This is only called when:

1. This node produced the block (was leader)
2. Replay finished processing this block already

core/src/sigverifier/bls_sigverifier.rs
bw-solana: I'm not sure if updating stats in sigverify makes the most sense. I'm wondering if we should add vote stats near `fn process_consensus_message`? I believe this would also allow us to observe our own votes, so we wouldn't need to call `record_vote` in two places

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,116 @@
...
+    skip_fallback: (u64, Histogram),
+    final_: (u64, Histogram),
+}
+
+impl VoteMetrics {
+    fn new() -> Self {
+        // TODO: pick sensible parameters for the buckets
+        let histogram = Histogram::new(10, 10).unwrap();
+        Self {
+            notar: (0, histogram.clone()),
```
bw-solana: struct here might be good to avoid using tuple notation elsewhere

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,116 @@
...
+/// The global metrics datastructure.
+///
+/// Each node has one of these.
+#[derive(Default)]
+pub struct AlpenglowMetrics {
+    /// Stores the node's view of how the nodes are voting.
+    vote_metrics: BTreeMap<Pubkey, VoteMetrics>,
+    /// Stores the node's view of how the nodes are producing blocks.
+    leader_metrics: BTreeMap<Pubkey, Histogram>,
+}
```
bw-solana: I think if we split these out (vote/leader metrics in their own, local structs/vars), we could avoid a lot of the locking and plumbing in this PR

akhi3030: Just to make sure I understand, I think you are suggesting something like the following:

```
struct VoteMetrics {
  metrics:  BTreeMap<Pubkey, VoteMetrics>,
}

struct LeaderMetrics {
  metrics: BTreeMap<Pubkey, Histogram>,
}
```

And then this way, we will just be plumbing one of the two structs through the rest of the codebase?

bw-solana: yeah, that's right.

For example, I think the VoteMetrics could be created at a much lower level and avoid having to get plumbed at TPU/TVU level

core/src/sigverifier/bls_sigverifier.rs
akhi3030: moved the recording to `process_consensus_message`.

votor/src/alpenglow_metrics.rs
akhi3030: Split into two structs.

votor/src/alpenglow_metrics.rs
akhi3030: switched to struct.

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,175 @@
...
+    std::{
+        collections::BTreeMap,
+        time::{Duration, Instant},
+    },
+};
+
+/// Tracks metrics for a single [`Vote`]
+#[derive(Debug, Clone)]
+struct Metric {
+    count: u64,
```
bw-solana: I think Histogram already maintains a count. Can we just use that directly?

votor/Cargo.toml
```diff
@@ -32,6 +32,7 @@ bs58 = { workspace = true }
 crossbeam-channel = { workspace = true }
 dashmap = { workspace = true, features = ["rayon", "raw-api"] }
 etcd-client = { workspace = true, features = ["tls"] }
+histogram = "0.11.3"
```
bw-solana: any reason why we want to maintain multiple versions? Could we bump the top level workspace version?

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,175 @@
...
+    notar: Metric,
+    notar_fallback: Metric,
+    skip: Metric,
+    skip_fallback: Metric,
+    final_: Metric,
+}
+
+impl Default for NodeVoteMetrics {
+    fn default() -> Self {
+        let histogram = Histogram::new(7, 20).unwrap();
```
bw-solana: would be good to explain why we chose these.

I'm also wondering if 20 is enough... I think that mean we will panic for any latency greater than 1 second

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl NodeVoteMetrics {
+    /// Records metrics for when `vote` was received after `elapsed` time has passed since the start of the slot.
+    fn record_vote(&mut self, vote: &Vote, elapsed: Duration) {
+        match vote {
+            Vote::Notarize(_) => {
+                self.notar
+                    .histogram
+                    .increment(elapsed.as_micros() as u64)
+                    .unwrap();
```
bw-solana: If we're going to unwrap, we should clamp the input first

votor/src/consensus_metrics.rs
```diff
@@ -0,0 +1,175 @@
...
+    SlotNotFound,
+}
+
+/// Tracks various Alpenglow related metrics.
+pub struct AgMetrics {
+    /// Used to track this node's view of how the other nodes on the network are voting.
+    node_metrics: BTreeMap<Pubkey, NodeVoteMetrics>,
+    /// Used to track when this node received blocks from different leaders in the network.
+    leader_metrics: BTreeMap<Pubkey, Histogram>,
+    /// Tracks when individual slots began.
```
bw-solana, line 97: might be nice to spell out exactly when we determine a slot starts (e.g. first shred observed or when we reset onto slot N-1)

votor/src/event_handler.rs
```diff
@@ -244,6 +244,18 @@ impl EventHandler {
             // Block has completed replay
             VotorEvent::Block(CompletedBlock { slot, bank }) => {
                 debug_assert!(bank.is_frozen());
+                {
+                    let mut metrics_guard = vctx.ag_metrics.write();
+                    match metrics_guard.record_block_hash_seen(*bank.collector_id(), slot) {
```
bw-solana, line 249: probably doesn't matter, but it's possible we could record multiple datapoints for the same slot if there is equivocation

votor/Cargo.toml
akhi3030: The new version is much more expressive and ergonomic.  I was trying to upgrade the version for the rest of agave but it is turning into quite a complicated PR.  I think I can live with the older version for this PR.  We can explore upgrading the crate for all of agave separately.

bw-solana: sounds good. Multiple versions is probably fine for Alpenglow, but we'll need to unify in upstream Agave

votor/src/alpenglow_metrics.rs
akhi3030: Yes, found something.  Will remove.

votor/src/alpenglow_metrics.rs
akhi3030: Increased to 10s and added documentation.

votor/src/alpenglow_metrics.rs
akhi3030: Removed the `unwrap`s in favour of `warn` logging.

votor/src/consensus_metrics.rs
akhi3030: improved the documentation.

votor/src/event_handler.rs
akhi3030: Good point.  I am hoping that that will be rare enough that it will not impact visibility ... 🤞

akhi3030:
@bw-solana: addressed your comments above.  PTAL.

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,192 @@
...
+        .max_value(10_000_000)
+        .build()
+        .unwrap()
+}
+
+/// Tracks metrics for a single [`Vote`]
+#[derive(Debug, Clone)]
+struct Metric {
+    histogram: Histogram,
+}
```
bw-solana: nit: could just use `Histogram` directly

votor/src/consensus_metrics.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub fn record_block_hash_seen(
+        &mut self,
+        leader: Pubkey,
+        slot: Slot,
+    ) -> Result<(), RecordBlockHashError> {
+        let Some(start) = self.start_of_slot.get(&slot) else {
+            return Err(RecordBlockHashError::SlotNotFound);
+        };
+        let elapsed = start.elapsed().as_micros();
+        let elapsed = match elapsed.try_into() {
```
bw-solana, line 137: nit: could probably collapse these lines

bw-solana (approved):
Left a couple small nits, but LGTM!

votor/src/alpenglow_metrics.rs
akhi3030: good spot.  It was leftover from when I was also tracking the count.  Will remove.

votor/src/consensus_metrics.rs
akhi3030: I wanted to but I am printing `elapsed` on line 149 below that's why it seems simpler to not collapse them.

votor/src/alpenglow_metrics.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Errors returned from [`AgMetrics::record_block_hash_seen`].
+#[derive(Debug)]
+pub enum RecordBlockHashError {
+    /// Could not find start of slot entry.
+    SlotNotFound,
+}
+
+/// Tracks various Alpenglow related metrics.
+pub struct AgMetrics {
```
AshwinSekar: very minor nit, would name this `Metrics` or `ConsensusMetrics` (same with the file). Essentially want to avoid calling anything that's already in the `votor` crate Ag - and one day in the future we won't have TowerBFT so everything will be alpenglow anyway.

akhi3030: makes sense.  Renamed.

PR #422, opened 2025-08-27
genesis: fix alpenglow feature activation

AshwinSekar:
Since `alpenglow` was removed from `FEATURE_NAMES` we need to activate it separately


bw-solana:
https://buildkite.com/anza/infra-invalidator/builds/2526

PR #426, opened 2025-08-28
event stats cleanup

bw-solana:
#### Problem
I was looking at some metrics from running 10 node cluster and deciding to polish some small things related to Event Handler stats while I was in the area.

#### Summary of Changes

- add `_us` postfix to some timing variables so it' obvious what the units are
- Add comments for each stat
- Add EventCountAndTime to make it obvious what each field is (count & time w/ units)

### DIFF
--- a/votor/src/event_handler/stats.rs
+++ b/votor/src/event_handler/stats.rs
@@ -225,13 +225,17 @@ impl EventHandlerStats {
                 "event_handler_received_event_count_and_timing",
                 ("event", format!("{:?}", event), String),
                 ("count", *count as i64, i64),
-                ("elapsed", *time_us as i64, i64)
+                ("elapsed_us", *time_us as i64, i64)
             );
         }
         datapoint_info!(
             "event_handler_timing",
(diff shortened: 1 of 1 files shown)


votor/src/event_handler/stats.rs
```diff
@@ -196,18 +219,19 @@ impl EventHandlerStats {
...
                 ("count", *count as i64, i64),
-                ("elapsed", *ms as i64, i64)
+                ("elapsed", *time_us as i64, i64)
             );
         }
         datapoint_info!(
             "event_handler_timing",
-            ("receive_event_time", self.receive_event_time as i64, i64),
-            ("send_vote_time", self.send_vote_time as i64, i64),
+            ("receive_event_time", self.receive_event_time_us as i64, i64),
```
wen-coding: Do you want to add _us to "receive_event_time" here as well?

wen-coding (dismissed):

votor/src/event_handler/stats.rs
bw-solana: yeah, good call

wen-coding (approved):

PR #427, opened 2025-08-28
parent_ready: choose parent with most skips

AshwinSekar:
#### Problem
When there are multiple parent readys available to produce our block off of we should pick the smallest by slot # parent, as that corresponds to the fork with the most skips.

This is important during duplicate block attacks, as blocks with multiple notarize-fallback certificates will eventually get a skip and switching forks in replay incurs a cost which eats into the skip timer.

#### Summary of Changes
Choose the smallest parent ready when selecting the parent for block production

### DIFF
--- a/votor/src/certificate_pool/parent_ready_tracker.rs
+++ b/votor/src/certificate_pool/parent_ready_tracker.rs
@@ -213,14 +213,10 @@ impl ParentReadyTracker {
             // and catching up. Either way we should not attempt to produce this slot
             return BlockProductionParent::MissedWindow;
         }
-        // TODO: for duplicate blocks we should adjust this to choose the
-        // parent with the least amount of duplicate blocks if possible.
-        // Notice that each scenario with multiple NotarFallbacks also will eventually
-        // have a skip for that slot, so prefer the skip if we've received it.
         match self
             .slot_statuses
             .get(&slot)
-            .and_then(|ss| ss.parents_ready.iter().max().copied())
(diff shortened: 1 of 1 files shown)


votor/src/certificate_pool/parent_ready_tracker.rs
```diff
@@ -213,14 +213,10 @@ impl ParentReadyTracker {
...
         }
-        // TODO: for duplicate blocks we should adjust this to choose the
-        // parent with the least amount of duplicate blocks if possible.
-        // Notice that each scenario with multiple NotarFallbacks also will eventually
-        // have a skip for that slot, so prefer the skip if we've received it.
         match self
             .slot_statuses
             .get(&slot)
-            .and_then(|ss| ss.parents_ready.iter().max().copied())
+            .and_then(|ss| ss.parents_ready.iter().min().copied())
```
bw-solana, line 219: I thought we wanted to look at the `parents_ready` parent?

AshwinSekar: "Pick the parent ready with the smallest parent (slot)"
It's a bit confusing naming wise, it doesn't mean the parent ready's parent but rather the smallest parent slot.

So `ss.parents_ready` holds all the blocks that are valid parents to build `slot` off of. We want to choose the smallest slot #.
E.g. if I'm free to build block 4 off of 3 or 2, I should choose to build it off of 2 (skipping 3).

bw-solana (approved):
LGTM

PR #428, opened 2025-08-28
repair: plumb shred resolver service, and send events from blockstore

AshwinSekar:
#### Problem
We need a controller to receive requests to repair blocks from the pool/replay and send requests to the network.
It needs to be blockstore aware so that it can send out additional repairs and reconcile invalid responses.

Additionally we don't want to poll blockstore every time to see if our requests have completed. I would like to set it up so that we have the minimal blockstore interactions to find out which shreds to request, and then can continue on notifications from blockstore.

#### Summary of Changes
Plumb together a new service for this (open to naming suggestions), impl in the next PR. Send events from blockstore for use as feedback in this service.

### DIFF
--- a/ledger/src/blockstore_meta.rs
+++ b/ledger/src/blockstore_meta.rs
@@ -518,8 +518,8 @@ impl Index {
     /// Checks the index to see if we have all the data shreds in this FEC set
     pub(crate) fn is_data_set_complete(fec_set_index: u32, index: &Index) -> bool {
         let data_indices =
-            u64::from(fec_set_index)..u64::from(fec_set_index) + (DATA_SHREDS_PER_FEC_BLOCK as u64);
-        index.data().range(data_indices).count() == DATA_SHREDS_PER_FEC_BLOCK
+            (fec_set_index as usize)..(fec_set_index as usize + DATA_SHREDS_PER_FEC_BLOCK);
+        index.data().count_ones_in_range(data_indices) == DATA_SHREDS_PER_FEC_BLOCK
     }
 }
 
@@ -636,6 +636,13 @@ impl ShredIndexV2 {
             .range((start, end))
             .iter_ones()
             .map(|idx| idx as u64)
+    }
+
+    pub(crate) fn count_ones_in_range<R>(&self, range: R) -> usize
+    where
+        R: RangeBounds<usize>,
+    {
+        self.index.range(range).count_ones()
     }
 
     fn iter(&self) -> impl Iterator<Item = u64> + '_ {


bw-solana:
Adding @cpubot just for awareness. He was going to look into some of the shred/blockstore --> replay flow performance

core/src/repair/shred_resolver_service.rs
```diff
@@ -0,0 +1,51 @@
+//! The shred resolver service listens to events emitted by
```
carllin, line 1: maybe `duplicate_shred_resolver_service`

core/src/repair/shred_resolver_service.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub fn new(
+        _blockstore: Arc<Blockstore>,
+        event_receiver: ShredEventReceiver,
+        _outstanding_repairs: Arc<RwLock<OutstandingShredRepairs>>,
+    ) -> Self {
+        let t_listen = thread::spawn(move || loop {
+            let Ok(_events) = event_receiver.recv() else {
+                break;
+            };
+        });
```
carllin, line 43: do we want to be consistent with the error handling in `block_creation_loop` and `event_handler`: https://github.com/anza-xyz/alpenglow/blob/b0eb8f05f516ade030159d8b80817b22006ab3f3/core/src/block_creation_loop.rs#L325-L343

ledger/src/shred_event.rs
```diff
@@ -0,0 +1,48 @@
...
+        conflicting_shred_type: ShredType,
+    },
+
+    ChainedMerkleRootConflict {
+        location: BlockLocation,
+        slot: Slot,
+        lower_fec_set_index: u32,
+        lower_merkle_root: Hash,
+        higher_fec_set_index: u32,
+        higher_chained_merkle_root: Hash,
```
carllin: what does lower vs higher mean in this context?

ledger/src/blockstore_meta.rs
```diff
@@ -852,6 +857,13 @@ impl ErasureMeta {
         self.config.num_data <= num_data + num_coding
     }
 
+    /// Checks the index to see if we have all the data shreds in this FEC set
+    pub(crate) fn is_data_set_complete(fec_set_index: u32, index: &Index) -> bool {
+        let data_indices =
+            u64::from(fec_set_index)..u64::from(fec_set_index) + (DATA_SHREDS_PER_FEC_BLOCK as u64);
+        index.data().range(data_indices).count() == DATA_SHREDS_PER_FEC_BLOCK
+    }
```
carllin: might be better as a method in `Index`

ledger/src/blockstore_meta.rs
```diff
@@ -836,6 +836,11 @@ impl ErasureMeta {
         self.fec_set_index.checked_add(num_data)
     }
 
+    /// Returns the index of the last data shred in this FEC set
+    pub(crate) fn last_data_shred_index(fec_set_index: u32) -> u32 {
+        fec_set_index + (DATA_SHREDS_PER_FEC_BLOCK as u32) - 1
+    }
+
```
carllin: might be better as a utility function in the shred module

ledger/src/blockstore.rs
```diff
@@ -1589,6 +1592,13 @@ impl Blockstore {
             handle_duplicate(shred);
         }
 
+        if let Some(shred_event_sender) = shred_event_sender {
+            for event in shred_events {
+                // TODO: handle error
```
carllin, line 1597: probably need to resolve this on startup right?

ledger/src/blockstore.rs
```diff
@@ -1568,6 +1569,7 @@ impl Blockstore {
         is_trusted: bool,
         retransmit_sender: &EvictingSender<Vec<shred::Payload>>,
         handle_duplicate: &F,
+        shred_event_sender: Option<&ShredEventSender>,
```
carllin, line 1572: starting to think it might be better to just return the `InsertResults` and have the caller handle the duplicates/sending the events now

ledger/src/shred_event.rs
```diff
@@ -0,0 +1,48 @@
...
+    },
```
carllin: maybe some comments on when each of these events are sent? For instance `CompletedFECSet` is only sent for the last FEC set

ledger/src/shred_event.rs
```diff
@@ -0,0 +1,48 @@
...
+    solana_clock::Slot,
+    solana_hash::Hash,
+};
+
+pub type ShredEventSender = Sender<ShredEvent>;
+pub type ShredEventReceiver = Receiver<ShredEvent>;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum ShredEvent {
+    CompletedFECSet {
```
carllin, line 18: nit: Should this be called `CompletedLastFECSet` because it's only sent on the last fec set?

ledger/src/blockstore.rs
```diff
@@ -2030,25 +2075,55 @@ impl Blockstore {
...
+            let is_last_in_slot =
+                self.is_complete_fec_set_last_in_slot(erasure_set, location, just_inserted_shreds);
+            shred_insertion_tracker
+                .shred_events
+                .push(ShredEvent::CompletedFECSet {
+                    location,
+                    slot,
+                    fec_set_index,
+                    is_last_in_slot,
+                });
```
carllin, line 2124: This can only be sent once because each shred in an FEC set only calls into this function `check_insert_data_shred()` once? i.e. we filter out existing shreds before this function is called

ledger/src/blockstore.rs
```diff
@@ -2267,6 +2352,15 @@ impl Blockstore {
...
 
+            shred_events.push(ShredEvent::ChainedMerkleRootConflict {
+                location,
+                slot,
+                lower_fec_set_index: erasure_set.fec_set_index(),
+                lower_merkle_root: merkle_root.unwrap_or_default(),
+                higher_fec_set_index: next_erasure_set.fec_set_index(),
+                higher_chained_merkle_root: chained_merkle_root.unwrap_or_default(),
+            });
+
```
carllin: the repair controller will be in charge of dumping these bad shreds? 

For future PR's, might be good to organize them per event handler. Have a feeling it's going to be complicated 🫡

core/src/repair/shred_resolver_service.rs
AshwinSekar: the idea was that should be a replacement to repair_service.rs so it sends out normal repair as well as duplicate scenarios. 

wanted to avoid calling it ag_repair_service 😅

core/src/repair/shred_resolver_service.rs
AshwinSekar: yeah definitely, i'll add proper error handling when i fill in the impl in the next PR

ledger/src/blockstore.rs
AshwinSekar: wdym?

ledger/src/blockstore.rs
AshwinSekar: exactly, we have a check up above (see PossibleDuplicateShred::Exists) that makes sure we don't already have the shred. So regardless of whether we naturally complete the FEC set or complete it as a result of FEC recovery, we'll only send this event out once.

ledger/src/blockstore.rs
AshwinSekar: Yeah exactly the MerkleRootConflict / ChainedMerkleRootConflict won't happen during normal operation, only if we're receiving duplicate blocks or getting a malicious repair response.

I wanted to avoid having to rescan blockstore after x ms and dump/re request repair in this case. Since it's a low prob event figured it's better to have it event driven - the repair controller gets notified when this happens and takes action.

ledger/src/shred_event.rs
AshwinSekar: it's actually sent for each completed fec set, the `is_last_in_slot` field indicates whether it's the last fec set

ledger/src/shred_event.rs
AshwinSekar: CMR conflict occurs when 2 adjacent FEC sets don't chain together correctly. the lower (by index) FEC sets' merkle root is not equal to the higher (by index) FEC sets' chained merkle root.

ledger/src/blockstore.rs
AshwinSekar: yeah definitely is strange right now... i'm inclined to punt on this because we're going to remove `handle_duplicate` for non TowerBFT slots anyway.

ledger/src/shred_event.rs
AshwinSekar: actually #437 will give us https://github.com/anza-xyz/agave/pull/7728 which enforces fixed 32:32 FEC sets.
Once that lands I can remove `higher_fec_set_index` as it will just be +32 (and some other places where I specify both index and fec_set_index). Will wait to merge this until then.

core/src/repair/shred_resolver_service.rs
carllin: we can just drag and drop it into repair_service when the time comes then

ledger/src/shred_event.rs
carllin: ah dope was thinking it would be clearer to call it previous/current merkle root instead of lower/higher

ledger/src/blockstore.rs
carllin: i.e. do we need to resend these events if we shut down, and then start back up again before these events are handled

AshwinSekar: it should be fine, essentially repair will have to perform a blockstore scan initially (when receiving a certificate to initiate catchup, or a request to repair an alternate version from replay/safe to notar). The idea behind these events are just so we don't have to rescan to see if we've finished.

So if we restart then we don't need to resend these events.

ledger/src/blockstore_meta.rs
```diff
@@ -514,6 +514,13 @@ impl Index {
     pub(crate) fn coding_mut(&mut self) -> &mut ShredIndex {
         &mut self.coding
     }
+
+    /// Checks the index to see if we have all the data shreds in this FEC set
+    pub(crate) fn is_data_set_complete(fec_set_index: u32, index: &Index) -> bool {
+        let data_indices =
+            u64::from(fec_set_index)..u64::from(fec_set_index) + (DATA_SHREDS_PER_FEC_BLOCK as u64);
+        index.data().range(data_indices).count() == DATA_SHREDS_PER_FEC_BLOCK
```
cpubot: nit: could take the `count_ones` fastpath here https://github.com/anza-xyz/alpenglow/blob/0c0c6ffc29fa06232168f523d9efb8cd2985065a/ledger/src/bit_vec.rs#L344

ledger/src/shred_event.rs
AshwinSekar: done I refactored now 5c7c2f550be52cba0782f3973162883f57986706 we only need to track the previous fec set

ledger/src/blockstore_meta.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/428/commits/7bc4555c0ea86be9fc7c94178876512b6d6f0eeb

PR #430, opened 2025-08-29
[core] add bls sigverifier logic

samkim-crypto:
#### Summary of Changes
First attempt at the BLS sigverifier logic. The implementation does leave some room for better optimization, but I think this can be a good start.

The `verify_batches` is the main function that does verifies packet signatures.
1. It first goes through the packets, deserializes them, and groups them into simple votes or certificates
2. It then verifies the batch of votes and certificates separately

The `verify_votes` function verifies the batch of votes.
1. It organizes the BLS public keys, signatures, and messages so that we can invoke the `par_verify_distinct`, which optimistically verifies the signatures in a batch.
2. It invokes `par_verify_distinct`
3. If the `par_verify_distinct` fails, then it individually verifies the packets

The `verify_certificate` function verifies the batch of certifies.
1. It first deserializes the bit vec, either resulting in a base2 encoding or base3
2. If base2 encoding, it simply constructs the aggregate BLS public key and verifies the cert
3. If base3 encoding, then it constructs two aggregate BLS public keys. Constructs messages to verify, and then invokes `par_verify_distinct`.

Fixes #


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -142,6 +156,88 @@ impl BLSSigVerifier {
...
+        let bank = self.root_bank.load();
+        let Some(key_to_rank_map) = get_key_to_rank_map(&bank, slot) else {
+            return false;
+        };
+
+        match &message {
+            ConsensusMessage::Vote(vote_message) => {
+                let signed_payload = match bincode::serialize(&vote_message.vote) {
+                    Ok(payload) => payload,
+                    Err(_) => return false,
```
wen-coding: We just successfully deserialized this message, so this theoretically shouldn't happen right? Do you think this is worth a warning maybe?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -142,6 +156,88 @@ impl BLSSigVerifier {
...
+                };
+
+                let Some((_, bls_pubkey)) = key_to_rank_map.get_pubkey(vote_message.rank.into())
+                else {
+                    return false;
```
wen-coding: nit: we need to update stats on every failure case, but we can do that in another PR as well.

votor/src/consensus_pool/vote_certificate_builder.rs
```diff
@@ -452,9 +458,18 @@ mod tests {
...
+        let message_refs: Vec<&[u8]> = all_messages.iter().map(|m| m.as_slice()).collect();
+
+        let verification_result = SignatureProjective::verify_distinct_aggregated(
+            &pubkey_refs,
+            &certificate_message.signature,
+            &message_refs,
+        );
+
+        assert!(
+            verification_result.unwrap_or(false),
```
wen-coding: nit: we expect Ok here right? Maybe just unwrap() so if it's an error, we can see the actual error?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -36,8 +60,69 @@ pub struct BLSSigVerifier {
...
+        let mut certs_to_verify = Vec::new();
+
+        let bank = self.root_bank.load();
+
+        for mut packet in batches.iter_mut().flatten() {
+            if packet.meta().discard() {
+                continue;
+            }
+
+            let message: ConsensusMessage = match packet.deserialize_slice(..) {
```
wen-coding, line 83: Maybe we should save the deserialized messages somewhere so we don't do it again in send_packets, hmm

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -36,8 +60,69 @@ pub struct BLSSigVerifier {
...
+                Ok(msg) => msg,
+                Err(_) => {
+                    packet.meta_mut().set_discard(true);
```
wen-coding, line 86: nit: we should add stats for verifications, but that can be done in a follow up PR.

core/benches/bls_sigverify.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    stage.join().unwrap();
+}
+
+fn bench_bls(c: &mut Criterion) {
+    solana_logger::setup();
+    c.bench_function("bls_sigverify_stage_all_valid", |b| {
+        bench_bls_sigverify_stage(b, true);
+    });
+    c.bench_function("bls_sigverify_stage_one_invalid", |b| {
```
wen-coding, line 163: We assume 20% malicious, so can we maybe test 20% invalid as well?

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -142,6 +227,218 @@ impl BLSSigVerifier {
...
+    vote_message: VoteMessage,
+    bls_pubkey: &'a BlsPubkey,
+    packet: PacketRefMut<'a>,
+}
+
+impl VoteToVerify<'_> {
+    fn payload_slice(&self) -> &[u8] {
+        // The bincode serialization of `ConsensusMessage::Vote(VoteMessage)` has the layout:
+        // [ 4-byte Enum Discriminant | Vote Payload | 96-byte Signature | 2-byte Rank ]
+        const BINCODE_ENUM_DISCRIMINANT_SIZE: usize = 4;
```
wen-coding, line 389: Err, this makes assumptions on payload layout so it seems a bit brittle. Are you trying to avoid another VoteMessage serialize here? We can cache serialized vote if needed.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -142,6 +227,218 @@ impl BLSSigVerifier {
...
+}
+
+struct CertToVerify<'a> {
+    cert_message: CertificateMessage,
+    packet: PacketRefMut<'a>,
+}
+
+fn certificate_to_vote_message_base2(certificate: &Certificate) -> Option<Vote> {
+    match certificate {
+        Certificate::Notarize(slot, hash) => Some(Vote::new_notarization_vote(*slot, *hash)),
```
wen-coding, line 412: I'm thinking whether we should add a helper function near `certificate_limits_and_vote_types`, so if this changes we remember to change everything together.

I feel this whole mapping is getting complicated enough it should be an independent object, but that can happen in a later PR.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -422,4 +720,361 @@ mod tests {
...
+        let aggregate_signature: BlsSignature = SignatureProjective::aggregate(&signature_refs)
+            .unwrap()
+            .into();
+
+        let mut bitmap = BitVec::<u8, Lsb0>::new();
+        bitmap.resize(num_signers, false);
+        for i in 0..num_signers {
+            bitmap.set(i, true);
+        }
+        let encoded_bitmap = encode_base2(&bitmap).unwrap();
```
wen-coding: nit: can we use `vote_certificate_builder` so we do end-to-end verification here?

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Yes, I was actually going to write this issue in the PR description, but I forgot. I can add a todo here and address it in a follow-up.

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Yes, I originally had the serialize logic here and I thought it was a little wasteful to deserialize (to distinguish vote and certificate) and then serialize again. By caching serialized vote, what do you precisely mean?

wen-coding: I mean, it's very likely the same kind of vote all pour in at the same time, for example:
Notarize(5, hash1)
Finalize(6)

So we can keep a cache if serialization turns out to waste a lot of time. But let's put this in first then improve performance.

samkim-crypto: Okay yeah that makes sense. Let me add a todo.

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: I added a todo in [5065967](https://github.com/anza-xyz/alpenglow/pull/430/commits/5065967163709d5dd3d5dd25ad371b665dffc78f) to figure out the best way to remove redundant serialization.

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Okay yep, I added a todo in [5065967](https://github.com/anza-xyz/alpenglow/pull/430/commits/5065967163709d5dd3d5dd25ad371b665dffc78f).

core/benches/bls_sigverify.rs
samkim-crypto: I added the 20% malicious case in the bench logic. I think we need more exhaustive cases for the bench in general, but I'll subsequently add these in a follow-up.

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Yes good idea. I added a todo in [5065967](https://github.com/anza-xyz/alpenglow/pull/430/commits/5065967163709d5dd3d5dd25ad371b665dffc78f).

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Done!

wen-coding (approved):

PR #431, opened 2025-08-29
Do not serialize/deserialize bls_pubkey_to_rank_map.

wen-coding:
This is a derived map, initialize on first use.

### DIFF
--- a/runtime/src/epoch_stakes.rs
+++ b/runtime/src/epoch_stakes.rs
@@ -16,6 +16,7 @@ pub type EpochAuthorizedVoters = HashMap<Pubkey, Pubkey>;
 pub type SortedPubkeys = Vec<(Pubkey, BLSPubkey)>;
 
 #[derive(Clone, Debug, Default)]
+#[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
 #[cfg_attr(feature = "dev-context-only-utils", derive(PartialEq))]
 pub struct BLSPubkeyToRankMap {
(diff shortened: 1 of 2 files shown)


runtime/src/epoch_stakes.rs
```diff
@@ -5,14 +5,17 @@ use {
...
+    },
 };
 
 pub type NodeIdToVoteAccounts = HashMap<Pubkey, NodeVoteAccounts>;
 pub type EpochAuthorizedVoters = HashMap<Pubkey, Pubkey>;
 pub type SortedPubkeys = Vec<(Pubkey, BLSPubkey)>;
 
-#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+#[derive(Clone, Debug, Default)]
 #[cfg_attr(feature = "frozen-abi", derive(AbiExample))]
```
bw-solana, line 19: we can delete this now, right?

bw-solana (approved):
LGTM. I think we can get rid of AbiExample derivation though

runtime/src/epoch_stakes.rs
wen-coding: Unfortunately no. The abi package is not smart enough to figure out we don't need this any more because of OnceLock, so it won't compile. But we won't use it.
See my previous attempt:
https://buildkite.com/anza/alpenglow/builds/1761#0198f781-2060-4f10-9cf7-9b934a1b11c6

PR #432, opened 2025-09-02
[votor] Fix Fallback certificate build in votor

samkim-crypto:
#### Summary of Changes

Fixes #


votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -83,23 +83,24 @@ impl VoteCertificateBuilder {
...
             if MAXIMUM_VALIDATORS <= rank {
                 return Err(CertificateError::ValidatorDoesNotExist(vote_message.rank));
             }
-            target_bitmap.set(rank, true);
+
+            let current_vote_type = VoteType::get_type(&vote_message.vote);
+
+            if current_vote_type == vote_types[0] {
+                self.input_bitmap_1.set(rank, true);
+            } else if vote_types.len() > 1 && current_vote_type == vote_types[1] {
```
wen-coding: nit: we currently don't support vote_types.len() other than 1 and 2, so maybe we can just test vote_types.len() == 2 here?

votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -83,23 +83,24 @@ impl VoteCertificateBuilder {
 
     /// Aggregates a slice of `VoteMessage`s into the builder.
     pub fn aggregate(&mut self, messages: &[VoteMessage]) -> Result<(), CertificateError> {
-        let Some(vote_type) = messages.first().map(|m| VoteType::get_type(&m.vote)) else {
```
wen-coding, line 86: I think this was working because we pass in messages in the same vote pool, so the vote type of first vote should be the same as rest of the votes. But I think your change makes it more robust.

votor/src/certificate_pool/vote_certificate_builder.rs
```diff
@@ -338,4 +342,119 @@ mod tests {
...
+            let signature = keypair.sign(&serialized_vote);
+            vote_messages.push(VoteMessage {
+                vote,
+                signature: signature.into(),
+                rank: i as u16,
+            });
+            keypairs.push(keypair);
+        }
+
+        // 2. Generation: Aggregate votes and build the certificate. This will use base2 encoding.
```
wen-coding: nit: maybe say this will use base2 encoding because it only contains one type of vote.

votor/src/certificate_pool/vote_certificate_builder.rs
samkim-crypto: Yep, good idea. Fixed!

votor/src/certificate_pool/vote_certificate_builder.rs
samkim-crypto: Hm okay, a single vote pool can technically contain two types of votes (e.g. `Notarize` and `NotarizeFallback`), right? I know that this case should pretty much never happen in practice, but I wonder if this case should be explicitly tested in our live tests.

votor/src/certificate_pool/vote_certificate_builder.rs
samkim-crypto: Yep, updated!

votor/src/certificate_pool/vote_certificate_builder.rs
wen-coding: Our current implementation will only put one type of votes in one pool. But what you have is reasonable, `get_type` is cheap so it's good we check it always.

samkim-crypto: Oh, so how are the `NotarizeFallback` and `Skip` certificates that can contain two types of votes constructed? Is the `VoteCertificateBuilder` actually not used for the certificate construction?

wen-coding: We go to two pools: first Notarize pool then NotarizeFallback pool. At the end we perform build to construct the final certificate.

wen-coding (approved):

votor/src/certificate_pool/vote_certificate_builder.rs
samkim-crypto: I see. I will probably need to look at the precise logic more precisely and update the tests accordingly, but I'll merge this PR and make any related updates in https://github.com/anza-xyz/alpenglow/pull/430.

PR #435, opened 2025-09-04
Create alpenglow_connection_cache and remove Alpenglow from fetch_stage

wen-coding:
- We should always use QUIC client in Alpenglow, not depend on vote_use_quic
- When we are using QUIC, we shouldn't plug Alpenglow port and sender into fetch_stage any more. The QUIC streamer is the equivalent of fetch_stage, plug into both places will only make them compete for packets (worse, fetch_stage will forward QUIC control packets over)

Future work: we can consider writing our own voting_service, not sharing much with old voting_service any more.


bw-solana (approved):
LGTM

PR #440, opened 2025-09-06
Split Alpenglow voting_service from Tower voting_service.

wen-coding:
Now that we don't share connection cache with TowerBFT votes, time to completely split the two voting_service.


votor/src/voting_service.rs
```diff
@@ -0,0 +1,366 @@
...
+        // will cause a packet spike and overwhelm the network. If we later find out that this is
+        // not an issue, we can optimize this by using multi_targret_send or similar methods.
+        for socket in sockets {
+            if let Err(e) = send_message(buf.clone(), socket, &connection_cache) {
+                warn!("Failed to send alpenglow message to {socket}: {e:?}");
+            }
+        }
+    }
+
+    pub fn handle_bls_vote(
```
AshwinSekar: maybe broadcast_bls_op or handle_bls_op since it's both votes and certificates

votor/src/voting_service.rs
```diff
@@ -0,0 +1,366 @@
...
+                        &additional_listeners,
+                        &mut staked_validators_cache,
+                    );
+                }
+            })
+            .unwrap();
+        Self { thread_hdl }
+    }
+
+    fn broadcast_alpenglow_message(
```
AshwinSekar: maybe broadcast_consensus_message

votor/src/voting_service.rs
```diff
@@ -0,0 +1,366 @@
+use {
+    crate::{
+        staked_validators_cache::StakedValidatorsCache, vote_history_storage::VoteHistoryStorage,
+        voting_utils::BLSOp,
```
AshwinSekar: maybe just move this into here since we're in the votor crate now

votor/src/voting_service.rs
wen-coding: Changed to handle_bls_op.

votor/src/voting_service.rs
wen-coding: Done.

votor/src/voting_service.rs
wen-coding: Done.

carllin (approved):

PR #443, opened 2025-09-10
Add timing stats and clean up bls sigverify stats a bit.

wen-coding:
Several changes in BLS Sigverifier stats:
- We probably don't StatsUpdater for a while, remove for now. I do need to change the counters to Atomic because `verify_batches()` takes `&self` instead of `&mut self`
- Now that a lot of malformed checks are done in `verify_batches`, move stats increment there, otherwise `verify_batches` mark those packets as discarded and we won't see malformed at all
- Add preliminary timing tracking

### DIFF
--- a/core/src/sigverifier/bls_sigverifier/stats.rs
+++ b/core/src/sigverifier/bls_sigverifier/stats.rs
@@ -1,51 +1,54 @@
-use std::time::{Duration, Instant};
+use std::{
+    sync::atomic::{AtomicU64, Ordering},
+    time::{Duration, Instant},
+};
 
 pub(super) const STATS_INTERVAL_DURATION: Duration = Duration::from_secs(1);
 
-#[derive(Debug, Default)]
-pub(super) struct StatsUpdater {
-    pub(super) sent: u64,
-    pub(super) sent_failed: u64,
-    pub(super) verified_votes_sent: u64,
-    pub(super) verified_votes_sent_failed: u64,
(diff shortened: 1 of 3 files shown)


core/src/sigverifier/bls_sigverifier/stats.rs
```diff
@@ -59,49 +70,106 @@ impl BLSSigVerifierStats {
...
+                i64
+            ),
+            (
+                "votes_batch_optimistic_elapsed_us",
+                self.votes_batch_optimistic_elapsed_us
+                    .load(Ordering::Relaxed) as i64,
+                i64
+            ),
+            (
+                "votes_batch_parallel_verify_count",
```
waldrobi, line 95: What is the difference between: 
`votes_batch`, `votes_batch_optimistic`, and `votes_batch_parallel_verify`? 

I understand the verification of votes might be processed concurrently, but what are optimistic votes in this context?

wen-coding: In BLS world, it's really cheap if all votes are valid, we can optimistically aggregate all pubkeys and all signatures, then verify once that the signatures match what we thought it should be. This should succeed in our world 99% of the time, since we only accept packets fro staked validators.

But in case there are invalid packets, then "optimistic" verification fails, we have to go to more costly "parallel_verify" to actually figure out which packets are invalid and to be discarded, so it's good to separate the two timings.

PR #445, opened 2025-09-11
[bls-sigverifier] Cache vote messages

samkim-crypto:
#### Summary of Changes
BLS optimistic batch verification of votes is much more efficient if we assume that the signatures are all based on the same message.

I added stats that record the number of distinct messages in a single batch. I also updated the verification logic so that if there is only one distinct message for the batch, I use the `par_verify_aggregate` function instead of the `par_verify_distinct` function.

In the process, I addressed a todo: cache serialized votes so that we don't serialize identical votes every time.

After we analyze the number of distinct messages in our cluster, we can add tailored optimizations for the specific number of messages.

Fixes #


core/src/sigverifier/bls_sigverifier/stats.rs
```diff
@@ -41,6 +42,7 @@ impl BLSSigVerifierStats {
             preprocess_count: AtomicU64::new(0),
             preprocess_elapsed_us: AtomicU64::new(0),
             votes_batch_count: AtomicU64::new(0),
+            votes_batch_distinct_messages_count: AtomicU64::new(0),
```
wen-coding, line 45: You need to add this to datapoint_info! below as well.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -305,41 +317,69 @@ impl BLSSigVerifier {
...
+        let (pubkeys, signatures): (Vec<_>, Vec<_>) = votes_to_verify
+            .iter()
+            .map(|v| (v.bls_pubkey, &v.vote_message.signature))
             .multiunzip();
 
         // Optimistically verify signatures; this should be the most common case
-        if SignatureProjective::par_verify_distinct(&pubkeys, &signatures, &messages)
-            .unwrap_or(false)
-        {
+        let verified_optimistically = if distinct_messages == 1 {
```
wen-coding: Unfortunately I think this will not happen very often, because we do push Notarize(slot, hash) and Finalize(slot) at the same time.

Can you sort messages based on associated vote and verify different type of votes in batch at the same time?

Because right now preprocessing doesn't take much time, I think the cost of memory copy will be small.

I pushed your change to Alpenglow test cluster, the numbers were 3.82k before your change and 3.18k after your change:
https://metrics.solana.com:8888/sources/16/chronograf/data-explorer?query=SELECT%20sum%28%22certs_batch_elapsed_us%22%29%20%2F%20sum%28%22certs_batch_count%22%29%20AS%20%22certs_elapsed_per_batch%22%2C%20sum%28%22votes_batch_optimistic_elapsed_us%22%29%20%2F%20sum%28%22votes_batch_count%22%29%20AS%20%22votes_elapsed_per_batch%22%20%2C%20sum%28%22preprocess_elapsed_us%22%29%20%2F%20sum%28%22preprocess_count%22%29%20AS%20%22preprocess_per_batch%22%2C%20sum%28%22votes_batch_elapsed_us%22%29%20%2F%20sum%28%22received_votes%22%29%20AS%20%22average_vote_time%22%2C%20sum%28%22certs_batch_elapsed_us%22%29%20%2F%20%28sum%28%22received%22%29%20-%20sum%28%22received_votes%22%29%20-%20sum%28%22received_discarded%22%29%20-%20sum%28%22received_malformed%22%29%29%20AS%20%22average_certs%22%20FROM%20%22testnet-alpenglow%22.%22autogen%22.%22bls_sig_verifier_stats%22%20WHERE%20time%20%3E%20%3AdashboardTime%3A%20AND%20time%20%3C%20%3AupperDashboardTime%3A%20GROUP%20BY%20time%28%3Ainterval%3A%29%20FILL%28null%29

core/src/sigverifier/bls_sigverifier/stats.rs
samkim-crypto: Okay fixed!

core/src/sigverifier/bls_sigverifier.rs
samkim-crypto: Okay, this is helpful. I just pushed some changes that optimizes batch verification when the number of distinct messages are small. It would be interesting to see if this results in any speed up.

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -82,6 +91,7 @@ pub struct BLSSigVerifier {
     root_bank: SharableBank,
     stats: BLSSigVerifierStats,
     verified_certs: RwLock<HashSet<Certificate>>,
+    vote_payload_cache: RwLock<HashMap<Vote, Arc<Vec<u8>>>>,
```
wen-coding, line 94: We should definitely do cache cleanup for votes older than roots, so this doesn't grow forever.

samkim-crypto: Oh yes I missed this. The clean up logic should be added now.

PR #446, opened 2025-09-11
feat: don't discard `BlockComponent::VersionedBlockMarker`s

ksn6:
#### Problem
To get fast leader handover rolling (https://github.com/anza-xyz/alpenglow/issues/293), we need to send out `ParentReadyUpdate` messages. These are of type `BlockComponent::VersionedBlockMarker`, which are currently entirely discarded.

#### Summary of Changes
- Rather than discarding block markers by turning shreds into a flattened `Vec<Entry>`, we turn shreds into a `Vec<BlockComponent>` vector
- For now, we preserve the remainder of Agave code. In the next PR, we'll:
  - Store special block marker types in blocks
  - Split logic between processing `Vec<Entry>` and `Vec<BlockComponent>` when necessary

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -4123,7 +4123,7 @@ impl Blockstore {
         let Some((&Range { start, .. }, &Range { end, .. })) =
             completed_ranges.first().zip(completed_ranges.last())
         else {
-            return Ok(vec![BlockComponent::default()]);
+            return Ok(vec![]);
         };
         let indices = u64::from(start)..u64::from(end);
         let keys = indices.clone().map(|index| (slot, index));


ledger/src/blockstore.rs
```diff
@@ -3070,97 +3070,109 @@ impl Blockstore {
...
-                                "Blockstore::get_block sanitize failed: {err:?}, slot: {slot:?}, \
-                                 {transaction:?}",
-                            );
-                        }
-                        transaction
+
+        if !slot_meta.is_full() {
+            trace!("do_get_complete_block_with_entries() failed for {slot} (slot not full)");
+            return Err(BlockstoreError::SlotUnavailable);
+        }
```
bw-solana: never nesters 🔥

entry/src/block_component.rs
```diff
@@ -372,6 +372,38 @@ impl BlockComponent {
...
+    /// If this component contains entries, returns a reference to them.
+    pub fn as_entries(&self) -> Option<&[Entry]> {
+        match self {
+            Self::Entries(entries) => Some(entries),
+            Self::BlockMarker(_) => None,
+        }
+    }
+
+    /// If this component contains a special marker, returns a reference to it.
+    pub fn as_block_marker(&self) -> Option<&VersionedBlockMarker> {
```
akhi3030, line 384: are these actually used anywhere? Should we define them only if we need them?

entry/src/block_component.rs
```diff
@@ -372,6 +372,38 @@ impl BlockComponent {
...
+    /// If this component contains entries, returns them as a vector.
+    pub fn into_entries(self) -> Option<Vec<Entry>> {
+        match self {
+            Self::Entries(entries) => Some(entries),
+            Self::BlockMarker(_) => None,
+        }
+    }
+
+    /// If this component contains a special marker, returns it.
+    pub fn into_block_marker(self) -> Option<VersionedBlockMarker> {
```
akhi3030, line 400: same as above, should we introduce this function when it is actually needed?

ledger/src/blockstore.rs
```diff
@@ -3922,20 +3922,36 @@ impl Blockstore {
...
+        &self,
+        slot: Slot,
+        shred_start_index: u64,
+    ) -> Result<Vec<BlockComponent>> {
+        self.get_slot_components_with_shred_info(slot, shred_start_index, false)
+            .map(|x| x.0)
+    }
+
     /// Returns the entry vector for the slot starting with `shred_start_index`
     pub fn get_slot_entries(&self, slot: Slot, shred_start_index: u64) -> Result<Vec<Entry>> {
```
akhi3030, line 3936: I tried to follow the call tree here to figure out how expensive this function might be.  

It seems like this function will call `get_slot_components_in_block()` which seems kinda expensive.  It looks up shreds from the DB; deshreds them; and then filters out the the `BlockComponent` and we then filter out the component with `Entry`s.  

Is it fair to say that this function can indeed be expensive?  Should add it to the documentation that that is the case?  Are there are other strategies possible to reduce the cost?

ledger/src/blockstore.rs
```diff
@@ -4087,7 +4123,7 @@ impl Blockstore {
         let Some((&Range { start, .. }, &Range { end, .. })) =
             completed_ranges.first().zip(completed_ranges.last())
         else {
-            return Ok(vec![]);
+            return Ok(vec![BlockComponent::default()]);
```
akhi3030: can we document why we cannot return an empty vec here?

entry/src/block_component.rs
ksn6: Copy-pasted these accessors over for now, but these accessors are all going away once we include them in a separate PR.

entry/src/block_component.rs
ksn6: same as above

ledger/src/blockstore.rs
ksn6: Thanks for the good questions.

Yes, this is expensive, although pretty much the same cost as what existed before. Functions that go from shreds to `Vec<Entry>` / `Vec<BlockComponent>` are generally expensive, as they involve parsing shreds.

There's currently active work to accelerate `Entry` serde, which should accelerate these expensive blockstore functions in the most common path, i.e., `BlockComponent::Entries`: https://github.com/anza-xyz/agave/pull/8141.

Probably not worth spending time on performance here until we benchmark.

ledger/src/blockstore.rs
ksn6: Ah, forgot to remove this. Thanks for catching.

ledger/src/blockstore.rs
```diff
@@ -4118,28 +4154,39 @@ impl Blockstore {
...
     }
 
+    pub fn get_components_in_data_block(
+        &self,
+        slot: Slot,
+        range: Range<u32>,
+        slot_meta: Option<&SlotMeta>,
+    ) -> Result<Vec<BlockComponent>> {
+        self.get_slot_components_in_block(slot, vec![range], slot_meta)
+    }
```
bw-solana, line 4174: does this just wrap range in a vec? Could we just do this inside `get_entries_in_data_block`?

ledger/src/blockstore.rs
```diff
@@ -4118,28 +4154,39 @@ impl Blockstore {
...
         slot_meta: Option<&SlotMeta>,
     ) -> Result<Vec<Entry>> {
-        self.get_slot_entries_in_block(slot, vec![range], slot_meta)
+        self.get_components_in_data_block(slot, range, slot_meta)
+            .map(|components| {
+                components
+                    .into_iter()
+                    .filter_map(|component| component.into_entries())
+                    .flatten()
+                    .collect()
```
bw-solana, line 4188: oof. The double alloc is pain. But I think we already do this upstream

ledger/src/blockstore.rs
```diff
@@ -3922,20 +3922,36 @@ impl Blockstore {
...
 
+    /// Returns the component vector for the slot starting with `shred_start_index`
+    pub fn get_slot_components(
+        &self,
+        slot: Slot,
+        shred_start_index: u64,
+    ) -> Result<Vec<BlockComponent>> {
+        self.get_slot_components_with_shred_info(slot, shred_start_index, false)
+            .map(|x| x.0)
+    }
```
bw-solana, line 3933: Could probably live without this shim layer as well, but I guess it's fine

ledger/src/blockstore.rs
ksn6: ^ ;)

PR #450, opened 2025-09-12
Skip verifying old votes/certs and certs which have been verified.

wen-coding:
Skip verifying these votes/certs:
- Any vote or cert with slot not greater than root slot
- Any cert which has been previously verified

On our 50 nodes cluster:
Before this PR: Every 10 seconds we verify 52 batches of votes, 50 batches of certs
After this PR: Every 10 seconds we verify 39 batches of votes, 3.5 batches of certs

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -30,8 +30,8 @@ use {
     },
     stats::BLSSigVerifierStats,
     std::{
-        collections::HashMap,
-        sync::{atomic::Ordering, Arc},
+        collections::{HashMap, HashSet},
+        sync::{atomic::Ordering, Arc, RwLock},
     },
     thiserror::Error,
 };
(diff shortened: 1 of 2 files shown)


samkim-crypto (commented):
Looks fine to me. Just need to fix the CI for the imports.

wen-coding:
> Looks fine to me. Just need to fix the CI for the imports.

Oops, fixed. Also added one missing counter.

samkim-crypto (approved):

qkniep:
How can it be that we would ever verify fewer than 50 (even more so, fewer than 25) batches of certs? 10s should be at least 20 slots, we have Notar + (Fast)Final for each one. Or am I misunderstanding these numbers?

qkniep:
But then you would still see at least verify once per slot, which would be 20+ times, not 3.5 times. I am very confused how this could happen.

wen-coding:
We don't verify the cert generated by ourselves, this verification only
verifies incoming certs. If you generated a FastFinalization cert first
then all incoming certs are dropped.

PR #456, opened 2025-09-18
Try switching to VoteStateV4.

wen-coding:


bw-solana (commented):
Can we add a PR description?

Also, it would be nice to create multiple, smaller PRs for things that can be separated. For example dropping `solana_bls_signatures::signature::Signature as BlsSignature` and just using `Signature` directly is unrelated cosmetic change and could be forked out

votor/src/voting_utils.rs
```diff
@@ -111,14 +115,19 @@ pub enum VoteError {
     SavedVoteHistoryError(#[from] VoteHistoryError),
 }
 
+pub(crate) struct BLSKeypairWithCompressedPubkey {
+    pub keypair: Arc<BLSKeypair>,
+    pub compressed_pubkey: [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
+}
```
bw-solana: feels a little funky having this since the `compressed_pubkey` can be derived from the keypair. Should we have separate APIs and let callers decide how they want to access and mix/match things?

wen-coding: Moving this to another PR later.

runtime/src/bank/tests.rs
```diff
@@ -731,13 +731,20 @@ where
...
-            *solana_votor_messages::state::VoteState::deserialize(vote_account.data()).unwrap();
+            VoteStateV4::deserialize(vote_account.data(), &Pubkey::default()).unwrap();
+        vote_state.epoch_credits.push((0, 0, 0));
         for _ in 0..MAX_LOCKOUT_HISTORY + 42 {
-            let mut epoch_credits = *vote_state.epoch_credits();
-            epoch_credits.set_credits(epoch_credits.credits() + 16);
-            vote_state.set_epoch_credits(epoch_credits);
-            vote_state.serialize_into(vote_account.data_as_mut_slice());
+            let current_credits = vote_state.epoch_credits.last_mut().unwrap();
+            current_credits.1 += 16;
```
ksn6: Rather than doing `current_credits.1`, could we do let `(x, y, ...) = vote_state.epoch_credits.last_mut().unwrap()`?

runtime/src/epoch_stakes.rs
```diff
@@ -31,9 +30,16 @@ impl BLSPubkeyToRankMap {
             .iter()
             .filter_map(|(pubkey, (stake, account))| {
                 if *stake > 0 {
-                    account
-                        .bls_pubkey()
-                        .map(|bls_pubkey| (*pubkey, *bls_pubkey, *stake))
+                    account.vote_state_view().and_then(|vote_state| {
```
ksn6, line 33: This is a bit hard to read - could we unnest / clean this a bit?

vote/src/vote_account.rs
```diff
@@ -194,14 +195,14 @@ impl VoteAccount {
         }
     }
 
-    pub fn bls_pubkey(&self) -> Option<&BLSPubkey> {
+    pub fn bls_pubkey(&self) -> Option<BLSPubkey> {
```
ksn6, line 198: This doesn't seem right - shouldn't invoking `bls_pubkey()` return `None` for `TowerBFT` and `Some(...)` for `Alpenglow`?

votor/src/voting_utils.rs
```diff
@@ -155,7 +158,7 @@ pub fn generate_vote_tx(
 ) -> GenerateVoteTxResult {
     let vote_account_pubkey = context.vote_account_pubkey;
     let authorized_voter_keypair;
-    let bls_pubkey_in_vote_account;
+    let bls_pubkey_in_vote_account: BLSPubkey;
```
ksn6: I don't think we need to forward declare this variable - could we move it to the site of initialization and remove the annotated type, unless really needed?

vote/src/vote_account.rs
```diff
@@ -194,14 +195,14 @@ impl VoteAccount {
...
-            VoteAccountState::TowerBFT(_) => None,
-            VoteAccountState::Alpenglow => Some(
-                AlpenglowVoteState::deserialize(self.0.account.data())
-                    .unwrap()
-                    .bls_pubkey(),
-            ),
+            VoteAccountState::TowerBFT(vote_state) => vote_state.bls_pubkey_compressed().map(|b| {
+                let bls_pubkey_compressed =
+                    bincode::deserialize::<BLSPubkeyCompressed>(&b).unwrap();
+                BLSPubkeyCompressed::try_as_affine(&bls_pubkey_compressed).unwrap()
```
carllin, line 203: @samkim-crypto just double checking this is the right transformation

runtime/src/stakes.rs
```diff
@@ -576,13 +576,13 @@ pub(crate) mod tests {
         let vote_pubkey = solana_pubkey::new_rand();
         let bls_keypair = BLSKeypair::new();
         let vote_account = if is_alpenglow {
-            AlpenglowVoteState::create_account_with_authorized(
-                &vote_pubkey,
+            vote_state::create_v4_account_with_authorized(
+                &solana_pubkey::new_rand(),
```
carllin, line 580: out of curiosity why does this need to change?

carllin (approved):

runtime/src/bank/tests.rs
wen-coding: Done.

runtime/src/epoch_stakes.rs
wen-coding: Changed

vote/src/vote_account.rs
wen-coding: This part is a mess and will go away in the next PR, we will no longer have TowerBFT or Alpenglow option.

votor/src/voting_utils.rs
wen-coding: This is mainly done to release context.authorized_voter_keypairs read lock asap

runtime/src/stakes.rs
wen-coding: AlpenglowVoteState::create_account_with_authorized will create a vote state belonging to our new vote program, we are ditching our new vote program everywhere and replacing it with VoteStateV4 accounts.

vote/src/vote_account.rs
wen-coding: It's pretty ugly, but should be correct because all local-cluster tests are passing :) We've turned on BLS verification now.

wen-coding: So VoteStateV4 only says "give me a 48 byte buffer as bls pubkey", it doesn't care what's in it or whether it decodes properly for now. This PR (and https://github.com/anza-xyz/agave/pull/8122 on Agave side) decides we serialize the compressed pubkey into the 48 byte buffer.

samkim-crypto: Yes, I confirm that this is the right transformation.

PR #459, opened 2025-09-21
feat(`blockstore`): efficiently detect `UpdateParent` markers on shred ingest

ksn6:
### Summary
Implements online detection and storage of `UpdateParent` `BlockMarker`s during shred ingestion. `UpdateParent` markers enable Alpenglow's fast leader handover by allowing blocks to signal parent block updates.

To do this efficiently, we first note that `BlockComponent`'s will be FEC set aligned. We then observe that an `UpdateParent` marker is a specific type of `BlockComponent::VersionedBlockMarker`.

Accordingly, we detect `UpdateParent` markers at FEC set boundaries as follows:

- **Case A**: When receiving a shred with `DATA_COMPLETE_SHRED = true`, signifying the last shred of a FEC set boundary, we check the 0th shred of the *next* FEC set for an `UpdateParent` marker

- **Case B**: When receiving a shred with `DATA_COMPLETE_SHRED = false`, we check whether the previous shred had `DATA_COMPLETE_SHRED = true`; if so, we examine the current shred for an `UpdateParent` marker

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -276,7 +276,7 @@ pub struct Blockstore {
     alt_index_cf: LedgerColumn<cf::AlternateIndex>,
     alt_data_shred_cf: LedgerColumn<cf::AlternateShredData>,
     alt_merkle_root_meta_cf: LedgerColumn<cf::AlternateMerkleRootMeta>,
-    update_parent_meta_cf: LedgerColumn<cf::UpdateParentMeta>,
+    parent_meta_cf: LedgerColumn<cf::ParentMeta>,
 
     highest_primary_index_slot: RwLock<Option<Slot>>,
     max_root: AtomicU64,
@@ -326,6 +326,9 @@ struct ShredInsertionTracker<'a> {
     // In-memory map that maintains the dirty copy of the index meta.  It will
     // later be written to `cf::Index` or `cf::AlternateIndex`
     index_working_set: HashMap<(BlockLocation, u64), IndexMetaWorkingSetEntry>,
+    // In-memory map that maintains the dirty copy of the parent meta.  It will
+    // later be written to `cf::UpdateParentMeta`
+    parent_metas: HashMap<(BlockLocation, u64), WorkingEntry<ParentMeta>>,
     duplicate_shreds: Vec<PossibleDuplicateShred>,
     // Collection of the current blockstore writes which will be committed
     // atomically.
@@ -344,6 +347,7 @@ impl ShredInsertionTracker<'_> {
             merkle_root_metas: HashMap::new(),
             slot_meta_working_set: HashMap::new(),
             index_working_set: HashMap::new(),
+            parent_metas: HashMap::new(),
             duplicate_shreds: vec![],
             write_batch,
             index_meta_time_us: 0,
@@ -472,7 +476,7 @@ impl Blockstore {
             alt_index_cf,
             alt_data_shred_cf,
             alt_merkle_root_meta_cf,
-            update_parent_meta_cf,
+            parent_meta_cf: update_parent_meta_cf,
 
             highest_primary_index_slot: RwLock::<Option<Slot>>::default(),
             new_shreds_signals: Mutex::default(),
@@ -1375,6 +1379,19 @@ impl Blockstore {
             )?;
         }
 
+        for (&(location, slot), working_parent_meta) in &shred_insertion_tracker.parent_metas {
+            if !working_parent_meta.should_write() {
+                // No need to rewrite the column
+                continue;
+            }
+
+            self.parent_meta_cf.put_in_batch(
+                &mut shred_insertion_tracker.write_batch,
+                (slot, location),
+                working_parent_meta.as_ref(),
+            )?;
+        }
+
         for (&(location, slot), index_working_set_entry) in
             shred_insertion_tracker.index_working_set.iter()
         {
@@ -1930,16 +1947,16 @@ impl Blockstore {
         current_shred: &Shred,
         slot: Slot,
         location: BlockLocation,
-        write_batch: &mut WriteBatch,
         just_inserted_shreds: &HashMap<(BlockLocation, ShredId), Cow<'_, Shred>>,
-    ) -> Result<()> {
+        parent_meta_working_set: &mut HashMap<(BlockLocation, u64), WorkingEntry<ParentMeta>>,
+    ) -> Result<Option<ParentMeta>> {
         // TODO(ksn): rather than reading and writing from the column directly, prefer to use an in
         // memory map and commit it at the end of the shred batch.
         //
         // If UpdateParentMeta already exists, mark the slot as dead and return early
-        if let Ok(Some(_)) = self.update_parent_meta_cf.get((slot, location)) {
+        if let Ok(Some(_)) = self.parent_meta_cf.get((slot, location)) {
             self.set_dead_slot(slot)?;
-            return Ok(());
+            return Ok(None);
         }
 
         let current_index = current_shred.index();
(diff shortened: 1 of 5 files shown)


qkniep:
If I understand correctly, you now consider multiple `UpdateParent`s to be valid. We probably don't want to handle that case (the block should be invalid) as it introduces unnecessary complexity for something no honest leader will ever do.

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + 1;
+            let next_shred_index = next_fec_set_index * DATA_SHREDS_PER_FEC_BLOCK as u32;
+
+            match self.get_data_shred_from_location(slot, next_shred_index as u64, location) {
+                Ok(Some(bytes)) => (Some(bytes), next_fec_set_index),
+                _ => (None, next_fec_set_index),
+            }
+        } else if current_index > 0 {
```
AshwinSekar: For Case B, we can restrict this to the start of FEC sets (idx % 32 == 0).

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+        let data_complete = current_shred.data_complete();
+
+        // Determine which shred to check for UpdateParent and which FEC set it belongs to
+        let (shred_bytes, target_fec_set) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + 1;
+            let next_shred_index = next_fec_set_index * DATA_SHREDS_PER_FEC_BLOCK as u32;
+
+            match self.get_data_shred_from_location(slot, next_shred_index as u64, location) {
```
AshwinSekar: Instead of `get_data_shred_from_location` we should use the version that prefers the in memory map before doing the slightly more expensive blockstore fetch 

https://github.com/anza-xyz/alpenglow/blob/5f0c96c2d1b88d4c682175cd9bdfcd94db95bb92/ledger/src/blockstore.rs#L2103

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+        &self,
+        current_shred: &Shred,
+        slot: Slot,
+        location: BlockLocation,
+        write_batch: &mut WriteBatch,
+    ) {
+        // Check if UpdateParentMeta already exists and exit early if duplicate found
+        if let Ok(Some(mut existing_meta)) = self.update_parent_meta_cf.get((slot, location)) {
+            if !existing_meta.duplicate_observed {
+                // Mark as duplicate and update
```
AshwinSekar: Agree with Quentin if we see multiple markers we should just mark the block as dead

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+                existing_meta.duplicate_observed = true;
+                let _ = self.update_parent_meta_cf.put_in_batch(
```
AshwinSekar: For these `put_in_batch` and other blockstore operations we should pass the error  up the calling chain and hard unwrap/expect  -  if blockstore is busted we should panic.

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+
+    fn check_for_update_parent(
+        &self,
+        current_shred: &Shred,
+        slot: Slot,
+        location: BlockLocation,
+        write_batch: &mut WriteBatch,
+    ) {
+        // Check if UpdateParentMeta already exists and exit early if duplicate found
+        if let Ok(Some(mut existing_meta)) = self.update_parent_meta_cf.get((slot, location)) {
```
AshwinSekar: Rather than reading and writing from the column directly, prefer to use an in memory map and commit it at the end of the shred batch. You can add it to the `ShredInsertionTracker`, for example check out how the `merkle_root_meta` column is fetched/written to https://github.com/anza-xyz/alpenglow/blob/5f0c96c2d1b88d4c682175cd9bdfcd94db95bb92/ledger/src/blockstore.rs#L1359

This will get us some performance, as we expect shreds from the same slot to arrive close together, so we can avoid having to hit the db so often.

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1901,121 @@ impl Blockstore {
...
+
+        let current_index = current_shred.index();
+        let fec_set_index = current_shred.fec_set_index();
+        let data_complete = current_shred.data_complete();
+
+        // Determine which shred to check for UpdateParent and which FEC set it belongs to
+        let (shred_bytes, target_fec_set) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + 1;
```
AshwinSekar: The fec set index is actually the first shred index, so we want + 32 here. (annoying I know but it's due to allowing variable sized sets)

AshwinSekar (commented):
As a note the math here assumes that SIMD-0317 and SIMD-0366 have been activated, 

We plan to activate these before Alpenglow of course, but we should make sure that this code doesn't cause any unexpected problems if it lands in a software version prior to these flags being activated. My initial scan doesn't see any problems - we'll quietly fail to deserialize, but something to keep in mind for this PR and upcoming changes.

ledger/src/blockstore.rs
ksn6: Done

ledger/src/blockstore.rs
ksn6: Nice - done.

ledger/src/blockstore.rs
ksn6: Ah, I thought we marked the slot as dead later on in the code, which is why I introduced this duplicate bool - just saw that there's a way to do this directly in blockstore.

Done.

ledger/src/blockstore.rs
ksn6: Done.

ledger/src/blockstore.rs
ksn6: Ah, I see - done.

FYI, we'll be adding in tests for all of this once we can turn `BlockComponent`'s into shreds. Will do this in the next PR to keep this one small.

ledger/src/blockstore.rs
ksn6: Ah, nice - I'll add a TODO here for now, and implement this in the next PR once we have tests going.

AshwinSekar: Would prefer making this change in this PR rather than a follow up since it deals with blockstore insertion

ledger/src/blockstore_meta.rs
```diff
@@ -979,6 +979,13 @@ impl OptimisticSlotMetaVersioned {
     }
 }
 
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+pub struct UpdateParentMeta {
```
AshwinSekar: We should instead name this ParentInfo/Meta or something similar and always populate it (can set fec_set_index to 0 for non updated blocks).

Then in the future we can retire the parent field of SlotMeta and switch to using this column exclusively for linking.

ledger/src/blockstore.rs
ksn6: Done - let me know if the updates work.

ledger/src/blockstore_meta.rs
ksn6: Could you elaborate a bit more? Are you suggesting that we keep the `new_parent_*`, add in `parent_*` fields here, and populate the `parent_*` fields everywhere? Or something else?

AshwinSekar: I would say have it be something like
parent slot+ block id
fec set index where the block starts
whether it has a parent update (so we can disqualify blocks with multiple updates

ksn6: Gotcha, I think I get the spirit.

PR #466, opened 2025-09-24
[core] Re-organize batches for bls-sigverify bench

samkim-crypto:
#### Summary of Changes

I used the logs in the current alpenglow test cluster re-organize the batches for testing in the bls-sigverify bench logic.

I first attempted to hard-code some of the vote and certificate batches in the benches, but it was hard to set up the test because the vote and certificate messages do not contain the actual validator keypairs. So I updated the code so that the bench logic freshly generates vote and certificate messages that model the batch of votes and certificate messages seen commonly in the logs.

### DIFF
--- a/core/benches/bls_sigverify.rs
+++ b/core/benches/bls_sigverify.rs
@@ -1,11 +1,13 @@
 #![allow(clippy::arithmetic_side_effects)]
 
 use {
-    criterion::{criterion_group, criterion_main, Bencher, Criterion},
-    crossbeam_channel::{unbounded, Receiver, Sender},
+    criterion::{criterion_group, criterion_main, BatchSize, Criterion, Throughput},
+    crossbeam_channel::unbounded,
     solana_bls_signatures::signature::Signature as BlsSignature,
-    solana_core::{sigverifier::bls_sigverifier::BLSSigVerifier, sigverify_stage::SigVerifyStage},
+    solana_core::{sigverifier::bls_sigverifier::BLSSigVerifier, sigverify_stage::SigVerifier},
+    solana_hash::Hash,
     solana_perf::packet::{Packet, PacketBatch, PinnedPacketBatch},
+    solana_pubkey::Pubkey,
     solana_runtime::{
         bank::Bank,
         bank_forks::BankForks,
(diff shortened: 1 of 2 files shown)


core/benches/bls_sigverify.rs
```diff
@@ -1,172 +1,260 @@
...
 
-struct BenchSetup {
-    stage: SigVerifyStage,
-    packet_sender: Sender<PacketBatch>,
-    consensus_message_receiver: Receiver<ConsensusMessage>,
-    validator_keypairs: Vec<ValidatorVoteKeypairs>,
-}
+const BENCH_SLOT: u64 = 70;
+// TODO(sam): use a small number for now to emulate the current test cluster
+const NUM_VALIDATORS: usize = 100;
```
wen-coding: nit: The current cluster has 50 nodes, we should probably simulate 2000 nodes cluster later

core/benches/bls_sigverify.rs
```diff
@@ -1,172 +1,260 @@
...
+    let msg1 = create_base2_cert_message(env, BENCH_SLOT, hash1);
+    // Cert 2 (Base3 - NotarizeFallback)
+    let msg2 = create_base3_cert_message(env, BENCH_SLOT + 1, hash2);
+
+    let packets = vec![message_to_packet(&msg1), message_to_packet(&msg2)];
+    vec![PinnedPacketBatch::new(packets).into()]
 }
 
-fn bench_bls(c: &mut Criterion) {
+fn bench_votes(c: &mut Criterion) {
```
wen-coding, line 203: nit: maybe we can put some rough numbers here how often we see these two scenarios in the log

core/benches/bls_sigverify.rs
samkim-crypto: Okay updated!

core/benches/bls_sigverify.rs
samkim-crypto: I commented on the distribution of vote batches. 

For certs, I think we need a slightly more sophisticated bench in a follow-up. Pretty much all certs are identical certs, but the numbers in a batch can vary (a single cert in a batch, all the way up to 6 identical certs in a batch). Currently, even if we have 6 identical certs in a single batch (but with different sigs), we do verify each of them.

In a follow-up, I'll address this and then add a more sophisticated bench logic for this case.

wen-coding (approved):

PR #467, opened 2025-09-24
refactor: remove dependency on `alpenglow-vote`

ksn6:
#### Problem
To upstream to `agave`, we need to remove dependencies on `alpenglow-vote`.

#### Summary of Changes
Remove dependencies on `alpenglow-vote`.

### DIFF
--- a/genesis/src/main.rs
+++ b/genesis/src/main.rs
@@ -668,11 +668,7 @@ fn main() -> Result<(), Box<dyn error::Error>> {
         .arg(
             Arg::with_name("alpenglow")
                 .long("alpenglow")
-                .takes_value(true)
-                .help(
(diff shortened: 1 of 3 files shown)


genesis/src/main.rs
```diff
@@ -910,10 +908,6 @@ fn main() -> Result<(), Box<dyn error::Error>> {
         }
     }
 
-    if let Some(alpenglow_so_path) = alpenglow_so_path {
```
wen-coding, line 913: nit: we should probably rename alpenglow_so_path, maybe make it only a boolean.

ksn6: Done.

wen-coding (approved):

PR #471, opened 2025-09-25
Add unittest for consensus_pool_service.

wen-coding:


votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+        solana_votor_messages::{
+            consensus_message::{
+                Certificate, CertificateType, VoteMessage, BLS_KEYPAIR_DERIVE_SEED,
+            },
+            vote::Vote,
+        },
+        std::sync::{Arc, Mutex},
+    };
+
+    struct SetupResult {
```
AshwinSekar: nit would call this ConsensusPoolComponents or something similar

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+    }
+
+    #[test]
+    fn test_standstill() {
+        let setup_result = setup();
+        // Do nothing for 10.1 seconds
+        thread::sleep(Duration::from_millis(10_100));
+        // Verify that we received a standstill event
+        wait_for_event(&setup_result.event_receiver, |event| {
+            if let VotorEvent::Standstill(slot) = event {
```
AshwinSekar: Can we just do something like `assert_matches!(VotorEvent::Standstill(0), event)` instead of the `if let`? Also prefer to use `matches` in the other instances of `wait_for_event`, think it's a little easier to read

wen-coding: Changed. Should be okay in this case, but in other instances we might get other events and I don't want tests to fail just because events arrive out of order.

wen-coding: Oops, actually it doesn't work in this case either. I got VotorEvent::ParentReady(1) before Standstill. Changing it back. Is there a reason we don't derive PartialEq on event?

votor/src/consensus_pool_service.rs
wen-coding: Changed to ConsensusPoolServiceTestComponents.

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+        });
+        setup_result.exit.store(true, Ordering::Relaxed);
+        setup_result.consensus_pool_service.join().unwrap();
+    }
+
+    #[test]
+    fn test_standstill() {
+        let setup_result = setup();
+        // Do nothing for 10.1 seconds
+        thread::sleep(Duration::from_millis(10_100));
```
carllin: CAn we avoid using magic numbers and use constants like `DELTA_STANDSTILL`

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+    #[test]
+    fn test_receive_and_send_consensus_message() {
+        solana_logger::setup();
+        let setup_result = setup();
+        let consensus_message_sender = setup_result.consensus_message_sender;
+        let consensus_pool_service = setup_result.consensus_pool_service;
+
+        // validator 0 to 7 send Notarize on slot 2
+        let block_id = Hash::new_unique();
+        let notarize_vote = Vote::new_notarization_vote(2, block_id);
```
carllin: For readability, could we define a variable `target_slot = 2` to avoid magic numbers

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+                signature: bls_keypair.sign(&vote_serialized).into(),
+                rank: my_rank as u16,
+            });
+            consensus_message_sender.send(consensus_message).unwrap();
+        }
+
+        // Listen on the bls_receiver for certificates
+        wait_for_event(&setup_result.bls_receiver, |event| {
+            if let BLSOp::PushCertificate { certificate } = event {
+                assert_eq!(certificate.certificate.slot(), 2);
```
carllin: can reuse that slot variable here and everywhere else we have an assert

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+        let setup_result = setup();
+        drop(setup_result.consensus_message_sender);
+        // Verify that the consensus pool service exits within 5 seconds
+        let start = Instant::now();
+        while !setup_result.exit.load(Ordering::Relaxed) && start.elapsed() < Duration::from_secs(5)
+        {
+            thread::sleep(Duration::from_millis(100));
+        }
+        assert!(setup_result.exit.load(Ordering::Relaxed));
+        setup_result.consensus_pool_service.join().unwrap();
```
carllin, line 724: can we make a generic test that takes a receiver and tests that this happens for all the  receiver channels as well, `bls_receiver`, `event_receiver`, `commitment_receiver`.

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+        // Verify that we received a produce block event
+        wait_for_event(&setup_result.event_receiver, |event| {
+            if let VotorEvent::ProduceWindow(LeaderWindowInfo { start_slot, .. }) = event {
+                assert_eq!(*start_slot, next_leader_slot.0);
+                true
+            } else {
+                false
+            }
+        });
+        setup_result.exit.store(true, Ordering::Relaxed);
```
carllin: Could we make these send -> receive tests more generic by handing a function `test_sender_receiver()` a set of (sender, receiver, event) and then checking that the receiver gets the event?

We can send construct an array of correct/expected (sender, receiver, event) combinations and loop through it running that function as a sanity check

votor/src/consensus_pool_service.rs
```diff
@@ -412,3 +412,269 @@ impl ConsensusPoolService {
...
+                }
+            }
+            false
+        });
+        setup_result.exit.store(true, Ordering::Relaxed);
+        consensus_pool_service.join().unwrap();
+    }
+
+    #[test]
+    fn test_produce_block_event() {
```
carllin: nit: maybe consistent `fn test_send_receive_` or `fn test_receive` prefix for all these types of tests

wen-coding: Done.

votor/src/consensus_pool_service.rs
wen-coding: Done.

votor/src/consensus_pool_service.rs
wen-coding: Done.

votor/src/consensus_pool_service.rs
wen-coding: Done.

votor/src/consensus_pool_service.rs
wen-coding: Done.

votor/src/consensus_pool_service.rs
wen-coding: Done. In the first case we send a bunch of votes and we should expect three different type of events fire at the same time on different receivers.

It seems to be too much hassle to unite receiver of different types, so I just put the first one into test_send_and_receive.

votor/src/consensus_pool_service.rs
AshwinSekar: ah if we don't care about order then we can just do `matches!(VotorEvent::Standstill(0), event)` right? and wait for event will skip the ones that don't match? 

And `Bank` only impls `PartialEq` for tests, so we could derive `PartialEq` for events in DCOU if needed, although I suspect using match is cleaner

wen-coding: Done.

PR #473, opened 2025-09-25
Add unittest for voting_utils and fix a bug.

wen-coding:
- Add unittest for voting_utils
- Fix a bug found in unittest: Right now we use root bank's epoch to get bls_pubkey_to_rank_map, we should really use the epoch of the slot being voted on.

### DIFF
--- a/perf/src/sigverify.rs
+++ b/perf/src/sigverify.rs
@@ -379,12 +379,10 @@ fn check_for_simple_vote_transaction(
         .checked_add(size_of::<Pubkey>())
         .ok_or(PacketError::InvalidLen)?;
 
-    let program_id = packet
+    if packet
         .data(instruction_program_id_start..instruction_program_id_end)
-        .ok_or(PacketError::InvalidLen)?;
-
-    if program_id == solana_sdk_ids::vote::id().as_ref()
-        || program_id == solana_votor_messages::id().as_ref()
+        .ok_or(PacketError::InvalidLen)?
+        == solana_sdk_ids::vote::id().as_ref()
     {
         packet.meta_mut().flags |= PacketFlags::SIMPLE_VOTE_TX;
     }
--- a/runtime-transaction/src/runtime_transaction/transaction_view.rs
+++ b/runtime-transaction/src/runtime_transaction/transaction_view.rs
@@ -38,24 +38,6 @@ fn is_simple_vote_transaction<D: TransactionData>(
     is_simple_vote_transaction_impl(signatures, is_legacy_message, instruction_programs)
 }
 
-fn is_alpenglow_simple_vote_transaction<D: TransactionData>(
-    transaction: &SanitizedTransactionView<D>,
-) -> bool {
-    let signatures = transaction.signatures();
-    let is_legacy_message = matches!(transaction.version(), TransactionVersion::Legacy);
-    let mut instruction_programs = transaction
-        .program_instructions_iter()
-        .map(|(program_id, _ix)| program_id);
-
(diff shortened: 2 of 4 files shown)


votor/src/voting_utils.rs
```diff
@@ -313,3 +311,293 @@ pub fn generate_vote_message(
...
+            own_vote_sender,
+            bls_sender: unbounded().0,
+            commitment_sender: unbounded().0,
+            wait_to_vote_slot: None,
+            root_bank,
+        }
+    }
+
+    #[test]
+    fn test_generate_vote_message() {
```
carllin: nit: test_generate_own_vote_message()

votor/src/voting_utils.rs
```diff
@@ -313,3 +311,293 @@ pub fn generate_vote_message(
...
+            panic!("Expected BLSOp::PushVote, got {result:?}");
+        }
+
+        // Check that own vote sender receives the vote
+        let received_message = own_vote_receiver.recv().unwrap();
+        assert_eq!(received_message, expected_message);
+    }
+
+    #[test]
+    fn test_wrong_vote_account() {
```
carllin: nit: this covers more than just the wrong vote account. Maybe  break the other edge cases into separate tests

votor/src/voting_utils.rs
```diff
@@ -313,3 +311,293 @@ pub fn generate_vote_message(
...
+        assert!(generate_vote_message(vote, false, &mut voting_context)
+            .unwrap()
+            .is_some());
+
+        // Wrong identity keypair
+        voting_context.identity_keypair = Arc::new(Keypair::new());
+        let vote = Vote::new_notarization_vote(6, Hash::new_unique());
+        assert!(generate_vote_message(vote, true, &mut voting_context)
+            .unwrap()
+            .is_none());
```
carllin, line 516: Do these types of mismatched keys not return VoteError's? Do we have tests to cover the different types of VoteErrors?

votor/src/voting_utils.rs
```diff
@@ -313,3 +311,293 @@ pub fn generate_vote_message(
...
+        let mut voting_context =
+            setup_voting_context_and_bank_forks(own_vote_sender, &validator_keypairs, my_index);
+
+        // If we try to vote for a slot in the future, we should panic
+        let vote = Vote::new_notarization_vote(1_000_000_000, Hash::new_unique());
+        let _ = generate_vote_message(vote, false, &mut voting_context);
+    }
+
+    #[test]
+    fn test_zero_staked_validator() {
```
carllin: nit: test_zero_staked_validator_fails_voting

votor/src/voting_utils.rs
wen-coding: Done.

votor/src/voting_utils.rs
wen-coding: Done.

votor/src/voting_utils.rs
wen-coding: Our theory is that these can all happen on a hot spare when validator operators are busy configuring their hot spare validators correctly. So right now it's all INFO logs, maybe we should upgrade the logs to warning/error instead?

votor/src/voting_utils.rs
wen-coding: Done.

votor/src/voting_utils.rs
carllin: yeah that makes sense to me as a warning

carllin (approved):

PR #475, opened 2025-09-26
Refactor BLS sigverifier out of the sigverify stage

samkim-crypto:
#### Problem

Currently, there is redundant serialization logic in the `verify_batches` and `send_packets` of the bls sigverifier.

#### Summary of Changes
I removed the redundant serialization logic. It wasn't possible to remove the redundant logic without tweaking the `SigVerifier` trait, so I added an associated `VerifiedOutput` type to the trait. This change made computing the stats a little more complicated, so I also added `VerifyPacketBatchesStats` to simplify the organization a bit.

From the perspective of the original ed25519 sigverifier, the changes should be the changes should be pretty minimal.

Fixes #


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -5,7 +5,7 @@ mod stats;
 use {
     crate::{
         cluster_info_vote_listener::VerifiedVoteSender,
-        sigverify_stage::{SigVerifier, SigVerifyServiceError},
+        sigverify_stage::{SigVerifyServiceError, VoteSigVerifier, VoteSigVerifierStats},
```
wen-coding: Since we have TowerBFT Votes as well, per our tradition this should probably be called ConsensusSigVerifier, but I still kind of like BLSSigverifier, because we do BLS verifications here.

core/src/sigverify_stage.rs
```diff
@@ -332,25 +332,41 @@ impl SigVerifyStage {
...
+        discard_time: &Measure,
+        excess_fail: usize,
+        pre_shrink_time_us: u64,
+        pre_shrink_total: usize,
+        num_valid_packets: usize,
+        post_shrink_total: usize,
+        post_shrink_time_us: u64,
+        verify_time: &Measure,
+        batches_len: usize,
+    ) {
```
wen-coding: Do you think it makes sense to take bls_sigverifier completely out of sigverify_stage.rs?

We are not processing transactions here, so there are no further "stages". We just inject directly into consensus pool.

So we can plug into quic streamer directly maybe? No further actions needed from SigVerifyStage.

We will do dedupe ourselves, because the original dedupe just compares bytes, I wonder whether that's much faster than deserialize and realize it's the same vote/cert.

We probably don't need discard_random,

We also don't need shrinking of batches that much.

The others we can easily replicate in our stats.

wen-coding:
Also, let's give this PR a new name, we are moving bls sigverify to our new location.

core/src/bls_sigverify/bls_sigverify_stage.rs
```diff
@@ -0,0 +1,267 @@
...
+
+type Result<T, SendType> = std::result::Result<T, BLSSigVerifyServiceError<SendType>>;
+
+pub struct BLSSigVerifyStage {
+    thread_hdl: JoinHandle<()>,
+}
+
+impl BLSSigVerifyStage {
+    // TODO(sam): This is inherited from regular sigverify for now
+    pub fn discard_excess_packets(batches: &mut [PacketBatch], mut max_packets: usize) {
```
wen-coding, line 72: We can keep this here for now to make this PR small, but I think we can get rid of PacketBatch structure later and fold whatever we still need here into bls_sigverifier.rs.

core/src/bls_sigverify/bls_sigverify_stage.rs
```diff
@@ -0,0 +1,267 @@
...
+            batches_len,
+            verify_time.as_ms(),
+            num_packets,
+            (num_packets as f32 / verify_time.as_s())
+        );
+
+        stats
+            .recv_batches_us_hist
+            .increment(recv_duration.as_micros() as u64)
+            .unwrap();
```
wen-coding, line 185: When we fold bls_sigverify_stage.rs into bls_sigverifier.rs we can fold these stats into stats.rs as well.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -388,60 +359,106 @@ impl BLSSigVerifier {
...
+        let verified_certs = self.verify_certificates(certs_to_verify, bank);
+        self.stats
+            .total_valid_packets
+            .fetch_add(verified_certs.len() as u64, Ordering::Relaxed);
+
+        for cert in verified_certs {
+            // Send the BLS certificate message to certificate pool.
+            match self
+                .message_sender
+                .try_send(ConsensusMessage::Certificate(cert.cert_message))
```
wen-coding, line 413: nit: we can fold the try_send in send_votes and send_certificates into one function maybe?

wen-coding (approved):

PR #485, opened 2025-09-27
feat: actually dump alpenglow metrics

akhi3030:
Closes: #408 

In the earlier PR, we collected the metrics but did not actually dump them.  In this PR, we are also dumping the metrics.

The `Datapoint` API is quite restrictive requiring `&'static str` for various arguments forcing a lot of code duplication.  On my long list of issues to explore is to see if we can make the `Datapoint` API a bit more friendlier to use.

### DIFF
--- a/votor/src/consensus_metrics.rs
+++ b/votor/src/consensus_metrics.rs
@@ -1,7 +1,7 @@
 use {
     histogram::Histogram,
     solana_clock::{Epoch, Slot},
-    solana_metrics::{datapoint::DataPoint, submit},
+    solana_metrics::datapoint_info,
     solana_pubkey::Pubkey,
     solana_votor_messages::vote::Vote,
     std::{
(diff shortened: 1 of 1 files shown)


votor/src/consensus_metrics.rs
```diff
@@ -167,7 +168,90 @@ impl ConsensusMetrics {
...
     /// Performs end of epoch reporting and reset all the statistics for the subsequent epoch.
     fn end_of_epoch_reporting(&mut self) {
-        // TODO: currently, just clearing the stats and not actually reporting
+        for (addr, metrics) in &self.node_metrics {
+            let addr = addr.to_string();
+            let mut datapoint = DataPoint::new("votor_consensus_metrics");
+            datapoint.add_tag("address", &addr);
+
+            datapoint.add_field_u64("notar_vote_count", metrics.notar.entries());
+            if let Ok(v) = metrics.notar.mean() {
```
AshwinSekar: If you use the macro version it will handle parsing options for you, https://docs.rs/solana-metrics/latest/src/solana_metrics/datapoint.rs.html#123

So you could write this as something like:
```rust
datapoint_info!(
  "votor-consensus-metrics",
  "address" => addr,
  ("notar_vote_mean", metrics.notar.mean().ok(), i64),
  ("notar_vote_stddev", metrics.notar.stddev(), i64),
  ...
```

AshwinSekar: Also I don't think `u64` works for influx, you'll need to use `i64`

akhi3030: thanks, much cleaner.

AshwinSekar (approved):

PR #487, opened 2025-09-27
Add event_handler unittests and fix a few problems.

wen-coding:
- Add event_handler unittests
- Found a few problems:
   - We seem to be missing try_final in try_notar (Alpenglow white paper v1, page 22, line 15)
   - Doesn't quite make sense to skip slot 0, so special case that
- Also added a few test only functions


votor/src/event_handler.rs
```diff
@@ -606,6 +617,8 @@ impl EventHandler {
         )?;
         pending_blocks.remove(&slot);
 
+        Self::try_final(my_pubkey, (slot, block_id), voting_context, votes)?;
```
AshwinSekar, line 610: nice catch!

votor/src/event_handler.rs
```diff
@@ -527,7 +527,15 @@ impl EventHandler {
...
             let my_old_pubkey = vctx.vote_history.node_pubkey;
             *my_pubkey = new_pubkey;
-            vctx.vote_history = VoteHistory::restore(ctx.vote_history_storage.as_ref(), my_pubkey)?;
+            if let Ok(vote_history) =
+                VoteHistory::restore(ctx.vote_history_storage.as_ref(), my_pubkey)
+            {
+                warn!("Loaded vote history for {my_pubkey}: {vote_history:?}");
+                vctx.vote_history = vote_history;
+            } else {
+                warn!("No vote history found for {my_pubkey}, starting fresh");
```
AshwinSekar: We shouldn't start fresh when calling set-identity as this could result in slashable votes. Unlike TowerBFT we don't have any information on chain to reconstruct our vote history - we should exit and the operator can fix it.

votor/src/event_handler.rs
```diff
@@ -598,6 +606,9 @@ impl EventHandler {
             voting_context,
         )? {
             votes.push(bls_op);
+        } else {
+            // We may be hot spare so don't do anything else and just return
```
AshwinSekar: We should still continue as normal, we could just be an RPC and returning here will cause us to not report the Processed commitment.

votor/src/event_handler.rs
wen-coding: Hmm, you mean, we should panic here, or return an error?

votor/src/event_handler.rs
wen-coding: Hmm, that covers the commitment part, but should we remove the block from pending_blocks? We basically did nothing, could we screw up the hot spare's states if we proceed here?

AshwinSekar: we shouldn't change anything if we're unstaked. We don't know why we're unstaked (RPC? hot spare that will switch soon/later?). Deviating parts of the protocol could have unintended consequences, e.g. if we don't remove the block from pending_blocks then we'll keep retrying this block later.

I assume you're worried about a case where we are a hot spare that will switch soon. Once the switch happens the new vote history will be loaded which might be different than the current one. We shouldn't deviate from the protocol, once the new vote history is loaded we'll be fine. There are expected to be some slots that we don't vote on during the brief switch period. 

In a V2 we could probably split out unstaked nodes entirely for a minimal voting loop - just rooting and reporting commitments.

votor/src/event_handler.rs
AshwinSekar: Just what the code was doing before, returning an error so the validator can shutdown.

As an improvement we could consider keeping the validator running with the old identity but indicating to the user that set-identity failed - but it needs to be extremely clear in the CLI so that they don't accidentally set-identity on the other box and end up with 2 validators with the same identity.

wen-coding: I'm confused how it's supposed to work for a hot spare:
- vote history storage is specified on command line
- so hot spare restore vote history storage on start up, I assume this one is empty
- when set-identity happens it should re-read the same storage with the new identity

Can the same vote history storage support two identities at the same time? Currently it just kills the event loop in my test.

AshwinSekar: The file should be keyed by identity, e.g. in tower we have:
```
tower-1_9-{$IDENTITY}.bin
```
so on startup  of a hot spare, we populate a vote history file. Before set-identity is called the operator copies in the vote history file of the staked identity. When set-identity is specified the validator should load the new vote history file and continue populating that one.

This is all untested at the moment, so you might have found a legitimate bug if this workflow isn't working

wen-coding: Hmm okay, I think after manually creating a vote history storage file every time  before I switch identity I was able to make the tests pass.

votor/src/event_handler.rs
wen-coding: Changed.

votor/src/event_handler.rs
AshwinSekar: Awesome! I think for a future follow up we can expose a `--require-vote-history` flag in the CLI similar to `--require-tower` and make it evident in that not using the flag when switching to a *staked* identity can cause us to send slashable votes.

If we do that, i'd feel comfortable relaxing the check here.

votor/src/event_handler.rs
```diff
@@ -684,7 +687,10 @@ impl EventHandler {
         // it's not necessary to vote skip prior to it and we won't
         // be able to check vote history if we've already voted on it
         let root_bank = voting_context.root_bank.load();
-        let start = first_of_consecutive_leader_slots(slot).max(root_bank.slot());
+        // No matter what happens, we should not vote skip for slot 0
```
qkniep, line 690: this is wrong and would break liveness!

votor/src/event_handler.rs
qkniep: yes, without this we would have never been able to slow finalize!

votor/src/event_handler.rs
qkniep, line 690: why not simply mark slot 0 as having a Notar vote from us? then we would also never vote skip, at the same time this would get rid of the special case for slot 1 in `try_notar()`

wen-coding: Hmm, I think we currently require a ParentReady on slot 1, I guess the point is to make sure everyone starts at roughly the same time.
But in reality the leader of slot 1 almost never makes it, because of our rolling restart model. So normally in rolling restart model the earlier restarted validators just keep sending skip 1 to 3 waiting for a skip certificate.
Are you suggesting that instead of special casing requiring a ParentReady on slot 1 in try_notar, on start up we actually send Notarize block 0 (even though it's already a root) to start the timer?
Or we should not really send out Notarize block 0 but special case 0 to start the timer?

qkniep: it shouldn't really matter if it works as is, I just thought the easiest way to handle first leader window is to have NO special handling apart from marking genesis as: notarized + finalized + ParentReady (for slot 0) + we voted notar (even though we didn't and no one did), then we would especially not have the special case of having a ParentReady that is not for the first slot in a window (and would still enter the first window with the correct timeouts and the ability to vote either skip or notar on slot 1)

wen-coding: Sounds good, if that's the case I suggest we check this in and complete votor/ upstream. If in the future we feel differently we can change the code in Agave/. @AshwinSekar what do you think?

AshwinSekar: Yeah i think that makes sense, when creating a fresh VoteHistory at genesis, we can additionally mark that we have voted notarize. 

For now this is fine, and we can adjust in a follow up

PR #488, opened 2025-09-28
Add unittest for votor.

wen-coding:


votor/src/votor.rs
```diff
@@ -296,3 +296,113 @@ impl Votor {
...
+            drop_bank_sender,
+            bank_notification_sender: None,
+            leader_window_notifier: Arc::new(LeaderWindowNotifier::default()),
+            event_sender,
+            event_receiver,
+            own_vote_sender,
+            consensus_message_receiver,
+        });
+
+        // Give some time to ensure the thread is waiting
```
AshwinSekar, line 393: Appreciate the effort, but is this test necessary? Feel like it's not really testing anything that the local-cluster tests miss.

I think once the migration is implemented we can have more nuanced tests for this file.

wen-coding: Sure, I do want most files in votor/ to have good test coverage, but since we haven't implemented migration yet guess we can punt on this for now.

PR #493, opened 2025-09-30
Keep genesis_utils.rs in sync with upstream.

wen-coding:
#### Summary of Changes
- We don't really need `bls_keypair` in `ValidatorVoteKeypairs`, they can be derived later (and `ValidatorVoteKeypairs::random()` is currently wrong)
- We don't really use `create_genesis_config_with_leader_ex` in Alpenglow now, let's add the variation when we need it.

### DIFF
--- a/core/src/sigverifier/bls_sigverifier.rs
+++ b/core/src/sigverifier/bls_sigverifier.rs
@@ -15,7 +15,7 @@ use {
     },
     solana_bls_signatures::{
         pubkey::{Pubkey as BlsPubkey, PubkeyProjective, VerifiablePubkey},
-        signature::{Signature as BlsSignature, SignatureProjective},
+        signature::{Signature, SignatureProjective},
     },
     solana_clock::Slot,
     solana_measure::measure::Measure,
@@ -348,7 +348,7 @@ impl BLSSigVerifier {
             .collect();
 
         let verified_optimistically = if let Ok(aggregate_pubkeys) = aggregate_pubkeys_result {
-            let all_signatures: Vec<&BlsSignature> = votes_to_verify
+            let all_signatures: Vec<&Signature> = votes_to_verify
                 .iter()
                 .map(|v| &v.vote_message.signature)
                 .collect();
@@ -595,7 +595,7 @@ mod tests {
     use {
         super::*,
         crossbeam_channel::Receiver,
-        solana_bls_signatures::{Signature, Signature as BLSSignature},
+        solana_bls_signatures::{keypair::Keypair as BLSKeypair, Signature},
         solana_hash::Hash,
         solana_perf::packet::{Packet, PinnedPacketBatch},
         solana_runtime::{
@@ -611,6 +611,7 @@ mod tests {
         solana_votor_messages::{
             consensus_message::{
                 Certificate, CertificateMessage, CertificateType, ConsensusMessage, VoteMessage,
+                BLS_KEYPAIR_DERIVE_SEED,
             },
             vote::Vote,
         },
@@ -918,10 +919,14 @@ mod tests {
 
         for (i, validator_keypair) in validator_keypairs.iter().enumerate().take(num_votes) {
             let rank = i as u16;
-            let bls_keypair = &validator_keypair.bls_keypair;
+            let bls_keypair = BLSKeypair::derive_from_signer(
+                &validator_keypair.vote_keypair,
+                BLS_KEYPAIR_DERIVE_SEED,
+            )
+            .unwrap();
 
             // Sign the payload with the BLS keypair.
-            let signature: BLSSignature = bls_keypair.sign(&vote_payload).into();
+            let signature: Signature = bls_keypair.sign(&vote_payload).into();
 
             // Construct the full message.
             let consensus_message = ConsensusMessage::Vote(VoteMessage {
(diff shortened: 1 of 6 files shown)


core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -595,7 +595,7 @@ mod tests {
     use {
         super::*,
         crossbeam_channel::Receiver,
-        solana_bls_signatures::{Signature, Signature as BLSSignature},
+        solana_bls_signatures::{keypair::Keypair as BLSKeypair, Signature},
```
bw-solana: It would be nice to:

1. separate cosmetic vs. functional changes into different PRs
2. do all of the cosmetic changes together to maintain consistency (e.g. add BLS* prefix to Keypair, Signature, etc. or remove it from all - except where we have conflicting definitions of course)

core/src/sigverifier/bls_sigverifier.rs
```diff
@@ -918,10 +919,14 @@ mod tests {
 
         for (i, validator_keypair) in validator_keypairs.iter().enumerate().take(num_votes) {
             let rank = i as u16;
-            let bls_keypair = &validator_keypair.bls_keypair;
+            let bls_keypair = BLSKeypair::derive_from_signer(
+                &validator_keypair.vote_keypair,
+                BLS_KEYPAIR_DERIVE_SEED,
+            )
```
bw-solana: this is called so many times, it might make sense to add some test helper function like `derive_bls_keypair_from_signer_with_default_seed`

runtime/src/genesis_utils.rs
```diff
@@ -360,21 +341,21 @@ pub fn create_genesis_config_with_leader_ex_no_features(
...
     validator_lamports: u64,
     fee_rate_governor: FeeRateGovernor,
     rent: Rent,
     cluster_type: ClusterType,
     mut initial_accounts: Vec<(Pubkey, AccountSharedData)>,
-    is_alpenglow: bool,
 ) -> GenesisConfig {
-    let validator_vote_account = if is_alpenglow {
+    let is_alpenglow = validator_bls_pubkey.is_some();
+    let validator_vote_account = if let Some(bls_pubkey_compressed) = validator_bls_pubkey {
```
bw-solana: could combine these

runtime/src/genesis_utils.rs
```diff
@@ -360,21 +341,21 @@ pub fn create_genesis_config_with_leader_ex_no_features(
     validator_pubkey: &Pubkey,
     validator_vote_account_pubkey: &Pubkey,
     validator_stake_account_pubkey: &Pubkey,
-    validator_bls_pubkey: Option<&BLSPubkey>,
+    validator_bls_pubkey: Option<[u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE]>,
```
bw-solana: might be nice to name this `bls_pubkey_compressed`

runtime/src/genesis_utils.rs
```diff
@@ -452,14 +433,13 @@ pub fn create_genesis_config_with_leader_ex(
...
```
bw-solana: similarly here, `bls_pubkey_compressed`

runtime/src/genesis_utils.rs
```diff
@@ -233,29 +242,6 @@ pub fn create_genesis_config_with_leader(
         mint_lamports,
         validator_pubkey,
         validator_stake_lamports,
-        false,
-    )
-}
-
-#[cfg(feature = "dev-context-only-utils")]
-pub fn create_genesis_config_with_leader_enable_alpenglow(
```
bw-solana, line 241: is this completely unused?

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Removed cosmetic changes.

core/src/sigverifier/bls_sigverifier.rs
wen-coding: Done.

runtime/src/genesis_utils.rs
wen-coding: Hmm what do you mean? The creation of stake account depends on vote account.

runtime/src/genesis_utils.rs
wen-coding: Changed.

runtime/src/genesis_utils.rs
wen-coding: Changed.

runtime/src/genesis_utils.rs
wen-coding: It's not used in our code currently.

runtime/src/genesis_utils.rs
bw-solana: I just meant this could be written like:
```rust
let (is_alpenglow , validator_vote_account) = if let Some(bls_pubkey_compressed) = validator_bls_pubkey {
```

but it's not a big deal

bw-solana (dismissed):

runtime/src/genesis_utils.rs
wen-coding: I'm fine either way, however this is already in Agave upstream, so I'll keep as is for now. We can later change it in Agave.

PR #495, opened 2025-10-01
refactor: minor improvements to `add_missing_parent_ready()`

akhi3030:
- The function gets the slot as input and then also returns it.  It does not need to return the slot as the caller already knows it.
- clarify that the parent of the `finalized_block` will be returned in the documentation.

### DIFF
--- a/votor/src/event_handler.rs
+++ b/votor/src/event_handler.rs
@@ -283,11 +283,11 @@ impl EventHandler {
                     received_shred,
                     stats,
                 )?;
-                if let Some((ready_slot, parent_block)) =
+                if let Some(parent_block) =
                     Self::add_missing_parent_ready(block, ctx, vctx, local_context)
                 {
                     Self::handle_parent_ready_event(
-                        ready_slot,
+                        slot,
                         parent_block,
                         vctx,
                         ctx,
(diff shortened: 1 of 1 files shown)


votor/src/event_handler.rs
```diff
@@ -283,11 +283,11 @@ impl EventHandler {
                     received_shred,
                     stats,
                 )?;
-                if let Some((ready_slot, parent_block)) =
+                if let Some(parent_block) =
```
wen-coding, line 286: Thanks for doing this. But for refactoring/renaming work like this, can we start to do them in the Agave repo so I don't need to keep syncing upstream?

We plan to fully move to agave in a few weeks (hopefully) and then we can only sync one way from Agave to Alpenglow.

wen-coding: Ah never mind, I haven't moved event_handler to agave yet, sorry

votor/src/event_handler.rs
```diff
@@ -409,12 +409,12 @@ impl EventHandler {
...
                     stats,
                 )?;
-                if let Some((slot, block)) =
+                if let Some(parent_block) =
                     Self::add_missing_parent_ready(block, ctx, vctx, local_context)
                 {
                     Self::handle_parent_ready_event(
-                        slot,
-                        block,
+                        block.0,
```
wen-coding, line 416: Hmm, maybe someday it makes sense to pass in block instead of slot here? We could have two different blocks in the same slot linking to different parent right?

But I agree right now this is not used.

akhi3030: indeed.  that was my thinking as well.

PR #496, opened 2025-10-01
feat: `ConsensusMetrics`: improvements to when the slots start

akhi3030:
For `ConsensusMetrics`: previously we were starting slots in the following manner:
- the first slot in the window started `DELTA_TIMEOUT` time after parent ready was seen.
- the remaining slots started when the skip timer for the previous slots fired

In the alpenglow cluster, we observed that this is too pessimistic and in vast majority of cases, the votes and the block hash arrived before the slots were started.

This PR improves when the slots start.  In particular:
- the first slot in the window starts when we see the parent ready
- the remaining slots start when either the skip timer for the previous slot fires; a block for the previous slot was seen; or a finalization certificate for the previous slot was seen.

Finally as `ConsensusMetrics` is owned by a single struct now, it updates it so that it is no longer stored inside a lock.

### DIFF
--- a/votor/src/event_handler.rs
+++ b/votor/src/event_handler.rs
@@ -335,7 +335,8 @@ impl EventHandler {
             VotorEvent::Timeout(slot) => {
                 info!("{my_pubkey}: Timeout {slot}");
                 if slot != last_of_consecutive_leader_slots(slot) {
-                    vctx.consensus_metrics.record_start_of_slot(slot + 1);
+                    vctx.consensus_metrics
+                        .record_start_of_slot(slot.saturating_add(1));
                 }
                 if vctx.vote_history.voted(slot) {
                     return Ok(votes);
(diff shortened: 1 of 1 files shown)


votor/src/event_handler.rs
```diff
@@ -246,6 +246,9 @@ impl EventHandler {
                 debug_assert!(bank.is_frozen());
                 {
                     let mut metrics_guard = vctx.consensus_metrics.write();
+                    if slot != last_of_consecutive_leader_slots(slot) {
```
akhi3030: note that this is only being captured after the block was replayed not when it was first seen.  Capturing when it was first seen is fairly complicated and we figured this is a close enough approximation.

qkniep: yes, this is what I meant, in the paper we usually ignore execution, but for synchronous execution that makes perfect sense

votor/src/event_handler.rs
```diff
@@ -246,6 +246,9 @@ impl EventHandler {
...
+                        metrics_guard.record_start_of_slot(slot + 1);
```
qkniep: this should be `slot`, not `slot + 1`, right?

votor/src/event_handler.rs
qkniep: however, the condition above should be `first_of_consecutive_leader_slots` not `last_of_consecutive_leader_slots`, right?

akhi3030: Yes!  Actually, I think we do not need the `if`.  In case we see the block before the parent ready, we should still start the timer.

votor/src/event_handler.rs
akhi3030: yes!  Now fixed.

PR #499, opened 2025-10-02
WIP: update `BlockComponent` to reflect SIMD-0337

ksn6:
A bunch of updates to `BlockComponent`:

- Remove `BlockMarkerV2` in favor of having new variants go into `BlockMarkerV1`
- Introduce `BlockHeader`, a new `BlockMarker` variant
- Support the ability to deserialize `Vec<BlockComponent>`
- Update serde given the above
- While we're at it, clean up comments

The first two changes pertain to changes in SIMD-0337.

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -9,8 +9,12 @@ use {
         ser::{Serialize, Serializer},
     },
     solana_account::{AccountSharedData, ReadableAccount},
+    solana_bls_signatures::{
+        keypair::Keypair as BLSKeypair, pubkey::PubkeyCompressed as BLSPubkeyCompressed,
+    },
     solana_instruction::error::InstructionError,
     solana_pubkey::Pubkey,
+    solana_vote_interface::authorized_voters::AuthorizedVoters,
     std::{
         cmp::Ordering,
         collections::{hash_map::Entry, HashMap},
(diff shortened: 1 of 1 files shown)


entry/src/block_component.rs
```diff
@@ -461,21 +482,61 @@ impl BlockComponent {
...
         match (entries.is_empty(), remaining_bytes.is_empty()) {
-            (true, true) => Ok(Self::Entries(Vec::new())),
+            (true, true) => {
+                // Special case: allow exactly 8 zero bytes as empty entry batch
+                if data.len() == ENTRY_COUNT_SIZE {
+                    Ok((Self::EntryBatch(vec![]), ENTRY_COUNT_SIZE))
+                } else {
+                    Err(BlockComponentError::EmptyEntries)
+                }
+            }
```
ksn6, line 497: I'm having to put this special case in to get tests here to pass: https://github.com/anza-xyz/alpenglow/blob/aef12a22bb9f33a0fbdc58b9fa3fd15a9e11ae40/turbine/src/broadcast_stage/standard_broadcast_run.rs#L738-L743

This isn't great, since we technically aren't supposed to allow deserialization of `[0_u8; 8]` into `Vec::<Entry>::empty()` according to [SIMD-0307](https://github.com/solana-foundation/solana-improvement-documents/blob/main/proposals/0307-add-block-footer.md?plain=1#L72). This should rather resolve to an error.

How important is it that we be able to deserialize 8 bytes of zeros into `Vec::<Entry>::empty()`, and can we avoid this entirely?

bw-solana:
We generally make quicker progress and get higher quality reviews when splitting into as many small PRs as possible

PR #500, opened 2025-10-02
feat: improvements to `ConsensusMetrics` based on observed data

akhi3030:
https://anza-xyz.slack.com/archives/C06GZ20BPQB/p1759399012138259?thread_ts=1759232075.506819&cid=C06GZ20BPQB also contains more information.

This PR does the following:

- does not start the the next slot when the finalization cert for the current slot is seen as this is too early.
- only records `block_hash_seen` for the first slot in the window as for the other slots, the metric should be approximately 0
- previously a node was only recording its own vote.  Moved the vote recording to a different place where hopefully it will record votes from all nodes including self

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -6,6 +6,7 @@ use {
             VoteType, MAX_ENTRIES_PER_PUBKEY_FOR_NOTARIZE_LITE,
             MAX_ENTRIES_PER_PUBKEY_FOR_OTHER_TYPES,
         },
+        consensus_metrics::ConsensusMetrics,
         consensus_pool::{
             parent_ready_tracker::ParentReadyTracker,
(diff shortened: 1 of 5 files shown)


votor/src/consensus_pool.rs
```diff
@@ -402,6 +410,17 @@ impl ConsensusPool {
             "Validator stake is zero for pubkey: {validator_vote_key}"
         );
 
+        match self
```
akhi3030, line 413: @qkniep : It seems like if I move this recording a few lines below after the check for the conflict votes, then we will not have the issue you noted of duplicate votes.  What do you think, should I move it?

qkniep: I would keep it as it is for now. If we move it down, we would also ignore all votes before the finalized slot, which we most definitely will see and it would make the metrics harder to analyze.

PR #501, opened 2025-10-02
Replace numbers in consensus_pool_service stats with Saturating<usize>.

wen-coding:
Replace u16/u32/u64 with Saturating<usize> so we can get rid of incr functions.

### DIFF
--- a/votor/src/consensus_pool_service/stats.rs
+++ b/votor/src/consensus_pool_service/stats.rs
@@ -1,96 +1,87 @@
 use {
-    solana_metrics::datapoint_info,
-    std::time::{Duration, Instant},
+    solana_metrics::create_datapoint,
+    std::{
+        num::Saturating,
+        time::{Duration, Instant},
+    },
 };
 
 const STATS_REPORT_INTERVAL: Duration = Duration::from_secs(10);
 
 #[derive(Debug)]
 pub(crate) struct ConsensusPoolServiceStats {
-    pub(crate) add_message_failed: u32,
-    pub(crate) certificates_sent: u16,
-    pub(crate) certificates_dropped: u16,
-    pub(crate) new_finalized_slot: u16,
-    pub(crate) parent_ready_missed_window: u16,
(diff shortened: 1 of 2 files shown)


votor/src/consensus_pool_service/stats.rs
```diff
@@ -1,96 +1,87 @@
...
+            new_finalized_slot: Saturating(new_finalized_slot),
+            parent_ready_missed_window: Saturating(parent_ready_missed_window),
+            parent_ready_produce_window: Saturating(parent_ready_produce_window),
+            received_votes: Saturating(received_votes),
+            received_certificates: Saturating(received_certificates),
+            standstill,
+            prune_old_state_called: Saturating(prune_old_state_called),
+            ..
+        } = self;
+        let datapoint = create_datapoint!(
```
AshwinSekar: any reason to not use `datapoint_info!` like before?

votor/src/consensus_pool_service/stats.rs
```diff
@@ -1,96 +1,87 @@
...
+                parent_ready_produce_window,
                 i64
             ),
-            ("received_votes", self.received_votes, i64),
-            ("received_certificates", self.received_certificates, i64),
-            ("standstill", self.standstill, i64),
-            ("prune_old_state_called", self.prune_old_state_called, i64),
+            ("received_votes", received_votes, i64),
+            ("received_certificates", received_certificates, i64),
+            ("standstill", standstill, i64),
```
AshwinSekar: should we use `bool` instead of `i64` here?

votor/src/consensus_pool_service/stats.rs
wen-coding: Changed.

votor/src/consensus_pool_service/stats.rs
wen-coding: Changed.

AshwinSekar (approved):

PR #503, opened 2025-10-02
block_component: add in a few accessors

ksn6:
`block_component.rs`: add in a few accessors

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -498,6 +498,8 @@ impl BlockComponent {
     /// Check if data looks like an entry batch (non-zero entry count). Returns `None` if we can't
     /// deduce whether the data is an entry batch.
     pub fn infer_is_entries(data: &[u8]) -> Option<bool> {
+        // Per documentation, the first 8 bytes denote the length of an entry batch, where a length
+        // of zero indicates a block marker.
         data.get(..8)
             .and_then(|bytes| bytes.try_into().ok())
             .map(|bytes| u64::from_le_bytes(bytes) != 0)


entry/src/block_component.rs
```diff
@@ -478,6 +486,34 @@ impl BlockComponent {
             (false, false) => Err(BlockComponentError::MixedData),
         }
     }
+
+    /// Check if data looks like an entry batch (non-zero entry count).
+    pub fn infer_is_entries(data: &[u8]) -> Option<bool> {
```
wen-coding, line 500: I assume associated tests are in a follow up PR?

entry/src/block_component.rs
```diff
@@ -478,6 +486,34 @@ impl BlockComponent {
...
+        data.get(..8)
+            .and_then(|bytes| bytes.try_into().ok())
+            .map(|bytes| u64::from_le_bytes(bytes) != 0)
+    }
+
+    /// Check if data looks like a block marker (zero entry count).
+    pub fn infer_is_block_marker(data: &[u8]) -> Option<bool> {
```
wen-coding, line 509: Why Option<bool> instead of bool?

entry/src/block_component.rs
```diff
@@ -478,6 +486,34 @@ impl BlockComponent {
...
+        Self::infer_is_entries(data).map(|is_entries| !is_entries)
+    }
+
+    /// Get entries if this is an entry batch.
+    pub fn as_entries(&self) -> Option<&Vec<Entry>> {
```
wen-coding: Is it possible to merge as_entries and into_entries? Or at least put them together.

entry/src/block_component.rs
```diff
@@ -478,6 +486,34 @@ impl BlockComponent {
...
+        match self {
+            Self::Entries(entries) => Some(entries),
+            _ => None,
+        }
+    }
+
+    /// Get marker if this is a block marker.
+    pub fn as_versioned_block_marker(&self) -> Option<&VersionedBlockMarker> {
+        match self {
+            Self::BlockMarker(marker) => Some(marker),
```
wen-coding, line 516: Sorry for a stupid question, why can't we just expose the enum?

entry/src/block_component.rs
```diff
@@ -478,6 +486,34 @@ impl BlockComponent {
             (false, false) => Err(BlockComponentError::MixedData),
         }
     }
+
+    /// Check if data looks like an entry batch (non-zero entry count).
+    pub fn infer_is_entries(data: &[u8]) -> Option<bool> {
+        data.get(..8)
```
bw-solana, line 503: might be nice to spell out why the 8 here in a comment or something

entry/src/block_component.rs
bw-solana: yeah, I think the bool part is obvious, but would be nice to indicate what Some/None mean in function headers

entry/src/block_component.rs
ksn6: Done - putting these together, as we can't merge (without doing something very unsafe), since one returns an owned type while another returns a shared reference.

entry/src/block_component.rs
ksn6: Not a stupid question :) This is syntactic sugar that saves users one level of nesting + allows users to easily use block markers in tandem with combinator/monad operations (e.g., `and_then`, `map`, `or_else`, etc.).

entry/src/block_component.rs
ksn6: `Some(true)` occurs when we're certain that we're working with a block marker.

`Some(false)` occurs when we're certain that we're working with an entry batch.

`None` occurs when we aren't sure of whether `data` is a block marker. Most commonly, `data` may not have eight bytes present. It doesn't seem correct to specify `false` in this case. The following two cases probably never happen but - in theory, it's possible that we aren't able to turn the first eight bytes into a `[u8; 8]` and then subsequently a `u64`.

entry/src/block_component.rs
ksn6: Yep, may as well test in this PR. Done.

entry/src/block_component.rs
wen-coding: Please put that into comments above the function.

ksn6: Done.

wen-coding (dismissed):

entry/src/block_component.rs
carllin: is it important to return None here instead of just false?

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: It's slightly cleaner / more general to do it this way, since invokers can just do `unwrap_or(false)` if they really want false.

wen-coding: I would slightly prefer `Result<bool, Error>`, but given you don't have a lot of failure cases now I don't really care about it that much. Approving for now and we can change it later if needed.

carllin: > It's slightly cleaner / more general to do it this way, since invokers can just do unwrap_or(false) if they really want false.

Hmm, i would say if the caller is mostly running integrity checks on completed data sets where the buffer is at least the 8 bytes, then they probably only care about the boolean. If there was a consistent risk of the buffer being complete maybe Option would be warranted

Not a huge issue though.

ksn6: Yep, totally agreed. It's the paranoia that's talking from me :) Another minor benefit - we can simply negate this function to infer whether something is a block marker.

PR #504, opened 2025-10-02
block component: introduce header and remove `BlockMarkerV2`

ksn6:
Introduce `BlockHeader` and remove `BlockMarkerV2`, per SIMD-0337.

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -45,17 +45,9 @@
 /// ```text
 /// ┌─────────────────────────────────────────┐
 /// │ Variant ID                   (1 byte)   │
-/// ├─────────────────────────────────────────┤
-/// │ Byte Length                  (2 bytes)  │
-/// ├─────────────────────────────────────────┤
-/// │ Variant Data              (variable)    │
-/// └─────────────────────────────────────────┘
-/// ```
-///
-/// ### BlockMarkerV2 Layout
-/// ```text
-/// ┌─────────────────────────────────────────┐
-/// │ Variant ID                   (1 byte)   │
+/// │   0 = BlockFooter                       │
+/// │   1 = BlockHeader                       │
+/// │   2 = UpdateParent                      │
 /// ├─────────────────────────────────────────┤
 /// │ Byte Length                  (2 bytes)  │
 /// ├─────────────────────────────────────────┤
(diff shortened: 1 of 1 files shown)


entry/src/block_component.rs
```diff
@@ -230,28 +230,7 @@ pub enum VersionedBlockMarker {
...
-/// │ Variant Data              (variable)    │
-/// └─────────────────────────────────────────┘
-/// ```
-///
-/// The byte length field indicates the size of the variant data that follows,
-/// allowing for proper parsing even if unknown variants are encountered.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum BlockMarkerV2 {
-    BlockFooter(VersionedBlockFooter),
+    BlockHeader(VersionedBlockHeader),
```
wen-coding, line 233: Should we do:
BlockFooter(VersionedBlockFooter) = 0,
BlockHeader(VersionedBlockHeader) = 1,
...
?

ksn6: Not sure I follow - could you elaborate on what this enables?

If this is for serialization purposes, we currently handle variant logic here:

```rust
impl BlockMarkerV1 {
    /// Serializes to bytes with variant ID and byte length prefix.
    fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
        let (variant_id, data_bytes) = match self {
            Self::BlockFooter(footer) => (0_u8, footer.to_bytes()?),
            Self::BlockHeader(header) => (1_u8, header.to_bytes()?),
            Self::UpdateParent(update) => (2_u8, update.to_bytes()?),
        };

        write_variant_with_length(variant_id, &data_bytes)
    }

    /// Deserializes from bytes, validating variant ID and byte length.
    fn from_bytes(data: &[u8]) -> Result<Self, BlockComponentError> {
        let (variant_id, payload) = read_variant_with_length(data)?;

        match variant_id {
            0 => Ok(Self::BlockFooter(VersionedBlockFooter::from_bytes(
                payload,
            )?)),
            1 => Ok(Self::BlockHeader(VersionedBlockHeader::from_bytes(
                payload,
            )?)),
            2 => Ok(Self::UpdateParent(VersionedUpdateParent::from_bytes(
                payload,
            )?)),
            _ => Err(BlockComponentError::UnknownVariant {
                variant_type: "BlockMarkerV1".to_string(),
                id: variant_id,
            }),
        }
    }
}
```

carllin:
What was the reason we had BlockMarkerV2 originally instead of folding everything into BlockMarkerV1?

carllin (approved):

PR #505, opened 2025-10-03
block component: `Entries` -> `EntryBatch`

ksn6:
Rename `Entries` -> `EntryBatch`, fix some comments, and improve some tests.

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -1,7 +1,7 @@
-// The following imports are only needed for dev-context-only-utils.
 use {
     crate::vote_state_view::VoteStateView,
     itertools::Itertools,
+    log::warn,
     serde::{
         de::{MapAccess, Visitor},
         ser::{Serialize, Serializer},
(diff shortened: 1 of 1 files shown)


entry/src/block_component.rs
```diff
@@ -1,21 +1,19 @@
...
+//! Block components and friends.
+//!
+//! A `BlockComponent` is either an entry batch or a block marker. Entry batches contain
+//! the transactions for a block, while markers contain metadata like block footers and
+//! parent updates.
+//!
+//! Supported marker types:
+//! - `BlockFooter`: Block production metadata
+//! - `BlockHeader`: Parent slot and block ID
+//! - `UpdateParent`: Fast leader handover support
```
wen-coding: Is this guaranteed to change with your enum definition? Should you refer the readers to the enum instead?

ksn6: It'll be easy to maintain this comment, since the enum definition changing shouldn't be very common. E.g., modifications to this enum each require approved SIMDs.

entry/src/block_component.rs
```diff
@@ -45,33 +45,32 @@
...
-/// │   2 = UpdateParent                      │
 /// ├─────────────────────────────────────────┤
 /// │ Byte Length                  (2 bytes)  │
 /// ├─────────────────────────────────────────┤
 /// │ Variant Data              (variable)    │
 /// └─────────────────────────────────────────┘
 /// ```
 ///
-/// ### BlockFooterV1 Layout
+/// ### BlockMarkerV2 Layout
```
wen-coding: Err, why do we have BlockMarkerV2 again? Is this a failed rebase?

PR #506, opened 2025-10-03
block component: parse `Vec<BlockComponent>`

ksn6:
Support the ability to parse `Vec<BlockComponent>`.

We will use this e.g., when disseminating the last two FEC sets:

(1) the second-to-last FEC set contains tx data
(2) the last FEC set contains a `BlockFooter`


carllin:
might need a rebase

PR #507, opened 2025-10-03
refactor: `do_get_complete_block_with_entries` in `blockstore`

ksn6:
`do_get_complete_block_with_entries` is hard to read in `blockstore.rs`.

I'm moving around some logic to reduce nesting + improve readability.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -3086,15 +3086,12 @@ impl Blockstore {
             allow_dead_slots,
         )?;
 
-        if slot_entries.is_empty() {
+        let Some(slot_entry) = slot_entries.last() else {
             trace!("do_get_complete_block_with_entries() failed for {slot} (slot not full)");
             return Err(BlockstoreError::SlotUnavailable);
-        }
+        };
 
-        let blockhash = slot_entries
-            .last()
-            .map(|entry| entry.hash)
-            .unwrap_or_else(|| panic!("Rooted slot {slot:?} must have blockhash"));
+        let blockhash = slot_entry.hash;
 
         // Build entries summary if requested
         let mut starting_transaction_index = 0;


ledger/src/blockstore.rs
```diff
@@ -3074,97 +3074,112 @@ impl Blockstore {
...
+
+        if slot_entries.is_empty() {
+            trace!("do_get_complete_block_with_entries() failed for {slot} (slot not full)");
+            return Err(BlockstoreError::SlotUnavailable);
+        }
+
+        let blockhash = slot_entries
+            .last()
+            .map(|entry| entry.hash)
+            .unwrap_or_else(|| panic!("Rooted slot {slot:?} must have blockhash"));
```
akhi3030:
```suggestion
        let blockhash = match slot_entries.last() {
            None => {
                trace!("do_get_complete_block_with_entries() failed for {slot} (slot_entries is empty)");
                return Err(BlockstoreError::SlotUnavailable);
            }
            Some(entry) => entry.hash,
        };

```

- We do not need to first check for empty and then panic
- Also fixes the trace message which I suspect was a copy paste from above.

ksn6: Nice, done (albeit a bit differently).

ledger/src/blockstore.rs
```diff
@@ -3074,97 +3074,109 @@ impl Blockstore {
...
+        }
+
+        let (slot_entries, _, _) = self.get_slot_entries_with_shred_info(
+            slot,
+            /*shred_start_index:*/ 0,
+            allow_dead_slots,
+        )?;
+
+        let Some(slot_entry) = slot_entries.last() else {
+            trace!("do_get_complete_block_with_entries() failed for {slot} (slot not full)");
```
akhi3030, line 3090:
```suggestion
            trace!("do_get_complete_block_with_entries() failed for {slot} (slot_entries is empty)");
```

PR #508, opened 2025-10-03
feat: retain votes for the previous 8 slots

akhi3030:
In order to pay out rewards properly, a node needs to observe votes for up to the previous 8 slots.  Otherwise, some vote that are received just after finalization will be dropped and their senders will not fairly rewarded.

Additionally separates the metrics for tracking old votes and old certs in bls_sigverifier.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -1375,11 +1375,21 @@ mod tests {
             &validator_keypairs,
             stakes_vec,
         );
-        let bank0 = Bank::new_for_tests(&genesis.genesis_config);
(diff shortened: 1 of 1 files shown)


qkniep:
We should probably only process those votes if we are in fact the leader for the corresponding slot s+8. Otherwise, we are adding a lot of useless BLS verifications.

akhi3030:
Absolutely agreed, we don't want to do additional BLS verifications unnecessarily.  It will still be interesting to capture the metrics for visibility purposes though.  So I propose the following:

- we should capture the voting metrics in bls_sigverifier before we drop the votes.  This does mean that we capture metrics from potentially unverified votes.  
- as suggested above, we only accept older vote messages if we are going to be a leader in slot+8.

PR #509, opened 2025-10-04
Create VAT Feature Gate then Filter Vote Accounts and deduct VAT when activated

wen-coding:
- Create VAT Feature Gate `alpenglow_vat_and_limit_validators`
- If `alpenglow_vat_and_limit_validators` is activated, filter vote accounts
- If both `alpenglow` and `alpenglow_vat_and_limit_validators` features are activated, actually deduct 1.6 SOL VAT (for tests it's configured to be 1 lamport) from identity accounts of all selected vote accounts once every epoch.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -11,9 +11,7 @@ use {
     },
     crossbeam_channel::Receiver,
     solana_clock::Slot,
-    solana_entry::block_component::{
-        BlockFooterV1, BlockMarkerV1, GenesisCertificate, VersionedBlockMarker,
-    },
+    solana_entry::block_component::{BlockFooterV1, GenesisCertificate, VersionedBlockMarker},
     solana_gossip::cluster_info::ClusterInfo,
     solana_hash::Hash,
     solana_ledger::{
@@ -692,12 +690,10 @@ fn maybe_include_genesis_certificate(parent_slot: Slot, ctx: &LeaderContext) {
     if parent_slot != ctx.genesis_cert.slot || parent_slot == 0 {
         return;
     }
-    let genesis_marker = VersionedBlockMarker::new(BlockMarkerV1::new_genesis_certificate(
-        ctx.genesis_cert.clone(),
-    ));
 
-    let mut poh_recorder = ctx.poh_recorder.write().unwrap();
     // Send the genesis certificate
+    let genesis_marker = VersionedBlockMarker::new_genesis_certificate(ctx.genesis_cert.clone());
+    let mut poh_recorder = ctx.poh_recorder.write().unwrap();
     poh_recorder
         .send_marker(genesis_marker)
         .expect("Max tick height cannot have been reached");
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -444,6 +444,29 @@ impl VersionedBlockMarker {
     pub const fn new(marker: BlockMarkerV1) -> Self {
         Self::V1(marker)
     }
+
+    pub fn new_block_footer(f: BlockFooterV1) -> Self {
+        let f = VersionedBlockFooter::V1(f);
+        let f = BlockMarkerV1::BlockFooter(LengthPrefixed::new(f));
+        VersionedBlockMarker::V1(f)
+    }
+
+    pub fn new_block_header(h: BlockHeaderV1) -> Self {
+        let h = VersionedBlockHeader::V1(h);
+        let h = BlockMarkerV1::BlockHeader(LengthPrefixed::new(h));
+        VersionedBlockMarker::V1(h)
+    }
+
+    pub fn new_update_parent(u: UpdateParentV1) -> Self {
+        let u = VersionedUpdateParent::V1(u);
+        let u = BlockMarkerV1::UpdateParent(LengthPrefixed::new(u));
+        VersionedBlockMarker::V1(u)
+    }
+
+    pub fn new_genesis_certificate(g: GenesisCertificate) -> Self {
+        let g = BlockMarkerV1::GenesisCertificate(LengthPrefixed::new(g));
+        VersionedBlockMarker::V1(g)
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -626,9 +649,7 @@ mod tests {
             wincode::deserialize::<GenesisCertificate>(&bytes).unwrap()
         );
 
-        let marker = VersionedBlockMarker::new(BlockMarkerV1::new_block_footer(
-            VersionedBlockFooter::V1(footer.clone()),
-        ));
+        let marker = VersionedBlockMarker::new_block_footer(footer.clone());
         let bytes = wincode::serialize(&marker).unwrap();
         assert_eq!(
             marker,
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -6345,10 +6345,7 @@ pub mod tests {
         },
         solana_clock::{DEFAULT_MS_PER_SLOT, DEFAULT_TICKS_PER_SLOT},
         solana_entry::{
-            block_component::{
-                BlockHeaderV1, BlockMarkerV1, UpdateParentV1, VersionedBlockHeader,
-                VersionedBlockMarker, VersionedUpdateParent,
-            },
+            block_component::{BlockHeaderV1, UpdateParentV1, VersionedBlockMarker},
             entry::{next_entry, next_entry_mut},
         },
         solana_hash::Hash,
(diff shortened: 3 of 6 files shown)


runtime/src/bank.rs
```diff
@@ -211,6 +211,15 @@ pub const SECONDS_PER_YEAR: f64 = 365.25 * 24.0 * 60.0 * 60.0;
...
 pub const MAX_LEADER_SCHEDULE_STAKES: Epoch = 5;
 
+const MAX_ALPENGLOW_VOTE_ACCOUNTS: usize = 2000;
+
+// Minimum balance for an identity account to be considered for Alpenglow VAT.
+#[cfg(not(feature = "dev-context-only-utils"))]
+const ALPENGLOW_VAT_TO_BURN_PER_EPOCH: u64 = 1_600_000_000; // 1.6 SOL
+
+#[cfg(feature = "dev-context-only-utils")]
+const ALPENGLOW_VAT_TO_BURN_PER_EPOCH: u64 = 1; // For tests only require 1 lamport
```
AshwinSekar, line 226: does this cause problem in local-cluster tests or just unit tests? Would it also make sense to check `ClusterType` and use 1 lamport in Development clusters?

runtime/src/bank.rs
```diff
@@ -5644,6 +5680,34 @@ impl Bank {
...
+        stakes
+    }
+
+    fn get_all_identity_account_balances(&self) -> HashMap<Pubkey, u64> {
+        self.vote_accounts()
+            .iter()
+            .map(|(_, (_, account))| {
+                let identity = account.node_pubkey();
+                let balance = self.get_balance(identity);
+                (*identity, balance)
```
AshwinSekar: should we be collecting collisions here for multiple vote accounts with the same identity? Or is that handled in `clone_and_filter_for_alpenglow`?

runtime/src/bank.rs
wen-coding: I think I do want 1.6SOL for devnet and testnet. The invalidator cluster seems to have ~1000 SOL in identity account so they are not affected.

This is mainly to make sure unittests and local-cluster tests don't break.

runtime/src/bank.rs
wen-coding: We plan to enforce the rule vote account and identity account are 1:1. It will be in another PR though.

runtime/src/epoch_stakes.rs
```diff
@@ -24,33 +25,40 @@ pub struct BLSPubkeyToRankMap {
...
 }
 
+pub(crate) fn bls_pubkey_compressed_bytes_to_bls_pubkey(
+    bls_pubkey_compressed_bytes: [u8; BLS_PUBLIC_KEY_COMPRESSED_SIZE],
+) -> Option<BLSPubkey> {
+    let bls_pubkey_compressed: BLSPubkeyCompressed =
+        bincode::deserialize(&bls_pubkey_compressed_bytes).ok()?;
+    BLSPubkey::try_from(bls_pubkey_compressed).ok()
+}
+
```
carllin, line 35: is it possible that somebody could register a BLS pubkey that doesn't deserialize? Seems like we should enforce this during registration of the key and filter out those possibilities.

Then we can hard unwrap here which greatly simplifies this

vote/src/vote_account.rs
```diff
@@ -132,6 +132,11 @@ impl VoteAccounts {
         if valid_entries.is_empty() {
             warn!("no valid alpenglow vote accounts found");
         }
+        info!(
+            "Out of {} vote accounts, {} are valid alpenglow vote accounts after filtering",
+            self.vote_accounts.len(),
+            valid_entries.len()
+        );
```
carllin, line 136: This function is a bit unwieldy, because it doesn't seem like  `identity_account_balances` needs to be passed in, it can be entirely derived from fields already belonging to `VoteAccounts`

If we pass in the minimum_identity_account_balance, we have all identity stakes via VoteAccounts::staked_nodes(), and we should be able to run the stake filter without being passed it.

runtime/src/bank.rs
carllin: This seems like it could be non-deterministic if multiple vote accounts have the same identity. Could we add a sum/filter by highest stake for now to make things consistent until vote account and identity are 1:1

On second thought, don't we already have this functionality in VoteAccounts::staked_nodes()?

runtime/src/epoch_stakes.rs
wen-coding: We would of course check when the BLS pubkey is registered. I didn't make it `unwrap` because I was worried that somehow one bad pubkey sneaking in can halt the whole chain (let's say we have a bug somewhere). It's not easy to modify other people's vote account for one bad pubkey.

vote/src/vote_account.rs
wen-coding: We are filtering on the balance in the identity account, not stake of the vote account. I don't think I have that information in `staked_nodes`, do I?

runtime/src/bank.rs
wen-coding: Hmm, I can actually filter out vote accounts sharing the same identity. Let me do that.

wen-coding: Wait, can't do that, you can attack others by setting node identity to other people's node identity. Hmm, let me just use the highest stake for now, err, this gets messy

wen-coding: Let me check with Justin what the plan is about making identity account and vote account 1:1. I think it will happen before Alpenglow launch.

jstarry: I'm not aware of any plan to make identity and vote accounts 1:1. It's not super trivial to do so I think. By the way, in order to set a node identity for a vote account, the node identity must sign the transaction, so it's not possible for someone else use your node identity unless they are able to sign for it.

wen-coding: That's a good point. I'll just filter out all vote accounts sharing identity account with others here and work with you to enforce 1:1 on identity and vote account then.

wen-coding: Done, removed all vote accounts sharing identity account with others in this PR. I'll follow up on enforcing 1:1 on identity and vote account with Justin separately.

vote/src/vote_account.rs
carllin: ah yeah messed that up, sleepy brained sorry

runtime/src/bank.rs
wen-coding: Also this part is also obsolete. Per trent's suggestion, we now deduct VAT out of vote account instead of identity account.

runtime/src/bank.rs
```diff
@@ -2265,6 +2283,32 @@ impl Bank {
...
+                        None
+                    }
+                })
+        {
+            // burn VAT from each staked vote account
+            let vat = ALPENGLOW_VAT_TO_BURN_PER_EPOCH;
+            let mut account = self.get_account(vote_pubkey).unwrap();
+            total_vat += vat;
+            account.set_lamports(account.lamports().saturating_sub(vat));
+            self.store_account(vote_pubkey, &account);
```
AshwinSekar: for performance i believe it's better to use `store_accounts` so that the accounts db thread can do some work in parallel.

I would suggest buffering the modified accounts and then storing them all at once

runtime/src/genesis_utils.rs
```diff
@@ -331,7 +332,10 @@ pub fn activate_all_features(genesis_config: &mut GenesisConfig) {
 pub fn do_activate_all_features<const IS_ALPENGLOW: bool>(genesis_config: &mut GenesisConfig) {
     // Activate all features at genesis in development mode
     for feature_id in FeatureSet::default().inactive() {
-        if IS_ALPENGLOW || *feature_id != agave_feature_set::alpenglow::id() {
+        if IS_ALPENGLOW
+            || (*feature_id != agave_feature_set::alpenglow::id()
+                && *feature_id != agave_feature_set::alpenglow_vat_and_limit_validators::id())
```
AshwinSekar, line 337: i don't think we need to remove this from default features, it would be good to start testing out the limiting portion on test clusters and such

wen-coding: The problem is that if I don't do it, then some local_cluster tests which writes no BLS pubkey in genesis config will instantly fail. I think we still need to test cases where vote accounts have no BLS pubkeys?

runtime/src/bank.rs
wen-coding: Done.

runtime/src/genesis_utils.rs
AshwinSekar: got it. i'm fine with this, and then closer to release we can make this the default

runtime/src/bank.rs
```diff
@@ -2265,6 +2283,30 @@ impl Bank {
...
+            .iter()
+            .filter_map(|(vote_pubkey, stake)| {
+                if *stake == 0u64 {
+                    None
+                } else {
+                    // burn VAT from each staked vote account
+                    let vat = ALPENGLOW_VAT_TO_BURN_PER_EPOCH;
+                    let mut account = self.get_account(vote_pubkey).unwrap();
+                    total_vat += vat;
+                    account.set_lamports(account.lamports().saturating_sub(vat));
```
AshwinSekar: nit: IIUC at this point all of these accounts should have `>= ALPENGLOW_VAT_TO_BURN_PER_EPOCH` in the vote account:
i would prefer to `checked_sub(vat).unwrap()`

AshwinSekar (dismissed):

runtime/src/bank.rs
wen-coding: Done.

AshwinSekar (approved):

PR #510, opened 2025-10-05
Add back the Alpenglow vote account tests.

wen-coding:
Deleted all Alpenglow vote account tests in https://github.com/anza-xyz/alpenglow/pull/461.

We should add back the tests to make sure we can perform all the operations on Alpenglow vote accounts.

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -506,24 +506,17 @@ mod tests {
             authorized_withdrawer: Pubkey::new_unique(),
             commission: rng.gen(),
         };
-
-        if is_alpenglow {
+        let vote_state_versions = if is_alpenglow {
             let bls_pubkey_compressed: BLSPubkeyCompressed =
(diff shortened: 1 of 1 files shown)


vote/src/vote_account.rs
```diff
@@ -480,46 +482,76 @@ mod tests {
...
+                epoch: rng.gen(),
+                leader_schedule_epoch: rng.gen(),
+                unix_timestamp: rng.gen(),
+            };
+
+            let vote_state = VoteStateV3::new(&vote_init, &clock);
+            AccountSharedData::new_data(
+                rng.gen(), // lamports
+                &VoteStateVersions::new_v3(vote_state),
+                &solana_sdk_ids::vote::id(), // owner
```
carllin: nit: Can factor this out of the `if/else` if we create a common `VoteStateVersions`

carllin (approved):

Issue #406, opened 2025-08-20
Need special handling for leaders at s+8


wen-coding:
For example
- cert pool can’t just clean up at root, need to keep the data longer
- need to keep verifying and packing votes even after a cert is constructed 
- should probably add a reasonable cutoff though

PR #513, opened 2025-10-06
feat: keep older vote messages for leaders in bls sigverifier

akhi3030:
#### Problem
If we discard vote messages for any slots earlier than root slot, then the node that will be the leader in 8 slots will not be able to include them in the certs for rewards.

#### Summary of Changes

Updates bls sigverifier to keep older votes.

Part of #406

### DIFF
--- a/core/benches/bls_sigverify.rs
+++ b/core/benches/bls_sigverify.rs
@@ -5,6 +5,7 @@ use {
     crossbeam_channel::unbounded,
     solana_bls_signatures::signature::Signature as BlsSignature,
     solana_core::bls_sigverify::bls_sigverifier::BLSSigVerifier,
+    solana_gossip::{cluster_info::ClusterInfo, contact_info::ContactInfo},
     solana_hash::Hash,
     solana_ledger::leader_schedule_cache::LeaderScheduleCache,
     solana_perf::packet::{Packet, PacketBatch, PinnedPacketBatch},
@@ -16,6 +17,8 @@ use {
             create_genesis_config_with_alpenglow_vote_accounts, ValidatorVoteKeypairs,
         },
     },
+    solana_signer::Signer,
+    solana_streamer::socket::SocketAddrSpace,
     solana_votor::consensus_pool::vote_certificate_builder::VoteCertificateBuilder,
     solana_votor_messages::{
         consensus_message::{Certificate, ConsensusMessage, VoteMessage},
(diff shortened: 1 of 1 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -135,8 +138,22 @@ impl BLSSigVerifier {
 
             match message {
                 ConsensusMessage::Vote(vote_message) => {
-                    // Only need votes newer than root slot
-                    if vote_message.vote.slot() <= root_bank.slot() {
+                    // If node will be a leader in 8 slots, it needs to keep older vote messages otherwise, discard older messages to avoid unnecessary verification compute.
+                    let vote_slot = vote_message.vote.slot();
+                    let slot_to_check = match self
+                        .leader_schedule_cache
+                        .slot_leader_at(vote_slot + 8, None)
```
akhi3030, line 149: The `8` should be defined as a const somewhere.  Not sure where is a good place though.  Suggestions from reviewers are welcome.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -135,8 +138,22 @@ impl BLSSigVerifier {
...
+                    {
+                        Some(addr) => {
+                            if addr == self.my_pubkey {
+                                vote_slot + 8
+                            } else {
+                                vote_slot
+                            }
+                        }
+                        None => vote_slot,
```
qkniep, line 158: when would this case occur? only in the edge case where we see a vote for the end of epoch `e` but haven't calculated the leader schedule for the epoch `e+1` yet?

akhi3030: It does seem to go into a bit of a rabbit hole.  

I assume we are not using a `fixed_schedule` in which case, we will never hit this case.

Otherwise, this seems to be related to `EpochSchedule::leader_schedule_slot_offset`.  This seems to be set to `DEFAULT_LEADER_SCHEDULE_SLOT_OFFSET` which seems to be set to `DEFAULT_SLOTS_PER_EPOCH`.  

So I would deduce that we are computing the schedule one entire epoch in advance.

akhi3030: It also seems important to ensure that we are sharing the same `LeaderScheduleCache` in various places.  I haven't validated it.  Otherwise, we will be missing some computations.

core/src/tpu.rs
```diff
@@ -337,10 +338,13 @@ impl Tpu {
 
         let alpenglow_sigverify_stage = {
             let sharable_banks = bank_forks.read().unwrap().sharable_banks();
+            let my_pubkey = cluster_info.id();
```
AshwinSekar: sorry to jut in, we'll need to pass `cluster_info` into the sigverifier so that it is aware of identity changes.
At some coarse interval we should check if our pubkey changes and if so update `my_pubkey` - otherwise we'll end up discarding votes if `set-identity` is called during operation.

akhi3030: that is great input.  Yes, I'll fix this shortly.  Please keep them coming!

PR #514, opened 2025-10-06
feat: keep older vote messages for leaders in consensus pool

akhi3030:
#### Problem

If we discard vote messages for any slots earlier than root slot, then the node that will be the leader in 8 slots will not be able to include them in the certs for rewards.

#### Summary of Changes

Similar to https://github.com/anza-xyz/alpenglow/pull/513, this updates the consensus pool to keep older votes.

Part of #406

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -21,6 +21,7 @@ use {
     solana_clock::{Epoch, Slot},
     solana_epoch_schedule::EpochSchedule,
     solana_hash::Hash,
+    solana_ledger::leader_schedule_cache::LeaderScheduleCache,
     solana_pubkey::Pubkey,
     solana_runtime::{bank::Bank, epoch_stakes::VersionedEpochStakes},
     solana_votor_messages::{
@@ -122,6 +123,7 @@ pub struct ConsensusPool {
     /// Slot stake counters, used to calculate safe_to_notar and safe_to_skip
     slot_stake_counters_map: BTreeMap<Slot, SlotStakeCounters>,
     pub(crate) consensus_metrics: Arc<PlRwLock<ConsensusMetrics>>,
+    leader_schedule_cache: Arc<LeaderScheduleCache>,
 }
 
 impl ConsensusPool {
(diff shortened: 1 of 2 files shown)


votor/src/consensus_pool.rs
```diff
@@ -422,7 +426,20 @@ impl ConsensusPool {
...
+            Some(addr) => {
+                if addr == self.my_pubkey {
+                    vote_slot.saturating_add(8)
+                } else {
+                    vote_slot
+                }
+            }
+            None => vote_slot,
+        };
+        if slot_to_check < root_slot {
```
qkniep, line 442: why is this `<` here, whereas we use `<=` in BLSVerifier?

akhi3030: good catch, I wanted to ask the original authors about that as well.  

Looking at: https://github.com/anza-xyz/alpenglow/commit/308af20343f79b9c246ccef69c4728df1ffb3900#diff-2077122be05c942393f87ed53352bb244607026554c88290d4d76b98c1124801R197 and https://github.com/anza-xyz/alpenglow/commit/7e8e58bc4afb7d980c7f48864724f79ea6f89f35#diff-ea37759cc14ea84779ba69701cf8244e8314ad695cb412f76a4e3f103d57ffa0R129, it seems like @wen-coding is the author in both cases.

@wen-coding: do you know why we use `<` here and `<=` in BLSVerifier?  I would've expected us to use `<=` in both cases.

wen-coding: I was worried that changing `<=` to `<` in BLSVerifier will block all the standstill messages related to root so we don't know how many others are agreeing with me. We don't use it for now, but maybe we want to output a warning "I'm in standstill and how many people are at the same slot with me".

rogerANZA:
> separate rewards container:

This was the original plan, to handle the rewards with a separate program. Somebody told us not to do that. In any case, I'm still a bit insecure about the hole one vote is enough for 100% rewards plan. I'm trying to cook something better.

akhi3030:
> > separate rewards container:
> 
> This was the original plan, to handle the rewards with a separate program. Somebody told us not to do that. In any case, I'm still a bit insecure about the hole one vote is enough for 100% rewards plan. I'm trying to cook something better.

@rogerANZA : by separate program do you mean a separate smart contract?  The proposal from Ashwin is to use a separate rust component inside anza.

> In any case, I'm still a bit insecure about the hole one vote is enough for 100% rewards plan. I'm trying to cook something better.


Cool, let us know how you proceed.

votor/src/consensus_pool.rs
qkniep: If you have finalized a block in slot `s`, even during standstill there is no relevant information gained by seeing a finalization from someone for slot `s`. Sending the finalization during standstill only matters for nodes that are actually behind to catch up to the latest finalized slot.

PR #516, opened 2025-10-06
Add clone and filter for alpenglow to vote accounts

wen-coding:
#### Summary of Changes
Implement `clone_and_filter_for_alpenglow` for Alpenglow VAT:
- Filter out any vote accounts without BLS pubkey
- Given `minimum_identity_account_balance` and `identity_account_balances`, filter out any vote account without required balance
- If we have more than `max_vote_accounts` vote accounts after above filtering, sort by stake and truncate
- If any vote account in the resulting list has the same stake as any truncated vote account, just remove this vote account as well (A and B have the same stake amount, it's unfair to keep A in and kick B out just because of Pubkey difference, because that can be grinded)
- If we end up with an empty list (could happen if everyone in the world has the same stake, happens in tests, doesn't happen in real world), log a warning

### DIFF
--- a/vote/src/vote_account.rs
+++ b/vote/src/vote_account.rs
@@ -1064,7 +1064,7 @@ mod tests {
             MIN_STAKE_FOR_STAKED_ACCOUNT,
             &identity_balances,
         );
-        assert!(filtered.len() <= num_alpenglow_nodes);
+        assert_eq!(filtered.len(), num_alpenglow_nodes);
         // Check that all filtered accounts have bls pubkey
         for (_pubkey, (_stake, vote_account)) in filtered.vote_accounts.iter() {
             assert!(vote_account


vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
     >,
 }
 
+impl VoteAccounts {
+    pub fn clone_and_filter_for_alpenglow(
```
bw-solana, line 84: maybe `clone_and_filter_on_balance`?

vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
     >,
 }
 
+impl VoteAccounts {
```
bw-solana, line 70: I think it would be good to add the full filtering rules as a function header here (basically what you put in the PR description)

vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
...
+        let mut entries_to_sort: Vec<(&Pubkey, &VoteAccount, u64)> = self
+            .vote_accounts
+            .iter()
+            .filter_map(|(pubkey, (stake, vote_account))| {
+                if vote_account
+                    .vote_state_view()
+                    .bls_pubkey_compressed()
+                    .is_some()
+                    && *stake != 0u64
+                {
```
bw-solana, line 106: could invert to reduce nesting
```rust
                if vote_account
                    .vote_state_view()
                    .bls_pubkey_compressed()
                    .is_none()
                    || *stake == 0u64
                {
                    // Invalid or unstaked vote account
                    return None;
                }
```

I think you could also suck the other check inside here:
```rust
                if vote_account
                    .vote_state_view()
                    .bls_pubkey_compressed()
                    .is_none()
                    || *stake == 0u64
                    || *identity_account_balances.get(pubkey)? < minimum_identity_account_balance
                {
                    // Invalid vote account, unstaked vote account, or insufficiently funded identity account
                    return None;
                }
```

vote/src/vote_account.rs
```diff
@@ -47,6 +48,14 @@ struct VoteAccountInner {
     vote_state_view: VoteStateView,
 }
 
+pub fn sort_pubkey_and_stake_pair<T>(
```
bw-solana: do we use this anywhere?

vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
...
+                    }
+                    Some((pubkey, vote_account, *stake))
+                } else {
+                    None
+                }
+            })
+            .collect();
+        // Sort by stake descending
+        if entries_to_sort.len() > max_vote_accounts {
+            entries_to_sort.sort_by(|a, b| b.2.cmp(&a.2));
```
bw-solana: should we be calling `sort_pubkey_and_stake_pair`?

vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
...
+            // Find the stake of the first one being in the truncated list (so it's not max_vote_accounts - 1)
+            let floor_stake = entries_to_sort.get(max_vote_accounts).unwrap().2;
+            entries_to_sort.truncate(max_vote_accounts);
+            // Per SIMD 357, we remove all vote accounts with stake equal to the first truncated one.
+            entries_to_sort.retain(|(_pubkey, _vote_account, stake)| *stake > floor_stake);
```
bw-solana: I think we could get rid of these lines (and the iterator below) in favor of something like:
```rust
entries_to_sort
    .into_iter()
    .take_while(|&&(_, _, stake)| stake > floor_stake)
    .map(|(pubkey, vote_account, stake)| (*pubkey, (stake, vote_account.clone())))
    .collect();
```

this would save the shift that comes with retain as well as double iteration

vote/src/vote_account.rs
```diff
@@ -65,6 +74,74 @@ pub struct VoteAccounts {
...
+            vote_accounts: Arc::new(valid_entries),
+            staked_nodes: OnceLock::new(),
+        }
+    }
+
+    pub fn identity_accounts_for_staked_nodes(&self) -> Vec<&Pubkey> {
+        self.vote_accounts
+            .iter()
+            .filter_map(|(_, (stake, vote_account))| {
+                if *stake != 0u64 {
```
bw-solana, line 146: do we care about `bls_pubkey_compressed` here?

vote/src/vote_account.rs
```diff
@@ -550,6 +627,30 @@ mod tests {
...
+        num_nodes: usize,
+        num_alpenglow_nodes: usize,
+        stake_per_node: Option<u64>,
+    ) -> (VoteAccounts, HashMap<Pubkey, u64>) {
+        let mut vote_accounts = VoteAccounts::default();
+        let mut identity_balances = HashMap::new();
+        for index in 0..num_nodes {
+            let is_alpenglow = index < num_alpenglow_nodes;
+            let pubkey = Pubkey::new_unique();
+            let stake = stake_per_node.unwrap_or_else(|| rng.gen_range(1..997));
```
bw-solana: should comment on the magic numbers or pull into a const

vote/src/vote_account.rs
```diff
@@ -550,6 +627,30 @@ mod tests {
...
+    // generated have non-zero stake.
+    // Also we return a default HashMap of identity account balances.
+    fn new_staked_vote_accounts<R: Rng>(
+        rng: &mut R,
+        num_nodes: usize,
+        num_alpenglow_nodes: usize,
+        stake_per_node: Option<u64>,
+    ) -> (VoteAccounts, HashMap<Pubkey, u64>) {
+        let mut vote_accounts = VoteAccounts::default();
+        let mut identity_balances = HashMap::new();
```
bw-solana: we could pre-alloc the HashMap

vote/src/vote_account.rs
```diff
@@ -891,4 +992,153 @@ mod tests {
...
+    #[test]
+    fn test_clone_and_filter_for_alpenglow_truncates() {
+        let mut rng = rand::thread_rng();
+        let current_limit = 3000;
+        let (vote_accounts, identity_balances) =
+            new_staked_vote_accounts(&mut rng, current_limit, current_limit, None);
+        // All vote accounts should be returned if the limit is high enough.
+        let filtered = vote_accounts.clone_and_filter_for_alpenglow(
+            current_limit + 500,
+            1,
```
bw-solana: might be nice to have some `const MIN_IDENTITY_ACCOUNT_BALANCE: usize = 1;` that can be re-used for these tests

vote/src/vote_account.rs
```diff
@@ -891,4 +992,153 @@ mod tests {
...
+        let mut rng = rand::thread_rng();
+        let num_alpenglow_nodes = 2000;
+        // Check that non-alpenglow accounts are kicked out, 2000 accounts with bls pubkey, 1000 accounts without.
+        let num_nodes = num_alpenglow_nodes + 1000;
+        let (vote_accounts, identity_balances) =
+            new_staked_vote_accounts(&mut rng, num_nodes, num_alpenglow_nodes, None);
+        let new_limit = num_alpenglow_nodes + 500;
+        let filtered =
+            vote_accounts.clone_and_filter_for_alpenglow(new_limit, 1, &identity_balances);
+        assert!(filtered.len() <= num_alpenglow_nodes);
```
bw-solana: should this be strictly `==` ?

vote/src/vote_account.rs
wen-coding: It's a little more than filter on balance, we also check BLS pubkey as well. Any other suggestions?

vote/src/vote_account.rs
wen-coding: Because we allocate account with random stake, we might hit the case that several accounts on the border have the same amount of stake, so we can't use == here.

vote/src/vote_account.rs
wen-coding: Done.

vote/src/vote_account.rs
wen-coding: Nice, changed.

vote/src/vote_account.rs
wen-coding: Oops, removed.

vote/src/vote_account.rs
wen-coding: We don't care about sorting pubkey here, because we remove everything on the border with the same stake. Added comments.

vote/src/vote_account.rs
wen-coding: It doesn't make a difference, this one is only for retrieving identity pubkeys from vote accounts so we can look up balances. I prefer to remove the `bls_pubkey_compressed` check here to keep the code simple, we will filter them out in `clone_and_filter` function above.

vote/src/vote_account.rs
wen-coding: Good point, changed into const.

vote/src/vote_account.rs

vote/src/vote_account.rs
wen-coding: done.

vote/src/vote_account.rs
bw-solana: yeah, good point. There's also the identity balance check. I'm not sure what the best name might be

vote/src/vote_account.rs
bw-solana: yeah, I guess the implication is we potentially have a slightly larger hashmap with entries that we never ask for. In exchange we can eliminate an additional check. I'm fine either way

vote/src/vote_account.rs
bw-solana: But we shouldn't have to drop any nodes based on stake here, right? My understanding is we filter out all the vote accounts that don't have a BLS pubkey, this drops us below the 2500 node limit, and we can just return everything that's still valid

vote/src/vote_account.rs
bw-solana: Bump on this one

vote/src/vote_account.rs
wen-coding: We could, because when the limit is 2000, we will drop everyone with the same stake as the 2001st one (given that they all have enough balance and BLS pubkey), so that can make the numbers sightly below 2000 when you assign stakes randomly.

vote/src/vote_account.rs
wen-coding: With a simple lookup on 2000+ vote accounts I think it should be okay. Theoretically someone could spam us by adding vote accounts, but they could as easily add a BLS pubkey and stake 1 lamport, so I don't think it can help us much if we have intentional attackers.

vote/src/vote_account.rs
bw-solana: But aren't these the parameters for the filtering associated w/ this assert?:
- 2k Alpenglow nodes
- 1k non-alpenglow nodes (will get filtered out)
- 2.5k node limit

So I'm assuming we should never enter this conditional: https://github.com/anza-xyz/alpenglow/pull/516/files/6ce59175c8d0647e88d29338c9b03046b07e7b2b#diff-a1bec2a8003e0433e167e7d8b9ac1f2b9b537ad2625e0c1a479187917555d04eR110

What am I missing?

vote/src/vote_account.rs
wen-coding: Oops, somehow missed this one, done.

vote/src/vote_account.rs
wen-coding: Hmm okay, I think this is the leftover when I made the stake 0 to 997. Thanks for catching this, changed.

bw-solana: yeah, I understand why the other tests can't have strict checks due to random stake amount variation. I was assuming some copy+paste thing

PR #518, opened 2025-10-07
wip: type for slice root

kobi-lizard:
#### Problem
Hash type is ubiquitous and a bit abused. For example, sometimes there's a comment next to some argument clarifying the the Hash type means slice merkle root. This makes it hard to introduce the double merkle block id and check for correctness.

#### Summary of Changes
Distinguish slice merkle roots with a new type.
Simplified the way the merkle tree implementation is indexed.

### DIFF
--- a/gossip/src/duplicate_shred.rs
+++ b/gossip/src/duplicate_shred.rs
@@ -344,6 +344,7 @@ pub(crate) mod tests {
         solana_hash::Hash,
         solana_keypair::Keypair,
         solana_ledger::shred::{ProcessShredsStats, ReedSolomonCache, Shredder},
+        solana_runtime::bank::SliceRoot,
         solana_signature::Signature,
         solana_signer::Signer,
         solana_system_transaction::transfer,
@@ -457,7 +458,7 @@ pub(crate) mod tests {
             &entries,
             is_last_in_slot,
             // chained_merkle_root
-            Some(Hash::new_from_array(rng.gen())),
+            Some(SliceRoot(Hash::new_from_array(rng.gen()))),
             next_shred_index,
             next_code_index, // next_code_index
             &ReedSolomonCache::default(),
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -48,7 +48,7 @@ use {
     solana_measure::measure::Measure,
     solana_metrics::datapoint_error,
     solana_pubkey::Pubkey,
-    solana_runtime::bank::Bank,
+    solana_runtime::bank::{Bank, SliceRoot},
     solana_signature::Signature,
     solana_signer::Signer,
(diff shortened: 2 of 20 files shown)


slice-root/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use solana_hash::Hash;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct SliceRoot(pub Hash);
```
akhi3030:
```suggestion
pub struct SliceRoot(Hash);
```

It will be cleaner to not make the internal field `pub`.  And to have a constructor with appropriate visibility to ensure that "valid" SliceRoots can be constructed.

slice-root/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+use solana_hash::Hash;
```
akhi3030: To reviewers, this crate introduced to break the cyclic dependency.  Once, the work here is done, the type definition ca be moved to solana-sdk which is probably the right place for defining `SliceRoot`.

AshwinSekar: Generally SDK represents types that will be used on chain, perhaps it would make sense for this to live in `votor-messages` or `ledger/src/shred.rs` instead

slice-root/src/lib.rs
AshwinSekar: Is there an advantage to using a struct here rather than a type alias?

akhi3030: a type alias serves just as documentation and doesn't offer actual type safety.  It is still possible to accidentally pass in the wrong type and the compiler will not complain.

slice-root/src/lib.rs
kobi-lizard: Tried shred.rs, there's a cyclic dep with runtime.
Can try votor-messages, but pretty sure there's a cyclic dependency :(

slice-root/src/lib.rs
kobi-lizard: Right, type alias would be like the mentioned comment that says "this argument means this". I want to make sure and trace correct methods get/return SliceRoot or AlpenglowBlockId and Rust to guide my hand. Is it just me or this is how it should be? 😅

kobi-lizard: With this codebase I will get a stroke without Rust holding my hand lol

AshwinSekar: Yeah that's fine, i'm cool with the struct approach

slice-root/src/lib.rs
AshwinSekar: `votor-messages` is designed to be a lightweight leaf crate for types that everyone can depend on. AFAIK it shouldn't depend on any non SDK crates, maybe give it a try?

PR #520, opened 2025-10-08
block component: parse `Vec<BlockComponent>`

ksn6:
Support the ability to parse `Vec<BlockComponent>`.

We will use this e.g., when disseminating the last two FEC sets:

(1) the second-to-last FEC set contains tx data
(2) the last FEC set contains a `BlockFooter`

Also we'll be using this in `blockstore_processor.rs`.

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -3207,4 +3207,96 @@ mod tests {
         );
         assert!(marker_component.as_entry_batch().is_none());
     }
+
+    #[test]
+    fn test_to_bytes_multiple_empty() {
+        // Empty slice
+        let empty: Vec<BlockComponent> = vec![];
+        let bytes = BlockComponent::to_bytes_multiple(&empty).unwrap();
+        assert!(bytes.is_empty());
+        let deserialized = BlockComponent::from_bytes_multiple(&bytes).unwrap();
+        assert_eq!(deserialized.len(), 0);
+    }
+
+    #[test]
+    fn test_to_bytes_multiple_entry_batches() {
+        // Multiple entry batches
+        let components = vec![
+            BlockComponent::new_entry_batch(create_mock_entry_batch(3)).unwrap(),
+            BlockComponent::new_entry_batch(create_mock_entry_batch(5)).unwrap(),
+            BlockComponent::new_entry_batch(create_mock_entry_batch(2)).unwrap(),
+        ];
+        let bytes = BlockComponent::to_bytes_multiple(&components).unwrap();
+        let deserialized = BlockComponent::from_bytes_multiple(&bytes).unwrap();
+        assert_eq!(deserialized.len(), 3);
+        assert_eq!(deserialized[0].entry_batch().len(), 3);
+        assert_eq!(deserialized[1].entry_batch().len(), 5);
+        assert_eq!(deserialized[2].entry_batch().len(), 2);
+
+        // Round-trip equality
+        assert_eq!(components, deserialized);
+    }
+
+    #[test]
+    fn test_to_bytes_multiple_markers() {
+        // Multiple markers
+        let components = vec![
+            BlockComponent::new_block_marker(VersionedBlockMarker::new(
+                BlockMarkerV1::BlockFooter(VersionedBlockFooter::new(BlockFooterV1 {
+                    block_producer_time_nanos: 111,
+                    block_user_agent: b"node1".to_vec(),
+                })),
+            )),
+            BlockComponent::new_block_marker(VersionedBlockMarker::new(
+                BlockMarkerV1::UpdateParent(VersionedUpdateParent::new(
+                    create_parent_ready_update_with_data(42, Hash::new_unique()),
+                )),
+            )),
+            BlockComponent::new_block_marker(VersionedBlockMarker::new(
+                BlockMarkerV1::BlockHeader(VersionedBlockHeader::new(BlockHeaderV1 {
+                    parent_slot: 100,
+                    parent_block_id: Hash::new_unique(),
+                })),
+            )),
+        ];
+        let bytes = BlockComponent::to_bytes_multiple(&components).unwrap();
+        let deserialized = BlockComponent::from_bytes_multiple(&bytes).unwrap();
+        assert_eq!(deserialized.len(), 3);
+        assert!(deserialized.iter().all(|c| c.is_marker()));
+
+        // Round-trip equality
+        assert_eq!(components, deserialized);
+    }
+
+    #[test]
+    fn test_to_bytes_multiple_mixed() {
+        // Mixed: entries and markers
+        let components = vec![
+            BlockComponent::new_entry_batch(create_mock_entry_batch(4)).unwrap(),
+            BlockComponent::new_block_marker(VersionedBlockMarker::new(
+                BlockMarkerV1::UpdateParent(VersionedUpdateParent::new(
+                    create_parent_ready_update(),
+                )),
+            )),
+            BlockComponent::new_entry_batch(create_mock_entry_batch(1)).unwrap(),
+            BlockComponent::new_block_marker(VersionedBlockMarker::new(
+                BlockMarkerV1::BlockFooter(VersionedBlockFooter::new(BlockFooterV1 {
+                    block_producer_time_nanos: 999,
+                    block_user_agent: b"test".to_vec(),
+                })),
+            )),
+        ];
+        let bytes = BlockComponent::to_bytes_multiple(&components).unwrap();
+        let deserialized = BlockComponent::from_bytes_multiple(&bytes).unwrap();
+        assert_eq!(deserialized.len(), 4);
+        assert!(deserialized[0].is_entry_batch());
+        assert!(deserialized[1].is_marker());
+        assert!(deserialized[2].is_entry_batch());
+        assert!(deserialized[3].is_marker());
+        assert_eq!(deserialized[0].entry_batch().len(), 4);
+        assert_eq!(deserialized[2].entry_batch().len(), 1);
+
+        // Round-trip equality
+        assert_eq!(components, deserialized);
+    }
 }


entry/src/block_component.rs
```diff
@@ -459,14 +459,28 @@ impl BlockComponent {
...
+        let mut components = Vec::new();
+        let mut cursor = 0;
+
+        while cursor < data.len() {
+            let remaining = &data[cursor..];
+            let (component, bytes_consumed) = Self::parse_single_from_bytes(remaining)?;
+            components.push(component);
+            cursor += bytes_consumed;
+        }
+
```
akhi3030, line 487:
```suggestion
assert_eq!(cursor, data.len());
```

just a nit safety check.

entry/src/block_component.rs
akhi3030:
```suggestion
        let entries = bincode::deserialize::<Vec<_>>(data)
```

nit: Not part of the PR but we do not need to specify the type here and let the compiler infer it.

akhi3030: We are using an ancient version of bicode.  Newer versions have [`decode_from_slice`](https://docs.rs/bincode/latest/bincode/fn.decode_from_slice.html) which also returns the number of bytes read.  We should consider bumping the version.  In a separate PR of course.

entry/src/block_component.rs
```diff
@@ -614,6 +648,50 @@ impl VersionedBlockMarker {
             _ => Err(BlockComponentError::UnsupportedVersion { version }),
         }
     }
+
+    /// Determines the size of a VersionedBlockMarker in bytes without fully parsing it.
+    fn get_versioned_marker_size(data: &[u8]) -> Result<usize, BlockComponentError> {
+        const VERSION_SIZE: usize = 2;
```
akhi3030, line 663: should we add some documentation about why this is 2?

akhi3030: or maybe define is as `std::mem::size_of::<u16>()`.  

Even better if we had a `struct Version(u16)`.

entry/src/block_component.rs
```diff
@@ -484,21 +498,33 @@ impl BlockComponent {
...
 
-        // Handle remaining data
         let remaining_bytes = data
             .get(cursor..)
             .ok_or(BlockComponentError::CursorOutOfBounds)?;
 
         match (entries.is_empty(), remaining_bytes.is_empty()) {
-            (true, true) => Ok(Self::EntryBatch(Vec::new())),
+            (true, true) => {
+                // Special case: allow exactly 8 zero bytes as empty entry batch
```
AshwinSekar: Do you see this in prod? This shouldn't be happening or allowed.
If it's a test that causes this issue we should update the test rather than special casing it here.

ksn6: Gotcha - yep, there were a couple of local cluster tests that weren't passing as a result of me disallowing this case, which is why I did this.

Will look into this and revert.

entry/src/block_component.rs
```diff
@@ -459,14 +459,28 @@ impl BlockComponent {
...
+    ///
+    /// This function can handle multiple BlockComponents serialized sequentially in the data.
     ///
     /// # Errors
     /// Returns an error if deserialization fails or data is invalid.
-    ///
-    /// TODO(karthik): fuzz test this function.
-    /// TODO(karthik): should we error out if there are remaining bytes after deserialization?
-    pub fn from_bytes(data: &[u8]) -> Result<Self, BlockComponentError> {
+    pub fn from_bytes(data: &[u8]) -> Result<Vec<Self>, BlockComponentError> {
```
carllin: nit: naming is a bit asymmetric, `from/to_bytes()` as a default return the single object I think, and we can have `to/from_bytes_multiple` for the vector

entry/src/block_component.rs
```diff
@@ -554,7 +580,15 @@ impl<'de> Deserialize<'de> for BlockComponent {
...
             {
-                BlockComponent::from_bytes(value).map_err(de::Error::custom)
+                let components = BlockComponent::from_bytes(value).map_err(de::Error::custom)?;
+                if components.len() == 1 {
+                    Ok(components.into_iter().next().unwrap())
+                } else {
+                    Err(de::Error::custom(format!(
+                        "expected 1 component, got {}",
+                        components.len()
+                    )))
```
carllin: Not quite sure I understand the design:

1. we always expect only 1 component on deserialization?
2. The serialize/to_bytes() function https://github.com/anza-xyz/alpenglow/blob/79e069d76fd05ffe3c46eb9b9e67314d5eb15b33/entry/src/block_component.rs#L555-L563 always serializes just one component, but the deserialize/from_bytes() function expects a vector?

ledger/src/blockstore.rs
```diff
@@ -4119,7 +4119,12 @@ impl Blockstore {
                         // TODO(karthik): if Alpenglow flag is disabled, return an error on special
                         // EntryBatches.
                         BlockComponent::from_bytes(&payload)
-                            .map(|eb| eb.entry_batch().to_vec())
+                            .map(|cs| {
+                                cs.into_iter()
+                                    .filter_map(|bc| bc.into_entry_batch())
```
carllin, line 4124: this filter on `into_entry_batch()` will just silently ignore if there's something like a header/footer in the middle of the block right? Do we want to tolerate this?

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs

ksn6: Yep, can explore this in a separate PR, although...

... upgrading bincode is going to be a bit of a mission. A ton of our code, agave, SPL libraries, etc., all depend on this very specific version of bincode. We'll have to be very careful that nothing breaks backward compatibility.

entry/src/block_component.rs
ksn6: Let's roll with `std::mem::size_of::<u16>()` for now. I do like the struct idea, though it's probably worth having separate version structs for each versioned struct here. Will address this in a separate PR.

entry/src/block_component.rs
ksn6: Done

entry/src/block_component.rs
ksn6: I implemented the serde traits for `BlockComponent` in case there was a use-case for someone to have a `BlockComponent` within another struct, and that person wanted to `#[derive(Serialize, Deserialize)]` on the parent struct.

Agreed - the earlier implementation I had was a bit confusing. serde should only ever deal with single instances of a struct. I now additionally have `from/to_bytes()` which are essentially identical to what serde do (and are more convenient to use in code) + `from/to_bytes_multiple()` which are for multiple components.

ledger/src/blockstore.rs
ksn6: Correct - this is intentional behavior in this PR, and will change in another PR (WIP). We aren't sending out headers / footers / update parents just yet.

entry/src/block_component.rs
ksn6: I updated this PR to force an error on deserializing `[0_u8; 8]`. The error isn't in a local cluster test, but rather:

https://buildkite.com/anza/alpenglow/builds/2169#0199cc97-b093-4393-82d2-25b2eb796e57/78-3416

Here's where the test fails:

https://github.com/anza-xyz/alpenglow/blob/fe5447408bc33d5d8bdc086458cb22f8f0f5cf31/turbine/src/broadcast_stage/standard_broadcast_run.rs#L739

The containing test attempts to broadcast a bank with slot 2, which interrupts slot 0. As a result, we trigger the following logic:

https://github.com/anza-xyz/alpenglow/blob/fe5447408bc33d5d8bdc086458cb22f8f0f5cf31/turbine/src/broadcast_stage/standard_broadcast_run.rs#L214

This is then the issue:

https://github.com/anza-xyz/alpenglow/blob/fe5447408bc33d5d8bdc086458cb22f8f0f5cf31/turbine/src/broadcast_stage/standard_broadcast_run.rs#L90-L104

In particular, it's the fact that we're serializing an empty slice of entries. Concretely:

- `make_merkle_shreds_from_entries` ends up bincode serializing `&[]`...
- ... which results in `[0_u8; 8]`...
- ... which then later ends up getting deserialized by `BlockComponent::from_bytes(...)`...
- ... resulting in an error...
- ... which is then unwrapped to the buildkite error seen above

Potential solution:
- Because every block is supposed to have a header + footer...
- ... what if the shredder takes in a `[BlockComponent; 2]` with a header and a footer...
- ...and we serialize these bytes rather than an empty entries list?

AshwinSekar: Got it yeah we should update the test in this case.

For context the `finish_prev_slot` is a pathway in TowerBFT for abandoning the current leader block. It will result in the block being dead (we don't finish the poh ticks). In order to abandon the current leader block we submit an empty entry batch with `is_last_in_slot` set (so we stop trying to repair the block). 

Either way this block will be marked as dead in replay - in this case we're failing a little earlier at deserialization - but the end result should be the same.

ksn6: That's a good point - fixed the test!

ledger/src/blockstore.rs
carllin: @AshwinSekar this should be safe because you can't get the same bank hash from a different set of shreds right, blockid is incorporated into the hash

AshwinSekar: it's the other way around block_id will contain the bank hash after https://github.com/solana-foundation/solana-improvement-documents/pull/298

entry/src/block_component.rs
```diff
@@ -433,6 +427,21 @@ impl BlockComponent {
         }
     }
 
+    ///
+    /// This function serializes each component in the slice and concatenates the results.
+    /// The resulting bytes can be deserialized using `from_bytes_multiple()`.
+    ///
+    /// # Errors
+    /// Returns an error if any component fails to serialize.
+    pub fn to_bytes_multiple(components: &[Self]) -> Result<Vec<u8>, BlockComponentError> {
```
carllin, line 436: Maybe a simple test that tests `to_bytes_multiple()` -> `from_bytes_multiple()` returns the original components

ksn6: Oops - done.

carllin (approved):

PR #523, opened 2025-10-08
refactor: eliminate a unnecessary copying

akhi3030:
#### Problem

Some `Vote` related APIs were taking the `Vote` related structs as reference and then having to copy / clone it later.  

#### Summary of Changes

We are able to pass the structs by value and avoid the copying

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -157,7 +157,7 @@ impl ConsensusPool {
         slot: Slot,
         vote_type: VoteType,
         block_id: Option<Hash>,
-        transaction: &VoteMessage,
+        transaction: VoteMessage,
         validator_vote_key: &Pubkey,
         validator_stake: Stake,
     ) -> Option<Stake> {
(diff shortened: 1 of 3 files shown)


votor/src/consensus_pool.rs
```diff
@@ -157,7 +157,7 @@ impl ConsensusPool {
         slot: Slot,
         vote_type: VoteType,
         block_id: Option<Hash>,
-        transaction: &VoteMessage,
+        transaction: VoteMessage,
```
wen-coding: We should rename these `transaction` vote.....

votor/src/consensus_pool.rs
```diff
@@ -744,7 +744,7 @@ mod tests {
                     bank.epoch_stakes_map(),
                     bank.slot(),
                     &Pubkey::new_unique(),
-                    &dummy_transaction(validator_keypairs, &vote, rank),
+                    dummy_transaction(validator_keypairs, &vote, rank),
```
wen-coding: Same, dummy_transaction->dummy_vote

wen-coding (approved):

Issue #522, opened 2025-10-08
Ideas on how to store votes for rewards


akhi3030:
Some ideas I had about packing aggregates into the block footer:
- These s - 8 votes are not required for consensus, and we don't need to update or track any thresholds on these old slots
- Ideally we could avoid requiring the consensus pool be shared between multiple threads
- Instead what if we had a separate rewards container:
  - If sigverifier receives a vote > root send it to the ConsensusPool, if it's a notar or skip also send it to the rewards container
  - If sigverifier receives a notar/skip vote where we are leader for vote_slot + 8 > root send it to rewards container
- This container doesn't have to do anything fancy, just store one vote per validator per slot
- When the block creation loop is about to finish producing the block, it can call the BLS aggregation fn for s - 8 and record the result in the block footer
- At this point we can clean up s - 8 from the rewards container

Separates the rewards tracking from the main consensus tracking so we don't have to mess with the ConsensusPool. Maybe wrap the `VoteMessage` in an `Arc` to avoid the overhead of sending it twice. Let me know what you think.

_Originally posted by @AshwinSekar in https://github.com/anza-xyz/alpenglow/issues/514#issuecomment-3372078633_

PR #525, opened 2025-10-09
RFC: setup for a container to track votes for paying rewards

akhi3030:
Introduces a basic container that can be used to track votes for paying out rewards.  

Part of https://github.com/anza-xyz/alpenglow/issues/522.

### DIFF
--- a/votor/src/consensus_rewards.rs
+++ b/votor/src/consensus_rewards.rs
@@ -1,0 +1,98 @@
+use {
+    crate::consensus_pool::vote_certificate_builder::VoteCertificateBuilder,
+    solana_clock::Slot,
+    solana_gossip::cluster_info::ClusterInfo,
+    solana_ledger::leader_schedule_cache::LeaderScheduleCache,
+    solana_runtime::bank_forks::SharableBanks,
+    solana_votor_messages::{
+        consensus_message::{Certificate, CertificateMessage, VoteMessage},
+        vote::Vote,
+    },
+    std::{collections::BTreeMap, sync::Arc},
+};
+
+/// A container for storing [`VoteMessage`]s to help pay rewards.
+pub struct ConsensusRewards {
+    /// Per [`Slot`], stores a list of received skip and notarization [`VoteMessage`]s.
+    //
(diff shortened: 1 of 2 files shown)


votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    /// Retrives notar and skip certificates for paying out rewards.
+    ///
+    /// Data for all slots <= to the slot being queried is discarded.
+    //
+    // TODO: fix the various panics below.
+    // TODO: this will probably be called somewhere in the block production loop.
+    pub fn get_aggregate(&mut self, slot: Slot) -> Vec<CertificateMessage> {
```
qkniep, line 76: nit: this should probably be called `get_aggregates()` (plural)

qkniep: also a bit unfortunate that these things are called certificates at the type level, ideally these aggregates would have a different type and a "certificate" is always something with the correct threshold of votes

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.votes = self.votes.split_off(&slot);
+        let (notar_votes, skip_votes) = self.votes.remove(&slot).unwrap_or_default();
+
+        let mut ret = vec![];
+        if let Some(vote_message) = notar_votes.last() {
+            let mut cert_builder = VoteCertificateBuilder::new(Certificate::Notarize(
```
qkniep, line 82: we have to keep in mind that aggregating here is very expensive

qkniep: also unfortunate that we will redo all the aggregation work that was already done in `ConsensusPool` before

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,98 @@
...
+                slot,
+                *vote_message.vote.block_id().unwrap(),
```
qkniep, line 84: have to keep in mind that we might see notar votes for many different blocks

PR #530, opened 2025-10-11
Switch ConsensusMetrics to use sender/receiver so we don't introduce write locks on critical path.

wen-coding:
#### Problem
The current write locks are actually fine because it's quick, but things might change in the future when metrics implementation changes, also we can't nicely parallelize vote handling if we have this write lock.

#### Summary of Changes
- Vote is small, copying votes for metrics should be cheap
- Replace write locks with sender/receiver, so metrics can happily process all events in its own thread

### DIFF
--- a/votor/src/consensus_metrics.rs
+++ b/votor/src/consensus_metrics.rs
@@ -84,6 +84,7 @@ impl NodeVoteMetrics {
             Vote::Skip(_) => self.skip.increment(elapsed),
             Vote::SkipFallback(_) => self.skip_fallback.increment(elapsed),
             Vote::Finalize(_) => self.final_.increment(elapsed),
+            Vote::Genesis(_) => Ok(()), // Only for migration, tracked elsewhere
         };
         match res {
             Ok(()) => (),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11889,6 +11889,7 @@ version = "3.1.0"
 dependencies = [
  "bitvec",
  "bytemuck",
+ "log",
  "num_enum",
  "serde",
  "solana-account",
--- a/votor-messages/Cargo.toml
+++ b/votor-messages/Cargo.toml
@@ -11,11 +11,16 @@ license = { workspace = true }
 edition = { workspace = true }
 
 [features]
-frozen-abi = ["dep:solana-frozen-abi", "dep:solana-frozen-abi-macro"]
+frozen-abi = [
+    "dep:solana-frozen-abi",
+    "dep:solana-frozen-abi-macro",
+    "solana-hash/frozen-abi",
+]
 
 [dependencies]
 bitvec = { workspace = true }
 bytemuck = { workspace = true }
+log = { workspace = true }
 num_enum = { workspace = true }
 serde = { workspace = true }
 solana-account = { workspace = true }
--- a/votor-messages/src/consensus_message.rs
+++ b/votor-messages/src/consensus_message.rs
@@ -37,6 +37,8 @@ pub enum Certificate {
     NotarizeFallback(Slot, Hash),
     /// Skip certificate
     Skip(Slot),
+    /// Genesis certificate
+    Genesis(Slot, Hash),
 }
 
 /// Certificate type
@@ -52,6 +54,8 @@ pub enum CertificateType {
     NotarizeFallback,
     /// Skip certificate
     Skip,
+    /// Genesis certificate
+    Genesis,
 }
 
 impl Certificate {
@@ -77,6 +81,7 @@ impl Certificate {
             Certificate::Notarize(_, _) => CertificateType::Notarize,
             Certificate::NotarizeFallback(_, _) => CertificateType::NotarizeFallback,
             Certificate::Skip(_) => CertificateType::Skip,
+            Certificate::Genesis(_, _) => CertificateType::Genesis,
         }
     }
 
@@ -87,6 +92,7 @@ impl Certificate {
             | Certificate::FinalizeFast(slot, _)
             | Certificate::Notarize(slot, _)
             | Certificate::NotarizeFallback(slot, _)
+            | Certificate::Genesis(slot, _)
             | Certificate::Skip(slot) => *slot,
         }
     }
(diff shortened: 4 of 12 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -92,7 +92,7 @@ pub struct BLSSigVerifier {
     stats: BLSSigVerifierStats,
     verified_certs: RwLock<HashSet<Certificate>>,
     vote_payload_cache: RwLock<HashMap<Vote, Arc<Vec<u8>>>>,
-    consensus_metrics: Option<Arc<PlRwLock<ConsensusMetrics>>>,
+    consensus_metrics_sender: Option<ConsensusMetricsEventSender>,
```
akhi3030: should we get rid of the `Option` then?  I think sending a message should not perturb the performance very much.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -216,6 +218,14 @@ impl BLSSigVerifier {
         votes_result?;
         certs_result?;
 
+        if let Some(consensus_metrics_sender) = self.consensus_metrics_sender.as_ref() {
+            if let Err(e) =
+                consensus_metrics_sender.send((Instant::now(), consensus_metrics_to_send))
+            {
+                warn!("Failed to send consensus metrics: {e}");
```
akhi3030: the error can only happen if the channel is disconnected and the error contains the the message to be sent.  So if this error happens, we will pollute the logs with a vec of messages.  I suggest the following.


```suggestion
                warn!("could not send consensus metrics, receive side of channel is closed");
```

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -163,9 +164,10 @@ impl BLSSigVerifier {
                     };
 
                     // Capture votes received metrics before old messages are potentially discarded below.
-                    if let Some(c) = self.consensus_metrics.as_ref() {
-                        c.write().record_vote(*solana_pubkey, &vote_message.vote);
-                    }
+                    consensus_metrics_to_send.push(ConsensusMetricsEvent::Vote {
```
akhi3030, line 167: Why not include the timestamp in the event itself instead of including it in the message below?  That way we can more flexible and record votes over arbitrary periods of time.

core/src/tpu.rs
```diff
@@ -345,7 +344,7 @@ impl Tpu {
                 sharable_banks,
                 verified_vote_sender.clone(),
                 verified_consensus_message_sender,
-                Some(consensus_metrics),
+                Some(consensus_metrics_sender.clone()),
```
akhi3030:
```suggestion
                Some(consensus_metrics_sender),
```

core/src/tvu.rs
```diff
@@ -385,7 +385,8 @@ impl Tvu {
             snapshot_controller,
             replay_highest_frozen,
             leader_window_notifier,
-            consensus_metrics,
+            consensus_metrics_sender: consensus_metrics_sender.clone(),
```
akhi3030:
```suggestion
            consensus_metrics_sender: consensus_metrics_sender,
```

core/src/validator.rs
```diff
@@ -1697,7 +1694,8 @@ impl Validator {
             config.voting_service_test_override.clone(),
             votor_event_sender.clone(),
             votor_event_receiver,
-            consensus_metrics.clone(),
+            consensus_metrics_sender.clone(),
```
akhi3030, line 1697: To absolutely isolate the performance for the `BLSSigVerifier`, we should give it an exclusive channel for sending the events.  Sharing a channel with other components will incur a cost.  `ConsensusMetrics` module can have a list of channels that it polls and we don't care if it is a bit slower.

votor/src/event_handler.rs
```diff
@@ -334,9 +347,12 @@ impl EventHandler {
             VotorEvent::Timeout(slot) => {
                 info!("{my_pubkey}: Timeout {slot}");
                 if slot != last_of_consecutive_leader_slots(slot) {
-                    vctx.consensus_metrics
-                        .write()
-                        .record_start_of_slot(slot.saturating_add(1));
+                    vctx.consensus_metrics_sender
+                        .send((
+                            Instant::now(),
+                            vec![ConsensusMetricsEvent::StartOfSlot { slot }],
```
akhi3030:
```suggestion
                            vec![ConsensusMetricsEvent::StartOfSlot { slot.saturating_add(1) }],
```

Note the plus 1.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: That's what I originally did, including one timestamp for the total batch seems more convenient since you don't have to put a timestamp on each enum variant. Also the timestamp here is an approximate time to begin with, since packet batching is done before BLS sigverifier gets the data.

That said, I don't have strong feelings about this, so I can change it to timestamp on each variant as well.

core/src/validator.rs
wen-coding: I would worry about the performance if we block the handling before the valid messages are sent out to consensus pool. Since we now send to consensus pool first then send to metrics, this should be fine, because we can spend some CPU to send the metrics while the QUIC streamer is preparing our next batch.

In other words, the send is now out of the critical path.

votor/src/event_handler.rs
wen-coding: Oops, thanks for catching it.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Done.

votor/src/consensus_metrics.rs
```diff
@@ -107,37 +128,86 @@ pub struct ConsensusMetrics {
...
+                                self.maybe_new_epoch(epoch);
+                            }
+                            ConsensusMetricsEvent::StartOfSlot { slot } => {
+                                self.record_start_of_slot(slot, received);
+                            }
+                        }
+                    }
+                }
+                Err(err) => match err {
+                    RecvTimeoutError::Timeout => (),
```
akhi3030: do we want a low level, e.g. `trace` logging here to help detect if for some reason metrics collection is not working?

wen-coding: Done.

PR #531, opened 2025-10-12
feat: get `PohRecorder` to send `EntryMarker`s to broadcast stage

ksn6:
### Summary

We need to push `BlockComponent::BlockMarker`s into blocks from various parts of the code. Key examples:

- `BlockFooter` and `BlockHeader` from `PohRecorder`
- `UpdateParent` from `votor`

### Approach

We modify `WorkingBankEntry` to work with `EntryMarker` types rather than `Entry`, such that `PohRecorder` can properly disseminate markers.

NOTE: rather than land this PR in, we're going to break this PR into multiple smaller PRs.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -4120,12 +4120,6 @@ impl Blockstore {
                         // EntryBatches.
                         BlockComponent::from_bytes_multiple(&payload)
                             .map(|cs| {
-                                for c in &cs {
-                                    if c.is_marker() {
-                                        println!("MARKER FOUND :: {c:?}");
-                                    }
-                                }
-
                                 cs.into_iter()
                                     .filter_map(|bc| bc.as_entry_batch_owned())
                                     .flatten()


ledger/src/shredder.rs
```diff
@@ -63,6 +63,39 @@ impl Shredder {
...
+        chained_merkle_root: Option<Hash>,
+        next_shred_index: u32,
+        next_code_index: u32,
+        reed_solomon_cache: &ReedSolomonCache,
+        stats: &mut ProcessShredsStats,
+    ) -> impl Iterator<Item = Shred> {
+        let now = Instant::now();
+        let data: Vec<u8> = components
+            .iter()
+            .flat_map(|component| component.to_bytes().unwrap())
```
AshwinSekar: hmm don't block markers need to be in their own FEC set? I think it would be better to have separate API to serialize entries and markers and then call the correct one in the caller, less chance of misusing the shredder.

poh/src/poh_recorder.rs
```diff
@@ -60,7 +61,86 @@ pub enum PohRecorderError {
 
 pub(crate) type Result<T> = std::result::Result<T, PohRecorderError>;
 
-pub type WorkingBankEntry = (Arc<Bank>, (Entry, u64));
+#[derive(Clone, Debug)]
+pub enum EntryMarker {
```
AshwinSekar: nit: seems better under `entry/src`?

poh/src/poh_recorder.rs
```diff
@@ -590,11 +671,14 @@ impl PohRecorder {
                 entry_count,
             );
 
-            for tick in &self.tick_cache[..entry_count] {
-                working_bank.bank.register_tick(&tick.0.hash);
+            for (entry, tick_height) in &self.tick_cache[..entry_count] {
+                working_bank.bank.register_tick(&entry.hash);
+
+                let tick = (entry.clone().into(), *tick_height);
```
AshwinSekar: nit: prefer `From` for ease of reading

ledger/src/shredder.rs
AshwinSekar: coalesce can return the `(Vec<Entry>, Vec<BlockMarker>)` and then broadcast can serialize them separately.

Also gives us optionality in the future if we want to enforce ordering things without having to wrangle banking stage threads, aka that the block footer must be the last FEC set.

poh/src/poh_recorder.rs
```diff
@@ -60,7 +64,86 @@ pub enum PohRecorderError {
...
+impl From<EntryMarker> for BlockComponent {
+    fn from(entry_marker: EntryMarker) -> Self {
+        match entry_marker {
+            EntryMarker::Entry(entry) => BlockComponent::EntryBatch(vec![entry]),
+            EntryMarker::Marker(marker) => BlockComponent::BlockMarker(marker),
+        }
+    }
+}
+
+pub type WorkingBankEntry = (Arc<Bank>, (EntryMarker, u64));
```
apfitzge, line 67: thinking out loud:

should the `u64` be integrated into the enum? is tick_height relavent for the markers or only the entry? does it even matter after AG is enabled?

poh/src/poh_recorder.rs
ksn6: sg done

poh/src/poh_recorder.rs
ksn6: post-alpenglow, we'd like to get rid of the alpentick / ticks entirely

I'm +epsilon in favor of keeping the u64 outside the enum to minimize code changes, but could pretty easily be convinced otherwise

poh/src/poh_recorder.rs
ksn6: done

ledger/src/shredder.rs
ksn6: ah, good catch - though, why not just invoke `Self::make_shreds_from_data_slice` once per component + chain the results? this seems simpler, and can't really be misused.

so - I don't know if we want `(Vec<Entry>, Vec<BlockMarker>)` since we might see:

```
BlockHeader -> EntryBatch -> UpdateParent -> EntryBatch -> BlockFooter
```

in particular, we'll want to see a clean break between the two separate entry batches since the last entry from pre-update parent and first entry from post-update parent shouldn't be in the same FEC set b/c of the UpdateParent taking its own FEC set

what do you think of invoking `Self::make_hreds_from_data_slice` once per component for now and then once Fast Leader Handover is in, we create something like the following:

```rust
struct BlockComponentCoalescer {
    header: VersionedBlockHeader,
    entries: Vec<Entry>,
    update_parent: Option<VersionedUpdateParent>,
    entries_take_two: Vec<Entry>,
    footer: VersionedBlockFooter,
}

impl BlockComponentCoalescer {
    pub fn create_from_components(component_stream: impl Iterator<BlockComponent>) -> Self {
        todo!()
    }

    pub fn to_shreds(self) -> impl Iterator<Shred> {
        todo!()
    }
}
```

where the reordering you're referring to would all be done in the `to_shreds` function here

poh/src/poh_recorder.rs
apfitzge: nah don't have a strong opinion

ksn6: sg - will close this for now then

ledger/src/shredder.rs
AshwinSekar: yeah i think this makes sense, you could make it even simpler and stop coalescing as soon as you see a block marker, e.g. in your example as soon as you see UpdateParent just send over the entry batch you currently have.

But overall i think this direction makes sense

ksn6:
looks like we all agree on design - closing this PR out in favor of breaking this up into smaller PRs

PR #532, opened 2025-10-13
migration: Add genesis vote tracking

AshwinSekar:
Split from #502 

#### Problem
As part of the migration we need a new type BLS vote/cert for determining the genesis slot for Alpenglow.

#### Summary of Changes
Add the new vote type and enable tracking in certificate pool.
Follow up PRs will plug this in and wire the BLSSigverifier to accept this vote type *only* during the migration period.

### DIFF
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -19,9 +19,9 @@
 //! 5) We initialize Votor with `G` as genesis, and disable TowerBFT for any slots past `G`
 //! 6) We exit vote only mode, and reenable rooting and commitment reporting
 //!
-//! If at any point during the migration we see:
+//! If at any point during the migration we see a:
 //! - A genesis certificate
-//! - A finalization certificate
+//! - or a finalization certificate (fast finalization or a slow finalization with notarization)
 //!
 //! It means the cluster has already switched to Alpenglow and our node is behind. We perform any appropriate
 //! repairs and immediately transition to Alpenglow at the certified block.


votor-messages/src/migration.rs
```diff
@@ -0,0 +1,42 @@
...
+//! 1) We track blocks which have `GENESIS_VOTE_THRESHOLD`% of stake's vote txs for the parent block.
+//!    The parent block is referred to as reaching super OC.
+//! 2) Notice that all super OC blocks that must be a part of the same fork in presence of
+//!    less than `MIGRATION_MALICIOUS_THRESHOLD` double voters
+//! 3) We find the latest ancestor of the super OC block < `s`, `G` and cast a BLS vote (the genesis vote) via all to all
+//! 4) If we observe `GENESIS_VOTE_THRESHOLD`% votes for the ancestor block `G`:
+//!    5a) We clear any TowerBFT blocks past `G`.
+//!    5b) We propagate the Genesis certificate for `G` via all to all
+//! 5) We initialize Votor with `G` as genesis, and disable TowerBFT for any slots past `G`
+//! 6) We exit vote only mode, and reenable rooting and commitment reporting
```
wen-coding, line 20: nit: should this explanation in `votor-messages` or in `votor` near your code?

votor-messages/src/vote.rs
```diff
@@ -261,3 +281,32 @@ impl SkipFallbackVote {
...
+    derive(AbiExample),
+    frozen_abi(digest = "2JAiHmnnKHCzhkyCY3Bej6rAaVkMHsXgRcz1TPCNqAJ9")
+)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Default, Serialize, Deserialize)]
+pub struct GenesisVote {
+    slot: Slot,
+    block_id: Hash,
+}
+
+impl GenesisVote {
```
wen-coding, line 297: Sorry I forgot, why do we need GenesisVote instead of just doing:
`Genesis(slot, block_id)`

Otherwise looks good to me.

votor-messages/src/migration.rs
```diff
@@ -0,0 +1,42 @@
...
+//!
+//! If at any point during the migration we see:
+//! - A genesis certificate
+//! - A finalization certificate
```
wen-coding: nit: specify whether it's `and` or `or` relationship between two certificates, and whether FastFinalize vs Finalize makes a difference here.

votor-messages/src/migration.rs
AshwinSekar: I think it's better here since this file will later contain the struct to be used by replay and the rest of the code. There shouldn't be any changes in votor specifically apart from tracking the GenesisVote

votor-messages/src/vote.rs
AshwinSekar: It serializes into the same thing, this just makes it more convenient to call `.slot()/.block_id()` instead

PR #533, opened 2025-10-13
refactor: remove some dead code

akhi3030:
#### Problem

The conversion is not used anywhere.

#### Summary of Changes

Removes the code related to conversion.


votor/src/consensus_pool/vote_certificate_builder.rs
```diff
@@ -46,31 +46,6 @@ pub struct VoteCertificateBuilder {
     input_bitmap_2: BitVec<u8, Lsb0>,
 }
 
-impl TryFrom<CertificateMessage> for VoteCertificateBuilder {
```
wen-coding, line 49: It was left here for the S+8 case, so you can inherit CertificateMessage from consensus pool and start building from there.

But since we don't go that route, we can remove it.

PR #534, opened 2025-10-13
refactor: various improvements to `VoteCertificateBuilder`

akhi3030:
#### Problem

- The logic in `VoteCertificateBuilder::build` is a bit obtuse.  In the scenario with a single bitmap, the actual operation is pretty straightforward.
- `VoteCertificateBuilder::aggregate` is silently ignoring potential bugs if there are somehow more than 2 types of bitmaps needed

#### Summary of Changes

refactors `VoteCertificateBuilder` to use betters types and catches potentially invalid states by panicking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8044,7 +8044,6 @@ dependencies = [
  "min-max-heap",
  "num_cpus",
  "num_enum",
- "parking_lot 0.12.4",
  "prio-graph",
  "qualifier_attr",
  "quinn",
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -70,7 +70,6 @@ lru = { workspace = true }
 min-max-heap = { workspace = true }
 num_cpus = { workspace = true }
 num_enum = { workspace = true }
-parking_lot = { workspace = true }
 prio-graph = { workspace = true }
 qualifier_attr = { workspace = true }
 quinn = { workspace = true }
--- a/core/benches/bls_sigverify.rs
+++ b/core/benches/bls_sigverify.rs
@@ -79,6 +79,7 @@ impl PregeneratedBatches {
 fn setup_environment() -> BenchEnvironment {
     let (verified_votes_s, _) = unbounded();
     let (consensus_msg_s, _) = unbounded();
+    let (consensus_metrics_sender, _) = unbounded();
 
     let validator_keypairs: Arc<Vec<_>> = Arc::new(
         (0..NUM_VALIDATORS)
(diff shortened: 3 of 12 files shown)


votor/src/consensus_pool/vote_certificate_builder.rs
```diff
@@ -85,33 +80,21 @@ impl VoteCertificateBuilder {
...
-        input_bitmap_2.resize(new_length, false);
-        let bitmap = if input_bitmap_2.count_ones() > 0 {
-            // If we have two bitmaps, use Base3 encoding
-            encode_base3(&input_bitmap_1, &input_bitmap_2).map_err(CertificateError::EncodeError)?
-        } else {
-            // If we only have one bitmap, use Base2 encoding
-            encode_base2(&input_bitmap_1).map_err(CertificateError::EncodeError)?
+    pub fn build(mut self) -> Result<CertificateMessage, CertificateError> {
+        let bitmap = match self.bitmap_1 {
+            None => {
```
samkim-crypto, line 84: I think this does introduce non-trivial breaking change.

In the old logic, it checked if there are actually two types of votes in the current bitmaps to decide whether to use base2 or base3 encodings.
In the new logic, it checks whether to use base2 or base3 encodings based on the certificate type.

For example, for skip certificates, even though there are only skip votes, it will always be encoded with base3 encodings.

I think the old logic probably makes more sense for efficiency since this would mean NotarizeFallback and Skip certificates will always be encoded in base3. We want base3 encoding to never be used in the average case. What do you think?

votor/src/consensus_pool/vote_certificate_builder.rs
```diff
@@ -34,47 +34,42 @@ pub enum CertificateError {
...
-            input_bitmap_1: BitVec::repeat(false, MAXIMUM_VALIDATORS),
-            input_bitmap_2: BitVec::repeat(false, MAXIMUM_VALIDATORS),
+            bitmap_0: BitVec::repeat(false, MAXIMUM_VALIDATORS),
+            bitmap_1,
         }
     }
 
     /// Aggregates a slice of `VoteMessage`s into the builder.
     pub fn aggregate(&mut self, messages: &[VoteMessage]) -> Result<(), CertificateError> {
-        if messages.is_empty() {
```
samkim-crypto, line 61: Just wanted to double check if there was a specific reason you wanted to remove this check. Do you think it makes sense to allow aggregation of empty messages?

akhi3030: the way I understood the code, if the length of messages is 0, then we will call `self.signature.aggregate_with` with an empty iterator which will be a NOP.  So removing this check doesn't hurt.

votor/src/consensus_pool/vote_certificate_builder.rs
akhi3030: Yes, that's a good point.  I'll update the PR.

akhi3030: the relevant changes are in the other functions actually.  We should now only use base3 if the second bitmap was actually used.  As a bonus, this also removed an `unwrap()`.

samkim-crypto: Nice, this looks clean and elegant.

votor/src/consensus_pool/vote_certificate_builder.rs
samkim-crypto: Oh sorry, I thought this check returned an error, but this check seems to have returned an `Ok(())` previously. Yeah in this case, this check is redundant. I wonder if returning an error on an empty message is a safer behavior here. But for the purpose of this PR, I think this is fine.

akhi3030: Yeah, I think we can make a case for both returning an error or return `Ok(())`.

PR #535, opened 2025-10-14
refactor: bump the `solana-bls-signatures` crate version (#8473)

akhi3030:
We recently did some work on the `solana-bls-signatures` crate to introduce iterators instead of slices.

This PR bumps the `solana-bls-signatures` crate version so we can use iterators.

Fixes #

#### Summary of Changes

Fixes #


akhi3030:
doing the work on a different PR.

PR #537, opened 2025-10-14
Add Migration tracker and update consumers

AshwinSekar:
Split from #502 

#### Problem
We currently use the alpenglow feature flag to determine whether to enable Alpenglow. This does not work with Carlgration.

#### Summary of Changes
Instead add a tracker `MigrationStatus` for use with all threads that need to be Alpenglow aware. This PR only updates some of the consumers, enabling alpenglow and updating the rest of the consumers will happen in a future PR.

`MigrationStatus` will track:
- Whether alpenglow is enabled (poh is shutdown and block creation loop is started)
- Our view of the genesis block - the ancestor of the super-oc block
- The cluster generated genesis certificate

In a future PR we will set our view of the genesis block based on super-oc tracking, and enable alpenglow once it matches the certificate.

Additionally we need to store the genesis block in `Bank` and in a future PR the snapshot. This is useful to distinguish the following scenario:
- We startup and the feature flag is active.
- We do not know if this means we have already completed the migration, or we are before/in the middle of migration.
- Instead we can check the root bank on startup to see if the genesis block is present.

Finally we update replay to read `MigrationStatus`:
- On startup check the root bank to distinguish whether we are pre or post migration.
- If the feature flag is activated during steady state, compute and store the migration slot in `MigrationStatus`
- Use `MigrationStatus` to determine if we should `maybe_start_leader`, send `VotorEvent::Block` or engage in any of the TowerBFT consensus protocols.

PohService similarly is updated to use `MigrationStatus` to determine when to shutdown and enable the block creation loop.

### DIFF
--- a/votor/src/consensus_pool_service.rs
+++ b/votor/src/consensus_pool_service.rs
@@ -12,7 +12,6 @@ use {
         },
         event::{LeaderWindowInfo, VotorEvent, VotorEventSender},
         voting_service::BLSOp,
-        votor::Votor,
     },
     crossbeam_channel::{select, Receiver, Sender, TrySendError},
     solana_clock::Slot,
@@ -23,12 +22,15 @@ use {
     },
     solana_pubkey::Pubkey,
     solana_runtime::{bank::Bank, bank_forks::SharableBanks},
-    solana_votor_messages::consensus_message::{Certificate, ConsensusMessage},
+    solana_votor_messages::{
+        consensus_message::{Certificate, ConsensusMessage},
+        migration::MigrationStatus,
+    },
     stats::ConsensusPoolServiceStats,
     std::{
         sync::{
             atomic::{AtomicBool, Ordering},
-            Arc, Condvar, Mutex,
+            Arc,
         },
         thread::{self, Builder, JoinHandle},
         time::{Duration, Instant},
@@ -38,7 +40,7 @@ use {
 /// Inputs for the certificate pool thread
 pub(crate) struct ConsensusPoolContext {
     pub(crate) exit: Arc<AtomicBool>,
-    pub(crate) start: Arc<(Mutex<bool>, Condvar)>,
+    pub(crate) migration_status: Arc<MigrationStatus>,
 
     pub(crate) cluster_info: Arc<ClusterInfo>,
     pub(crate) my_vote_pubkey: Pubkey,
@@ -182,24 +184,28 @@ impl ConsensusPoolService {
         let mut events = vec![];
         let mut my_pubkey = ctx.cluster_info.id();
         let root_bank = ctx.sharable_banks.root();
-        let mut consensus_pool = ConsensusPool::new_from_root_bank(my_pubkey, &root_bank);
 
-        // Wait until migration has completed
-        info!("{}: Certificate pool loop initialized", &my_pubkey);
-        Votor::wait_for_migration_or_exit(&ctx.exit, &ctx.start);
+        // Unlike the other votor threads, consensus pool starts even before alpenglow is enabled
+        // As it is required to track the Genesis Vote.
+        let mut consensus_pool = if ctx.migration_status.is_alpenglow_enabled() {
+            ConsensusPool::new_from_root_bank(my_pubkey, &root_bank)
+        } else {
+            ConsensusPool::new_from_root_bank_pre_migration(
+                my_pubkey,
+                &root_bank,
+                ctx.migration_status.clone(),
+            )
+        };
+
         info!("{}: Certificate pool loop starting", &my_pubkey);
         let mut stats = ConsensusPoolServiceStats::new();
+        let mut highest_parent_ready = root_bank.slot();
 
         // Standstill tracking
         let mut standstill_timer = Instant::now();
 
         // Kick off parent ready
-        let root_block = (root_bank.slot(), root_bank.block_id().unwrap_or_default());
-        let mut highest_parent_ready = root_bank.slot();
-        events.push(VotorEvent::ParentReady {
-            slot: root_bank.slot().checked_add(1).unwrap(),
-            parent_block: root_block,
-        });
+        let mut kick_off_parent_ready = false;
 
         // Ingest votes into certificate pool and notify voting loop of new events
         while !ctx.exit.load(Ordering::Relaxed) {
@@ -209,6 +215,26 @@ impl ConsensusPoolService {
                 my_pubkey = new_pubkey;
                 consensus_pool.update_pubkey(my_pubkey);
                 warn!("Certificate pool pubkey updated to {my_pubkey}");
+            }
+
+            // Kick off parent ready event, this either happens:
+            // - When we first migrate to alpenglow from TowerBFT - kick off with genesis block
+            // - If we startup post alpenglow migration - kick off with root block
+            if !kick_off_parent_ready && ctx.migration_status.is_alpenglow_enabled() {
+                let genesis_block = ctx
+                    .migration_status
+                    .genesis_block()
+                    .expect("Alpenglow enabled");
+                // can expect once we have block id in snapshots (SIMD-0333)
+                let root_block = (root_bank.slot(), root_bank.block_id().unwrap_or_default());
+                let kick_off_block = genesis_block.max(root_block);
+
+                highest_parent_ready = kick_off_block.0;
+                events.push(VotorEvent::ParentReady {
+                    slot: kick_off_block.0.checked_add(1).unwrap(),
+                    parent_block: kick_off_block,
+                });
+                kick_off_parent_ready = true;
             }
 
             Self::add_produce_block_event(
@@ -221,9 +247,14 @@ impl ConsensusPoolService {
             );
 
             if standstill_timer.elapsed() > DELTA_STANDSTILL {
-                events.push(VotorEvent::Standstill(
-                    consensus_pool.highest_finalized_slot(),
-                ));
+                // No reason to pollute channel with Standstill before the
+                // migration is complete. We still need standstill to refresh the
+                // Genesis cert though.
+                if kick_off_parent_ready {
+                    events.push(VotorEvent::Standstill(
+                        consensus_pool.highest_finalized_slot(),
+                    ));
+                }
                 stats.standstill = true;
                 standstill_timer = Instant::now();
                 match Self::send_certificates(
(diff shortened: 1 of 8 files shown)


AshwinSekar:
This PR does not change any behavior by itself - just defines the API. Will also likely be torn apart during upstream review so going to merge for now to unblock #555.

votor-messages/src/migration.rs
```diff
@@ -40,3 +54,244 @@ pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
 ///
 /// Using 38% as the `SWITCH_FORK_THRESHOLD` gives us 82% for `GENESIS_VOTE_THRESHOLD`.
 pub const GENESIS_VOTE_THRESHOLD: f64 = 82.0 / 100.0;
+
+/// The interval at which we refresh our genesis vote
+pub const GENESIS_VOTE_REFRESH: Duration = Duration::from_millis(400);
```
bw-solana, line 59: intentionally set to slot time? Should we derive it?

votor-messages/src/migration.rs
```diff
@@ -40,3 +54,244 @@ pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
...
+
+/// Tracks the details about the migrationary period and eligble genesis block
```
bw-solana, line 61: eligible*

votor-messages/src/migration.rs
```diff
@@ -40,3 +54,244 @@ pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
...
+
+    /// Flag indicating whether PoH has shutdown
+    pub shutdown_poh: AtomicBool,
+
+    /// Flag indicating whether the block creation loop has started
+    pub block_creation_loop_started: AtomicBool,
+
+    /// The slot in which we entered the migrationary period
+    /// u64::MAX if we do not yet know when the migration will start
+    pub(crate) start_migration_slot: AtomicSlot,
```
bw-solana, line 84: where do we need this outside of migration? Might be good to keep private and funnel everyone through the APIs to access this so that we can make the `u64::MAX` --> `None` translation

votor/src/consensus_pool_service.rs
```diff
@@ -182,24 +184,28 @@ impl ConsensusPoolService {
         let mut events = vec![];
         let mut my_pubkey = ctx.cluster_info.id();
         let root_bank = ctx.sharable_banks.root();
-        let mut consensus_pool = ConsensusPool::new_from_root_bank(my_pubkey, &root_bank);
 
-        // Wait until migration has completed
-        info!("{}: Certificate pool loop initialized", &my_pubkey);
-        Votor::wait_for_migration_or_exit(&ctx.exit, &ctx.start);
```
bw-solana, line 189: where is the wait happening now for `ConsensusPoolService` and `EventHandler` loops?

AshwinSekar: `ConsensusPoolService` no longer waits as it is required in order to aggregate the `GenesisVote`. 
The only change that I had to make was to not send `Standstill` events, the thread doesn't do anything else meaningful pre genesis.

`EventHandler` still waits here:
```rust
let Some(genesis_block) = migration_status.wait_for_migration_or_exit(&exit) else {
```
I basically moved the condvar to `MigrationStatus` instead of having it nested in `Votor`

votor-messages/src/migration.rs
AshwinSekar: yeah good catch i'll put a PR up addressing your comments later today

votor-messages/src/migration.rs
AshwinSekar: I just picked something random here, open to suggestion. 

Ideally we want to spam / refresh this vote ASAP to minimize user perceived downtime, but obviously I don't want to overwhelm the cluster. block time seemed reasonable because that's roughly the frequency at which we expect to send A2A during Alpenglow operation. 

On the other hand if we've viewed the strong-OC block and are ready to cast our genesis vote there's a high likelihood the same is true of the entire cluster. The genesis vote *should* succeed immediately, however if it doesn't something is wrong. Probably indicates some widespread turbine/gossip failure requiring repair or more blocks to be made. In that case there's not really a point to spamming the vote as we're waiting on other nodes to repair so we could be more conservative here.

core/src/validator.rs
```diff
@@ -1567,10 +1561,11 @@ impl Validator {
                     vote_history
                 }
                 Err(e) => {
+                    // TODO(ashwin): we need to be viligant about this and add a CLI option to panic here.
```
carllin, line 1564: don't think the current tower history panics on failure either lol

runtime/src/bank.rs
```diff
@@ -900,6 +904,9 @@ pub struct Bank {
     /// This is used to avoid recalculating the same epoch rewards at epoch boundary.
     /// The hashmap is keyed by parent_hash.
     epoch_rewards_calculation_cache: Arc<Mutex<HashMap<Hash, Arc<PartitionedRewardsCalculation>>>>,
+
+    /// The alpenglow genesis block, and the certificate from the genesis vote
+    alpenglow_genesis: Option<(Block, Certificate)>,
```
carllin, line 909: does this have to live in the bank/children banks forever if it also lives in the genesis tracker? 

Maybe eventually just a method to read this info from the account?

votor-messages/src/migration.rs
carllin: Yeah can test this on testnet with people refreshing the same vote over and over.

1. is this vote also going over gossip or are we nixing that completely?

2. Are we planning on switching mainnet votes over to all-to-all before Alpenglow launches, or is that blocked on the validator set limitations first (would be nice to derisk this first)

votor-messages/src/migration.rs
```diff
@@ -40,3 +54,244 @@ pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
...
+    /// The first slot of the migrationary period
+    pub fn migration_slot(&self) -> Option<Slot> {
+        let slot = self.start_migration_slot.load(Ordering::Relaxed);
+        (slot != u64::MAX).then_some(slot)
+    }
+
+    /// Has the migration completed, and alpenglow enabled
+    pub fn is_alpenglow_enabled(&self) -> bool {
+        self.shutdown_poh.load(Ordering::Relaxed)
+            && self.block_creation_loop_started.load(Ordering::Relaxed)
```
carllin, line 161: might be safer require acquire-release semantics as well since we're using this to synchronize with the poh_service thread/event handler/consensus pool service. 

will have to take a more detailed look at where replay is shutting down, but just need to make sure replay shutdown -> `shutdown_poh = true` ->  `block_creation_loop = true` -> block_creation_loop is strictly consistent

carllin, line 161: is this equivalent to just checking `block_creation_loop_started` since thats only set after `shutdown_poh`

votor/src/event_handler.rs
```diff
@@ -124,18 +124,21 @@ impl EventHandler {
 
         // Wait until migration has completed
         info!("{}: Event loop initialized", local_context.my_pubkey);
-        Votor::wait_for_migration_or_exit(&exit, &start);
-        info!("{}: Event loop starting", local_context.my_pubkey);
-
-        if exit.load(Ordering::Relaxed) {
+        let Some(genesis_block) = migration_status.wait_for_migration_or_exit(&exit) else {
```
carllin, line 127: The block creation loop is started by poh_service after migration

Similarly, can this event loop be started before the block creation loop by poh_service after the migration has completed? Would be great to remove as many of these concurrent points of synchronization across services as possible

core/src/validator.rs
AshwinSekar: hello :) 
yeah definitely which I think is totally fine - however in Tower we're able to reconstruct our Tower from VoteAccounts which is not the case with Votor.

My concern (and i've accidentally done this while testing hence the change here) is that we would mess up a set-identity or restart and silently go with a default VoteHistory and submit slashable votes. I think similar to the `--require-tower` arg we should have a `--do-not-require-vote-history` arg and make it the default to panic if we're missing a VoteHistory on startup

runtime/src/bank.rs
AshwinSekar: yeah this is just here out of convenience, in a future PR this will be populated from the account. It's not going to be serialized and can be deleted (along with the MigrationStatus / genesis tracker) once AG succeeds. 

But in the interest of less code for steve to review, i'll remove it here and access the account directly when checking startup

votor-messages/src/migration.rs
AshwinSekar: 1) nixing gossip completely 🪓 
2) My understanding was that the 2k capped validator set / 1.6 sol per epoch VAT would roll out first (wen has it keyed as a separate feature flag). At that point we could run an A2A test, but not sure there's any value considering we're running it on testnet rn uncapped?

votor-messages/src/migration.rs
AshwinSekar: yeah that's fine, this is all only used by one thread:
- dday is if maybe_start_leader and block_creation_loop run at the same time OR handle_votable_bank and votor cast votes together
- Replay is the only thread that will enable alpenglow (to come in #563)
- It blocks until poh is shutdown
- maybe_start_leader /handle_votable_bank etc. is *only* called if poh is not shutdown

But I think just making this acquire release is safer just in case.

AshwinSekar: yeah, I'm just going to rename this to "poh_is_shutdown" and only check that as it's the thing we want to block on

votor/src/event_handler.rs
AshwinSekar: I'm actually going the other way and moving block creation loop to the tpu 😅 
Reasoning is:
- PohService will be deleted
- MigrationStatus will be deleted post migration, so good to have everything independent
- There's only a one way sync here - the "notifier" is replay stage telling poh to shutdown. After PohService is killed, everyone is woken up and it's go time. 

There's no back and forth so I figured this is the cleanest and makes it easy to delete all the code once the migration has succeeded.

votor-messages/src/migration.rs
carllin: Is this roughly the flow?

```
Thread 1 (replay stage):
if !block_creation_loop_started {
    poh_shutdown -> true
}

if !block_creation_loop_started {
    maybe_start_leader()
}

Thread 2 (poh service):
if poh_shutdown {
    block_creation_loop_started -> true
    poh_service::block_creation_loop()
}
```

And goal is to avoid `block_creation_loop()` and `maybe_start_leader` from running concurrently?

PR #538, opened 2025-10-15
feat: bump the `solana-bls-signatures` crate version

akhi3030:
#### Problem
The new version of `solana-bls-signatures` introduces some iterators that will eliminate some vec constructions.

#### Summary of Changes

Similar to https://github.com/anza-xyz/agave/pull/8473.  Bumps the crate version and fixes various vec constructions.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -344,11 +344,10 @@ impl BLSSigVerifier {
             .collect();
 
         let verified_optimistically = if let Ok(aggregate_pubkeys) = aggregate_pubkeys_result {
-            if let Ok(aggregate_signature) = SignatureProjective::par_aggregate(
-                votes_to_verify
-                    .par_iter()
-                    .map(|v| &v.vote_message.signature),
-            ) {
+            let signatures = votes_to_verify
+                .par_iter()
+                .map(|v| &v.vote_message.signature);
+            if let Ok(aggregate_signature) = SignatureProjective::par_aggregate(signatures) {
                 if distinct_messages == 1 {
                     let payload_slice = distinct_payloads[0].as_slice();
                     aggregate_pubkeys[0]


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -61,7 +61,7 @@ fn aggregate_keys_from_bitmap(
         .map(|(_, bls_pubkey)| PubkeyProjective::try_from(*bls_pubkey))
         .collect();
     let pubkeys = pubkeys.ok()?;
-    PubkeyProjective::par_aggregate(&pubkeys.iter().collect::<Vec<_>>()).ok()
+    PubkeyProjective::par_aggregate(pubkeys.par_iter()).ok()
```
akhi3030, line 64: I tried some techniques to see if I can get away without constructing the vec above.  However, because it doesn't seem possible to do error checking without `collect()`ing.  Happy to receive feedback on improvements.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -336,35 +333,33 @@ impl BLSSigVerifier {
             .votes_batch_distinct_messages_count
             .fetch_add(distinct_messages as u64, Ordering::Relaxed);
 
-        let distinct_data: Vec<_> = grouped_pubkeys.into_iter().collect();
-        let aggregate_pubkeys_result: Result<Vec<PubkeyProjective>, _> = distinct_data
-            .iter()
-            .map(|(_, pubkeys)| PubkeyProjective::par_aggregate(pubkeys))
+        let (distinct_payloads, distinct_pubkeys): (Vec<_>, Vec<_>) =
```
akhi3030, line 339: I think it is possible to skip the collection into `distinct_pubkeys` here but I'll leave that optimisation for a future PR.

core/src/bls_sigverify/bls_sigverifier.rs
ksn6: Is `par_aggregate` + `par_iter` okay to use here? Would `iter()` alone work?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -336,35 +333,33 @@ impl BLSSigVerifier {
...
 
         let verified_optimistically = if let Ok(aggregate_pubkeys) = aggregate_pubkeys_result {
-            let all_signatures: Vec<&BlsSignature> = votes_to_verify
-                .iter()
-                .map(|v| &v.vote_message.signature)
-                .collect();
-
-            if let Ok(aggregate_signature) = SignatureProjective::par_aggregate(&all_signatures) {
+            if let Ok(aggregate_signature) = SignatureProjective::par_aggregate(
+                votes_to_verify
```
ksn6: nit: for readability, could we keep the definition of `all_signatures` above minus the `.collect()`?

Is `.par_iter()` + `par_aggregate()` okay to use here, or should we just go with `.iter()` here?

core/src/bls_sigverify/bls_sigverifier.rs
akhi3030: we were doing a `par_aggregate` before.  I assume because we want to do the aggregation in parallel.  `iter()` would not work then.  Or maybe I am misunderstanding the question.

core/src/bls_sigverify/bls_sigverifier.rs
akhi3030: reintroduced the signatures binding.

akhi3030: we are doing the `par_aggregate()` here to get parallelism.

PR #539, opened 2025-10-16
feat: block component fuse and serialized size

ksn6:
This PR solves two problems:

- Fuse two `BlockComponent`s together if both are `EntryBatch`es - i.e., return a single `BlockComponent` with a single concatenated `EntryBatch`.

- Given a `BlockComponent`, determine its serialized size reasonably efficiently (i.e., without having to actually serialize the component).

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -966,6 +966,8 @@ impl BlockFooterV1 {
         }
 
         // Read timestamp
+        // Unwrap: HEADER_SIZE = TIMESTAMP_SIZE + USER_AGENT_LEN_SIZE > TIMESTAMP_SIZE, so this will
+        // never fail.
         let time_bytes = data[..Self::TIMESTAMP_SIZE].try_into().unwrap();
         let block_producer_time_nanos = u64::from_le_bytes(time_bytes);
 


entry/src/block_component.rs
```diff
@@ -355,6 +355,7 @@ impl BlockComponent {
     ///
     /// TODO(karthik): lower this to a reasonable value.
     const MAX_ENTRIES: usize = u32::MAX as usize;
+    const ENTRY_COUNT_SIZE: usize = 8;
```
akhi3030, line 360: can we get some documentation on this please?

entry/src/block_component.rs
```diff
@@ -379,15 +380,37 @@ impl BlockComponent {
...
             Self::EntryBatch(entries) => entries,
             Self::BlockMarker(_) => &[],
         }
     }
 
+    /// Returns a reference to the entry batch vector.
+    ///
+    /// # Panics
+    /// Panics if this component is not an entry batch.
+    pub fn entry_batch(&self) -> &Vec<Entry> {
```
akhi3030:
```suggestion
    pub fn entry_batch(&self) -> &[Entry] {
```

Usually a slice is considered more ergonomic than a reference to a Vec.

akhi3030: I don't see much difference between this function and the `entries` function above.  Do we need both?

entry/src/block_component.rs
```diff
@@ -379,15 +380,37 @@ impl BlockComponent {
...
+            Self::EntryBatch(entries) => entries,
+            _ => panic!("BlockComponent isn't an EntryBatch."),
+        }
+    }
+
+    /// Returns a mutable reference to the entry batch vector.
+    ///
+    /// # Panics
+    /// Panics if this component is not an entry batch.
+    pub fn entry_batch_mut(&mut self) -> &mut Vec<Entry> {
```
akhi3030: doesn't seem to be used anywhere.  Maybe we add it when we need it.

entry/src/block_component.rs
```diff
@@ -379,15 +380,37 @@ impl BlockComponent {
...
+        match self {
+            Self::EntryBatch(entries) => entries,
+            _ => panic!("BlockComponent isn't an EntryBatch."),
+        }
+    }
+
     /// Consumes this component and returns the entries if it's an entry batch.
-    pub fn into_entry_batch(self) -> Option<Vec<Entry>> {
+    pub fn as_entry_batch_owned(self) -> Option<Vec<Entry>> {
```
akhi3030, line 396: also doesn't seem to be used.

entry/src/block_component.rs
```diff
@@ -415,6 +438,23 @@ impl BlockComponent {
         matches!(self, Self::BlockMarker(_))
     }
 
+    /// Fuses another BlockComponent into this one if both are entry batches.
+    ///
+    /// Returns None if fusion occurred, Some(other) if no fusion was possible.
+    pub fn maybe_fuse(&mut self, other: Self) -> Option<Self> {
```
akhi3030:
```suggestion
    pub fn try_fuse(&mut self, other: Self) -> Option<Self> {
```

nit: I think this is a more intuitive name for this function.

entry/src/block_component.rs
```diff
@@ -415,6 +438,23 @@ impl BlockComponent {
...
+            Self::EntryBatch(entries) => {
+                if let Self::EntryBatch(other_entries) = other {
+                    entries.extend(other_entries);
+                    None
+                } else {
+                    Some(other)
+                }
+            }
+            _ => Some(other),
+        }
```
akhi3030:
```suggestion
        match (self, &other) {
            (Self::EntryBatch(self_entries), Self::EntryBatch(other_entries)) => {
                self_entries.extend_from_slice(other_entries);
                None
            }
            _ => Some(other),
        }
    }
```

I find the suggestion to be shorter, more readable, and it uses extend_from_slice which can be faster in many cases.

entry/src/block_component.rs
```diff
@@ -975,6 +1053,14 @@ impl VersionedBlockFooter {
         let footer = BlockFooterV1::from_bytes(remaining)?;
         Ok(Self::Current(footer))
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
akhi3030: this function doesn't need to return a `Result`.

entry/src/block_component.rs
```diff
@@ -1077,6 +1173,14 @@ impl VersionedBlockHeader {
         let header = BlockHeaderV1::from_bytes(remaining)?;
         Ok(Self::Current(header))
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
akhi3030: this function also doesn't need to return a `Result`.

entry/src/block_component.rs
```diff
@@ -610,6 +662,8 @@ impl<'de> Deserialize<'de> for BlockComponent {
 // ============================================================================
 
 impl VersionedBlockMarker {
+    const VERSION_SIZE: usize = 2;
```
akhi3030, line 645: documentation please.

entry/src/block_component.rs
```diff
@@ -699,7 +749,15 @@ impl VersionedBlockMarker {
         };
 
         // Total size includes the version field
-        Ok(VERSION_SIZE + marker_inner_size)
+        Ok(Self::VERSION_SIZE + marker_inner_size)
+    }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
akhi3030: this can be updated to not return a `Result` if the functions it is calling are also updated.

entry/src/block_component.rs
```diff
@@ -802,6 +860,9 @@ fn read_variant_with_length(data: &[u8]) -> Result<(u8, &[u8]), BlockComponentEr
 }
 
 impl BlockMarkerV1 {
+    const VARIANT_ID_SIZE: u64 = 1;
+    const LENGTH_FIELD_SIZE: u64 = 2;
```
akhi3030, line 846: documentation.

entry/src/block_component.rs
```diff
@@ -833,6 +894,16 @@ impl BlockMarkerV1 {
             }),
         }
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
akhi3030: this can also be updated to not return a `Result`.

entry/src/block_component.rs
```diff
@@ -878,6 +949,9 @@ impl<'de> Deserialize<'de> for BlockMarkerV1 {
 impl BlockFooterV1 {
     /// Maximum length for user agent bytes.
     const MAX_USER_AGENT_LEN: usize = 255;
+    const TIMESTAMP_SIZE: usize = 8;
+    const LENGTH_SIZE: usize = 1;
+    const HEADER_SIZE: usize = Self::TIMESTAMP_SIZE + Self::LENGTH_SIZE;
```
akhi3030, line 939: documentation.

entry/src/block_component.rs
```diff
@@ -902,39 +976,43 @@ impl BlockFooterV1 {
...
-        const HEADER_SIZE: usize = TIMESTAMP_SIZE + LENGTH_SIZE;
-
-        if data.len() < HEADER_SIZE {
+        if data.len() < Self::HEADER_SIZE {
             return Err(BlockComponentError::InsufficientData);
         }
 
         // Read timestamp
-        let time_bytes: [u8; TIMESTAMP_SIZE] = data[..TIMESTAMP_SIZE]
+        let time_bytes: [u8; Self::TIMESTAMP_SIZE] = data[..Self::TIMESTAMP_SIZE]
```
akhi3030:
```suggestion
        let time_bytes = data[..Self::TIMESTAMP_SIZE]
```

inference works here.

entry/src/block_component.rs
```diff
@@ -902,39 +976,43 @@ impl BlockFooterV1 {
...
             .try_into()
             .map_err(|_| BlockComponentError::InsufficientData)?;
```
akhi3030:
```suggestion
            .unwrap();
```

nit: this cannot fail because of the check above.  As a reader, I think using unwrap is cleaner.

entry/src/block_component.rs
```diff
@@ -902,39 +976,43 @@ impl BlockFooterV1 {
...
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> u64 {
+        let user_agent_size = self.block_user_agent.len().min(Self::MAX_USER_AGENT_LEN) as u64;
+        Self::HEADER_SIZE as u64 + user_agent_size
+    }
 }
 
 impl VersionedBlockFooter {
+    const VERSION_SIZE: u64 = 1;
```
akhi3030, line 1000: documentation.

entry/src/block_component.rs
```diff
@@ -902,39 +976,43 @@ impl BlockFooterV1 {
...
 
         Ok(Self {
             block_producer_time_nanos,
             block_user_agent,
         })
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> u64 {
+        let user_agent_size = self.block_user_agent.len().min(Self::MAX_USER_AGENT_LEN) as u64;
```
akhi3030, line 993: I am wondering if we should add documentation and checks to ensure that `self.block_user_agent` cannot be set to larger than `MAX_USER_AGENT_LEN`.  And then we can use an `assert` here instead.

entry/src/block_component.rs
```diff
@@ -902,39 +976,43 @@ impl BlockFooterV1 {
...
+        let block_user_agent = data[Self::HEADER_SIZE..Self::HEADER_SIZE + user_agent_len].to_vec();
 
         Ok(Self {
             block_producer_time_nanos,
             block_user_agent,
         })
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> u64 {
```
akhi3030, line 992: nit: sizes are usually `usize`.  Doesn't matter so much as we use `u64` interchangeably.

entry/src/block_component.rs
```diff
@@ -1018,6 +1104,9 @@ impl<'de> Deserialize<'de> for VersionedBlockFooter {
 // ============================================================================
 
 impl BlockHeaderV1 {
+    const SLOT_SIZE: u64 = 8;
+    const HASH_SIZE: u64 = 32;
```
akhi3030, line 1096: documentation

entry/src/block_component.rs
```diff
@@ -1034,9 +1123,16 @@ impl BlockHeaderV1 {
...
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> u64 {
+        Self::SLOT_SIZE + Self::HASH_SIZE
+    }
 }
 
 impl VersionedBlockHeader {
+    const VERSION_SIZE: u64 = 1;
```
akhi3030, line 1123: documentation.

entry/src/block_component.rs
```diff
@@ -1120,6 +1224,9 @@ impl<'de> Deserialize<'de> for VersionedBlockHeader {
 // ============================================================================
 
 impl UpdateParentV1 {
+    const SLOT_SIZE: u64 = 8;
+    const HASH_SIZE: u64 = 32;
```
akhi3030, line 1219: documentation.

entry/src/block_component.rs
```diff
@@ -1136,9 +1243,16 @@ impl UpdateParentV1 {
...
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> u64 {
+        Self::SLOT_SIZE + Self::HASH_SIZE
+    }
 }
 
 impl VersionedUpdateParent {
+    const VERSION_SIZE: u64 = 1;
```
akhi3030, line 1246: documentation.

entry/src/block_component.rs
```diff
@@ -1179,6 +1293,14 @@ impl VersionedUpdateParent {
         let update = UpdateParentV1::from_bytes(remaining)?;
         Ok(Self::Current(update))
     }
+
+    /// Returns the serialized size in bytes without actually serializing.
+    fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
akhi3030: can be updated to not return a `Result`.

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: done - yep for context there was a case where I wanted `entries` to return the empty slice rather than panic; I don't think we need this anymore

entry/src/block_component.rs
ksn6: we use this here:

https://github.com/anza-xyz/alpenglow/pull/539/files#diff-01b72272fd1033dcbb8fed13821a6232d35913119af6be20a7c2d1e9c772feb9R4124

entry/src/block_component.rs

entry/src/block_component.rs

entry/src/block_component.rs

entry/src/block_component.rs

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs
ksn6: done

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: yep, a good idea; for now, the various structs here all expose internals - let's fix this in a separate PR

ksn6:
> sorry about excessive comments. I hope they are useful.

No worries about the quantity - the code looks cleaner now, thanks. Just a note: one summary comment grouping similar comments (e.g. documentation) would make things easier to address.

entry/src/block_component.rs
```diff
@@ -902,39 +961,42 @@ impl BlockFooterV1 {
...
-        if data.len() < HEADER_SIZE {
+        if data.len() < Self::HEADER_SIZE {
             return Err(BlockComponentError::InsufficientData);
         }
 
         // Read timestamp
-        let time_bytes: [u8; TIMESTAMP_SIZE] = data[..TIMESTAMP_SIZE]
-            .try_into()
-            .map_err(|_| BlockComponentError::InsufficientData)?;
+        let time_bytes = data[..Self::TIMESTAMP_SIZE].try_into().unwrap();
```
akhi3030, line 971: nit: I should've mentioned it in the previous comment but forgot.  We could add a comment about why the unwrap is safe here.

entry/src/block_component.rs
```diff
@@ -1018,6 +1088,11 @@ impl<'de> Deserialize<'de> for VersionedBlockFooter {
 // ============================================================================
 
 impl BlockHeaderV1 {
+    /// Size in bytes of the slot field.
+    const SLOT_SIZE: u64 = 8;
```
akhi3030, line 1094:
```suggestion
    const SLOT_SIZE: u64 = std::mem::size_of::<Slot>();
```

akhi3030: is this cleaner?  Similar comments below for hash and other slot fields.

ksn6: would rather stick with explicit sizing for now rather than `std::mem::size_of` everywhere; slots are generally 8 bytes everywhere in the code already, etc.

entry/src/block_component.rs
ksn6: done

PR #540, opened 2025-10-16
RFC: another proposal for the rewards container

akhi3030:
This combines feedback and suggestions received in previous proposals into one.  

- Introduces a `ConsensusRewards` container that tracks certificate builders and vote messages relevant for rewards
- Updates certificate builder so they can be reused for rewards
- updates consensus pool to send the builders to the rewards containers
- TODO: update BLSSigVerifier to send msgs to rewards container if they are no longer needed for the consensus pool
- TODO: update block producer to actually generate the certificates and use them in the block markers
- TODO: various nits in the PR.

### DIFF
--- a/votor/src/consensus_pool/consensus_rewards.rs
+++ b/votor/src/consensus_pool/consensus_rewards.rs
@@ -16,6 +16,7 @@ use {
 
 mod rewards_entry;
 
+/// Returns true if we are interested in storing entries for this slot.
 fn should_record(
     cluster_info: &ClusterInfo,
     leader_schedule_cache: &LeaderScheduleCache,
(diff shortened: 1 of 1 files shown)


votor/src/consensus_pool.rs
```diff
@@ -235,6 +243,10 @@ impl ConsensusPool {
                     };
                 }
             });
+
+            self.consensus_rewards
```
akhi3030, line 247: feedback from Quentin: this is only getting called if we actually build a cert.  If we do not hit the stake limit above, we will not call this.

votor/src/consensus_pool/vote_certificate_builder.rs
```diff
@@ -71,27 +72,48 @@ impl VoteCertificateBuilder {
...
+                })
             }
-        };
+        }
+    }
+
+    /// Builds a [`CertificateMessage`] from the builder that only considers the first bitmap for inclusion.
+    ///
+    /// This builds a certificate for rewards and as per the SIMD does not include fallback votes so does not include the second bitmap.
+    pub fn build_for_rewards(self) -> Result<CertificateMessage, CertificateError> {
```
akhi3030, line 109: Feedback from Quentin.  This doesn't really work.  Above we have aggregated signatures from both types of votes but now we are using just one bitmap to build the cert.

PR #541, opened 2025-10-17
Send Alpenglow Last Voted Slot from BLS Sigverifier to RPC.

wen-coding:
#### Problem
We don't have last_voted slot in vote_state in Alpenglow any more.

#### Summary of Changes

The correct fix is to send certificates to block footer and then consume them for delinquency reporting.

Before we have block footer implemented let's just quickly send last_voted_slot from BLS Sigverifier to RPC, so we don't get spammed by delinquency alerts any more.

### DIFF
--- a/rpc/src/alpenglow_last_voted.rs
+++ b/rpc/src/alpenglow_last_voted.rs
@@ -39,11 +39,7 @@ impl AlpenglowLastVoted {
     }
 
     pub fn get_last_voted(&self, pubkey: &Pubkey) -> Option<Slot> {
-        self.last_voted_map
-            .read()
-            .unwrap()
-            .get(pubkey)
-            .copied()
+        self.last_voted_map.read().unwrap().get(pubkey).copied()
     }
 }
 


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -291,6 +295,10 @@ impl BLSSigVerifier {
             }
         }
 
+        // Send to RPC service for last voted tracking
+        self.alpenglow_last_voted
```
akhi3030, line 299: Is this potentially in the critical path?  This function can be potentially expensive as it is using `retain`.

wen-coding: The critical path is between message receive to sending out to consensus pool. But you raised a good point about `retain`. I changed it to refuse to updating the HashMap beyond 2000 pubkeys. We should only have 50 pubkeys in test cluster. And by the time we move to real cluster we should have the real thing implemented.

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar, line 275: nit: could store a different map in `self.latest_notarize_fallback_or_stronger: HashMap<Pubkey, Slot>` where we only store the max to avoid having to iterate again for the `max()` in `update_last_voted`

rpc/src/alpenglow_last_voted.rs
```diff
@@ -0,0 +1,64 @@
...
+// will have the real thing implemented.
+const MAX_ENTRIES: usize = 2000;
+
+#[derive(Default)]
+pub struct AlpenglowLastVoted {
+    last_voted_map: RwLock<HashMap<Pubkey, Slot>>,
+}
+
+impl AlpenglowLastVoted {
+    pub fn update_last_voted(&self, verified_votes_by_pubkey: &HashMap<Pubkey, Vec<Slot>>) {
```
akhi3030, line 19: it seems we are only using the largest slot so I wonder if we can compute that instead of sending a list of slots and having to iterate over it again and again.

akhi3030 (approved):
nit on performance.  But LGTM as this is a short term hack.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Do you not need all verified votes? Then we can change this part. But let's address that in another PR and push this to silence the delinquent alerts.

rpc/src/alpenglow_last_voted.rs
wen-coding: Sure, let me see how Ashwin wants to change verified votes. But let me submit this and push to invalidator first to silence the annoying alerts.

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar: oh i mean just add a separate map for RPC that only tracks the highest. we can keep verified votes the same.

PR #543, opened 2025-10-17
feat: get `PohRecorder` to send `EntryMarker`s

ksn6:
### Summary

We need to push `BlockComponent::BlockMarker`s into blocks from various parts of the code, e.g., `BlockFooter`, `BlockHeader`, and `UpdateParent` for Alpenglow Fast Leader Handover.

### Approach

We modify `WorkingBankEntry` to work with `EntryMarker` types rather than `Entry`, such that `PohRecorder` can properly disseminate markers.

### DIFF
--- a/ledger/src/shredder.rs
+++ b/ledger/src/shredder.rs
@@ -17,14 +17,6 @@ use {
     },
 };
 
-/// Result of shredding components, including the shreds and updated state.
-pub struct ShredderComponentsResult {
-    pub shreds: Vec<Shred>,
-    pub next_shred_index: u32,
-    pub next_code_index: u32,
-    pub chained_merkle_root: Option<Hash>,
-}
-
 static PAR_THREAD_POOL: std::sync::LazyLock<ThreadPool> = std::sync::LazyLock::new(|| {
     rayon::ThreadPoolBuilder::new()
         .num_threads(get_thread_count())
@@ -700,30 +692,6 @@ mod tests {
         {
             assert_eq!(e.payload(), c.payload());
         }
-    }
-
-    #[test]
-    fn test_component_empty_entry_batch() {
-        let keypair = Keypair::new();
-        let shredder = Shredder::new(500, 495, 25, 400).unwrap();
-
-        let component = BlockComponent::EntryBatch(vec![]);
-        let initial_merkle_root = Hash::new_from_array(rand::thread_rng().gen());
-
-        let (data_shreds, coding_shreds) = shredder.component_to_merkle_shreds_for_tests(
-            &keypair,
-            &component,
-            false,
-            Some(initial_merkle_root),
-            10,
-            5,
-            &ReedSolomonCache::default(),
-            &mut ProcessShredsStats::default(),
-        );
-
-        // Verify we still get shreds even with empty entry batch
-        assert!(!data_shreds.is_empty());
-        assert!(!coding_shreds.is_empty());
     }
 
     #[test_matrix([true, false])]
--- a/ledger/src/shred.rs
+++ b/ledger/src/shred.rs
@@ -60,7 +60,7 @@ pub use {
         shred_data::ShredData,
         stats::{ProcessShredsStats, ShredFetchStats},
     },
-    crate::shredder::{ReedSolomonCache, Shredder, ShredderComponentsResult},
+    crate::shredder::{ReedSolomonCache, Shredder},
 };
 use {
     self::{shred_code::ShredCode, traits::Shred as _},


core/src/tpu_entry_notifier.rs
```diff
@@ -91,7 +97,7 @@ impl TpuEntryNotifier {
         }
         *current_transaction_index += entry.transactions.len();
 
-        if let Err(err) = broadcast_entry_sender.send((bank, (entry, tick_height))) {
+        if let Err(err) = broadcast_entry_sender.send((bank, (entry_marker, tick_height))) {
```
apfitzge: this doesn't seem quite right to me?

This channel goes to broadcast, do we not want it to send entry markers? or are they sent through elsewhere?

I think we probably either want this broadcast channel here to ONLY send entries if that's what we expect (i.e. send entry not entry_marker here) or we cannot just return Ok immediately above.

ksn6: Yep, sounds good - was originally planning on sending out just entries and then updating this later, though can send out all `EntryMarker`-s now.

EDIT: @apfitzge am I doing the index-updating logic correctly, i.e., only on entries and not on markers?

entry/src/entry_marker.rs
```diff
@@ -0,0 +1,119 @@
+//! Entry marker types for the POH recording pipeline.
```
AshwinSekar:
```suggestion
//! Entry marker types for the PoH recording pipeline.
```

entry/src/entry_marker.rs
```diff
@@ -0,0 +1,119 @@
...
+//!
+//! This module defines `EntryMarker`, a wrapper type that allows both regular entries and block
+//! markers (headers, footers) to flow through the same POH recording channel.
```
AshwinSekar:
```suggestion
//! markers (headers, footers) to flow through the same PoH recording channel.
```

entry/src/entry_marker.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+/// Converts a BlockComponent into an EntryMarker.
+///
+/// # Panics
+///
+/// Panics if the BlockComponent is an EntryBatch containing more than one entry,
+/// as EntryMarker can only wrap a single entry.
+impl From<BlockComponent> for EntryMarker {
```
AshwinSekar: nit: can we make this a TryFrom instead? And then the caller can expect

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -76,19 +100,19 @@ fn max_coalesce_time(serialized_batch_byte_count: u64, max_batch_byte_count: u64
 }
 
 pub(super) fn recv_slot_entries(
```
AshwinSekar, line 88: I think it would be simpler for here and the shredding code if we didn't mix the components together.
Essentially change `ReceiveResults` to either return an entry batch or a block marker.

Then the shredding code becomes really simple as broadcast stage already does the scan to update the next_*_index and cmr - you just have to change entries_to_shreds to either call `make_merkle_shreds_from_entries` or `make_merkle_shreds_from_block_marker` (or just use make from data slice for both), and we can avoid the  tracking in `make_merkle_shreds_from_components`

ledger/src/shredder.rs
```diff
@@ -63,6 +72,105 @@ impl Shredder {
...
+            next_shred_index,
+            next_code_index,
+            reed_solomon_cache,
+            stats,
+        )
+        .unwrap()
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn make_merkle_shreds_from_components(
```
AshwinSekar: See the other comment, but I think if we go with that approach we dont' need this function at all, since broadcast stage already takes care of looping through the receiver

ledger/src/shredder.rs
```diff
@@ -63,6 +72,105 @@ impl Shredder {
...
+        component: &BlockComponent,
+        is_last_in_slot: bool,
+        chained_merkle_root: Option<Hash>,
+        next_shred_index: u32,
+        next_code_index: u32,
+        reed_solomon_cache: &ReedSolomonCache,
+        stats: &mut ProcessShredsStats,
+    ) -> impl Iterator<Item = Shred> {
+        let now = Instant::now();
+        let bytes = component.to_bytes().unwrap();
```
AshwinSekar, line 79: not in this pr but master has been updated to use wincode instead of bincode, so we'll want to make sure we're using that for this serialization as well.

turbine/src/broadcast_stage/broadcast_utils.rs
AshwinSekar: The coalesce code here can be:
- If the first thing you recieve is an Entry, drain until you either reach a block marker or reach the target limit
- Return the `Vec<Entry>`
- OR if the first thing you pull off is a block marker, immediately return the block marker

Should hopefully make it simpler and avoid the allo in fuse or push

AshwinSekar (commented):
Overall looks pretty good just some ideas for simplifying the coalesce and shredding code

core/src/tpu_entry_notifier.rs
```diff
@@ -54,47 +54,53 @@ impl TpuEntryNotifier {
...
     pub(crate) fn send_entry_notification(
         exit: Arc<AtomicBool>,
-        entry_receiver: &Receiver<WorkingBankEntry>,
+        entry_receiver: &Receiver<WorkingBankEntryMarker>,
         entry_notification_sender: &EntryNotifierSender,
-        broadcast_entry_sender: &Sender<WorkingBankEntry>,
+        broadcast_entry_sender: &Sender<WorkingBankEntryMarker>,
         current_slot: &mut u64,
         current_index: &mut usize,
         current_transaction_index: &mut usize,
```
apfitzge, line 62: this c-code is not your fault...but can we refactor this to return these values. Why are we doing mutable refs just to change integers.

Seems should be 2 fns:
1. update indexes (infallible)
2. send -> Result

apfitzge: separate from this PR btw

core/src/tpu_entry_notifier.rs
apfitzge: I guess not sure how we want to handle the markers.
Block header/footer, doesn't matter that much obviously if we update indexes.

The mid-slot parent-switcharoo is up for debate I think. If we're pretending the previous txs didn't happen or aren't "really" part of the block it might make sense to reset.
But probably I think we'd still want a way to index those txs included before the switch, right? Which means not resetting is the right approach

ksn6: Yep, makes sense. Given this, let's:

- Increment indices on markers + entries
- Not reset indices after an `UpdateParent` message.

apfitzge: err i don't think we need to increment on the markers, right? at least for the transaction indexes. entry index probably best to

entry/src/entry_marker.rs
ksn6: done

entry/src/entry_marker.rs
ksn6: done

core/src/tpu_entry_notifier.rs
ksn6: Yep, this is the plan ^; my bad, should've been more clear

entry/src/entry_marker.rs
ksn6: done

core/src/tpu_entry_notifier.rs
ksn6: yep, makes sense - will add a todo for now and fix in a follow-up PR

turbine/src/broadcast_stage/broadcast_utils.rs
ksn6: nice, yep - this simplifies things

ledger/src/shredder.rs
ksn6: yep, done

ledger/src/shredder.rs
ksn6: nothing should change by much here - it's mostly stuff under `BlockComponent`; can address that separately

core/src/banking_stage.rs
```diff
@@ -828,7 +831,10 @@ mod tests {
         trace!("getting entries");
         let entries: Vec<_> = entry_receiver
             .iter()
-            .map(|(_bank, (entry, _tick_height))| entry)
+            .filter_map(|(_bank, (entry_marker, _tick_height))| match entry_marker {
```
apfitzge: nit: just use as_entry()

apfitzge: several places where this could be done to just avoid repeating this same match

entry/src/entry_marker.rs
```diff
@@ -0,0 +1,129 @@
...
+/// Converts a BlockComponent into an EntryMarker.
+///
+/// # Errors
+///
+/// Returns `EntryMarkerError::MultipleEntries` if the BlockComponent is an EntryBatch
+/// containing more than one entry, as EntryMarker can only wrap a single entry.
+impl TryFrom<BlockComponent> for EntryMarker {
+    type Error = EntryMarkerError;
+
+    fn try_from(component: BlockComponent) -> Result<Self, Self::Error> {
```
apfitzge: for the uninitiated, could you give a brief explanation of what a BlockComponent is? why can it represent multiple entries if we only allow it to have one?

core/src/banking_stage.rs
ksn6: done (here and elsewhere)

entry/src/entry_marker.rs
ksn6: we can entirely eliminate this `TryFrom` thanks to the refactor that ashwin suggested

in case folks are curious as to what BlockComponents are, see [here](https://github.com/anza-xyz/alpenglow/blob/master/entry/src/block_component.rs)

ledger/src/shredder.rs
```diff
@@ -63,6 +72,71 @@ impl Shredder {
...
+            chained_merkle_root,
+            next_shred_index,
+            next_code_index,
+            reed_solomon_cache,
+            stats,
+        )
+        .unwrap()
+        .collect();
+
+        let mut current_shred_index = next_shred_index;
```
AshwinSekar: This stuff below is best left to the caller, I don't think it should be the responsibility of the Shredder to return this information.

The caller will likely already iterate or otherwise manipulate the shreds, so adding this extra tracking isn't helpful. And sometimes especially in tests we don't care about this stuff.

poh/src/poh_recorder.rs
```diff
@@ -363,7 +364,8 @@ impl PohRecorder {
                                     num_hashes: entry.num_hashes,
                                     hash: entry.hash,
                                     transactions,
-                                },
+                                }
+                                .into(),
```
AshwinSekar, line 368: Really prefer using From vs into wherever possible as it makes it difficult to figure out what's going on outside of an IDE

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -76,38 +87,65 @@ fn max_coalesce_time(serialized_batch_byte_count: u64, max_batch_byte_count: u64
...
+    let (mut bank, (entry_marker, mut last_tick_height)) = match carryover_entry.take() {
+        Some((bank, (entry_marker, tick_height))) => (bank, (entry_marker, tick_height)),
         None => receiver.recv_timeout(Duration::new(1, 0))?,
     };
     assert!(last_tick_height <= bank.max_tick_height());
-    let mut entries = vec![entry];
 
-    // Drain the channel of entries.
+    // If the first thing is a block marker, return it immediately
+    if let Some(marker) = entry_marker.as_marker().cloned() {
```
AshwinSekar: do we need the clone here, can we add maybe an `is_marker` and then call `into_marker`?

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -76,38 +87,65 @@ fn max_coalesce_time(serialized_batch_byte_count: u64, max_batch_byte_count: u64
...
+            break;
+        }
+
+        // Add the entry to our batch
+        entries.push(next_marker.into_entry().expect("must be Entry"));
         assert!(last_tick_height <= bank.max_tick_height());
     }
 
-    let mut serialized_batch_byte_count = serialized_size(&entries)?;
+    let mut serialized_batch_byte_count = bincode::serialized_size(&entries).map_err(|e| {
```
AshwinSekar: nit why did this line need to change?

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -142,24 +181,33 @@ pub(super) fn recv_slot_entries(
...
+        }
+
+        let entry = entry_marker.into_entry().expect("must be Entry");
         let entry_bytes = serialized_size(&entry)?;
+
         if serialized_batch_byte_count + entry_bytes > max_batch_byte_count {
             // This entry will push us over the batch byte limit. Save it for
             // the next batch.
-            *carryover_entry = Some((try_bank, (entry, tick_height)));
+            *carryover_entry = Some((try_bank, (entry.into(), tick_height)));
```
AshwinSekar, line 192: pls From

AshwinSekar (commented):
Almost there just some minor comments

poh/src/poh_recorder.rs
ksn6: Hm, would rather use `.into()` , as we don't annotate variables with types everywhere anyways, and this doesn't clutter code.

turbine/src/broadcast_stage/broadcast_utils.rs
ksn6: yep, works

turbine/src/broadcast_stage/broadcast_utils.rs
ksn6: reverted - keeping the `bincode::` for explicitness though

turbine/src/broadcast_stage/broadcast_utils.rs
ksn6: see https://github.com/anza-xyz/alpenglow/pull/543#discussion_r2455881685

turbine/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -110,6 +110,54 @@ impl StandardBroadcastRun {
...
+        if self.next_code_index > max_code_shreds_per_slot {
+            return Err(BroadcastError::TooManyShreds);
+        }
+
+        Ok(shredder_result.shreds)
+    }
+
+    #[allow(dead_code)]
     #[allow(clippy::too_many_arguments)]
     fn entries_to_shreds(
```
AshwinSekar: can we delete this now? or if used in tests make it call component to shreds?

ledger/src/shredder.rs
ksn6: sg, done

turbine/src/broadcast_stage/standard_broadcast_run.rs
ksn6: yep, done

poh/src/poh_recorder.rs
apfitzge: nah Ashwin is right we definitely try to use From where we can

ledger/src/shredder.rs
```diff
@@ -17,6 +17,14 @@ use {
     },
 };
 
+/// Result of shredding components, including the shreds and updated state.
+pub struct ShredderComponentsResult {
```
AshwinSekar: can we delete this now?

ledger/src/shredder.rs
```diff
@@ -572,6 +637,95 @@ mod tests {
...
+        for (e, c) in coding_shreds_entries
+            .iter()
+            .zip(coding_shreds_components.iter())
+        {
+            assert_eq!(e.payload(), c.payload());
+        }
+    }
+
+    #[test]
+    fn test_component_empty_entry_batch() {
```
AshwinSekar: nit: this shouldn't be allowed right?

ledger/src/shredder.rs
ksn6: oops, yep

ledger/src/shredder.rs
ksn6: done

PR #544, opened 2025-10-17
feat: update certificate builder to handle generate certificates for rewards

akhi3030:
#### Problem

- Current certificate builder does not handle building certs for rewards properly.
- As pointed out by @qkniep in https://github.com/anza-xyz/alpenglow/pull/540#discussion_r2438869071, we need to aggregate the signatures for different types of messages separately.  

#### Summary of Changes

- Renames `VoteCertificateBuilder` to just `CertificateBuilder`.
- Overhauls the builder so that we can track the building of different types of certificates separately.
- Improves error handling the builder.
- Enables ability to build certificates for rewards

### DIFF
--- a/votor/src/consensus_pool/certificate_builder.rs
+++ b/votor/src/consensus_pool/certificate_builder.rs
@@ -21,8 +21,8 @@ const MAXIMUM_VALIDATORS: usize = 4096;
 pub enum AggregateError {
     #[error("BLS error: {0}")]
     Bls(#[from] BlsError),
-    #[error("Validator does not exist for given rank: {0}")]
-    ValidatorDoesNotExist(u16),
+    #[error("rank {0} is invalid")]
+    InvalidRank(u16),
     #[error("validator already included")]
     ValidatorAlreadyIncluded,
 }
@@ -89,7 +89,7 @@ fn build_cert_from_bitmaps(
 fn try_set_bitmap(bitmap: &mut BitVec<u8, Lsb0>, rank: u16) -> Result<(), AggregateError> {
     let mut ptr = bitmap
         .get_mut(rank as usize)
-        .ok_or(AggregateError::ValidatorDoesNotExist(rank))?;
+        .ok_or(AggregateError::InvalidRank(rank))?;
     if *ptr {
         return Err(AggregateError::ValidatorAlreadyIncluded);
     }
@@ -445,9 +445,7 @@ mod tests {
         };
         assert_eq!(
             builder.aggregate(&[message_out_of_bounds]),
-            Err(AggregateError::ValidatorDoesNotExist(
-                rank_out_of_bounds as u16
-            ))
+            Err(AggregateError::InvalidRank(rank_out_of_bounds as u16))
         );
 
         // Test bls error


akhi3030:
the test failures seem to be happening because I have not updated them to the new builder.

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
-            signature: self.signature.into(),
-            bitmap,
-        })
+            Self::Skip {
+                mut signature0,
+                bitmap0,
+                sig_and_bitmap1,
+            } => match sig_and_bitmap1 {
+                None => build_cert_from_bitmap(certificate, signature0, bitmap0)
+                    .map_err(BuildError::Encode),
```
qkniep, line 230: why do `build_cert_from_bitmap()` and `build_cert_from_bitmaps()` return different error types? they both seem to only ever return the same encoding error

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
-            if MAXIMUM_VALIDATORS <= rank {
-                return Err(CertificateError::ValidatorDoesNotExist(vote_message.rank));
+    /// Aggregates new [`VoteMessage`]s into the builder.
+    fn aggregate(
+        &mut self,
+        certificate: &Certificate,
+        msgs: &[VoteMessage],
+    ) -> Result<(), AggregateError> {
+        let vote_types = certificate_limits_and_vote_types(certificate).1;
+        assert_eq!(vote_types.len(), 2);
```
qkniep: for `SingleVote` (i.e. notar, final, fast final certs) this does not hold

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+        match self {
+            Self::Skip {
+                signature0,
+                bitmap0,
+                sig_and_bitmap1,
+            } => {
+                for msg in msgs {
+                    let rank = msg.rank as usize;
+                    if MAXIMUM_VALIDATORS <= rank {
```
qkniep: nit: this would be the more intuitive ordering imo
```suggestion
                    if rank >= MAXIMUM_VALIDATORS {
```

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
-                    .get_or_insert(BitVec::repeat(false, MAXIMUM_VALIDATORS))
-                    .set(rank, true);
+            Self::NotarFallback {
+                signature,
+                bitmap0,
+                bitmap1,
+            } => {
+                for msg in msgs {
+                    let rank = msg.rank as usize;
+                    if MAXIMUM_VALIDATORS <= rank {
```
qkniep: same nit: order
```suggestion
                    if rank >= MAXIMUM_VALIDATORS {
```

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
             }
-        }
 
-        Ok(self
-            .signature
-            .aggregate_with(messages.iter().map(|m| &m.signature))?)
+            Self::SingleVote { signature, bitmap } => {
+                for msg in msgs {
+                    let rank = msg.rank as usize;
+                    if MAXIMUM_VALIDATORS <= rank {
```
qkniep: same nit: order
```suggestion
                    if rank >= MAXIMUM_VALIDATORS {
```

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+                        return Err(AggregateError::ValidatorDoesNotExist(msg.rank));
+                    }
+                    let vote_type = VoteType::get_type(&msg.vote);
+                    assert_eq!(vote_type, vote_types[0]);
+                    bitmap.set(rank, true);
```
qkniep: I think we should check here whether this bit was already `true` and either (1) return an error (making it the callers responsibility to never do that), or (2) ignore the corresponding `msg` during aggregation below

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+                bitmap1,
+            } => {
+                for msg in msgs {
+                    let rank = msg.rank as usize;
+                    if MAXIMUM_VALIDATORS <= rank {
+                        return Err(AggregateError::ValidatorDoesNotExist(msg.rank));
+                    }
+                    let vote_type = VoteType::get_type(&msg.vote);
+                    if vote_type == vote_types[0] {
+                        bitmap0.set(rank, true);
```
qkniep: same here as in `SingleVote`-case

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+                sig_and_bitmap1,
+            } => {
+                for msg in msgs {
+                    let rank = msg.rank as usize;
+                    if MAXIMUM_VALIDATORS <= rank {
+                        return Err(AggregateError::ValidatorDoesNotExist(msg.rank));
+                    }
+                    let vote_type = VoteType::get_type(&msg.vote);
+                    if vote_type == vote_types[0] {
+                        bitmap0.set(rank, true);
```
qkniep: same here as in `SingleVote`-case

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+            },
+            Self::NotarFallback { .. } => Err(BuildForRewardsError::InvalidCertType),
+        }
+    }
+}
+
+/// Builder for creating [`CertificateMessage`] by using BLS signature aggregation.
+pub struct CertificateBuilder {
+    builder_type: BuilderType,
+    certificate: Certificate,
```
qkniep, line 277: not part of this PR, but the type name `Certificate` is *very* confusing to me, because it does *not* include the aggregate signature, right?

samkim-crypto: Yes initially, I was quite confused myself because I always think of certificate as the aggregate sig itself. The way the current code is structured, certificate is the essentially the message that the aggregate sig is certfiying.

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,288 @@ use {
...
+            Self::Skip {
+                mut signature0,
+                bitmap0,
+                sig_and_bitmap1,
+            } => match sig_and_bitmap1 {
+                None => build_cert_from_bitmap(certificate, signature0, bitmap0)
+                    .map_err(BuildError::Encode),
+                Some((signature1, bitmap1)) => {
+                    signature0.aggregate_with([signature1].iter())?;
+                    build_cert_from_bitmaps(certificate, signature0, bitmap0, bitmap1)
```
samkim-crypto, line 233:
```suggestion
                    build_cert_from_bitmaps(certificate, signature0, bitmap0, bitmap1).map_err(BuilderError::Encode)
```

As @qkniep suggested, it seems like it would be more consistent to add error mapping here and have `build_cert_from_bitmaps` return an `EncodeError` for consistency.

votor/src/consensus_pool/certificate_builder.rs
akhi3030: good point.  Moved the assertion below.

votor/src/consensus_pool/certificate_builder.rs
akhi3030: I have an outstanding branch where I will rename some of these types.

samkim-crypto (commented):
These are good changes. Other than the wrong assert statement `assert_eq!(vote_types.len(), 2);` on `aggregate` as @qkniep suggested, and the consistency in the error handling in `build_cert_from_bitmap` and `build_cert_from_bitmaps`, I think everything else seems fine on my side.

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,294 @@ use {
...
+pub enum AggregateError {
     #[error("BLS error: {0}")]
-    BlsError(#[from] BlsError),
-    #[error("solana-signer-store decode error: {0:?}")]
-    DecodeError(DecodeError),
-    #[error("solana-signer-store encode error: {0:?}")]
-    EncodeError(EncodeError),
+    Bls(#[from] BlsError),
     #[error("Validator does not exist for given rank: {0}")]
     ValidatorDoesNotExist(u16),
```
qkniep: nit and not part of this PR: but is this the wrong name? we don't check if there actually is a validator for that rank here, just if it exceeds the max, so maybe `ValidatorRankOutOfBounds` or something would be more accurate

votor/src/consensus_pool/certificate_builder.rs
```diff
@@ -16,87 +16,294 @@ use {
...
+    ValidatorAlreadyIncluded,
 }
 
-/// A builder for creating a `CertificateMessage` by efficiently aggregating BLS signatures.
-#[derive(Clone)]
-pub struct VoteCertificateBuilder {
+/// Different types of errors that can be returned from the [`CertificateBuilder::build()`] function.
+#[derive(Debug, Error, PartialEq)]
+pub enum BuildError {
+    #[error("Encoding failed: {0:?}")]
```
qkniep, line 33: nit: is there a good reason why `EncodeError` only implements `Debug`, not `Display`? I thought in Rust, generally error types implement `Display`

qkniep (dismissed):
LGTM now, just two more nits

votor/src/consensus_pool/certificate_builder.rs
akhi3030: I agree... I think I can fix it while I'm at it.

votor/src/consensus_pool/certificate_builder.rs
akhi3030: Hmm... I didn't dig into that one.  Fixing it will require making a change in the sdk repo and making a new release, etc.

PR #546, opened 2025-10-17
For AlpenglowLastVoted, count votes from self and those older than root.

wen-coding:
#### Problem
Previously we didn't count votes from self and those older than root, these caused some fake delinquent alerts.

#### Summary of Changes
- count votes from self
- count votes older than root, this means we are counting unverified votes, but we should switch to block footer certs before going to testnet, so this should be fine.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -7,7 +7,6 @@ use {
             bls_sigverify_service::BLSSigVerifyServiceError, stats::BLSSigVerifierStats,
         },
         cluster_info_vote_listener::VerifiedVoteSender,
-        consensus,
     },
     bitvec::prelude::{BitVec, Lsb0},
     crossbeam_channel::{Sender, TrySendError},


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -166,6 +168,12 @@ impl BLSSigVerifier {
                     };
 
                     // Capture votes received metrics before old messages are potentially discarded below.
+                    let slot = vote_message.vote.slot();
+                    if let Some(existing_slot) = last_voted_slots.get_mut(solana_pubkey) {
+                        *existing_slot = (*existing_slot).max(slot);
+                    } else {
+                        last_voted_slots.insert(*solana_pubkey, slot);
+                    }
```
akhi3030:
```suggestion
                    let existing_slot = last_voted_slots.entry(*solana_pubkey).or_insert(slot);
                    *existing_slot = (*existing_slot).max(slot);
```

This seems more readable and has only one lookup.

akhi3030 (approved):
seems to be getting more and more ingrained =).  

LGTM apart from one comment.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -166,6 +168,9 @@ impl BLSSigVerifier {
                     };
 
                     // Capture votes received metrics before old messages are potentially discarded below.
+                    let slot = vote_message.vote.slot();
```
AshwinSekar, line 170: should we disclude skip votes here? presumably a validator that only votes skip (due to some replay failure for example) should be considered delinquent?

wen-coding: Hmm, that's a good point, we didn't have skip votes before. Then how about SkipFallback and NotarizeFallback?

wen-coding: I think not counting skip is fine, the default DELINQUENT_VALIDATOR_SLOT_DISTANCE is 128, so about a minute of skipping everything before we send alerts. I guess this should be rare during normal circumstances. By 10 seconds we are already in cluster standstill.

So do we only count Notarize then?

AshwinSekar: I would only count notarize and finalize here. The case that I have in mind is that something has gone wrong with block ingestion/replay - in such a scenario the validator will still be casting notarize-fallback and skip* votes

wen-coding: Hmm, maybe we only need notarize? Because we will only vote finalize if we voted Notarize right. So counting Finalize only helps if somehow Notarize vote has been lost.

AshwinSekar: yeah that seems fine

wen-coding: Done.

wen-coding: I made it count Notarize and Finalize, not too much cost, and we can catch the case where Notarize is lost.

qkniep: I think we actually need to include the skip votes here! A perfectly correct node (for example in Australia) might be correctly voting skip on many blocks in a row, e.g. because the leaders for these blocks are proposing them aggressively (with the target of reaching only 60% or 80% of nodes). There is no misbehavior on the part of the node in Australia, if anything it's the leader who is misbehaving.

PR #550, opened 2025-10-21
refactor: various minor refactorings in `vote_pool.rs`

akhi3030:
#### Problem

- The `VotePool` trait defines abstracts over two types of pools but the consumer of the trait is breaking the abstraction boundary in various places.
- The trait defines two functions but the `total_stake()` function is only called on on the `SimplePool` so the abstraction provided is quite minimal.

#### Summary of Changes

I plan on doing a series of PRs to improve the code around `vote_pool.rs`.  In this PR:

- Removes the `VotePool` trait
- Renames the `VotePoolType` enum to `VotePool`
- Restricts the visibility of various types to the minimum needed.

### DIFF
--- a/votor/src/consensus_pool/vote_pool.rs
+++ b/votor/src/consensus_pool/vote_pool.rs
@@ -7,24 +7,10 @@ use {
     std::collections::{HashMap, HashSet},
 };
 
-#[derive(Debug)]
-pub(crate) struct VoteEntry {
-    pub(crate) vote_messages: Vec<VoteMessage>,
-    pub(crate) total_stake_by_key: Stake,
-}
-
-impl VoteEntry {
-    pub fn new() -> Self {
-        Self {
-            vote_messages: Vec::new(),
-            total_stake_by_key: 0,
-        }
-    }
-}
-
-pub(crate) trait VotePool {
-    fn total_stake(&self) -> Stake;
-    fn has_prev_validator_vote(&self, validator_vote_key: &Pubkey) -> bool;
+#[derive(Default)]
+struct VoteEntry {
+    vote_messages: Vec<VoteMessage>,
+    total_stake_by_key: Stake,
 }
 
 /// There are two types of vote pools:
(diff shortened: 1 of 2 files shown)


votor/src/consensus_pool/vote_pool.rs
```diff
@@ -7,51 +7,37 @@ use {
...
-    }
-}
-
-pub(crate) trait VotePool {
-    fn total_stake(&self) -> Stake;
-    fn has_prev_validator_vote(&self, validator_vote_key: &Pubkey) -> bool;
+#[derive(Default)]
+struct VoteEntry {
+    vote_messages: Vec<VoteMessage>,
+    total_stake_by_key: Stake,
```
qkniep, line 13: maybe more a comment to the original authors: what does the suffix `_by_key` mean here? why not just `total_stake`?

qkniep (commented):
good refactor imo

votor/src/consensus_pool/vote_pool.rs
qkniep: looks like this change is part of #551 anyways

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -7,51 +7,37 @@ use {
...
 }
 
 /// There are two types of vote pools:
 /// - SimpleVotePool: Tracks all votes of a specfic vote type made by validators for some slot N, but only one vote per block.
 /// - DuplicateBlockVotePool: Tracks all votes of a specfic vote type made by validators for some slot N,
 ///   but allows votes for different blocks by the same validator. Only relevant for VotePool's that are of type
 ///   Notarization or NotarizationFallback
-pub(crate) enum VotePoolType {
+pub(super) enum VotePool {
```
wen-coding, line 21: I don't have strong feelings about this, and I agree that `vote_pools: BTreeMap<PoolId, VotePool>,` sounds more natural, but I think we originally named it `VotePoolType` because `VotePoolType::SimpleVotePool` sounds a bit more natural.

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -7,51 +7,37 @@ use {
...
+pub(super) struct SimpleVotePool {
     /// Tracks all votes of a specfic vote type made by validators for some slot N.
-    pub(crate) vote_entry: VoteEntry,
+    vote_entry: VoteEntry,
     prev_voted_validators: HashSet<Pubkey>,
 }
 
 impl SimpleVotePool {
-    pub fn new() -> Self {
+    pub(super) fn new() -> Self {
```
wen-coding, line 33: We can derive Default on this one as well?

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -162,13 +144,13 @@ impl DuplicateBlockVotePool {
             .get(validator_vote_key)
             .is_some_and(|vs| vs.contains(block_id))
     }
-}
 
-impl VotePool for DuplicateBlockVotePool {
-    fn total_stake(&self) -> Stake {
+    #[cfg(test)]
+    pub(super) fn total_stake(&self) -> Stake {
```
wen-coding, line 149: Hmm, does total_stake ever make sense for DuplicateBlockVotePool? If not, we can remove total_stake() from trait and fix tests as necessary.

votor/src/consensus_pool/vote_pool.rs
akhi3030: Done in #551

votor/src/consensus_pool/vote_pool.rs
akhi3030: Done in #551.

PR #551, opened 2025-10-21
refactor: more `vote_pool.rs` cleanups

akhi3030:
#### Problem

See #550 

#### Summary of Changes

builds on top of #550.

- vote pools do not need to know about the certificate builder
- renames some variables in simple pool and uses API to reduce amount of code
- `prev_voted_block_ids` in duplicate pool uses a hash set instead of vec to make lookups cheaper
- duplicate pool's tracking of total stake is not needed by production code
- various APIs were taking a vote message and fields derived from it, passing just vote message

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -165,21 +165,21 @@ impl ConsensusPool {
 
     fn update_vote_pool(
         &mut self,
-        msg: VoteMessage,
+        vote: VoteMessage,
         validator_vote_key: Pubkey,
         validator_stake: Stake,
     ) -> Option<Stake> {
-        let vote_type = msg.vote.get_type();
+        let vote_type = vote.vote.get_type();
         let pool = self
             .vote_pools
-            .entry((msg.vote.slot(), vote_type))
+            .entry((vote.vote.slot(), vote_type))
             .or_insert_with(|| Self::new_vote_pool(vote_type));
         match pool {
             VotePool::SimpleVotePool(pool) => {
-                pool.add_vote(validator_vote_key, validator_stake, msg)
+                pool.add_vote(validator_vote_key, validator_stake, vote)
             }
             VotePool::DuplicateBlockVotePool(pool) => {
-                pool.add_vote(validator_vote_key, msg, validator_stake)
+                pool.add_vote(validator_vote_key, validator_stake, vote)
             }
         }
     }
@@ -229,11 +229,11 @@ impl ConsensusPool {
                 if let Some(pool) = self.vote_pools.get(&(slot, *vote_type)) {
                     match pool {
                         VotePool::SimpleVotePool(pool) => {
-                            cert_builder.aggregate(pool.messages()).unwrap();
+                            cert_builder.aggregate(pool.votes()).unwrap();
                         }
                         VotePool::DuplicateBlockVotePool(pool) => {
-                            if let Some(msgs) = pool.messages(block_id.as_ref().unwrap()) {
-                                cert_builder.aggregate(msgs).unwrap();
+                            if let Some(votes) = pool.votes(block_id.as_ref().unwrap()) {
+                                cert_builder.aggregate(votes).unwrap();
                             }
                         }
                     };
--- a/votor/src/consensus_pool/vote_pool.rs
+++ b/votor/src/consensus_pool/vote_pool.rs
@@ -3,7 +3,7 @@ use {
     solana_hash::Hash,
     solana_pubkey::Pubkey,
     solana_votor_messages::consensus_message::VoteMessage,
-    std::collections::{HashMap, HashSet},
+    std::collections::{BTreeMap, BTreeSet},
 };
 
 /// There are two types of vote pools:
@@ -18,9 +18,9 @@ pub(super) enum VotePool {
 
 #[derive(Default)]
 pub(super) struct SimpleVotePool {
-    msgs: Vec<VoteMessage>,
+    votes: Vec<VoteMessage>,
     total_stake: Stake,
-    prev_voted_validators: HashSet<Pubkey>,
+    prev_voted_validators: BTreeSet<Pubkey>,
 }
 
 impl SimpleVotePool {
@@ -28,18 +28,18 @@ impl SimpleVotePool {
         &mut self,
         validator_vote_key: Pubkey,
         validator_stake: Stake,
-        msg: VoteMessage,
+        vote: VoteMessage,
     ) -> Option<Stake> {
         if !self.prev_voted_validators.insert(validator_vote_key) {
             return None;
         }
-        self.msgs.push(msg);
+        self.votes.push(vote);
         self.total_stake = self.total_stake.saturating_add(validator_stake);
         Some(self.total_stake)
     }
 
-    pub(super) fn messages(&self) -> &[VoteMessage] {
-        &self.msgs
+    pub(super) fn votes(&self) -> &[VoteMessage] {
+        &self.votes
     }
 
     pub(super) fn total_stake(&self) -> Stake {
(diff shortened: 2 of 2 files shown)


votor/src/consensus_pool/vote_pool.rs
```diff
@@ -23,101 +16,87 @@ pub(super) enum VotePool {
     DuplicateBlockVotePool(DuplicateBlockVotePool),
 }
 
+#[derive(Default)]
 pub(super) struct SimpleVotePool {
-    /// Tracks all votes of a specfic vote type made by validators for some slot N.
-    vote_entry: VoteEntry,
+    msgs: Vec<VoteMessage>,
```
qkniep: call this `votes` instead? it's the same problematic naming that we had with certificates before, where the `XYZMessage` is the `XYZ` together with the signature, whereas in cryptography a "message" is always what the signature is over

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -23,101 +16,87 @@ pub(super) enum VotePool {
...
-        }
-    }
-
     pub(super) fn add_vote(
         &mut self,
-        validator_vote_key: &Pubkey,
+        validator_vote_key: Pubkey,
         validator_stake: Stake,
-        vote_message: VoteMessage,
+        msg: VoteMessage,
```
qkniep: same, this should be a `vote`

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -23,101 +16,87 @@ pub(super) enum VotePool {
...
+    vote_messages: Vec<VoteMessage>,
+    total_stake_by_key: Stake,
+}
+
 pub(super) struct DuplicateBlockVotePool {
     max_entries_per_pubkey: usize,
     votes: HashMap<Hash, VoteEntry>,
-    total_stake: Stake,
-    prev_voted_block_ids: HashMap<Pubkey, Vec<Hash>>,
+    prev_voted_block_ids: HashMap<Pubkey, HashSet<Hash>>,
```
qkniep: wouldn't it be more straightforward to instead have a `HashSet<Pubkey>` within the `VoteEntry`?

qkniep: Also, for performance, this and `votes` would maybe better be a `BTreeSet`, because `len()` is small?

votor/src/consensus_pool.rs
```diff
@@ -145,33 +145,28 @@ impl ConsensusPool {
...
-            .entry((slot, vote_type))
+            .entry((msg.vote.slot(), vote_type))
             .or_insert_with(|| Self::new_vote_pool(vote_type));
         match pool {
             VotePool::SimpleVotePool(pool) => {
-                pool.add_vote(validator_vote_key, validator_stake, vote_message)
+                pool.add_vote(validator_vote_key, validator_stake, msg)
+            }
+            VotePool::DuplicateBlockVotePool(pool) => {
+                pool.add_vote(validator_vote_key, msg, validator_stake)
```
qkniep: different order of arguments between the two `add_vote()` methods

votor/src/consensus_pool/vote_pool.rs
akhi3030: makes sense.

votor/src/consensus_pool/vote_pool.rs
akhi3030: > wouldn't it be more straightforward to instead have a `HashSet<Pubkey>` within the `VoteEntry`?

Do you mean:

```
struct VoteEntry {
    vote_messages: Vec<VoteMessage>,
    total_stake_by_key: Stake,
    prev_voted_block_ids: BTreeSet<Pubkey>,
}

pub(super) struct DuplicateBlockVotePool {
    max_entries_per_pubkey: usize,
    votes: BTreeMap<Hash, VoteEntry>,
}
```

I don't see how this works.  The existing way allows us to look up how many `Hash` a validator has voted for.  This would tell us how many validators have voted for a given `Hash`.  

Or maybe I misunderstood your suggestion?

qkniep: nvm, what you are saying makes sense, when I wrote this I missed that we care about the *number* of hashes

qkniep: even more so then, these `HashMap`s are all incredibly small (at most 3 elements) so hashing is a huge overhead, a simple `Vec` would probably be better for performance

akhi3030: I've updated the containers to be `BTree*` in the PR now.  I don't think we need to fret too much about it.  I have a branch on top of this PR where I am removing this pool entirely.

qkniep (approved):

votor/src/consensus_pool.rs
```diff
@@ -229,24 +224,22 @@ impl ConsensusPool {
             if accumulated_stake as f64 / (total_stake as f64) < limit {
                 continue;
             }
-            let mut certificate_builder = CertificateBuilder::new(cert_type);
+            let mut cert_builder = CertificateBuilder::new(cert_type);
             vote_types.iter().for_each(|vote_type| {
-                if let Some(vote_pool) = self.vote_pools.get(&(slot, *vote_type)) {
-                    match vote_pool {
+                if let Some(pool) = self.vote_pools.get(&(slot, *vote_type)) {
+                    match pool {
```
wen-coding, line 230: I don't have strong feelings about this, but I do think `vote_pool` here is better, because we are inside ConsensusPool, which is also a pool.

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -23,116 +16,100 @@ pub(super) enum VotePool {
...
+        self.total_stake = self.total_stake.saturating_add(validator_stake);
+        Some(self.total_stake)
     }
 
-    pub(super) fn add_to_certificate(&self, output: &mut CertificateBuilder) {
-        output
-            .aggregate(&self.vote_entry.vote_messages)
-            .expect("Incoming vote message signatures are assumed to be valid")
+    pub(super) fn votes(&self) -> &[VoteMessage] {
+        &self.votes
```
wen-coding, line 42: I intentionally didn't do this because I was thinking that maybe we will add more logic here, like intentionally bypassing some slashed parties.

If it's a simple vec forever then what you have works.

Next time maybe let's talk before making all these decisions.

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -23,116 +16,100 @@ pub(super) enum VotePool {
...
+    total_stake_by_key: Stake,
+}
+
 pub(super) struct DuplicateBlockVotePool {
     max_entries_per_pubkey: usize,
-    votes: HashMap<Hash, VoteEntry>,
-    total_stake: Stake,
-    prev_voted_block_ids: HashMap<Pubkey, Vec<Hash>>,
+    vote_entries: BTreeMap<Hash, VoteEntry>,
+    prev_voted_block_ids: BTreeMap<Pubkey, BTreeSet<Hash>>,
```
wen-coding, line 63: I believe we will have up to 3 Hashes here per the protocol, why is BTreeSet better than Vec on 3 elements? (And under most cases this should be 1-element array I think)

votor/src/consensus_pool/vote_pool.rs
akhi3030: OK... it's only a few lines of code and we evolve it as needed.

votor/src/consensus_pool/vote_pool.rs
akhi3030: my thinking is that for a small collection, the performance on a btreeset is comparable to vec and btreeset guarantees at the type level that we do not have duplicates.  It also makes it easier for readers of the code to understand what to expect without inspecting all the call sites to see that we indeed only storing unique elements.

votor/src/consensus_pool.rs
akhi3030: changed in https://github.com/anza-xyz/alpenglow/pull/567.  PTAL.

PR #554, opened 2025-10-21
refactor: remove unnecessary accessor functions

akhi3030:
The different types of vote structs are fairly simple data structures, we can simply access the fields directly and do not need to implement accessor functions on them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9489,6 +9489,7 @@ dependencies = [
  "log",
  "qualifier_attr",
  "rand 0.8.5",
+ "shuttle",
  "solana-clock",
  "solana-entry",
  "solana-hash",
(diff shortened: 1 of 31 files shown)


votor-messages/src/vote.rs
```diff
@@ -48,52 +48,54 @@ pub enum VoteType {
...
     }
 
     /// Create a new genesis vote
     pub fn new_genesis_vote(slot: Slot, block_id: Hash) -> Self {
-        Self::from(GenesisVote::new(slot, block_id))
+        Self::from(GenesisVote { slot, block_id })
     }
+}
 
+impl Vote {
```
AshwinSekar: is there a reason to split up the impl here?

AshwinSekar (approved):

votor-messages/src/vote.rs
akhi3030: whoops that was an accident.  Now removed.

PR #558, opened 2025-10-21
migration: go dark on OC RPC during migrationary period

AshwinSekar:
Split from #502 

#### Problem
Although migrationary period blocks will not contain user transactions, we would like to not report confirmation during this time as it will be confusing to users - and we intend to reuse the same slot #s post migration

#### Summary of Changes
For any slot > migration slot, do not report OC notifications. We don't need to reenable this post genesis as Votor doesn't have a notion of OC and instead we set this confirmation level via finalized slots. 

No more TowerBFT votes so this thread in CIVL will just idle until it's deleted in 4.1.

### DIFF
--- a/core/src/cluster_info_vote_listener.rs
+++ b/core/src/cluster_info_vote_listener.rs
@@ -546,8 +546,11 @@ impl ClusterInfoVoteListener {
                     new_optimistic_confirmed_slots.push((slot, hash));
                     // Notify subscribers about new optimistic confirmation
                     if let Some(sender) = bank_notification_sender {
-                        // We do not report OC during the migration period. Post genesis, "OC" is faked by votor
-                        if migration_status.migration_slot().is_none_or(|ms| slot < ms) {
+                        // We do not report OC during the Alpenglow migration period. Post Alpenglow genesis, "OC" is faked by votor
+                        if migration_status
+                            .migration_slot()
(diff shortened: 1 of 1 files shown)


bw-solana:
> until it's deleted in 4.1.

love the optimism

core/src/cluster_info_vote_listener.rs
```diff
@@ -538,19 +546,22 @@ impl ClusterInfoVoteListener {
...
-                        sender
-                            .sender
-                            .send((
-                                BankNotification::OptimisticallyConfirmed(slot),
-                                dependency_work,
-                            ))
-                            .unwrap_or_else(|err| {
-                                warn!("bank_notification_sender failed: {err:?}")
-                            });
+                        // We do not report OC during the migration period. Post genesis, "OC" is faked by votor
```
bw-solana: Might want to specify *alpenglow* migration period here

core/src/cluster_info_vote_listener.rs
```diff
@@ -538,19 +546,22 @@ impl ClusterInfoVoteListener {
...
+                        if migration_status.migration_slot().is_none_or(|ms| slot < ms) {
```
bw-solana: ms is a rough acronym here

bw-solana (commented):
couple nits

PR #559, opened 2025-10-21
migration: only send FirstShred and Block for Alpenglow slots

AshwinSekar:
Split from #502 

#### Problem
We indiscriminately send the `FirstShred` and `Block` events for all slots currently. Pre migration & in TowerBFT clusters we are throwing these events away in replay stage.

#### Summary of Changes
Instead only send these events for slots past the alpenglow genesis and remove the hack in replay

### DIFF
--- a/votor/src/consensus_pool_service.rs
+++ b/votor/src/consensus_pool_service.rs
@@ -222,6 +222,7 @@ impl ConsensusPoolService {
             // - If we startup post alpenglow migration - kick off with root block
             if !kick_off_parent_ready && ctx.migration_status.is_alpenglow_enabled() {
                 let genesis_block = ctx.migration_status.genesis_block();
+                let root_bank = ctx.sharable_banks.root();
                 // can expect once we have block id in snapshots (SIMD-0333)
                 let root_block = (root_bank.slot(), root_bank.block_id().unwrap_or_default());
                 let kick_off_block @ (kick_off_slot, _) = genesis_block.max(root_block);


turbine/src/retransmit_stage.rs
```diff
@@ -901,14 +909,18 @@ fn notify_subscribers(
...
     }
-    votor_event_sender
-        .send(VotorEvent::FirstShred(slot))
-        .map_err(|err| {
-            warn!(
-                "Sending {:?} failed as channel became disconnected.  Ignoring.",
-                err.into_inner()
-            );
-        })
+    if slot >= migration_status.genesis_slot().unwrap_or(Slot::MAX) {
```
AshwinSekar: @akhi3030 just wanted to confirm if this was safe:

If the genesis slot ends up being 102, votor logic will send out `VotorEvent::ProduceWindow` for the mini window of just 103 and we will produce the first AG block as 103. 
However do we need a `FirstShred` event on 100 to ensure we don't fast crash timeout on 103 and end up voting skip?

akhi3030: https://github.com/anza-xyz/alpenglow/blob/cf32f91770c4dae6e086f5ae5d092d49f4c200d6/votor/src/event_handler.rs#L341.

Yes, when we receive the `TimeoutCrashedLeader` event, if we haven't seen any shreds for that slot, we will call `try_skip_window()` and try to skip the whole window.

So this would only work if the genesis slot is guaranteed to also be the first slot in the leader window.  Could we not make it a requirement that the genesis slot is the first slot in a leader window?

AshwinSekar: > Could we not make it a requirement that the genesis slot is the first slot in a leader window?

We setup the migration slot to maximize the chance that this is the case, however no guarantees unfortunately.

All good, i'll fake the `FirstShred` event for the genesis window

qkniep: don't we have the same problem with the regular timeout as well? (not just the crashed leader timeout)

AshwinSekar: I took a deeper look and actually it looks like we don't have any issues.
`TimerState` already handles the window not starting at a multiple of 4:
https://github.com/anza-xyz/alpenglow/blob/c3d497595c41079710b3ca36601f33a1cc142afb/votor/src/timer_manager/timers.rs#L38-L39

So in the example above, our `window` will just be `[103]` and we don't need to worry about `FirstShred` on `100`. This also means there won't be a timeout event for `100` to `102`.

akhi3030: yes, indeed, I didn't want to make any assumptions about window sizes.

core/src/banking_simulation.rs
```diff
@@ -833,6 +833,7 @@ impl BankingSimulator {
             sender,
             None,
             completed_block_sender,
+            migration_status.clone(),
```
bw-solana: nit: probably don't need to clone here

core/src/replay_stage.rs
```diff
@@ -873,10 +873,6 @@ impl ReplayStage {
 
                 let forks_root = bank_forks.read().unwrap().root();
                 let start_leader_time = if !migration_status.is_alpenglow_enabled() {
-                    // TODO(ashwin): This will be moved to the event coordinator once we figure out
-                    // migration
-                    for _ in votor_event_receiver.try_iter() {}
```
bw-solana, line 878: maybe we should add a debug assert here than this receiver is empty?

votor/src/consensus_pool_service.rs
```diff
@@ -224,13 +224,14 @@ impl ConsensusPoolService {
                 let genesis_block = ctx.migration_status.genesis_block();
                 // can expect once we have block id in snapshots (SIMD-0333)
                 let root_block = (root_bank.slot(), root_bank.block_id().unwrap_or_default());
```
bw-solana, line 227: is this just a snapshot of the root bank? Or does this update dynamically?

core/src/banking_simulation.rs
AshwinSekar: done [06ce068](https://github.com/anza-xyz/alpenglow/pull/559/commits/06ce068775660886b6a9c309b700848c720c23a7)

core/src/replay_stage.rs
AshwinSekar: yeah doesn't hurt, done

votor/src/consensus_pool_service.rs
AshwinSekar: nice catch updated it [46db152](https://github.com/anza-xyz/alpenglow/pull/559/commits/46db152bde664427046883a2784ec782c27b867f)

AshwinSekar: a TLDR on this logic, we need a parent ready to kick off everything:
- Alpenglow is enabled from startup -> use the root bank 
- We had to run the migration -> use the genesis from migration
- Weird case where we restart right after migration succeeded but havent' rooted anything yet -> use genesis

this simplifies into just max(root_bank, genesis)

PR #560, opened 2025-10-21
migration: turn off popular & ancestor hashes repair requests

AshwinSekar:
Split from #502 

#### Problem
Although we do not engage in TowerBFT duplicate block consensus post AG genesis, we still respond to TowerBFT duplicate block consensus repair requests.

#### Summary of Changes
Do not respond to any TowerBFT duplicate block repair requests on slots post AG genesis.
Do not take action on popular pruned forks post AG genesis.
Just kill the ancestor hashes thread since we're not going to be sending any requests when alpenglow is active.
Store duplicate shred proofs in blockstore for slashing, but do not gossip or send them to the duplicate block state machine

4.1 we can delete all tower bft duplicate block consensus

### DIFF
--- a/core/src/repair/ancestor_hashes_service.rs
+++ b/core/src/repair/ancestor_hashes_service.rs
@@ -32,6 +32,7 @@ use {
     solana_runtime::bank::Bank,
     solana_streamer::streamer::{self, PacketBatchReceiver, StreamerReceiveStats},
     solana_time_utils::timestamp,
+    solana_votor_messages::migration::MigrationStatus,
     std::{
         collections::HashSet,
         io::{Cursor, Read},
@@ -157,6 +158,7 @@ impl AncestorHashesService {
         ancestor_hashes_request_socket: Arc<UdpSocket>,
         ancestor_hashes_channels: AncestorHashesChannels,
         repair_info: RepairInfo,
+        migration_status: Arc<MigrationStatus>,
     ) -> Self {
         let outstanding_requests = Arc::<RwLock<OutstandingAncestorHashesRepairs>>::default();
         let (response_sender, response_receiver) = unbounded();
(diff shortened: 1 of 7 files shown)


core/src/repair/serve_repair.rs
```diff
@@ -575,11 +580,15 @@ impl ServeRepair {
...
-                        self.repair_handler
-                            .run_ancestor_hashes(recycler, from_addr, *slot, *nonce),
-                        "AncestorHashes",
-                    )
+                    if *slot > self.migration_status.genesis_slot().unwrap_or(Slot::MAX) {
+                        (None, "AncestorHashes")
+                    } else {
+                        (
+                            self.repair_handler
+                                .run_ancestor_hashes(recycler, from_addr, *slot, *nonce),
```
AshwinSekar, line 588: it's important that we still respond to ancestor hashes requests pre genesis, as it might be important to help stragglers join us

core/src/repair/ancestor_hashes_service.rs
```diff
@@ -239,6 +243,7 @@ impl AncestorHashesService {
     }
 
     /// Listen for responses to our ancestors hashes repair requests
+    #[allow(clippy::too_many_arguments)]
```
bw-solana, line 246: naughty

core/src/repair/ancestor_hashes_service.rs
```diff
@@ -591,6 +597,7 @@ impl AncestorHashesService {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
```
bw-solana, line 600: double homicide

AshwinSekar: shh `rm core/src/repair/ancestor_hashes_service.rs` coming soon

PR #561, opened 2025-10-21
migration: turn off epoch slots post alpenglow genesis

AshwinSekar:
Split from #502 

#### Problem
It's finally time 

#### Summary of Changes
Note this DOES NOT break repair, epoch-slots is only one component of repair peer selection:
https://github.com/anza-xyz/alpenglow/blob/426b9d075290838e94b6eaaeb24aeac1aeb3e428/core/src/cluster_slots_service/cluster_slots.rs#L489

necromancy of https://github.com/anza-xyz/agave/pull/3246 will clean up peer selection

### DIFF
--- a/core/src/cluster_slots_service.rs
+++ b/core/src/cluster_slots_service.rs
@@ -8,6 +8,7 @@ use {
     solana_ledger::blockstore::Blockstore,
     solana_measure::measure::Measure,
     solana_runtime::bank_forks::BankForks,
+    solana_votor_messages::migration::MigrationStatus,
     std::{
         sync::{
             atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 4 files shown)


core/src/cluster_slots_service.rs
```diff
@@ -79,16 +82,25 @@ impl ClusterSlotsService {
...
         let mut epoch_specs = EpochSpecs::from(bank_forks.clone());
+        let mut alpenglow_genesis_epoch = None;
 
         // initialize cluster slots with the current root bank
         let root_bank = bank_forks.read().unwrap().root_bank();
         cluster_slots.update(&root_bank, &cluster_info);
 
-        while !exit.load(Ordering::Relaxed) {
+        while !exit.load(Ordering::Relaxed)
+            && alpenglow_genesis_epoch.is_none_or(|ge| ge >= epoch_specs.current_epoch())
```
AshwinSekar, line 97: after we stop sending in `replay_stage.rs` this thread doesn't process any new slots. 
However we still want to propagate epoch-slots for the TowerBFT slots as it might help stragglers resolve a duplicate block situation and catchup to us.

There's not really a science to when we can kill this thread, but after observing a root in the new epoch seems good enough.

AshwinSekar: Note this also kills the propagation of the `LowestSlot` message, but afaict this gossip message is deprecated and repair works fine without it.

PR #562, opened 2025-10-21
migration: drop genesis votes & certs in sigverify post migration

AshwinSekar:
Split from #502 

#### Problem
The new vote/cert type `Genesis` is only used pre migration to determine the genesis slot.
AG votes should not be sent during TowerBFT slots

#### Summary of Changes
Once the migration succeeds, ignore any `Genesis` votes or certs in sigverify
Pre migration, ignore all AG votes in sigverify.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -26,6 +26,7 @@ use {
     solana_votor::consensus_metrics::{ConsensusMetricsEvent, ConsensusMetricsEventSender},
     solana_votor_messages::{
         consensus_message::{Certificate, CertificateType, ConsensusMessage, VoteMessage},
+        migration::MigrationStatus,
         vote::Vote,
     },
     std::{
@@ -92,6 +93,7 @@ pub struct BLSSigVerifier {
     consensus_metrics_sender: ConsensusMetricsEventSender,
     last_checked_root_slot: Slot,
     alpenglow_last_voted: Arc<AlpenglowLastVoted>,
+    migration_status: Arc<MigrationStatus>,
 }
 
 impl BLSSigVerifier {
@@ -120,6 +122,8 @@ impl BLSSigVerifier {
                 .retain(|vote, _| vote.slot() > root_bank.slot());
         }
 
+        let genesis_slot = self.migration_status.genesis_slot().unwrap_or(Slot::MAX);
+
         for mut packet in batches.iter_mut().flatten() {
             self.stats.received.fetch_add(1, Ordering::Relaxed);
             if packet.meta().discard() {
@@ -142,6 +146,24 @@ impl BLSSigVerifier {
 
             match message {
                 ConsensusMessage::Vote(vote_message) => {
+                    let slot = vote_message.vote.slot();
+
+                    // Only allow genesis votes before the alpenglow genesis
+                    if vote_message.vote.is_genesis_vote() && slot > genesis_slot {
+                        self.stats
+                            .received_invalid_genesis_vote
+                            .fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
+                    // Only allow regular alpenglow votes past genesis
+                    if !vote_message.vote.is_genesis_vote() && slot <= genesis_slot {
+                        self.stats
+                            .received_pre_migration_message
+                            .fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
                     // Missing epoch states
                     let Some(key_to_rank_map) =
                         get_key_to_rank_map(&root_bank, vote_message.vote.slot())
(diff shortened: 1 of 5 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -120,6 +122,8 @@ impl BLSSigVerifier {
                 .retain(|vote, _| vote.slot() > root_bank.slot());
         }
 
+        let genesis_slot = self.migration_status.genesis_slot().unwrap_or(Slot::MAX);
+
```
wen-coding: Would this slot ever change?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -120,6 +122,8 @@ impl BLSSigVerifier {
                 .retain(|vote, _| vote.slot() > root_bank.slot());
         }
 
+        let genesis_slot = self.migration_status.genesis_slot().unwrap_or(Slot::MAX);
```
wen-coding, line 125: nit: I don't know whether you have it elsewhere, but we should probably specify which slot `genesis_slot` refers to in the SIMD (Is it `G`)?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -142,6 +146,24 @@ impl BLSSigVerifier {
 
             match message {
                 ConsensusMessage::Vote(vote_message) => {
+                    let slot = vote_message.vote.slot();
+
+                    // Only allow genesis votes before the alpenglow genesis
+                    if vote_message.vote.is_genesis_vote() && slot > genesis_slot {
```
wen-coding, line 152: Is it possible that you got the genesis_slot too late (it was `None` for you but `Some(slot)` for others) so you are already discarding some of the valid genesis votes?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -142,6 +146,24 @@ impl BLSSigVerifier {
...
+                        self.stats
+                            .received_invalid_genesis_vote
+                            .fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
+                    // Only allow regular alpenglow votes past genesis
+                    if !vote_message.vote.is_genesis_vote() && slot <= genesis_slot {
```
wen-coding, line 160: It is possible we miss a few votes because genesis_slot changed between line 125 and here?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -1478,6 +1521,7 @@ mod tests {
             message_sender,
             consensus_metrics_sender,
             Arc::new(AlpenglowLastVoted::default()),
+            Arc::new(MigrationStatus::post_migration_status()),
```
wen-coding, line 1524: Should we maybe add one unittest to make sure we are dropping packets correctly pre and post migration?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -142,6 +146,24 @@ impl BLSSigVerifier {
...
+                            .received_invalid_genesis_vote
+                            .fetch_add(1, Ordering::Relaxed);
+                        continue;
+                    }
+
+                    // Only allow regular alpenglow votes past genesis
+                    if !vote_message.vote.is_genesis_vote() && slot <= genesis_slot {
+                        self.stats
+                            .received_pre_migration_message
+                            .fetch_add(1, Ordering::Relaxed);
```
wen-coding, line 163: nit: in case the migration test failed, should we record the slot as well here? We can remove the slot once we are confident the migration always succeeds. Same for certs.

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar: yeah it's `G` there's some documentation in migration.rs

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar: If it's `None` for you, you'll accept all genesis votes as `genesis_slot` is set to `Slot::MAX`

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar: yeah will do

core/src/bls_sigverify/bls_sigverifier.rs
AshwinSekar: yeah that's a good point, more broadly if your node is slow you could observe the genesis slot late and the cluster will have already started sending alpenglow votes.

The idea was to not waste sigverify resources if a malicious guy sends alpenglow votes way too early, however we don't want to miss any valid votes. I think either say: 
- it's okay to always sigverify, if someone is spamming us the quic control logic will time them out anyway
- or add a buffer - we start accepting alpenglow votes X slots before the migration starts

let me know what you think

qkniep: I think either of the two would make sense, just to be clear worst case that should happen if this fails should be that we trigger Alpenglow standstill

wen-coding: At this point we have the list of staked validators in the Epoch. I think we should treat attacking now same as attacking later, just blacklist for the whole Epoch.

Verifying valid packets is not that bad. How would you enable migration? You would have a feature flag saying we start Alpenglow migration at next Epoch right? (Even though the actual migration slot is not start of the Epoch.)

I think we can just start verifying the packets once we are in that Epoch.

wen-coding: I mean, let's just always sigverify in the new epoch, don't care about the slots.

AshwinSekar: thanks for the feedback i agree. i'll resubmit this with the changes later

PR #563, opened 2025-10-21
migration: enable alpenglow in replay once migration succeds

AshwinSekar:
Split from #502 

#### Problem
We don't actually do anything once we receive the genesis certificate on our super OC ancestor.

#### Summary of Changes
Enable alpenglow when this happens:
- Kill any in progress leader blocks
- Clear all banks past genesis
- Clear all shreds past genesis
- Shutdown poh

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -1372,10 +1372,9 @@ impl ReplayStage {
     /// - Clears any in progress leader blocks
     /// - Clears any TowerBFT blocks past the genesis block
     /// - Shutdown poh
-    /// - Start block creation loop
+    /// - Start block creation loop and Votor
     ///
     /// Should only be called if we have received a genesis certificate on our view of the genesis block.
-    /// A false return value means that something has gone seriously wrong.
     #[allow(clippy::too_many_arguments)]
     fn enable_alpenglow(
         exit: &AtomicBool,
@@ -1389,28 +1388,27 @@ impl ReplayStage {
         ancestors: &mut HashMap<Slot, HashSet<Slot>>,
         descendants: &mut HashMap<Slot, HashSet<Slot>>,
         progress: &mut ProgressMap,
-    ) -> bool {
+    ) {
         let root_bank = bank_forks.read().unwrap().root_bank();
-        warn!("{my_pubkey}: Alpenglow genesis vote has succeeded enabling alpenglow");
 
         let genesis_block @ (genesis_slot, block_id) = migration_status.genesis_block();
-        warn!("{my_pubkey}: Alpenglow genesis block {genesis_block:?}");
+        warn!(
+            "{my_pubkey}: Alpenglow genesis vote has succeeded enabling alpenglow. Genesis block \
+             {genesis_block:?}"
+        );
 
-        let Some(genesis_bank) = bank_forks.read().unwrap().get(genesis_slot) else {
-            error!(
-                "{my_pubkey}: Attempting to enable alpenglow before receiving the genesis block"
-            );
-            return false;
-        };
+        let genesis_bank = bank_forks.read().unwrap().get(genesis_slot).expect(
+            "{my_pubkey}: Attempting to enable alpenglow before receiving the genesis block",
+        );
+        assert!(genesis_bank.is_frozen());
 
         if genesis_bank.block_id() != Some(block_id) {
-            error!(
+            panic!(
                 "{my_pubkey}: Attempting to enable alpenglow but we have the wrong version of the \
                  genesis block our version: ({genesis_slot}, {:?}), certified version \
                  ({genesis_slot}, {block_id})",
                 genesis_bank.block_id()
             );
-            return false;
         }
 
         // Reset poh to the genesis block. This has to be done first before we can clear banks to
@@ -1436,7 +1434,7 @@ impl ReplayStage {
             .collect();
         for slot in slots_to_purge.into_iter() {
             warn!("{my_pubkey}: Purging poh block in slot {slot}");
-            Self::purge_unconfirmed_duplicate_slot(
+            Self::purge_unconfirmed_slot(
                 slot,
                 ancestors,
                 descendants,
@@ -1466,7 +1464,6 @@ impl ReplayStage {
             "migration-complete",
             ("genesis_slot", genesis_slot as i64, i64),
         );
-        true
     }
 
     /// Loads the tower from `tower_storage` with identity `node_pubkey`.
@@ -1831,7 +1828,7 @@ impl ReplayStage {
                         );
                     }
 
-                    Self::purge_unconfirmed_duplicate_slot(
+                    Self::purge_unconfirmed_slot(
                         *duplicate_slot,
                         ancestors,
                         descendants,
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -871,6 +873,23 @@ impl ReplayStage {
                 }
                 replay_active_banks_time.stop();
 
+                // Check if we've completed the migration conditions
+                if migration_status.is_ready_to_enable() {
```
AshwinSekar, line 877: So cluster_info_vote_listener will do discovery of the super OC block in a follow up PR and set the genesis_block
Consensus Pool listens for the genesis certificate and sets the genesis certificate https://github.com/anza-xyz/alpenglow/blob/c3d497595c41079710b3ca36601f33a1cc142afb/votor/src/consensus_pool.rs#L346

When they match we move to ready to enable https://github.com/anza-xyz/alpenglow/blob/c3d497595c41079710b3ca36601f33a1cc142afb/votor-messages/src/migration.rs#L467-L468

AshwinSekar: Although other threads are doing the discovery, it's important that Replay is the thread that actually enables alpenglow, otherwise we might have weird cases where we are in the middle of replaying or producing a block.

core/src/replay_stage.rs
```diff
@@ -1352,6 +1371,107 @@ impl ReplayStage {
...
+                ancestors,
+                descendants,
+                progress,
+                root_bank.as_ref(),
+                bank_forks,
+                blockstore,
+            );
+        }
+
+        // Purge any partial shreds greater than the genesis slot
```
AshwinSekar, line 1448: Note here, this is best effort. If someone was in progress of producing a TowerBFT block, we could end up ingesting shreds from it after we've purged here.

In testing on the 50 node cluster I haven't been able to get such a desync to happen but it is possible. We would end up relying on Alpenglow's repair mechanisms if that were to happen.

The other alternative is to bump the shred version, but this becomes very messy on a live cluster as it also partitions gossip.

core/src/replay_stage.rs
```diff
@@ -1352,6 +1371,107 @@ impl ReplayStage {
...
     }
 
+    /// Enables alpenglow
+    /// - Clears any in progress leader blocks
+    /// - Clears any TowerBFT blocks past the genesis block
+    /// - Shutdown poh
+    /// - Start block creation loop
+    ///
+    /// Should only be called if we have received a genesis certificate on our view of the genesis block.
+    /// A false return value means that something has gone seriously wrong.
```
bw-solana: Should we return `Err` and let replay loop handle or panic?

core/src/replay_stage.rs
```diff
@@ -1352,6 +1371,107 @@ impl ReplayStage {
...
+        leader_schedule_cache: &LeaderScheduleCache,
+        ancestors: &mut HashMap<Slot, HashSet<Slot>>,
+        descendants: &mut HashMap<Slot, HashSet<Slot>>,
+        progress: &mut ProgressMap,
+    ) -> bool {
+        let root_bank = bank_forks.read().unwrap().root_bank();
+        warn!("{my_pubkey}: Alpenglow genesis vote has succeeded enabling alpenglow");
+
+        let genesis_block @ (genesis_slot, block_id) = migration_status.genesis_block();
+        warn!("{my_pubkey}: Alpenglow genesis block {genesis_block:?}");
```
bw-solana: maybe combine these into one line/log?

core/src/replay_stage.rs
```diff
@@ -1352,6 +1371,107 @@ impl ReplayStage {
...
+
+        let Some(genesis_bank) = bank_forks.read().unwrap().get(genesis_slot) else {
```
bw-solana: another option is we could just `expect` on all of these (and let testing inform of us if any of these can happen and if/how we might recover)

core/src/replay_stage.rs
```diff
@@ -1352,6 +1371,107 @@ impl ReplayStage {
...
+        let slots_to_purge: Vec<Slot> = bank_forks
+            .read()
+            .unwrap()
+            .banks()
+            .iter()
+            .filter_map(|(slot, _)| (*slot > genesis_slot).then_some(*slot))
+            .collect();
+        for slot in slots_to_purge.into_iter() {
+            warn!("{my_pubkey}: Purging poh block in slot {slot}");
+            Self::purge_unconfirmed_duplicate_slot(
```
bw-solana: I wonder if we should rename `fn purge_unconfirmed_duplicate_slot` now that we're leveraging it for a very different purpose

core/src/replay_stage.rs
bw-solana: My understanding is we're not changing the leader schedule at all w/ Alpenglow, so the nice thing is that a node is only punishing itself if it does anything too weird to get shreds from a Tower block in after the purge. But obviously don't want this to happen w/ honest nodes.. 

We could try and create some quiesce period here.. I was thinking about waiting for new shreds to stop getting inserted, but that seems dangerous and exploitable. Some arbitrary sleep doesn't sound great either.

Is the way to test this to hide/delay the second super confirmation block from the next leader so it keeps building?

AshwinSekar: yeah i would like to avoid any intentional delay, also doesn't handle any sort of malicious case.

This can happen entirely honestly if there's some network delay in a scenario like this:
```
4999 -> 5000 (migration slot) -> 5001 (81% votes for 5000) -> 5002 (82% votes for 5001) -> (5003)
```
5002 causes 5001 to be super-oc which means 4999 should be the genesis block.

Everyone sends their BLS vote for 4999. Some of the nodes (group A) immediately see 82%, and clear banks and shreds > 4999.

Now part of the cluster (group B) and the Leader for 5003 were slow to ingest the BLS votes for 4999. So thinking that the vote has not yet succeeded, leader continues on to build 5003. A little while later this group B sees the BLS vote succeed and clears banks and shreds > 4999.

The situation is that group A has some partial shreds for 5003, while group B has no shreds for 5003.

When Alpenglow gets to 5003 the leader (who is in group B) cleanly creates the block and group B votes notarize. Perhaps group B >= 60% and the block gets finalized. Group A must now perform Alpenglow repair to get this block.

Not the end of the world this is a pathway that needs to work anyway, just a bit interesting that we could end up testing repair out of the gate. 

> Is the way to test this to hide/delay the second super confirmation block from the next leader so it keeps building?

Yep exactly, if we can add an artificial delay on the BLS ingest of the genesis vote to create a group B we should be able to repro

bw-solana: yeah, seems reasonable that we just live with this possibility and test it explicitly

core/src/replay_stage.rs
AshwinSekar: yeah good call I just ended up expect / panicking here

core/src/replay_stage.rs
AshwinSekar: renamed it but didn't go too deep to update comments around the call chain, since we can just delete this once migration succeeds

bw-solana (approved):

PR #566, opened 2025-10-22
migration: validate that migrationary blocks are VoM

AshwinSekar:
Split from #502 

#### Problem
Blocks in the migration period should not contain user transactions. #565 handles this on the block packing side, but we need verification on the replay side.

#### Summary of Changes
Mark migration period blocks as vote only. During replay mark vote only blocks which contain user transactions as dead.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -809,6 +809,7 @@ impl ReplayStage {
                     &slot_status_notifier,
                     &mut progress,
                     &mut replay_timing,
+                    migration_status.as_ref(),
                 );
                 generate_new_bank_forks_time.stop();
 
@@ -4359,6 +4360,7 @@ impl ReplayStage {
         slot_status_notifier: &Option<SlotStatusNotifier>,
         progress: &mut ProgressMap,
         replay_timing: &mut ReplayLoopTiming,
+        migration_status: &MigrationStatus,
     ) {
         // Find the next slot that chains to the old slot
         let mut generate_new_bank_forks_read_lock =
@@ -4406,6 +4408,13 @@ impl ReplayStage {
                     parent_slot,
                     forks.root()
                 );
+                // Migration period banks are VoM
+                let options = NewBankOptions {
+                    vote_only_bank: migration_status.should_bank_be_vote_only(child_slot),
+                };
+                if options.vote_only_bank {
+                    info!("Replaying block in slot {child_slot} in VoM");
+                }
                 let child_bank = Self::new_bank_from_parent_with_notify(
                     parent_bank.clone(),
                     child_slot,
@@ -4413,7 +4422,7 @@ impl ReplayStage {
                     &leader,
                     rpc_subscriptions,
                     slot_status_notifier,
-                    NewBankOptions::default(),
+                    options,
                 );
                 blockstore_processor::set_alpenglow_ticks(&child_bank);
                 let empty: Vec<Pubkey> = vec![];
@@ -4433,7 +4442,6 @@ impl ReplayStage {
         let mut generate_new_bank_forks_write_lock =
             Measure::start("generate_new_bank_forks_write_lock");
 
-        // TODO(ksn): should we have this if-statement check?
         if !new_banks.is_empty() {
             let mut forks = bank_forks.write().unwrap();
             let root = forks.root();
@@ -4805,6 +4813,7 @@ pub(crate) mod tests {
             &None,
             &mut progress,
             &mut replay_timing,
+            &MigrationStatus::default(),
         );
         assert!(bank_forks
             .read()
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore_processor.rs
```diff
@@ -1673,14 +1676,30 @@ fn confirm_slot_entries(
...
     let mut replay_timer = Measure::start("replay_elapsed");
+    let is_vote_only_bank = bank.vote_only_bank();
     let replay_entries: Vec<_> = entries
         .into_iter()
         .zip(entry_tx_starting_indexes)
-        .map(|(entry, tx_starting_index)| ReplayEntry {
-            entry,
-            starting_index: tx_starting_index,
+        .map(|(entry, tx_starting_index)| {
+            // If bank is in vote-only mode, validate that entries contain only vote transactions
```
AshwinSekar, line 1684: for some context the reason we never validated this before was because VoM is based on a local-trigger (tower distance) so we could end up rejecting blocks from honest nodes.

I believe that this is safe, since the migration slot is known via a rooted bank. The only scenario where this could be an issue is:
- Migration happens very quickly, first vote only bank is S
- Everyone purges S and it's descendants and we remake S and further slots as alpenglow
- Our node is really slow / has network issues somehow misses the tower bft blocks, and instead sees the alpenglow version of S and its descendants.

In this scenario we'd mark the Alpenglow block as dead because it's not vote only - however we would already mark it as dead because it's missing PoH ticks. 

The only way to remedy this situation is to wait for our slow node to eventually see the genesis certificate / a finalization certificate and then we'd purge, repair, and re-replay the Alpenglow blocks as not vote only.

core/src/replay_stage.rs
```diff
@@ -4454,14 +4459,21 @@ impl ReplayStage {
                     parent_slot,
                     forks.root()
                 );
+                // Migration period banks are VoM
+                let options = NewBankOptions {
+                    vote_only_bank: child_slot >= migration_slot && genesis_slot.is_none(),
```
bw-solana: Along the lines of this comment: https://github.com/anza-xyz/alpenglow/pull/560#pullrequestreview-3377312971

I think it would be nice to have a `MigrationStatus` API that just answered this question for us.

Something like `migration_status.slot_in_tower_to_alpenglow_transition(child_slot)`

ledger/src/blockstore_processor.rs
bw-solana: For the current, Tower case, we only ever mark are our own leader slot banks as VoM, right?

So they wouldn't actually go through this filtering process IIUC, which seems fine.

ledger/src/blockstore_processor.rs
```diff
@@ -1673,14 +1676,30 @@ fn confirm_slot_entries(
...
+            if let EntryType::Transactions(ref transactions) = entry {
+                if is_vote_only_bank
+                    && transactions
+                        .iter()
+                        .any(|tx| !tx.is_simple_vote_transaction())
```
bw-solana, line 1689: thinking about performance, this seems fine. `is_vote_only_bank` always false except during the transition (in which case load is already lighter. And even then, I believe these are `RuntimeTransaction` types, so the is_simple_vote_transaction is super cheap

ledger/src/blockstore_processor.rs
AshwinSekar: yep we don't validate (and with this change won't validate) tower blocks as VoM.

Only concern is if this validation could screw up something in Alpenglow which I don't believe it can

core/src/replay_stage.rs
AshwinSekar: Agreed, i'll put up a PR improving the MigrationStatus API and then merge this after

ledger/src/blockstore_processor.rs
carllin: > Our node is really slow / has network issues somehow misses the tower bft blocks, and instead sees the alpenglow version of S and its descendants.

If we check for the migration certificate in the block header, could we avoid this?

ledger/src/blockstore_processor.rs
carllin: Also, is the check on number of transactions is bounded at this point based on shred count and compute units?

ledger/src/blockstore_processor.rs
AshwinSekar: > If we check for the migration certificate in the block header, could we avoid this?

Yep, since we will have the genesis certificate in the BlockMarker, at this point we can get to `ReadyToEnable` and then progress with the migration.

core/src/replay_stage.rs
AshwinSekar: Updated it with the new API

ledger/src/blockstore_processor.rs
AshwinSekar: Yes we have block limits in form of CUs and maximum number of shreds amongst other things

PR #568, opened 2025-10-23
wip: refactor: `VotePool` accumulate checking for valid votes into one place

akhi3030:
#### Problem

- The checking for valid votes (i.e. not slashable or duplicate) is spread between three different files (consensus_pool.rs, vote_pool.rs, votor/src/common.rs).  And not always super straightforward to reason about.
- Some checks e.g. ensuring that a validator votes notar on only one block relies on a seemingly unrelated constant `MAX_ENTRIES_PER_PUBKEY_FOR_OTHER_TYPES` being set to 1 and not increasing.

#### Summary of Changes

- Combines simple and duplicate vote pools into one type
- handles each type of vote more explicitly and in a single place: `vote_pool::Votes::add_vote()` hopefully making it clear that we have caught all invalid behaviour.

TODO: If this approach is well received then add tests

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -92,7 +92,7 @@ fn get_key_and_stakes(
 
 pub struct ConsensusPool {
     my_pubkey: Pubkey,
-    // Vote pools to do bean counting for votes.
+    /// Vote pools to do bean counting for votes.
     vote_pools: BTreeMap<Slot, VotePool>,
     /// Completed certificates
     completed_certificates: BTreeMap<CertificateType, Arc<Certificate>>,
--- a/votor/src/consensus_pool/vote_pool.rs
+++ b/votor/src/consensus_pool/vote_pool.rs
@@ -10,7 +10,7 @@ use {
     thiserror::Error,
 };
 
-#[derive(Debug, Error, PartialEq)]
+#[derive(Debug, PartialEq, Eq, Error)]
 pub(super) enum AddVoteError {
     #[error("duplicate vote")]
     Duplicate,
@@ -29,7 +29,7 @@ struct Votes {
     finalize: BTreeMap<Pubkey, VoteMessage>,
     /// Notar votes are stored in map indexed by validator.
     notar: BTreeMap<Pubkey, VoteMessage>,
-    /// A validator can vote notar fallback on multiple blocks.
+    /// A validator can vote notar fallback on upto 3 blocks.
     ///
     /// Per validator, we store a map of which block ids the validator has voted notar fallback on.
     notar_fallback: BTreeMap<Pubkey, BTreeMap<Hash, VoteMessage>>,
@@ -72,10 +72,15 @@ impl Votes {
                     }
                     Entry::Occupied(mut e) => {
                         let map = e.get_mut();
+                        let len = map.len();
                         match map.entry(nf.block_id) {
                             Entry::Vacant(map_e) => {
-                                map_e.insert(vote);
-                                Ok(())
+                                if len == 3 {
+                                    Err(AddVoteError::Slash)
+                                } else {
+                                    map_e.insert(vote);
+                                    Ok(())
+                                }
                             }
                             Entry::Occupied(_) => Err(AddVoteError::Duplicate),
                         }
@@ -111,9 +116,8 @@ impl Votes {
                     return Err(AddVoteError::Slash);
                 }
                 if let Some(map) = self.notar_fallback.get(&voter) {
-                    if !map.is_empty() {
-                        return Err(AddVoteError::Slash);
-                    }
+                    assert!(!map.is_empty());
+                    return Err(AddVoteError::Slash);
                 }
                 match self.finalize.entry(voter) {
                     Entry::Occupied(_) => Err(AddVoteError::Duplicate),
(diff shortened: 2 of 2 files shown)


votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
 
-/// There are two types of vote pools:
-/// - SimpleVotePool: Tracks all votes of a specfic vote type made by validators for some slot N, but only one vote per block.
-/// - DuplicateBlockVotePool: Tracks all votes of a specfic vote type made by validators for some slot N,
-///   but allows votes for different blocks by the same validator. Only relevant for VotePool's that are of type
-///   Notarization or NotarizationFallback
-pub(super) enum VotePool {
-    SimpleVotePool(SimpleVotePool),
-    DuplicateBlockVotePool(DuplicateBlockVotePool),
+#[derive(Debug, Error, PartialEq)]
```
qkniep:
```suggestion
#[derive(Debug, Error, PartialEq, Eq)]
```

qkniep: nit: also, I would put `Error` at the end b/c it's not an `std` derive in this case, but not sure how this is usually done in agave

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
-    pub(super) fn total_stake(&self) -> Stake {
-        self.total_stake
-    }
-
-    pub(super) fn has_prev_validator_vote(&self, validator_vote_key: &Pubkey) -> bool {
-        self.prev_voted_validators.contains(validator_vote_key)
+    /// Get votes for the corresponding [`VoteType`] and block id.
+    ///
+    /// Depending on the type of of vote, block_id is expected to be Some().
+    fn get_votes(&self, vote_type: VoteType, block_id: Option<&Hash>) -> Vec<VoteMessage> {
```
qkniep, line 141: in the end, we seem to only use a reference to the returned `Vec`, so maybe there is a way to do this without cloning, e.g. by returning an iterator here

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
+                        let mut map = BTreeMap::new();
+                        map.insert(nf.block_id, vote);
+                        e.insert(map);
+                        Ok(())
+                    }
+                    Entry::Occupied(mut e) => {
+                        let map = e.get_mut();
+                        match map.entry(nf.block_id) {
+                            Entry::Vacant(map_e) => {
+                                map_e.insert(vote);
```
qkniep: if the length would become more than 3 here, that's also a slashable offence

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
+                        Ok(())
+                    }
+                }
+            }
+            Vote::Finalize(_) => {
+                if self.skip.contains_key(&voter) || self.skip_fallback.contains_key(&voter) {
+                    return Err(AddVoteError::Slash);
+                }
+                if let Some(map) = self.notar_fallback.get(&voter) {
+                    if !map.is_empty() {
```
qkniep: for my understanding: the case that `map` is empty should never happen, i.e. this could be an assertion as well?

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
+    skip_fallback: Stake,
+    /// Total stake that has voted finalize.
+    finalize: Stake,
+    /// Stake that has voted notar.
+    ///
+    /// Different validators may vote notar for different block ids, so this tracks stake per block id.
+    notar: BTreeMap<Hash, Stake>,
+    /// Stake that has voted notar fallback.
+    ///
+    /// Different validators may vote notar fallback for different block ids, so this tracks stake per block id.
```
qkniep: in this case even the same validator (up to 3 times)

votor/src/consensus_pool/vote_pool.rs
```diff
@@ -2,239 +2,375 @@ use {
...
-impl DuplicateBlockVotePool {
-    pub(super) fn new(max_entries_per_pubkey: usize) -> Self {
-        Self {
-            max_entries_per_pubkey,
-            vote_entries: BTreeMap::new(),
-            prev_voted_block_ids: BTreeMap::new(),
+impl Stakes {
+    /// Updates the corresponding stake after a vote has been successfully added to the pool.
+    ///
+    /// Returns the total stake of the corresponding type after the update.
```
qkniep: should maybe mention again that this is per block id for notar(-fallback)

votor/src/consensus_pool.rs
```diff
@@ -101,7 +93,7 @@ fn get_key_and_stakes(
 pub struct ConsensusPool {
     my_pubkey: Pubkey,
     // Vote pools to do bean counting for votes.
```
qkniep:
```suggestion
    /// Vote pools to do bean counting for votes.
```

qkniep (commented):
looks more robust and easier to reason about to me, added some comments

votor/src/consensus_pool/vote_pool.rs
akhi3030: I don't see a particular standard in agave but happy to do things in a recommended way where possible.

votor/src/consensus_pool/vote_pool.rs
akhi3030: good catch, fixed.

votor/src/consensus_pool/vote_pool.rs
akhi3030: makes sense.

votor/src/consensus_pool/vote_pool.rs
akhi3030: Yeah, I wanted to investigate this in a separate PR maybe.  Feeding an iterator into `CertBuilder::aggregate()` is not great API either as it currently iterates over the list multiple times.  I'll add a todo for now.

PR #570, opened 2025-10-23
feat: build and send block footer

ksn6:
Construct and disseminate the block footer. In a follow-up PR, we'll parse the footer in replay.

This PR addresses https://github.com/solana-foundation/solana-improvement-documents/pull/307.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -38,7 +38,7 @@ use {
             atomic::{AtomicBool, Ordering},
             Arc, Condvar, Mutex, RwLock,
         },
-        time::{Duration, Instant, UNIX_EPOCH},
+        time::{Duration, Instant},
     },
     thiserror::Error,
 };
@@ -294,7 +294,7 @@ fn produce_window(
 /// Afterwards:
 /// - Shutdown the record receiver
 /// - Clear any inflight records
-/// - TODO: insert the block footer
+/// - Insert the block footer
 /// - Insert the alpentick
 /// - Clear the working bank
 fn record_and_complete_block(
@@ -335,13 +335,7 @@ fn record_and_complete_block(
 
     // Construct and send the block footer
     let mut w_poh_recorder = poh_recorder.write().unwrap();
-    let block_producer_time_nanos = w_poh_recorder
-        .working_bank()
-        .unwrap()
-        .start
-        .duration_since(UNIX_EPOCH)
-        .unwrap()
-        .as_nanos() as u64;
+    let block_producer_time_nanos = w_poh_recorder.working_bank_block_producer_time_nanos();
     let footer = produce_block_footer(block_producer_time_nanos);
     w_poh_recorder.send_marker(footer)?;
 
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -41,7 +41,7 @@ use {
             atomic::{AtomicBool, AtomicU64, Ordering},
             Arc, Mutex, RwLock,
         },
-        time::{Instant, SystemTime},
+        time::{Instant, SystemTime, UNIX_EPOCH},
     },
     thiserror::Error,
 };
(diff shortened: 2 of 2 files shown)


core/src/block_creation_loop.rs
```diff
@@ -317,10 +338,12 @@ fn record_and_complete_block(
         )?;
     }
 
-    // TODO: insert block footer
+    // Construct and send the block footer
+    let footer = produce_block_footer();
```
AshwinSekar: awesome, this turned out very clean! FYI @akhi3030 you can plug in your rewards container here to populate the aggregates in the footer.

For the finalized certificate I imagine we can have something lightweight that can be updated on `VotorEvent::Finalized`. Can probably even live inside the rewards container and we can add it to the footer

core/src/block_creation_loop.rs
AshwinSekar: Can remove the TODO 🥳 
```suggestion
/// - Insert the block footer
```

ksn6: 😄

core/src/block_creation_loop.rs
```diff
@@ -106,6 +110,23 @@ enum StartLeaderError {
     ),
 }
 
+fn produce_block_footer() -> VersionedBlockMarker {
+    let block_producer_time_nanos = SystemTime::now()
```
AshwinSekar: > `block_producer_time_nanos: u64` is a nanosecond UNIX timestamp representing the time when the block producer started constructing the block. "started
constructing" is the point at which, from the perspective of the leader, all of
the consensus checks required for assuming leadership have "just passed". For
example, in Agave's pre-alpenglow implementations, this would be in
replay/maybe_start_leader. In a post-Alpenglow implementation, this would be
after receiving the proper vote/skip certificate for the previous slot.

So I think what we actually want is 
```rust
    poh_recorder_w.working_bank.unwrap().start
```

ksn6: makes sense

FYI - I changed `start` within `WorkingBank` to use `SystemTime`, since we can't turn an `Instant` into a UNIX timestamp

core/src/block_creation_loop.rs
akhi3030: perfect, this looks nice and straightforward.

core/src/block_creation_loop.rs
```diff
@@ -106,6 +110,18 @@ enum StartLeaderError {
     ),
 }
 
+fn produce_block_footer(block_producer_time_nanos: u64) -> VersionedBlockMarker {
+    let footer = BlockFooterV1 {
+        block_producer_time_nanos,
+        block_user_agent: format!("agave/{}", version!()).into_bytes(),
```
apfitzge, line 116: will there be a way to specify this via cli or something?
if people are using bindings is there a reasonable way for them to set this as something that's not agave?

ksn6: seems reasonable to do - can enable this in a follow-up PR

core/src/block_creation_loop.rs
```diff
@@ -317,10 +333,19 @@ fn record_and_complete_block(
...
 
-    // TODO: insert block footer
+    // Construct and send the block footer
+    let mut w_poh_recorder = poh_recorder.write().unwrap();
+    let block_producer_time_nanos = w_poh_recorder
+        .working_bank()
+        .unwrap()
+        .start
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
```
AshwinSekar: expect here and indicate that the operator has a misconfigured system clock

AshwinSekar: would be nice to make this a fn on poh recorder

core/src/block_creation_loop.rs
AshwinSekar: bump on this

poh/src/poh_recorder.rs
```diff
@@ -489,7 +507,12 @@ impl PohRecorder {
             datapoint_info!(
                 "leader-slot-start-to-cleared-elapsed-ms",
                 ("slot", bank.slot(), i64),
-                ("elapsed", start.elapsed().as_millis(), i64),
+                (
+                    "elapsed",
+                    // This errors if the clock drifts backwards, in which case we just return 0.
+                    start.elapsed().map(|dur| dur.as_millis()).unwrap_or(0),
```
AshwinSekar: would be nice to error! here as well. If the system clock is having problems it can screw up other parts of the client so better to make the operator aware.

core/src/block_creation_loop.rs
apfitzge: I think start was unused anyway, so this seems fine.

core/src/block_creation_loop.rs
ksn6: done, accidentally reverted the hunk before

core/src/block_creation_loop.rs

poh/src/poh_recorder.rs
ksn6: done

PR #571, opened 2025-10-24
feat: build and send block header

ksn6:

Construct and disseminate the block header. In a follow-up PR, we'll parse the header in replay.

We discuss block headers in detail here: https://github.com/solana-foundation/solana-improvement-documents/pull/337.

### DIFF
--- a/turbine/src/broadcast_stage/broadcast_fake_shreds_run.rs
+++ b/turbine/src/broadcast_stage/broadcast_fake_shreds_run.rs
@@ -4,7 +4,7 @@ use {
     solana_gossip::contact_info::ContactInfo,
     solana_hash::Hash,
     solana_keypair::Keypair,
-    solana_ledger::shred::{self, ProcessShredsStats, ReedSolomonCache, Shredder},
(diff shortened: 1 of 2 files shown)


core/src/block_creation_loop.rs
```diff
@@ -513,7 +529,16 @@ fn create_and_insert_leader_bank(slot: Slot, parent_bank: Arc<Bank>, ctx: &mut L
     ctx.poh_recorder.write().unwrap().set_bank(tpu_bank);
     ctx.record_receiver.restart(slot);
 
-    // TODO: fill in the block header
+    // Send block header marker
+    // TODO: place parent double merkle root here
```
AshwinSekar: For now you can use `tpu_bank.parent().block_id()`, this currently points to the CMR but will be updated the DM.

This will allow us to at least test parent logic with the CMR rather than using `Hash::default()`

ksn6: moving the logic over to broadcast as discussed, so that we can rely on the parent block id existing

AshwinSekar (approved):

PR #573, opened 2025-10-29
migration: update migration api to be a state machine

AshwinSekar:
#### Problem
Migration API is a bit hard to follow

#### Summary of Changes
Make it a state machine and add a ton of comments
Update callers
Use the off curve account to store the genesis certificate instead of polluting bank fields

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -253,6 +253,8 @@ fn start_loop(config: BlockCreationLoopConfig) {
         ctx.metrics.loop_count += 1;
         ctx.metrics.report(1000);
     }
+
+    info!("{my_pubkey}: Block creation loop shutting down");
 }
 
 /// Resets poh recorder
--- a/poh/src/poh_service.rs
+++ b/poh/src/poh_service.rs
@@ -111,9 +111,7 @@ impl PohService {
             .spawn(move || {
                 if migration_status.is_alpenglow_enabled() {
                     // We've started up post alpenglow migration. Don't bother starting PohService
-                    migration_status
-                        .is_poh_shutdown
-                        .store(true, Ordering::Release);
+                    info!("Post Alpenglow migration, not starting PohService");
                     return;
                 }
                 if poh_config.hashes_per_tick.is_none() {
@@ -171,10 +169,9 @@ impl PohService {
                     return;
                 }
 
-                // Notify that we have shutdown poh service, which lets the block creation loop and votor start
-                migration_status
-                    .is_poh_shutdown
-                    .store(true, Ordering::Release);
+                // Notify that we have shutdown poh service, which enables Alpenglow
+                // and lets the block creation loop and votor start
+                migration_status.poh_service_is_shutting_down();
                 info!("PohService shutdown");
             })
             .unwrap();
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -209,6 +209,11 @@ impl MigrationPhase {
     fn is_pre_feature_activation(&self) -> bool {
         matches!(self, MigrationPhase::PreFeatureActivation)
     }
+
+    /// Check if we are ready to enable
+    fn is_ready_to_enable(&self) -> bool {
+        matches!(self, MigrationPhase::ReadyToEnable { .. })
+    }
 }
 
 /// Keeps track of the current migration status
@@ -220,14 +225,12 @@ pub struct MigrationStatus {
     /// Communication with PohService
     /// Flag indicating whether we should shutdown Poh
     pub shutdown_poh: AtomicBool,
-    /// Flag indicating whether Poh has been shutdown
-    pub is_poh_shutdown: AtomicBool,
 
     /// The current phase of the migration we are in
     phase: RwLock<MigrationPhase>,
 
-    /// Used to notify threads that are waiting for alpenglow to be enabled
-    migration_wait: (Mutex<()>, Condvar),
+    /// Used to notify threads that are waiting for Poh to be shutdown and alpenglow to be enabled
+    migration_wait: (Mutex<bool>, Condvar),
 }
 
 impl Default for MigrationStatus {
@@ -257,9 +260,8 @@ impl MigrationStatus {
         Self {
             my_pubkey,
             shutdown_poh: AtomicBool::new(is_alpenglow_enabled),
-            is_poh_shutdown: AtomicBool::new(is_alpenglow_enabled),
             phase: RwLock::new(phase),
-            migration_wait: (Mutex::new(()), Condvar::new()),
+            migration_wait: (Mutex::new(is_alpenglow_enabled), Condvar::new()),
         }
     }
 
@@ -466,33 +468,47 @@ impl MigrationStatus {
         *phase = MigrationPhase::ReadyToEnable { genesis_cert: cert };
     }
 
-    /// Enable alpenglow only to be used during `ReadyToEnable`:
+    /// Enable alpenglow only to be used during `ReadyToEnable` from replay_stage:
     /// - Tell PoH to shutdown
     /// - Wait for Poh to shutdown
     /// - Notify all threads that are waiting for alpenglow to be enabled
     ///
     /// Transitions from `ReadyToEnable` to `AlpenglowEnabled`
-    pub fn enable_alpenglow(&self) {
+    pub fn enable_alpenglow(&self, exit: &AtomicBool) {
+        assert!(self.phase.read().unwrap().is_ready_to_enable());
+
+        // Tell PohService to shutdown and bump the phase to `MigrationPhase::AlpenglowEnabled`
+        self.shutdown_poh.store(true, Ordering::Release);
+        // Wait for PohService
+        self.wait_for_migration_or_exit(exit);
+
+        if exit.load(Ordering::Relaxed) {
+            warn!(
+                "{}: Validator shutdown before Alpenglow could be enabled",
+                self.my_pubkey
+            );
+            return;
+        }
+
+        warn!("{}: Alpenglow enabled!", self.my_pubkey);
+    }
+
+    /// PohService is shutting down after being asked to by replay_stage via `enable_alpenglow`.
+    ///
+    /// Transition the phase from `ReadyToEnable` to `AlpenglowEnabled`
+    pub fn poh_service_is_shutting_down(&self) {
         let MigrationPhase::ReadyToEnable { genesis_cert } = self.phase.read().unwrap().clone()
         else {
             unreachable!(
-                "{}: Programmer error, enabling alpenglow before we are ReadyToEnable",
+                "{}: Programmer error, PohService is shutting down before we are ReadyToEnable",
                 self.my_pubkey
             );
         };
 
-        self.shutdown_poh.store(true, Ordering::Release);
-        while !self.is_poh_shutdown.load(Ordering::Acquire) {
-            // Wait for PohService to shutdown poh
-            std::hint::spin_loop();
-        }
-
         *self.phase.write().unwrap() = MigrationPhase::AlpenglowEnabled { genesis_cert };
-
-        let (_lock, cvar) = &self.migration_wait;
-        cvar.notify_all();
-
-        warn!("{}: Alpenglow enabled!", self.my_pubkey);
+        let (is_alpenglow_enabled, condvar) = &self.migration_wait;
+        *is_alpenglow_enabled.lock().unwrap() = true;
+        condvar.notify_all();
     }
 
     /// Alpenglow has rooted a block in a new epoch. This indicates the migration epoch has completed.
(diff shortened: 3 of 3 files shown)


core/src/block_creation_loop.rs
```diff
@@ -197,6 +197,16 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
+    // leaders call `set_bank()`, otherwise, the old PoH
+    // tick producer will still tick in that alpenglow bank
+    // Since `wait_for_migration_or_exit` succeeded above we know that PohService has shutdown
+    // AND that replay no longer touches poh recorder. At this point BlockCreationLoop is the sole
+    // modifier of poh_recorder
+    {
+        let mut w_poh_recorder = ctx.poh_recorder.write().unwrap();
+        w_poh_recorder.enable_alpenglow();
+    }
     reset_poh_recorder(&ctx.bank_forks.read().unwrap().working_bank(), &ctx);
```
bw-solana, line 210: nit: maybe tuck this in the braces above?

core/src/cluster_info_vote_listener.rs
```diff
@@ -546,10 +546,9 @@ impl ClusterInfoVoteListener {
                     new_optimistic_confirmed_slots.push((slot, hash));
                     // Notify subscribers about new optimistic confirmation
                     if let Some(sender) = bank_notification_sender {
-                        // We do not report OC during the Alpenglow migration period. Post Alpenglow genesis, "OC" is faked by votor
                         if migration_status
-                            .migration_slot()
-                            .is_none_or(|migration_slot| slot < migration_slot)
+                            .phase()
```
bw-solana: I think we could abstract away phase for some of these calls. I.e. just call `migration_status.should_report_commitment_or_root(slot)` here

and then have something like:
```rust
impl MigrationStatus {
    pub fn should_report_commitment_or_root(&self, slot: Slot) -> bool {
        self.phase().should_report_commitment_or_root(slot)
    }
}
```rust

File: core/src/cluster_slots_service.rs [hunk 3]
```diff
@@ -87,20 +87,17 @@ impl ClusterSlotsService {
...
         // initialize cluster slots with the current root bank
         let root_bank = bank_forks.read().unwrap().root_bank();
         cluster_slots.update(&root_bank, &cluster_info);
 
         while !exit.load(Ordering::Relaxed)
-            && alpenglow_genesis_epoch.is_none_or(|ge| ge >= epoch_specs.current_epoch())
+            // Once we are in a full Alpenglow epoch, we can fully shutdown the cluster slots service.
+            // It is important to keep running while in the mixed migration epoch, as EpochSlots can
+            // still be useful for the TowerBFT slots.
+            && !matches!(migration_status.phase(), MigrationPhase::FullAlpenglowEpoch { .. })
```
bw-solana: make this an API in `MigrationStatus`

`fn in_full_alpenglow_epoch` or something

core/src/replay_stage.rs
```diff
@@ -2668,22 +2633,21 @@ impl ReplayStage {
             )?;
 
             // Check if we've rooted a bank that will tell us the migration slot
-            if migration_status.migration_slot().is_none() {
+            if matches!(
+                migration_status.phase(),
+                MigrationPhase::PreFeatureActivation
```
bw-solana: same here. Can make it an API

core/src/validator.rs
```diff
@@ -3030,6 +3033,51 @@ pub fn is_snapshot_config_valid(snapshot_config: &SnapshotConfig) -> bool {
     }
 }
 
+/// Based on the current feature flag activation and genesis certificate account in the root bank,
+/// determine which phase of the migration we are in and initialize accordingly.
+fn initialize_migration_status(my_pubkey: Pubkey, root_bank: Arc<Bank>) -> MigrationStatus {
```
bw-solana, line 3054: move to `migration.rs`?

poh/src/poh_service.rs
```diff
@@ -109,74 +109,71 @@ impl PohService {
...
+                // Migrate to alpenglow PoH
+                if !poh_exit.load(Ordering::Relaxed)
+                    // Should be set by migration status once we have progressed past the ReadyToEnable phase
+                    && migration_status.shutdown_poh.load(Ordering::Acquire)
                 {
-                    let mut w_poh_recorder = poh_recorder.write().unwrap();
-                    w_poh_recorder.enable_alpenglow();
+                    info!("Migrating poh service to alpenglow tick producer");
+                } else {
+                    poh_exit.store(true, Ordering::Relaxed);
```
bw-solana, line 168: when could we hit this case where `poh_exit` isn't already `true`?

runtime/src/bank.rs
```diff
@@ -904,9 +906,6 @@ pub struct Bank {
     /// This is used to avoid recalculating the same epoch rewards at epoch boundary.
     /// The hashmap is keyed by parent_hash.
     epoch_rewards_calculation_cache: Arc<Mutex<HashMap<Hash, Arc<PartitionedRewardsCalculation>>>>,
-
-    /// The alpenglow genesis block, and the certificate from the genesis vote
-    alpenglow_genesis: Option<(Block, Certificate)>,
```
bw-solana, line 909: 🔥 🔥

bw-solana: Note to self: this is now stored in an onchain account w/ a pubkey that is off-curve

runtime/src/bank.rs
```diff
@@ -1153,6 +1151,29 @@ impl Bank {
...
+            };
+            let cert_size = bincode::serialized_size(&cert).unwrap();
+            let lamports = Rent::default().minimum_balance(cert_size as usize);
+            let cert_account_data =
+                AccountSharedData::new_data(lamports, &cert, &system_program::ID).unwrap();
+            bank.store_account_and_update_capitalization(
+                &GENESIS_CERTIFICATE_ACCOUNT,
+                &cert_account_data,
+            );
+        }
```
bw-solana: maybe move all of this to some function in `migration.rs`? Or some sub-function in `Bank`?

votor-messages/src/migration.rs
```diff
@@ -58,239 +75,418 @@ pub const GENESIS_VOTE_THRESHOLD: f64 = 82.0 / 100.0;
...
-    /// Checks whether our view of the Alpenglow genesis block has been certified
-    pub fn is_genesis_certified(&self) -> bool {
-        let genesis_tracker_r = self.genesis_tracker.read().unwrap();
-        let Some(g_block) = genesis_tracker_r.genesis_block else {
-            return false;
+    /// Set the genesis certificate.
+    /// This should only be called with certificates that have passed signature verification
+    ///
+    /// Transitions to `ReadyToEnable` if we have already received a genesis block and it matches.
+    pub fn set_genesis_certificate(&self, cert: Arc<Certificate>) {
```
bw-solana, line 437: should we be writing the Genesis Account here?

core/src/validator.rs
AshwinSekar: it's tough because I want votor-messages to be as minimal dependency as possible so trying to not have it depend on `solana-runtime` (for `Bank`).

I can break it up though where we load the genesis cert from the account + check the feature flag here and then pass it into migration.rs?

votor-messages/src/migration.rs
AshwinSekar: I'm assuming you're talking about the off-curve account?
We don't actually have a bank to write to at this point, the flow I envision is:
- `MigrationPhase::Migration`, we detect a genesis block, vote & construct a genesis certificate
- We call `set_genesis_block` and `set_genesis_certificate`, these match so we go into `MigrationPhase::ReadyToEnable`
- Replay sees `MigrationPhase::ReadyToEnable`, calls `enable_alpenglow` and purges all the TowerBFT blocks > genesis
- `BlockCreationLoop` now starts - the first Alpenglow leader puts the genesis certificate into a `BlockMarker` on this block descended from genesis
- In replay when processing this `BlockMarker`, we insert the certificate to the Genesis Account

lmk if that makes sense

core/src/validator.rs
bw-solana: ah, yeah, makes sense. I'm not going to be too picky about it given this should get ripped out from the code in less than a year anyways

votor-messages/src/migration.rs
bw-solana: cool, makes sense to me

core/src/block_creation_loop.rs
AshwinSekar: `reset_poh_recorder` ends up write locking the poh recorder so we need to drop the lock via the braces

poh/src/poh_service.rs
AshwinSekar: The only case (annoyingly) is in tests that use the `short_lived_low_power_tick_producer`. Both of the other tick producers will only exit if the exit flag is set or shutdown_poh is set.

Bit ugly to have this extra case because of tests, but believe it's necessary.

bw-solana: I figured it would be something dumb like that

runtime/src/bank.rs
AshwinSekar: Actually realized I can do this even simpler since genesis config lets you specify initial accounts https://github.com/anza-xyz/alpenglow/pull/573/commits/5021a14f02f0d0a2429b423dde24e1db0148e3df
Avoids having to touch bank here at all

core/src/block_creation_loop.rs
```diff
@@ -197,6 +197,16 @@ fn start_loop(config: BlockCreationLoopConfig) {
     info!("{my_pubkey}: Block creation loop starting");
 
     // Setup poh
+    // Important this is called *before* any new alpenglow
+    // leaders call `set_bank()`, otherwise, the old PoH
+    // tick producer will still tick in that alpenglow bank
+    // Since `wait_for_migration_or_exit` succeeded above we know that PohService has shutdown
+    // AND that replay no longer touches poh recorder. At this point BlockCreationLoop is the sole
+    // modifier of poh_recorder
```
carllin, line 205: 👍

votor-messages/src/migration.rs
```diff
@@ -58,239 +76,490 @@ pub const GENESIS_VOTE_THRESHOLD: f64 = 82.0 / 100.0;
...
+    /// Alpenglow has rooted a block in a new epoch. This indicates the migration epoch has completed.
+    ///
+    /// Transitions from `AlpenglowEnabled` to `FullAlpenglowEpoch`
+    pub fn alpenglow_rooted_new_epoch(&self, full_alpenglow_epoch: Epoch) {
+        let mut phase = self.phase.write().unwrap();
+        let MigrationPhase::AlpenglowEnabled { genesis_cert } = &*phase else {
+            unreachable!(
+                "{}: Programmer error, Alpenglow rooted a block before it was enabled",
+                self.my_pubkey
+            );
```
carllin, line 523: it's impossible for a lagging validator to start in tower in epoch 1, miss seeing the transition block, and then jump into seeing an alpenglow finalized block in epoch 2 right?

 They would have to repair the migration block

votor-messages/src/migration.rs
```diff
@@ -58,239 +76,490 @@ pub const GENESIS_VOTE_THRESHOLD: f64 = 82.0 / 100.0;
...
-
-        warn!("{}: Shutting down poh", self.my_pubkey);
+        };
 
-        self.shutdown_poh.store(true, Ordering::Relaxed);
-        while !self.block_creation_loop_started.load(Ordering::Relaxed) {
-            // Wait for PohService to shutdown poh and start the block creation loop
+        self.shutdown_poh.store(true, Ordering::Release);
+        while !self.is_poh_shutdown.load(Ordering::Acquire) {
+            // Wait for PohService to shutdown poh
```
carllin: wonder if we should use a condvar here instead of spin loop as well

votor-messages/src/migration.rs
AshwinSekar: yeah exactly, haven't implemented the "Validator was offline during genesis certificate broadcast but sees a finalization certificate when it rejoins" logic yet, but it would involve having to repair the genesis block anyway and enable alpenglow. 

So impossible for the scenario you mentioned to happen.

votor-messages/src/migration.rs
AshwinSekar: Good point, i've removed `is_poh_shutdown` and instead moved it into the Condvar we already have https://github.com/anza-xyz/alpenglow/pull/573/commits/b2917c2493c9584c3b67631df9d93a90888b8c38

This means that PohService will actually store `true` when it's shutting down and notify all the threads, we don't need replay to wait on `is_poh_shutdown` and then notify all the threads.  Removes the extra hop and spin loop

AshwinSekar: mind updating your comment below?

carllin (approved):

AshwinSekar:
just adding some more color to carl's comment, here's    roughly what the enable model looks like


poh_service:
```rust
while !shutdown_poh::load(relaxed) { 
    run_poh_stuff() 
} 
migration_status.phase = AlpenglowEnabled
condvar.notify_all()
```

block_creation_loop & votor:

```rust
wait_for_condvar() 
start leader & vote for blocks > migration_status.genesis_block()
```

replay stage
```rust
loop {
   generate_new_bank_forks()
   replay_active_banks()
   if migration_status.phase() < AlpenglowEnabled {
     tower_bft_consensus()
     maybe_start_leader()
   }
   if migration_status.phase() == ReadyToEnable {
      poh_recorder.reset()
      clear all tower bft blocks > migration_status.genesis_block()
      shutdown_poh.store(true) && wait_for_condavr()
   }
}
 ```

AshwinSekar:
Yeah that's correct was changed in https://github.com/anza-xyz/alpenglow/pull/583

PR #574, opened 2025-10-29
block_creation_loop: startup separately from PohService

AshwinSekar:
#### Problem
Want to upstream the block creation loop, last bit of cleanup.
Currently it's a bit messy because it's started via PohService and poh can't depend on core so it's smuggled in as a closure.
Eventually we will also delete PohService.

#### Summary of Changes
Separate it out into its own thread, it waits until PohService is shutdown before starting block building.

### DIFF
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -6,7 +6,7 @@ use {
         admin_rpc_post_init::{AdminRpcRequestMetadataPostInit, KeyUpdaterType, KeyUpdaters},
         banking_stage::BankingStage,
         banking_trace::{self, BankingTracer, TraceError},
-        block_creation_loop::{BlockCreationLoop, BlockCreationLoopConfig, ReplayHighestFrozen},
+        block_creation_loop::ReplayHighestFrozen,
         cluster_info_vote_listener::VoteTracker,
         completed_data_sets_service::CompletedDataSetsService,
         consensus::{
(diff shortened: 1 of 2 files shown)


poh/src/poh_service.rs
```diff
@@ -173,24 +166,19 @@ impl PohService {
...
-                // TODO: Can essentially replace this with no ticks
-                // once we properly remove poh/entry verification in replay
                 {
                     let mut w_poh_recorder = poh_recorder.write().unwrap();
                     w_poh_recorder.enable_alpenglow();
                 }
-                info!("Starting alpenglow block creation loop");
                 migration_status
                     .block_creation_loop_started
                     .store(true, Ordering::Release);
```
bw-solana, line 180: delete this?

core/src/validator.rs
```diff
@@ -1437,8 +1436,10 @@ impl Validator {
             record_receiver,
             poh_service_message_receiver,
             migration_status.clone(),
-            block_creation_loop,
         );
+
+        let block_creation_loop = BlockCreationLoop::new(block_creation_loop_config);
```
bw-solana: should it be part of the TPU?

poh/src/poh_service.rs
AshwinSekar: This is needed for now as it's the sync point to let block_creation_loop / votor know that it can start. 
I'll rework this slightly in #573 after this PR lands.

PR #575, opened 2025-10-30
feat: enforce "one `BlockComponent` per `CompletedRange`" + replay

ksn6:
We:

- Enforce a bijection between `BlockComponent`s and `CompletedRange`s
- Introduce `BlockComponentProcessor`, allowing us to parse `BlockComponent`s and enforce consensus rules in replay

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -906,7 +906,9 @@ pub struct Bank {
     /// The hashmap is keyed by parent_hash.
     epoch_rewards_calculation_cache: Arc<Mutex<HashMap<Hash, Arc<PartitionedRewardsCalculation>>>>,
 
-    /// Block component processor for validating block headers/footers and clock bounds
+    /// Block component processor for validating block headers/footers and clock bounds. We
+    /// currently write to this during replay, as we process block components one at a time, and
+    /// read from this once replay is complete.
     pub block_component_processor: RwLock<BlockComponentProcessor>,
 }
 


local-cluster/tests/local_cluster.rs
```diff
@@ -7414,6 +7414,7 @@ fn test_alpenglow_ensure_liveness_after_second_notar_fallback_condition() {
 }
 
 #[test]
+#[ignore = "wip"]
```
ksn6: Debugging why this is not working atm.

ksn6: No longer relevant.

core/src/block_creation_loop.rs
```diff
@@ -377,14 +377,19 @@ fn record_and_complete_block(
...
     // Construct and send the block footer
     let mut w_poh_recorder = poh_recorder.write().unwrap();
+
+    // This is the "current" time
     let block_producer_time_nanos = w_poh_recorder.working_bank_block_producer_time_nanos();
-    let footer = produce_block_footer(block_producer_time_nanos);
-    w_poh_recorder.send_marker(footer)?;
 
-    // Alpentick and clear bank
+    // Update the bank's clock timestamp with the value from the block footer
```
AshwinSekar: revert the changes to this file

ledger/src/blockstore_processor.rs
```diff
@@ -1525,10 +1530,35 @@ pub fn confirm_slot(
...
+    // Process block component markers for Alpenglow slots
+    if let Some(parent_bank) = bank.parent() {
+        if let Some(first_alpenglow_slot) = bank
+            .feature_set
+            .activated_slot(&agave_feature_set::alpenglow::id())
+        {
+            if bank.parent_slot() >= first_alpenglow_slot {
+                let mut verifier = bank.block_component_verifier.write().unwrap();
+                for marker in slot_components.iter().filter_map(|bc| bc.as_marker()) {
+                    verifier.on_marker(
```
AshwinSekar: How will this work for UpdateParent? I'm thinking we should keep the ordering relative to the entries when processing

ledger/src/blockstore_processor.rs
```diff
@@ -1525,10 +1530,35 @@ pub fn confirm_slot(
         load_result
     }?;
 
+    // Process block component markers for Alpenglow slots
+    if let Some(parent_bank) = bank.parent() {
+        if let Some(first_alpenglow_slot) = bank
+            .feature_set
+            .activated_slot(&agave_feature_set::alpenglow::id())
```
AshwinSekar: you'll need to plumb the migration_status and check if is_alpenglow_enabled. Might just be easier to do in _new_from_parent only setting a verifier if it's an alpenglow bank.

AshwinSekar: need to also ensure we have parity between software with / without this change if someone sends a block component early

core/src/block_creation_loop.rs

ledger/src/blockstore_processor.rs
ksn6: `UpdateParent` won't have any processing in `block_component_verifier.rs`; all `UpdateParent`-related checks will happen on shred ingest.

> I'm thinking we should keep the ordering relative to the entries when processing

not sure I follow

AshwinSekar: ah right, that includes killing this bank if needed right?

And for the second point, maybe it doesn't matter for the current BlockMarkers e.g. header / footer, but if we had new block markers that changed some sort of bank state it seems the state change would be non-deterministic based on how the completed ranges filled in?

ksn6: > ah right, that includes killing this bank if needed right?

correct; it's unclear as to whether we'll want to have the direct invocation to kill the bank in blockstore, although blockstore would certainly initiate some sort of signal to kill the bank

> And for the second point, maybe it doesn't matter for the current BlockMarkers e.g. header / footer, but if we had new block markers that changed some sort of bank state it seems the state change would be non-deterministic based on how the completed ranges filled in?

it's case-dependent, though we could abstractly consider:

- ensuring that `BlockComponentVerifier`'s state is identical irrespective of the order in which block components are added (this property is currently true)

- placing bank-modifying logic in the `finish()` function, which purely operates on `BlockComponentVerifier`'s state and is currently invoked right before the bank is frozen

this is unless we don't want to just process the new bank-modifying block component on shred ingest

core/src/block_creation_loop.rs
AshwinSekar: still seeing changes in this file

runtime/src/block_component_verifier.rs
```diff
@@ -0,0 +1,268 @@
...
+#[derive(Default)]
+pub struct BlockComponentVerifier {
+    has_header: bool,
+    has_footer: bool,
+}
+
+impl BlockComponentVerifier {
+    pub fn new() -> Self {
+        Self::default()
+    }
```
akhi3030: we don't need a `new` when we have a default.

runtime/src/block_component_verifier.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    pub fn finish(&self) -> result::Result<(), BlockComponentVerifierError> {
```
akhi3030:
```suggestion
    pub fn finish(&self) -> Result<(), BlockComponentVerifierError> {
```

`Result` is auto included.  Do not need this here and in the rest of the module.

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> result::Result<(), BlockComponentVerifierError> {
+        let marker = match marker {
+            VersionedBlockMarker::V1(marker) | VersionedBlockMarker::Current(marker) => marker,
+        };
+
+        match marker {
+            BlockMarkerV1::BlockFooter(footer) => self.on_footer(bank, parent_bank, footer),
+            BlockMarkerV1::BlockHeader(header) => self.on_header(header),
+            // We process UpdateParent messages on shred ingest, so no callback needed here
+            BlockMarkerV1::UpdateParent(_) => Ok(()),
```
akhi3030, line 70: is there a check happening somewhere that we saw at most one update parent in the block?

ledger/src/blockstore_processor.rs
```diff
@@ -836,6 +836,11 @@ pub enum BlockstoreProcessorError {
 
     #[error("user transactions found in vote only mode bank at slot {0}")]
     UserTransactionsInVoteOnlyBank(Slot),
+
+    #[error("block component verifier error: {0}")]
+    BlockComponentVerifier(
+        #[from] solana_runtime::block_component_verifier::BlockComponentVerifierError,
```
akhi3030:
```suggestion
        #[from] BlockComponentVerifierError,
```

Why not do this and include the error in the top of the module?

runtime/src/block_component_processor.rs
ksn6: going to be doing that on shred ingest + marking slots dead there if bad things happen

ledger/src/blockstore_processor.rs
AshwinSekar: My suggestion would be to process the block marker inline, also ensures that we don't have to wait until the end to throw away a block if some marker is invalid

ledger/src/blockstore_processor.rs
ksn6: as discussed offline, block component verifier now takes in `migration_status`; we error if migration hasn't happened yet and observed a block component (excepting for `UpdateParent`, which we'll handle separately in replay)

ledger/src/blockstore_processor.rs
ksn6: not sure what you mean by "process the block marker inline" - do you mean "online?"

if so, `BlockComponentVerifier` accomplishes this via `on_marker()`, which is invoked in `blockstore_processor.rs`; i.e., at the first sight of a mistake (e.g., a double header / footer), we error out

AshwinSekar: Yeah sorry online.
Basically what i'm saying is assume you have FEC sets that look like this:
```
BlockMarker1 -> EntryBatch1 -> BlockMarker2 -> EntryBatch2
```

It seems we could have all sort of execution depending on the completed data ranges at the time.
I would expect to process these in the order that they appear in the block but:
- If I receive these FEC sets one at a time this is true
- If I receive all 4 FEC sets, the code written would process BM1 -> BM2 -> EB1 -> EB2

Again maybe this is not an issue because for the current BlockMarkers order doesn't matter, but I think it's a brittle assumption. For example I'm in the process of adding a new BlockMarker for migration that stores the Alpenglow genesis information in an onchain account - If EB1 interacts with this account then we could have divergent behavior

ksn6: nice, makes sense! just fixed this

ksn6:
for readers, to address https://github.com/anza-xyz/alpenglow/pull/575#discussion_r2504317013 easily, we'll be modifying https://github.com/solana-foundation/solana-improvement-documents/pull/337 to enforce that one `CompletedRange` exactly corresponds to one `BlockComponent`

this is equivalent to saying that the `DATA_COMPLETE_SHRED` flag on shreds serves as a delimiter between `BlockComponent`

core/src/replay_stage.rs
```diff
@@ -3681,6 +3681,34 @@ impl ReplayStage {
...
                 }
 
+                // Verify and process block components (e.g., header, footer) before freezing
+                // Only verify blocks that were replayed from blockstore (not leader blocks)
+                if !is_leader_block {
+                    if let Err(err) = bank
+                        .block_component_processor
+                        .read()
+                        .unwrap()
+                        .finish(migration_status)
```
wen-coding, line 3700: nit: I don't like the name finish, it sounds like you are trying to finish migration. But you are really checking whether the block format is consistent with migration status right?

ledger/src/blockstore.rs
```diff
@@ -4116,35 +4169,80 @@ impl Blockstore {
...
-                        BlockComponent::from_bytes_multiple(&payload)
-                            .map(|cs| {
-                                cs.into_iter()
-                                    .filter_map(|bc| bc.as_entry_batch_owned())
-                                    .flatten()
-                                    .collect_vec()
-                            })
-                            .map_err(|e| {
+                        let components =
+                            BlockComponent::from_bytes_multiple(&payload).map_err(|e| {
```
ksn6, line 4173: We can entirely remove `from_bytes_multiple` here and just go with `from_bytes`. This PR is getting large, though - will do this in a separate PR, as indicated in the below TODO.

ledger/src/blockstore.rs
```diff
@@ -4116,35 +4169,80 @@ impl Blockstore {
...
                                 BlockstoreError::InvalidShredData(Box::new(
                                     bincode::ErrorKind::Custom(format!(
-                                        "could not reconstruct entries: {e:?}"
+                                        "could not reconstruct block components: {e:?}"
                                     )),
                                 ))
-                            })
+                            })?;
+
+                        // Enforce that completed ranges have precisely one BlockComponent.
```
wen-coding, line 4181: nit: can you clarify how the caller can ensure this is true? And it should be specified in the doc as well.

ledger/src/blockstore.rs
```diff
@@ -4116,35 +4169,80 @@ impl Blockstore {
...
+                        // TODO(ksn): we can substantially clean up the parsing logic because of
+                        // this assumption.
+                        if components.len() != 1 {
+                            return Err(BlockstoreError::InvalidShredData(Box::new(
+                                bincode::ErrorKind::Custom(format!(
```
wen-coding, line 4186: nit: Is it worth introducing new error type rather than fitting into this one?

ledger/src/blockstore.rs
```diff
@@ -4116,35 +4169,80 @@ impl Blockstore {
...
             .collect()
     }
 
+    /// Fetch the components corresponding to all of the shred indices in `completed_ranges`
+    /// This function takes advantage of the fact that `completed_ranges` are both
+    /// contiguous and in sorted order. To clarify, suppose completed_ranges is as follows:
+    ///   completed_ranges = [..., (s_i..e_i), (s_i+1..e_i+1), ...]
+    /// Then, the following statements are true:
+    ///   s_i < e_i == s_i+1 < e_i+1
+    fn get_slot_components_in_block(
```
wen-coding, line 4210: nit: explain the difference from get_slot_entries_in_block maybe?

ledger/src/blockstore.rs
```diff
@@ -4116,35 +4169,80 @@ impl Blockstore {
...
+    /// Fetch the entries corresponding to all of the shred indices in `completed_ranges`
+    /// This function takes advantage of the fact that `completed_ranges` are both
+    /// contiguous and in sorted order. To clarify, suppose completed_ranges is as follows:
+    ///   completed_ranges = [..., (s_i..e_i), (s_i+1..e_i+1), ...]
+    /// Then, the following statements are true:
+    ///   s_i < e_i == s_i+1 < e_i+1
+    fn get_slot_entries_in_block(
+        &self,
+        slot: Slot,
+        completed_ranges: &CompletedRanges,
```
wen-coding, line 4228: Why are we changing CompletedRanges to &CompletedRanges here?

ledger/src/blockstore_processor.rs
```diff
@@ -837,6 +842,9 @@ pub enum BlockstoreProcessorError {
 
     #[error("user transactions found in vote only mode bank at slot {0}")]
     UserTransactionsInVoteOnlyBank(Slot),
+
+    #[error("block component verifier error: {0}")]
+    BlockComponentVerifier(#[from] BlockComponentProcessorError),
```
wen-coding: nit: is it verifier or processor?

ledger/src/blockstore_processor.rs
```diff
@@ -1506,21 +1514,66 @@ pub fn confirm_slot(
...
+    for (ix, (completed_range, component)) in completed_ranges
+        .iter()
+        .zip(slot_components.into_iter())
+        .enumerate()
+    {
+        let num_shreds = completed_range.end - completed_range.start;
+
+        match component {
+            BlockComponent::EntryBatch(entries) => {
+                let slot_full = slot_full && ix == last_entry_batch_index.unwrap();
```
wen-coding, line 1547: Why don't we unwrap when last_entry_batch_index is defined?

Also: is it always safe to unwrap?

ledger/src/blockstore.rs
```diff
@@ -3954,10 +3953,58 @@ impl Blockstore {
...
+
+        // TODO(ksn): get rid of this once we remove BlockComponent::from_bytes_multiple
+        if completed_ranges.len() != components.len() {
+            return Err(BlockstoreError::BlockComponentMisalignment(
+                slot,
+                start_index,
+            ));
+        }
+
+        Ok((components, completed_ranges, slot_meta.is_full()))
```
wen-coding, line 4005: nit: Maybe we should do the zipping inside this function, less confusion to caller.

ledger/src/blockstore_processor.rs
```diff
@@ -1506,21 +1514,66 @@ pub fn confirm_slot(
...
+    //     ingest, run through this verifying code here, and then error out when finish() is invoked
+    //     during replay, resulting in the slot being marked as dead.
+    let mut verifier = bank.block_component_processor.write().unwrap();
+
+    // Find the index of the last EntryBatch in slot_components
+    let last_entry_batch_index = slot_components
+        .iter()
+        .rev()
+        .position(|bc| matches!(bc, BlockComponent::EntryBatch(_)))
+        .map(|ix| slot_components.len() - 1 - ix);
```
wen-coding:
```suggestion
    let last_entry_batch_index = slot_components
        .iter()
        .rposition(|bc| matches!(bc, BlockComponent::EntryBatch(_)));
```

ledger/src/blockstore_processor.rs
```diff
@@ -1506,21 +1514,66 @@ pub fn confirm_slot(
...
+                    recyclers,
+                    log_messages_bytes_limit,
+                    prioritization_fee_cache,
+                    migration_status,
+                )?;
+            }
+            BlockComponent::BlockMarker(marker) => {
+                if let Some(parent_bank) = bank.parent() {
+                    verifier.on_marker(bank.clone_without_scheduler(), parent_bank, &marker)?;
+                }
```
wen-coding, line 1569: nit: add a comment we don't care about block marker for block 0 maybe?

runtime/src/bank.rs
```diff
@@ -903,6 +905,9 @@ pub struct Bank {
     /// This is used to avoid recalculating the same epoch rewards at epoch boundary.
     /// The hashmap is keyed by parent_hash.
     epoch_rewards_calculation_cache: Arc<Mutex<HashMap<Hash, Arc<PartitionedRewardsCalculation>>>>,
+
+    /// Block component processor for validating block headers/footers and clock bounds
+    pub block_component_processor: RwLock<BlockComponentProcessor>,
```
wen-coding, line 912: nit: maybe say when does write happen (or which thread writes and which reads). Would it be a problem if I call this before block components are processed?

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,374 @@
+use {
```
wen-coding, line 1: next time I would suggest putting this file into its own PR

core/src/replay_stage.rs
```diff
@@ -3686,6 +3686,34 @@ impl ReplayStage {
                     let _ = cluster_slots_update_sender.send(vec![bank_slot]);
                 }
 
+                // Verify and process block components (e.g., header, footer) before freezing
+                // Only verify blocks that were replayed from blockstore (not leader blocks)
+                if !is_leader_block {
```
wen-coding, line 3695: Is there ever a case I should pay attention to my own block_components?

For example, a standby sending out block in my name, putting in block component I don't agree with?

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,374 @@
...
+            BlockMarkerV1::BlockFooter(footer) => self.on_footer(bank, parent_bank, footer),
+            BlockMarkerV1::BlockHeader(header) => self.on_header(header),
+            // We process UpdateParent messages on shred ingest, so no callback needed here
+            BlockMarkerV1::UpdateParent(_) => Ok(()),
+        }
+    }
+
+    fn on_footer(
+        &mut self,
+        _bank: Arc<Bank>,
```
wen-coding, line 76: Hmm, this is because we will fill in the details later?

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,374 @@
...
+        _parent_bank: Arc<Bank>,
+        _footer: &VersionedBlockFooter,
+    ) -> Result<(), BlockComponentProcessorError> {
+        if self.has_footer {
+            return Err(BlockComponentProcessorError::MultipleBlockFooters);
+        }
+
```
wen-coding, line 83: Where do we plan to put the parsed content?

turbine/src/broadcast_stage.rs
```diff
@@ -170,7 +170,7 @@ impl BroadcastStageType {
                 bank_forks,
                 quic_endpoint_sender,
                 votor_event_sender.clone(),
-                BroadcastFakeShredsRun::new(0, shred_version),
+                BroadcastFakeShredsRun::new(0, shred_version, migration_status),
```
wen-coding, line 173: err, I do wish we put migration_status into another PR as well.

turbine/src/broadcast_stage/broadcast_duplicates_run.rs
```diff
@@ -107,7 +113,7 @@ impl BroadcastRun for BroadcastDuplicatesRun {
             self.prev_entry_hash = None;
             self.num_slots_broadcasted += 1;
 
-            true
+            self.migration_status.is_alpenglow_enabled()
```
wen-coding, line 116: nit: the above code seems to not be too related to send_header calculation.

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,374 @@
...
+}
+
+#[derive(Default)]
+pub struct BlockComponentProcessor {
+    has_header: bool,
+    has_footer: bool,
+}
+
+impl BlockComponentProcessor {
+    pub fn finish(
```
AshwinSekar, line 32: So rn we have to wait until the block has completed replay in order to do these checks.
A suggestion to make this more online:
- add an `on_entry_batch` which is called right before `confirm_slot_entries`
- `confirm_slot` has a MigrationStatus available for you to pass in to `on_marker` / `on_entry_batch`
- If `on_entry_batch` is called before `self.has_header`, we immediately know there's no header and can die 
- If `on_entry_batch` is called with `slot_is_full` and `!self.has_footer`, we immediately know there's no footer and can die

So `BlockComponentProcessor` processes entry batches and markers online, and we don't need to separately call finish in replay. lmk if that works

ledger/src/blockstore_processor.rs
```diff
@@ -1506,21 +1514,66 @@ pub fn confirm_slot(
...
+    //
+    // Pre-Alpenglow migration, suppose a validator receives a BlockMarker:
+    //
+    // (1) validators *incapable* of processing BlockMarkers will mark the slot as dead on shred
+    //     ingest in blockstore.
+    //
+    // (2) validators *capable* of processing BlockMarkers will store the BlockMarkers in shred
+    //     ingest, run through this verifying code here, and then error out when finish() is invoked
+    //     during replay, resulting in the slot being marked as dead.
+    let mut verifier = bank.block_component_processor.write().unwrap();
```
AshwinSekar: nit: should we call this `processor`?

runtime/src/block_component_processor.rs
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        if !self.has_header {
+            return Err(BlockComponentProcessorError::MissingBlockHeader);
+        }
+
+        Ok(())
+    }
+
+    pub fn on_marker(
```
AshwinSekar, line 56: Don't feel too strongly about this but for the transaction processor we invoke through bank rather than exposing to blockstore_processor directly, e.g:

https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/runtime/src/bank.rs#L3278-L3283

Could be nice to follow the same style for marker processing, and then blockstore_processor doesn't have to know anything about finding the parent bank or grabbing the write lock.

AshwinSekar: Makes it easier once we get to implementing the footer logic, by invoking through bank the code might look cleaner for updating accounts and bls verification.

core/src/replay_stage.rs
ksn6: given a block, `BlockComponentProcessor` accomplishes two things:

- returns an error, resulting in the block marked dead, if the block violates an invariant; sample invariants include # of block headers == 1, # of block footers == 1, etc. a follow-up PR will enforce the clock checks, e.g.

- soon, apply state changes that certain block components opine on; e.g., update account rewards, the clock sysvar, etc.

the `finish()` function reasons about the sufficient statistics gathered by the `BlockComponentProcessor` during accumulation in `confirm_slot`, and helps accomplish the two above goals. for now, all we do is to enforce that both a block header and footer are present.

right now, the specific invariants we enforce depend on `migration_status`, which is why we need to pass `migration_status` in; this being said, I think the name is reasonable, since we'll eventually be throwing `migration_status` entirely away.

ledger/src/blockstore.rs
ksn6: once this PR lands, I'll be modifying SIMD-0337 to ensure that this statement is true

in a follow-up PR, I'll be replacing the block component parsing with parsing a single component rather than a `Vec<BlockComponent>` and ensuring that the length is 1

ledger/src/blockstore.rs
ksn6: this code is going away in a follow-up PR; see https://github.com/anza-xyz/alpenglow/pull/575#discussion_r2508606384

ledger/src/blockstore.rs
ksn6: could you clarify? return type for getting slot components is `Result<Vec<BlockComponent>>` while the that of entries is `Result<Vec<Entry>>`; I think the comments elaborate a bit further on this

ledger/src/blockstore.rs
ksn6: `get_slot_components_with_shred_info` returns an owned `completed_ranges`, but also needs to invoke `get_slot_components_in_block`, so `get_slot_components_in_block` must take in `&CompletedRanges`

`get_slot_entries_in_block` invokes `get_slot_components_in_block`, so we might as well pass in a ref rather than an owned version of `CompletedRanges`, since an owned version here is more restrictive unnecessarily

ledger/src/blockstore_processor.rs
ksn6: ah, forgot to change this - processor, thx

ledger/src/blockstore_processor.rs
ksn6: `last_entry_batch_index` exists if and only if an `EntryBatch` exists in `slot_components`; as such, we cannot always invoke `.unwrap()`

however, if we've entered the match statement in line 1548, `last_entry_batch_index` must exist, since we're currently processing an entry batch

ledger/src/blockstore_processor.rs
ksn6: didn't know this existed - nice!

ledger/src/blockstore_processor.rs
ksn6: sure, done

runtime/src/block_component_processor.rs
ksn6: makes sense!

runtime/src/block_component_processor.rs
ksn6: yep, I'll be making use of the bank var here in a follow-up PR

core/src/replay_stage.rs
ksn6: no reason to do this - if a validator wishes to modify block component dissemination, they should be doing so either in block creation loop or in broadcast

could you clarify what you mean by "standby sending out block in my name?"

runtime/src/block_component_processor.rs
ksn6: wdym by the parsed content?

eventually, `on_footer` will be the place where process rewards, address the clock, etc.

turbine/src/broadcast_stage.rs
ksn6: will def do this in the future, although we're quite close to getting this PR across the finish line; would be ideal to keep the test components in for now

turbine/src/broadcast_stage/broadcast_duplicates_run.rs
ksn6: it is related, right? we only send headers if alpenglow is enabled

ledger/src/blockstore_processor.rs
ksn6: yep, done

runtime/src/bank.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: could you clarify what you mean by "the zipping inside this function?"

EDIT: nvm, I see what you mean - a lot of this code is going to become much cleaner after a bunch of small refactors. would rather we get this PR checked in asap to unblock a number of other things - I promise to make this change, though!

runtime/src/block_component_processor.rs
ksn6: this is a great suggestion; in the interest of parallelization, though, I'd like to implement this in a follow-up PR (I promise to get this out today!)

runtime/src/block_component_processor.rs
ksn6: yep, makes sense - could we do this in a follow-up PR rather than in this one?

core/src/replay_stage.rs
wen-coding: It is okay for now, but it gets tricky when we apply state changes I guess.

core/src/replay_stage.rs
wen-coding: Let's say I run two validators A and B, A is B's standby, now I'm switching identity so A becomes the active one, but it's still processing blocks produced by B, in this case do I need to process the block_components for the blocks owned by "me"?

runtime/src/block_component_processor.rs
wen-coding: I mean, the certs in blocks, in the future how do I access it?
self.bank.block_components().... ?

ledger/src/blockstore.rs
wen-coding: I mean, we now seem to almost identical comments between the two functions, would be good to mention difference, but I'm fine either way.

core/src/replay_stage.rs
```diff
@@ -3690,6 +3690,34 @@ impl ReplayStage {
...
+                            progress,
+                            duplicate_slots_to_repair,
+                            ancestor_hashes_replay_update_sender,
+                            purge_repair_slot_counter,
+                            &mut tbft_structs,
+                        );
+                        continue;
+                    }
+                }
+
```
carllin, line 3720: is there any reason not to move this check right after https://github.com/ksn6/alpenglow/blob/abb42d57cf609c64823b53ef39888e7aee001a0b/core/src/replay_stage.rs#L3606-L3639

That way all the failures are directly blocked one after another and don't have to reason about whether logic in between like https://github.com/ksn6/alpenglow/blob/abb42d57cf609c64823b53ef39888e7aee001a0b/core/src/replay_stage.rs#L3675-L3691 is necessary

Could even group the failure checks into a single function `check_block_failures`

ksn6: ashwin came up with this nice idea to entirely eliminate the finish() function; we pushed this a few days ago:

https://github.com/anza-xyz/alpenglow/pull/594

PR #578, opened 2025-10-31
feat: include bank hash in block footer

ksn6:
Alpenglow votes do not contain bank hashes. Accordingly, we add a `bank_hash` field to the block footer.

There's a bit of cleanup that we do here as well.

We additionally move block footer creation to poh recorder (see discussion in this thread).

This PR addresses https://github.com/solana-foundation/solana-improvement-documents/pull/298/.

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -1365,7 +1365,7 @@ mod tests {
     fn create_parent_ready_update() -> UpdateParentV1 {
         UpdateParentV1 {
             new_parent_slot: 42,
-            new_parent_block_id: Hash::new_from_array([42u8; 32]),
+            new_parent_block_id: Hash::new_from_array([96u8; 32]),
         }
     }
 
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -607,9 +607,10 @@ impl PohRecorder {
 
                 if is_alpentick {
                     // Wait for the bank to be frozen with timeout
+                    // TODO: change this to use DELTA_BLOCK from votor instead.
                     let start = Instant::now();
                     while !working_bank.bank.is_frozen() {
-                        if start.elapsed() > Duration::from_secs(1) {
+                        if start.elapsed() > Duration::from_millis(400) {
                             break;
                         }
                         std::hint::spin_loop();
@@ -617,11 +618,15 @@ impl PohRecorder {
 
                     // If the bank still isn't frozen, we've timed out
                     if !working_bank.bank.is_frozen() {
+                        error!(
+                            "slot = {} block production failure. bank freezing timed out.",
+                            working_bank.bank.slot()
+                        );
                         break;
                     }
 
                     // Send out the block footer
-                    let footer = self.produce_block_footer()?;
+                    let footer = self.produce_block_footer(working_bank);
                     let footer_entry_marker = (
                         EntryMarker::Marker(footer),
                         working_bank.max_tick_height - 1,
@@ -632,6 +637,10 @@ impl PohRecorder {
                         .send((working_bank.bank.clone(), footer_entry_marker));
 
                     if send_result.is_err() {
+                        error!(
+                            "slot = {} block production failure. failed to broadcast footer",
+                            working_bank.bank.slot()
+                        );
                         break;
                     }
                 }
@@ -977,12 +986,7 @@ impl PohRecorder {
             .as_nanos() as u64
     }
 
-    fn produce_block_footer(&self) -> Result<VersionedBlockMarker> {
-        let working_bank = self
-            .working_bank
-            .as_ref()
-            .ok_or(PohRecorderError::MaxHeightReached)?;
-
+    fn produce_block_footer(&self, working_bank: &WorkingBank) -> VersionedBlockMarker {
         if !working_bank.bank.is_frozen() {
             let slot = working_bank.bank.slot();
             error!("slot = {slot} creating a block footer with a non-frozen bank! ");
@@ -997,7 +1001,7 @@ impl PohRecorder {
         let footer = VersionedBlockFooter::Current(footer);
         let footer = BlockMarkerV1::BlockFooter(footer);
 
-        Ok(VersionedBlockMarker::Current(footer))
+        VersionedBlockMarker::Current(footer)
     }
 
     pub fn tick_alpenglow(&mut self, slot_max_tick_height: u64) {


entry/src/block_component.rs
```diff
@@ -272,6 +273,29 @@ pub struct BlockFooterV1 {
...
+/// ├─────────────────────────────────────────┤
+/// │ User Agent Length            (1 byte)   │
+/// ├─────────────────────────────────────────┤
+/// │ User Agent Bytes          (0-255 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Bank Hash                   (32 bytes)  │
+/// └─────────────────────────────────────────┘
+/// ```
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct BlockFooterV2 {
```
AshwinSekar: Could we just add the field to V1? Or are there any plans to release V1 on mainnet?

ksn6: SIMD-0298 specifies that we upgrade the footer to V2.

wen-coding: Hmm, I actually agree with Ashwin here, this hasn't launched yet, can we modify the SIMD? :)

ksn6: We'd have to modify the SIMD to say something along the lines of "as of the writing of this SIMD, block footers haven't launched yet, so we're going to toss the bank hash into V1."

If that's acceptable, then I'm not opposed.

If this is allowed, I'd rather we include Alpenglow certificates in BlockFooterV1 as well.

wen-coding: Let's try that. No need to increase version if the previous one has never launched.

ksn6: Done - fixed the SIMD + everything's now in `BlockFooterV1`.

core/src/block_creation_loop.rs
```diff
@@ -377,14 +379,16 @@ fn record_and_complete_block(
 
     // Construct and send the block footer
     let mut w_poh_recorder = poh_recorder.write().unwrap();
+    let bank = w_poh_recorder
+        .bank()
+        .expect("Bank cannot have been cleared as BlockCreationLoop is the only modifier");
+
+    let bank_hash = bank.hash();
```
wen-coding: nit: can you add comments here why we definitely have bank frozen here?

entry/src/block_component.rs
```diff
@@ -924,8 +931,12 @@ impl BlockFooterV1 {
 
     /// Serializes to bytes with user agent length capping.
     fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
-        let mut buffer =
-            Vec::with_capacity(8 + 1 + self.block_user_agent.len().min(Self::MAX_USER_AGENT_LEN));
+        let mut buffer = Vec::with_capacity(
+            32 + 8 + 1 + self.block_user_agent.len().min(Self::MAX_USER_AGENT_LEN),
```
wen-coding: nit: do you think we should have 32 and 8 defined as constant somewhere?

entry/src/block_component.rs
```diff
@@ -944,14 +955,18 @@ impl BlockFooterV1 {
             return Err(BlockComponentError::InsufficientData);
         }
 
-        // Read timestamp
-        // Unwrap: HEADER_SIZE = TIMESTAMP_SIZE + USER_AGENT_LEN_SIZE > TIMESTAMP_SIZE, so this will
-        // never fail.
-        let time_bytes = data[..Self::TIMESTAMP_SIZE].try_into().unwrap();
+        // Read bank hash; HEADER_SIZE > HASH_SIZE, so this unwrap will never fail.
+        let hash_bytes: [u8; 32] = data[..Self::HASH_SIZE].try_into().unwrap();
```
wen-coding: nit: I personally prefer to forget the knowledge `HEADER_SIZE > HASH_SIZE `and just do
`.map_err(|_| BlockComponentError::InsufficientData)?;`
everywhere, be dumb so code will not fail when other engineers come in and change it.

entry/src/block_component.rs
```diff
@@ -944,14 +955,18 @@ impl BlockFooterV1 {
...
+        // Read timestamp; HEADER_SIZE > HASH_SIZE + TIMESTAMP_SIZE, so this unwrap will never fail.
+        let time_bytes = data[Self::HASH_SIZE..Self::HASH_SIZE + Self::TIMESTAMP_SIZE]
+            .try_into()
+            .map_err(|_| BlockComponentError::InsufficientData)?;
         let block_producer_time_nanos = u64::from_le_bytes(time_bytes);
 
-        // Read user agent length
-        let user_agent_len = data[Self::TIMESTAMP_SIZE] as usize;
+        // Read user agent length; HEADER_SIZE > HASH_SIZE + TIMESTAMP_SIZE, so this access works.
+        let user_agent_len = data[Self::HASH_SIZE + Self::TIMESTAMP_SIZE] as usize;
```
wen-coding: Have you considered things like this?
```
const HASH_SIZE: usize = 32;
const TIMESTAMP_SIZE: usize = 8;
const USER_AGENT_LEN_OFFSET: usize = HASH_SIZE + TIMESTAMP_SIZE;

let user_agent_len = data[USER_AGENT_LEN_OFFSET] as usize;
```

We can put all offsets at the beginning of the file.

Also read a bit `vote/src/vote_state_view/field_frames.rs` to see whether any of the methods used there makes sense.

Of course, we can just do code refactoring in another PR.

entry/src/block_component.rs
```diff
@@ -3396,6 +3454,7 @@ mod tests {
         // Test BlockComponent with BlockFooter marker (various user agent sizes)
         for user_agent_len in [0, 10, 100, 255, 300] {
             let footer = BlockFooterV1 {
+                bank_hash: Hash::default(),
```
wen-coding: Hmm, if you don't add an unique bank_hash somewhere in the test, how can you make sure you are encoding and decoding correctly?

entry/src/block_component.rs

entry/src/block_component.rs
ksn6: yep, done

entry/src/block_component.rs
ksn6: Yep, the plan is to clean this all up once we rewrite large swaths of `block_component.rs` with `wincode`. I'm going to be working on this with @cpubot .

I've cleaned this up slightly in the meantime.

entry/src/block_component.rs
ksn6: ah, hadn't realized that `Hash::default()` was all zeros; updated this everywhere to be `Hash::new_unique()`

core/src/block_creation_loop.rs
AshwinSekar: we don't.
Since we're no longer doing parent bank hash but the current bank hash, this becomes tricky.
We'll have to produce the footer in broadcast when detecting the alpentick.

AshwinSekar: another alternative is to have `tick_alpenglow` hold on sending the last tick until replay has frozen the bank. At that point we send the footer + last tick to broadcast.

wen-coding: Hmm, we plan to have Double Merkle root in block footer later right? Does Double Merkle depend on block footer being finalized? Trying to figure out the dependencies we will introduce.

AshwinSekar: it's self referential to put the block id in the footer. the parent block id is in the header / UpdateParent.

wen-coding: Okay, so the dependency here is mostly:
block frozen->get hash->finalize block footer
I would slightly prefer `tick_alpenglow` holding on sending the last tick until replay has frozen the bank, so we don't put this logic into broadcast, so we can make broadcast as dumb as possible. How hard is it to do that?

AshwinSekar: shouldn't be too bad, need to wait here for the bank to be frozen, only if we're flushing cache due to alpentick
https://github.com/anza-xyz/alpenglow/blob/c3d497595c41079710b3ca36601f33a1cc142afb/poh/src/poh_recorder.rs#L603-L609

then along with sending the alpentick we send the footer

ksn6: nice catch - confirmed that the bank is in fact not frozen at this point.

I'll implement the wait + send on the `poh_recorder.rs` side.

poh/src/poh_recorder.rs
```diff
@@ -943,6 +977,24 @@ impl PohRecorder {
             .as_nanos() as u64
     }
 
+    fn produce_block_footer(&self) -> Result<VersionedBlockMarker> {
```
wen-coding: Hmm, it's a pretty big decision to move block footer production from block_creation_loop to poh_recorder (from now on everything you need for BlockFooter needs to come here).

Can we maybe at least split it into two PRs? Adding bank has in block_component as one PR.

ksn6: we can't easily modify `block_component.rs` to add in the bank hash field, unless we populate it with a dummy bank hash, which doesn't seem ideal.

I guess we could break this into:

- PR 1: move block footer, without bank hash, to poh recorder
- PR 2: include the bank hash in block component + populate this in poh recorder

the benefits of the above are minimal though (this PR maybe becomes 40 lines smaller); accordingly, I'd rather keep this as one PR.

poh/src/poh_recorder.rs
```diff
@@ -943,6 +977,24 @@ impl PohRecorder {
...
+        let working_bank = self
+            .working_bank
+            .as_ref()
+            .ok_or(PohRecorderError::MaxHeightReached)?;
+
+        let footer = BlockFooterV1 {
+            bank_hash: working_bank.bank.hash(),
```
wen-coding, line 996: nit: can you somehow check bank is frozen above and emit a warning if it isn't?

poh/src/poh_recorder.rs
wen-coding: I'm still not too happy about putting everything block footer depends on into poh_recorder. But we can check this in for now to unblock others.

entry/src/block_component.rs
```diff
@@ -1328,7 +1365,7 @@ mod tests {
     fn create_parent_ready_update() -> UpdateParentV1 {
         UpdateParentV1 {
             new_parent_slot: 42,
-            new_parent_block_id: Hash::default(),
+            new_parent_block_id: Hash::new_from_array([42u8; 32]),
```
bw-solana: might be nice to make this something that's not 42.

Just on the off chance we introduce some parsing bug, seems more likely we would catch it if value is different from `new_parent_slot`

poh/src/poh_recorder.rs
ksn6: sure, done

poh/src/poh_recorder.rs
```diff
@@ -602,6 +605,37 @@ impl PohRecorder {
...
                 working_bank.bank.register_tick(&entry.hash);
 
+                if is_alpentick {
+                    // Wait for the bank to be frozen with timeout
+                    let start = Instant::now();
+                    while !working_bank.bank.is_frozen() {
+                        if start.elapsed() > Duration::from_secs(1) {
+                            break;
+                        }
+                        std::hint::spin_loop();
```
wen-coding, line 616: Should we yield here rather than doing busy spin?

poh/src/poh_recorder.rs
```diff
@@ -602,6 +605,37 @@ impl PohRecorder {
...
+                    let footer_entry_marker = (
+                        EntryMarker::Marker(footer),
+                        working_bank.max_tick_height - 1,
+                    );
+
+                    send_result = self
+                        .working_bank_sender
+                        .send((working_bank.bank.clone(), footer_entry_marker));
+
+                    if send_result.is_err() {
```
wen-coding, line 639: Should we warn here? How about all other break above?

poh/src/poh_recorder.rs
bw-solana: Sorry, I'm still not getting it.. seems like we just need to get a handle on the bank hash.

Why can't we go poh recorder --> working bank --> bank hash in the current location (`block_creation_loop.rs`)?

poh/src/poh_recorder.rs
```diff
@@ -602,6 +605,37 @@ impl PohRecorder {
             for (entry, tick_height) in &self.tick_cache[..entry_count] {
                 working_bank.bank.register_tick(&entry.hash);
 
+                if is_alpentick {
+                    // Wait for the bank to be frozen with timeout
+                    let start = Instant::now();
+                    while !working_bank.bank.is_frozen() {
+                        if start.elapsed() > Duration::from_secs(1) {
```
wen-coding: Where does the 1 second constant come from?

poh/src/poh_recorder.rs
```diff
@@ -602,6 +605,37 @@ impl PohRecorder {
...
+                            break;
+                        }
+                        std::hint::spin_loop();
+                    }
+
+                    // If the bank still isn't frozen, we've timed out
+                    if !working_bank.bank.is_frozen() {
+                        break;
```
bw-solana, line 625: is this recoverable?

poh/src/poh_recorder.rs
```diff
@@ -602,6 +605,37 @@ impl PohRecorder {
...
+                        EntryMarker::Marker(footer),
+                        working_bank.max_tick_height - 1,
+                    );
+
+                    send_result = self
+                        .working_bank_sender
+                        .send((working_bank.bank.clone(), footer_entry_marker));
+
+                    if send_result.is_err() {
+                        break;
```
bw-solana, line 644: is this recoverable?

poh/src/poh_recorder.rs
```diff
@@ -943,6 +977,24 @@ impl PohRecorder {
             .as_nanos() as u64
     }
 
+    fn produce_block_footer(&self) -> Result<VersionedBlockMarker> {
+        let working_bank = self
+            .working_bank
+            .as_ref()
+            .ok_or(PohRecorderError::MaxHeightReached)?;
```
wen-coding: When would this happen?

entry/src/block_component.rs
ksn6: yep, done

poh/src/poh_recorder.rs
wen-coding: Hmm, you mean wait for block frozen in the block_creation_loop? I assume when we are shipping 4 blocks in a row in the window, we want block_creation_loop to start working on the next block once this one reaches time?

bw-solana: chatted offline. This is the required sequence:
poh tick --> bank complete --> bank freeze --> make bank hash --> make block footer

poh tick happens as part of alpentick process, so we need to make block footer after that, but we also want the block footer to be inserted before the last tick, so we can't create block footer after alpentick (unless we want to shuffle entries after the fact or in broadcast)

poh/src/poh_recorder.rs
ksn6: sadly no, since no footer = no way of producing a valid block, since the footer is the only place where the bank hash will be stored (since votes are no longer on-chain in Alpenglow)

poh/src/poh_recorder.rs
ksn6: ^ agreed with this

btw - one way to avoid sending rewards aggregates directly to poh recorder for footer creation would be to:

- (1) start constructing the block footer in block creation loop with everything besides the bank hash
- (2) block creation loop hands the footer to poh recorder, which then updates the bank hash prior to sending to broadcast

it's "builder pattern"-esque

bw-solana: I think long term, it would be nice to break fn tick_alpenglow into two pieces:

1. tick
2. flush

then we could do something like:

1. tick
2. wait for freeze
3. create block footer
4. flush

2 and 3 could remain in `block_creation_loop.rs`

poh/src/poh_recorder.rs
ksn6: it's a long enough constant to the point where if we've waited this long, we're definitely not producing a block; can probably shorten this to the block time - need to bench this; will add a TODO here for now

poh/src/poh_recorder.rs
ksn6: hopefully we entirely get rid of ticks haha

poh/src/poh_recorder.rs
ksn6: sadly no - at this point, we've timed out, and in practice, the leader's full window ends up getting skipped

will add logging here though so we don't silently error

poh/src/poh_recorder.rs
ksn6: would rather stick with a spin loop here rather than a yield, since the working bank will be frozen imminently

poh/src/poh_recorder.rs
ksn6: yep, done

poh/src/poh_recorder.rs
bw-solana: Log is definitely good.

But by recoverable I meant can the validator continue to function after this? Like will it be able to validate blocks and maybe even produce its next leader window?

If yes, error log and break seems good.

If not, would rather log and shut down the validator.

poh/src/poh_recorder.rs
ksn6: probably never - it's cleaner to just pass `working_bank` directly into this function to avoid these lines; done

poh/src/poh_recorder.rs
wen-coding: Do we need to worry about some poisonous transaction sneaking in which drags replay on forever here?

ksn6: > But by recoverable I meant can the validator continue to function after this? Like will it be able to validate blocks and maybe even produce its next leader window?

yes - the validator can continue to function after this; i.e., validate blocks from other leaders, produce upcoming leader windows, etc.

what I meant earlier is that this particular leader window is not recoverable by the leader; i.e., there's nothing the leader can do to produce blocks specifically for this leader window

sg - will keep the log and break, then

> Do we need to worry about some poisonous transaction sneaking in which drags replay on forever here?

not today, since solana txs currently specify a CU limit + it's the scheduler's responsibility to ensure that the algorithm used produces a reasonable sequence of txs that can run in time and fit within a block (both from a CU perspective as well as a timing perspective)

jump has a proposal to eliminate CUs entirely, at which point the scheduling problem gets a bit trickier

PR #579, opened 2025-11-02
fast leader handover: send `LeaderWindowInfo` over a channel

ksn6:
For fast leader handover, we'd like to listen to the first event that arrives from two channels in the block creation loop:

- From replay stage: a block that has completed replay
- From `votor`'s event handler: a parent ready update

Right now, `votor` communicates with block creation loop via a `condvar` in `LeaderWindowNotifier`. Instead, let's use a channel.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -81,12 +81,12 @@ pub struct BlockCreationLoopConfig {
     pub slot_status_notifier: Option<SlotStatusNotifier>,
 
     // Receivers / notifications from banking stage / replay / votor
-
-    // Channel to receive RecordReceiver from PohService
-    pub record_receiver_receiver: Receiver<RecordReceiver>,
     pub leader_window_info_receiver: Receiver<LeaderWindowInfo>,
     pub replay_highest_frozen: Arc<ReplayHighestFrozen>,
     pub highest_parent_ready: Arc<RwLock<(Slot, (Slot, Hash))>>,
+
+    // Channel to receive RecordReceiver from PohService
+    pub record_receiver_receiver: Receiver<RecordReceiver>,
 }
 
 struct LeaderContext {


votor/src/event_handler.rs
```diff
@@ -403,21 +399,7 @@ impl EventHandler {
...
-                        window_info.start_slot,
-                        window_info.end_slot,
-                        old_window_info.start_slot,
-                        old_window_info.end_slot,
-                    );
-                }
-                *l_window_info = Some(window_info);
-                ctx.leader_window_notifier.window_notification.notify_one();
+                ctx.leader_window_info_sender.send(window_info).unwrap();
             }
```
ksn6, line 403: @AshwinSekar - is this fine to do?

Rather than replacing a slow leader window, because we use a channel, we instead now wait for the slow leader window to finish.

It may be worth doing a `.send_timeout(Duration::from_secs(0))` and display some sort of error; but we'd still have to wait rather than replace the leader window.

core/src/validator.rs
```diff
@@ -1409,7 +1408,10 @@ impl Validator {
             !waited_for_supermajority && !config.no_wait_for_vote_to_start_leader;
 
         let replay_highest_frozen = Arc::new(ReplayHighestFrozen::default());
-        let leader_window_notifier = Arc::new(LeaderWindowNotifier::default());
+
+        let (leader_window_info_sender, leader_window_info_receiver) = bounded(7);
```
ksn6, line 1419: To prevent the event handler thread from blocking in the off-chance the block producer's really slow, I'm setting this to 7: assuming a validator with ~4% stake and 108k leader slots, it takes ~10k epochs to see a single leader get 7 leader slots in a row, which is what it'd take for this buffer to become full.

This being said, I'd imagine that block production timeouts are likely highly autocorrelated; if the first window times out, chances are that the second does, and so-forth - so maybe there's no real point in doing the "window clobbering" that we currently do. If so, this is an argument to set this channel to have a maximum length of 1.

votor/src/event_handler.rs
AshwinSekar: This is fine, but in block creation loop we should drain the channel if there are multiple leader windows and build the highest one.

Also since we have the preemption logic, if we're slow and in the process of building a window already, we give up in block creation loop.

ksn6: done

core/src/block_creation_loop.rs
```diff
@@ -221,23 +227,29 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
             parent_block: (parent_slot, _),
             skip_timer,
         } = {
-            let window_info = leader_window_notifier.window_info.lock().unwrap();
-            let (mut guard, timeout_res) = leader_window_notifier
-                .window_notification
-                .wait_timeout_while(window_info, Duration::from_secs(1), |wi| wi.is_none())
-                .unwrap();
-            if timeout_res.timed_out() {
+            // Drain all pending messages and keep the one with the highest start_slot
```
AshwinSekar: We're fine just taking the latest message since we know they'll come to order, might be simpler like this
```rust
let Some(window) = ctx.leader_window_info_receiver
    .recv_timeout(Duration::from_secs(1))
    .ok()
    .and_then(|window| {
        ctx.leader_window_info_receiver.try_iter().last().or(Some(window))
    }) else {
  continue;
}
```

ksn6: makes sense

core/src/block_creation_loop.rs
```diff
@@ -80,17 +81,19 @@ pub struct BlockCreationLoopConfig {
     pub slot_status_notifier: Option<SlotStatusNotifier>,
 
     // Receivers / notifications from banking stage / replay / votor
-    pub leader_window_notifier: Arc<LeaderWindowNotifier>,
-    pub replay_highest_frozen: Arc<ReplayHighestFrozen>,
 
     // Channel to receive RecordReceiver from PohService
     pub record_receiver_receiver: Receiver<RecordReceiver>,
+    pub leader_window_info_receiver: Receiver<LeaderWindowInfo>,
```
AshwinSekar, line 84: Move these back with the comment on line 83

AshwinSekar (commented):
Looks good just a minor nit

core/src/block_creation_loop.rs
ksn6: done

PR #580, opened 2025-11-03
migration: Send the GenesisVote when we discover the genesis block

AshwinSekar:
Split from #502 

#### Problem
Once cluster info vote listener discovers the genesis block (future PR) we need to send out the `GenesisVote` for it.

#### Summary of Changes
In replay if alpenglow is yet to be enabled and we have discovered a genesis block, send the vote.
Refactored `voting_utils` to be accessible from replay by passing in arguments rather than the `VotingContext`

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -86,12 +86,16 @@ use {
         event::{CompletedBlock, VotorEvent, VotorEventReceiver, VotorEventSender},
         root_utils,
         vote_history::VoteHistory,
-        vote_history_storage::VoteHistoryStorage,
+        vote_history_storage::{SavedVoteHistory, VoteHistoryStorage},
         voting_service::BLSOp,
-        voting_utils::GenerateVoteTxResult,
+        voting_utils::{self, GenerateVoteTxResult},
         votor::{LeaderWindowNotifier, Votor, VotorConfig},
     },
-    solana_votor_messages::{consensus_message::ConsensusMessage, migration::MigrationStatus},
+    solana_votor_messages::{
+        consensus_message::ConsensusMessage,
+        migration::{MigrationStatus, GENESIS_VOTE_REFRESH},
+        vote::Vote,
+    },
     std::{
         collections::{HashMap, HashSet},
         num::{NonZeroUsize, Saturating},
@@ -675,7 +679,7 @@ impl ReplayStage {
             leader_window_notifier,
             event_sender: votor_event_sender.clone(),
             event_receiver: votor_event_receiver.clone(),
-            own_vote_sender,
+            own_vote_sender: own_vote_sender.clone(),
             consensus_message_receiver,
             consensus_metrics_sender,
             consensus_metrics_receiver,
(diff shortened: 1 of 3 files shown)


core/src/replay_stage.rs
```diff
@@ -876,6 +881,21 @@ impl ReplayStage {
                     // TODO(ashwin): This will be moved to the event coordinator once we figure out
                     // migration
                     for _ in votor_event_receiver.try_iter() {}
+                    // Check if we should vote / refresh our genesis vote
+                    if last_genesis_vote_refresh_time.elapsed() > GENESIS_VOTE_REFRESH
+                        && migration_status.is_in_migration()
+                        && Self::maybe_send_genesis_vote(
```
AshwinSekar, line 887: This doesn't actually have to occur in replay_stage, it was just convenient to slot it in here.

If preferred we can do this elsewhere:
- ClusterInfoVoteListener, since that's the thread that finds the genesis block to vote on in the first place
- ConsensusPoolService, to avoid having to separately send to `own_vote_sender`, we can just insert into the pool directly
- EventHandler, this is the only thread that sends BLS votes currently, so might make sense to keep this BLS vote there as well. This involves a slight modification as EventHandler is asleep until the migration succeeds - instead it would minimally send out GenesisVotes while waiting. Another benefit here is that it has access to the VotingContext so we don't have to refactor voting_utils in this PR. 

Open to suggestions

core/src/replay_stage.rs
```diff
@@ -1352,6 +1372,59 @@ impl ReplayStage {
...
+                    identity_keypair.pubkey()
+                );
+                // If sending fails that means the channel is disconnected and we are shutting down
+                let _ = own_vote_sender.send(message.clone());
+                let _ = bls_sender.send(BLSOp::PushVote {
+                    message: Arc::new(message),
+                    slot,
+                    saved_vote_history:
+                        solana_votor::vote_history_storage::SavedVoteHistoryVersions::Current(
+                            SavedVoteHistory::default(),
```
AshwinSekar, line 1414: There's not really any point in saving this vote in vote history. If the validator dies and restarts during the migration period we don't want it to send out a GenesisVote until it's frozen the genesis block again, and there's no relevant interactions between GenesisVote and the other vote types.

core/src/replay_stage.rs
wen-coding: I would slightly prefer EventHandler, because that's the brain of the whole protocol, it makes sense to ask it to calculate when to send GenesisVote as well. But I can accept putting it into other locations as well.

How hard is it to modify EventHandler so it starts a bit early?

AshwinSekar: yeah that makes sense, it's not bad to modify it. I'll go ahead and do that

AshwinSekar: Addressed it here https://github.com/anza-xyz/alpenglow/pull/587.
Slight clarification on the original message, finding the genesis block actually happens in replay not ClusterInfoVoteListener. There was some confusion about allowing gossip votes which is why I originally had it in ClusterInfoVoteListener.

PR #581, opened 2025-11-04
fast leader handover: transmit optimistic parent to block creation loop

ksn6:
For fast leader handover, in addition to `votor` sending over a confirmed parent thanks to a parent ready update, we need replay stage to optimistically send block creation loop a parent to build on. This PR gets replay stage to send the optimistic parent when the next slot is the beginning of the validator's leader window.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -3418,6 +3418,13 @@ impl ReplayStage {
             return;
         }
 
+        // TODO(ksn): if the leader has, e.g., two consecutive leader windows, this line here
+        // prevents fast leader handover from the first leader window to the second. This is because
+        // the block id is only computed after shredding a block, which occurs in broadcast, which
+        // happens after replay. This function, on the other hand, is invoked prior to replay.
+        //
+        // To address this, we should additionally add in an optimistic parent channel that goes
+        // from broadcast to replay.
         let Some(block_id) = bank.block_id() else {
             return;
         };


core/src/replay_stage.rs
```diff
@@ -3431,6 +3439,31 @@ impl ReplayStage {
...
+
+                        let leader_window_info = LeaderWindowInfo {
+                            start_slot,
+                            end_slot,
+                            parent_block,
+                            skip_timer: Instant::now(), // can ignore
+                        };
+
+                        let _ = optimistic_parent_sender
+                            .send_timeout(leader_window_info, Duration::from_secs(1));
```
bw-solana: can you explain this one to me?

Seems like we should either just wait for the event to go through, or panic (for testing)/log when we timeout or something

core/src/replay_stage.rs
```diff
@@ -3431,6 +3439,31 @@ impl ReplayStage {
...
+                    }
+                }
+
```
bw-solana: this function (`process_replay_results`) is already such a disaster. For readability, might be nice to pull this into a sub-function like `fn maybe_notify_of_optimistic_parent`

core/src/replay_stage.rs
```diff
@@ -3431,6 +3439,31 @@ impl ReplayStage {
...
 
+                // Fast leader handover: if we're going to be the next leader, and our leader window
+                // starts on the next slot, then send the bank through this channel.
+                let is_next_leader = leader_schedule_cache
+                    .slot_leader_at(bank.slot() + 1, Some(bank))
+                    .is_some_and(|leader| &leader == my_pubkey);
+                let next_slot = bank.slot().saturating_add(1);
+
+                if is_next_leader && next_slot == first_of_consecutive_leader_slots(next_slot) {
+                    if let Some(block_id) = block_id {
```
bw-solana: I believe `block_id` is the only thing we really need from this function... otherwise we could just iterate over what is returned (frozen slots) in the upper layer and avoid passing `leader_schedule` and the `optimistic_parent_sender` down here.

Maybe we could expand what is returned here?

core/src/replay_stage.rs
AshwinSekar: Adding onto this, `replay_active_banks` returns a Vec of newly frozen slots. Think it would be better for `maybe_notify_of_optimistic_parent` to be a top level function called in the replay loop immediately after `replay_active_banks`. 

EDIT: nvm I see brennan has already pointed this out below

core/src/replay_stage.rs
ksn6: agreed - I'll log an error if we timeout within a second

I don't think we have to wait for the event to go through / panic, though, since this doesn't technically hard-block consensus, given that the parent ready channel can also supply window info

this being said - it should almost never be the case that this error is triggered, though agreed we shouldn't let it fail silently if it does happen

core/src/replay_stage.rs
ksn6: done - factored out `maybe_notify_of_optimistic_parent`; will address the top-level function part in the separate thread: https://github.com/anza-xyz/alpenglow/pull/581#discussion_r2496371499

core/src/replay_stage.rs
ksn6: nice! done

core/src/replay_stage.rs
```diff
@@ -856,6 +860,19 @@ impl ReplayStage {
                 );
                 let did_complete_bank = !new_frozen_slots.is_empty();
                 if migration_status.is_alpenglow_enabled() {
+                    let bank_forks_r = bank_forks.read().unwrap();
+                    for frozen_slot in &new_frozen_slots {
+                        let bank = bank_forks_r.get(*frozen_slot).unwrap();
+                        Self::maybe_notify_of_optimistic_parent(
+                            &bank,
+                            bank.block_id(),
```
bw-solana: could just grab this inside `maybe_notify_of_optimistic_parent` since you're passing `bank`

core/src/replay_stage.rs
```diff
@@ -3376,6 +3393,44 @@ impl ReplayStage {
...
 
+    /// Fast leader handover: if we're going to be the next leader, and our leader window
+    /// starts on the next slot, then send the bank through the optimistic parent channel.
+    fn maybe_notify_of_optimistic_parent(
+        bank: &Arc<Bank>,
+        block_id: Option<Hash>,
+        my_pubkey: &Pubkey,
+        leader_schedule_cache: &LeaderScheduleCache,
+        optimistic_parent_sender: &Sender<LeaderWindowInfo>,
+    ) {
```
AshwinSekar: would go early return here, and hit `leader_schedule_cache` last e.g:

```rust
let Some(block_id) = block_id else {
  return;
}
if next_slot != first_of_consecutive_leader_slots(next_slot) {
  return;
}
if !is_next_leader {
  return;
}

optimistic_parent_sender.send()
```

core/src/replay_stage.rs
ksn6: yep, done

core/src/replay_stage.rs
ksn6: sg - I'm rewriting the full thing with less nesting

core/src/replay_stage.rs
```diff
@@ -3377,6 +3393,47 @@ impl ReplayStage {
...
+
+        if !is_next_leader {
+            return;
+        }
+
+        if next_slot != first_of_consecutive_leader_slots(next_slot) {
+            return;
+        }
+
+        let Some(block_id) = bank.block_id() else {
```
AshwinSekar: So in relation to this and the `end_slot` below, we'll need to figure out what happens when a leader has 2 consecutive leader windows in a row.
Lets say leader is producing slot 100 - 108:

Here we notify optimistically on freezing 99 -> Optimistic(100 - 103)
In EventHandler we'll notify on ParentReady -> ProduceWindow(100 - 103).

Then once 103 is done production - this code *will not* notify Optimstic(104 - 107), as since 103 is a leader block we won't have the block id here. Instead we'll be waiting on the ProduceWindow(104 - 107).

AshwinSekar: Doesn't have to in this PR but I think what we should do is adjust `end_slot` here and when we generate `ProduceWindow` to find the actual end slot rather than just adding 4

AshwinSekar:
r+ pending fixing the conflicts

core/src/replay_stage.rs
AshwinSekar: Other option is to additionally send the Optimistic notification via broadcast to account for consecutive leader windows

AshwinSekar: Checked in with Kobi https://discord.com/channels/428295358100013066/1377667311174946816/1436382131646496939 think the second option is better, i.e.:
Do not adjust end_slot, instead also add an optimistic sender in broadcast to notify Optimistic on 103 once we have the block id - can handle in this PR or a followup

ksn6: yep, makes sense - will add in a TODO for now + address in a follow-up PR to keep this one small

PR #582, opened 2025-11-04
feat: generates rewards certs and adds them to the block footer

akhi3030:
See https://docs.google.com/document/d/1jyycl6EwP8eVqdrYj3tmFHfG5U86SPr91WXG3DvNyWQ/edit?tab=t.0 for a design doc.

The flow at a high level is the following:

- bls_sigverifier: sends relevant votes to consensus_rewards
- consensus_rewards aggregates the votes and if all the votes are seen, pre-generates the reward certs
- block creation loop requests the consensus_rewards to geneate the reward certs and then inserts them into the block footer

### DIFF
--- a/votor-messages/src/rewards_certificate.rs
+++ b/votor-messages/src/rewards_certificate.rs
@@ -1,9 +1,7 @@
 //! Defines aggregates used for vote rewards.
 
-#[cfg(feature = "dev-context-only-utils")]
-use solana_bls_signatures::keypair::Keypair as BLSKeypair;
 use {
-    solana_bls_signatures::Signature as BLSSignature,
+    solana_bls_signatures::SignatureCompressed as BLSSignatureCompressed,
     solana_clock::Slot,
     solana_hash::Hash,
     wincode::{containers::Pod, SchemaRead, SchemaWrite},
@@ -17,8 +15,8 @@ pub struct SkipRewardCertificate {
     /// The slot the certificate is for.
     pub slot: Slot,
     /// The signature
-    #[wincode(with = "Pod<BLSSignature>")]
-    pub signature: BLSSignature,
+    #[wincode(with = "Pod<BLSSignatureCompressed>")]
+    pub signature: BLSSignatureCompressed,
     /// The bitmap for validators, see solana-signer-store for encoding format
     pub bitmap: Vec<u8>,
 }
@@ -27,8 +25,7 @@ impl SkipRewardCertificate {
     /// Creates a new [`SkipRewardCertificate`] for test purposes.
     #[cfg(feature = "dev-context-only-utils")]
     pub fn new_for_tests() -> Self {
-        let bls_keypair = BLSKeypair::new();
-        let signature = bls_keypair.sign(b"hello").into();
+        let signature = BLSSignatureCompressed::default();
         Self {
             slot: 1234,
             signature,
@@ -46,8 +43,8 @@ pub struct NotarRewardCertificate {
     #[wincode(with = "Pod<Hash>")]
     pub block_id: Hash,
     /// The signature
-    #[wincode(with = "Pod<BLSSignature>")]
-    pub signature: BLSSignature,
+    #[wincode(with = "Pod<BLSSignatureCompressed>")]
+    pub signature: BLSSignatureCompressed,
     /// The bitmap for validators, see solana-signer-store for encoding format
     pub bitmap: Vec<u8>,
 }
@@ -56,8 +53,7 @@ impl NotarRewardCertificate {
     /// Creates a new [`NotarRewardCertificate`] for test purposes.
     #[cfg(feature = "dev-context-only-utils")]
     pub fn new_for_tests() -> Self {
-        let bls_keypair = BLSKeypair::new();
-        let signature = bls_keypair.sign(b"hello").into();
+        let signature = BLSSignatureCompressed::default();
         Self {
             slot: 1234,
             block_id: Hash::new_unique(),
--- a/votor/src/consensus_rewards/entry/partial_cert.rs
+++ b/votor/src/consensus_rewards/entry/partial_cert.rs
@@ -1,7 +1,10 @@
 use {
     super::{AddVoteError, BuildCertError},
     bitvec::{order::Lsb0, vec::BitVec},
-    solana_bls_signatures::{Signature as BLSSignature, SignatureProjective},
+    solana_bls_signatures::{
+        Signature as BLSSignature, SignatureCompressed as BLSSignatureCompressed,
+        SignatureProjective,
+    },
     solana_signer_store::encode_base2,
     solana_votor_messages::consensus_message::VoteMessage,
 };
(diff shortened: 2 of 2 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -172,8 +177,17 @@ impl BLSSigVerifier {
                         id: *solana_pubkey,
                         vote: vote_message.vote,
                     });
-                    // Only need votes newer than root slot
+
+                    // consensus pool does not need votes for slots other than root slot however the rewards container may still need them.
                     if vote_message.vote.slot() <= root_bank.slot() {
+                        // the only module that takes a write lock on consensus_rewards is this one and it does not take the write lock while it is verifying votes so this should never block.
```
akhi3030: feedback from Brennan: maybe we can do this inline if we can demonstrate that we have enough compute resources to handle the additional verification.  This will also require ensuring that the dedup can handle the DDoS vectors we are worried about.

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,90 @@
...
+            return false;
+        }
+        let Some(entry) = self.votes.get(&vote.vote.slot()) else {
+            return true;
+        };
+        entry.wants_vote(vote)
+    }
+
+    /// Adds received [`VoteMessage`]s from other nodes.
+    pub fn add_vote_message(&mut self, root_slot: Slot, vote: VoteMessage) {
```
akhi3030: Potential optimisation: if we look up how many validators we have for this epoch, then we can detect here if we have received votes from all the validators and pre-generate the certificates in this function instead of `build_rewards_certs`.  `build_rewards_certs` is more in the critical path than this function is.  Not sure how often we will actually hit this case though.

votor/src/consensus_rewards/entry.rs
```diff
@@ -0,0 +1,114 @@
...
+    Some((signature.into(), encode_base2(&bitmap).unwrap()))
+}
+
+/// Per slot container for storing notar and skip votes for creating rewards certificates.
+#[derive(Default)]
+pub(super) struct Entry {
+    /// map from validator rank to the skip vote.
+    skip: BTreeMap<u16, VoteMessage>,
+    /// notar votes are indexed by block id as different validators may vote for different blocks.
+    notar: BTreeMap<Hash, BTreeMap<u16, VoteMessage>>,
```
steviez:
```rust
pub struct ConsensusRewards {
    /// Per [`Slot`], stores skip and notar votes.
    votes: BTreeMap<Slot, Entry>,
    /// Stores the latest pubkey for the current node.
    cluster_info: Arc<ClusterInfo>,
    /// Stores the leader schedules.
    leader_schedule_cache: Arc<LeaderScheduleCache>,
}
```
I have some concerns about allocations, this is several layers deep of `BTreeMap` nesting. It looks like we'd also be doing this with a lock held

core/src/block_creation_loop.rs
```diff
@@ -409,9 +424,17 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
     block_timer: Instant,
     block_timeout: Duration,
 ) -> Result<(), PohRecorderError> {
+    // Taking a read lock on consensus_rewards can contend with the write lock in bls_sigverifier.
+    // We are ready to produce the block footer now, while we gather other bits of data, we can block on the consensus_rewards lock in a separate thread to minimise contention.
+    let handle = std::thread::spawn(move || {
+        // XXX: how to look up the slot.
+        let slot = u64::MAX;
+        consensus_rewards.read().build_rewards_certs(slot)
+    });
```
steviez: I know this isn't final, but spawning unnamed threads like this is not how we typically like to do stuff in parallel

core/src/block_creation_loop.rs
```diff
@@ -10,6 +10,7 @@ use {
         replay_stage::{Finalizer, ReplayStage},
     },
     crossbeam_channel::Receiver,
+    parking_lot::RwLock as PLRwLock,
```
steviez: Why this over `std::sync::RwLock` ?

akhi3030: A couple of reasons:

- My understanding is that parking lot is fairer to writers and ensures that writers do not starve.  In particular, if there is a writer waiting, then additional readers also have to wait
- for small critical sections, [benchmarks](https://blog.cuongle.dev/p/inside-rusts-std-and-parking-lot-mutexes-who-win) seems to suggest that parking lot does better.

core/src/block_creation_loop.rs
akhi3030: absolutely, this was just to show what I propose to do.  Is there an example that I can follow on how I could do something like this?

votor/src/consensus_rewards/entry.rs
akhi3030: Agreed.  What if I switch to `HashMap` and use `with_capacity()`?  In almost all of the instances, we have a good idea of what the expected size of the containers will be.

akhi3030 (commented):
thanks for the feedback Steve.  I'll push a change to use `HashMap` and `with_capacity` to minimise allocations.

votor/src/consensus_rewards/entry.rs
akhi3030: replaced most of the btreemaps with hashmaps now.  One remains where we are actually relying on the ordering or keys so not so easy to remove.

core/src/block_creation_loop.rs
```diff
@@ -409,9 +424,17 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
 fn record_and_complete_block(
     poh_recorder: &RwLock<PohRecorder>,
     record_receiver: &mut RecordReceiver,
+    consensus_rewards: Arc<PLRwLock<ConsensusRewards>>,
     block_timer: Instant,
     block_timeout: Duration,
 ) -> Result<(), PohRecorderError> {
+    // Taking a read lock on consensus_rewards can contend with the write lock in bls_sigverifier.
```
akhi3030: Feedback from @AshwinSekar: https://github.com/anza-xyz/alpenglow/pull/607#discussion_r2554446280

akhi3030: @AshwinSekar : I've reworked the PR to use channels and a dedicated thread now.  There is also a "stateless" version of `wants_vote` that we could call from the BLS sigverifier to do some filtering and avoid cloning all votes.  I'll add support for that as a separate commit.

core/src/block_creation_loop.rs
```diff
@@ -490,7 +519,14 @@ fn record_and_complete_block(
 
     // Produce the footer with the current timestamp
     let working_bank = w_poh_recorder.working_bank().unwrap();
-    let footer = produce_block_footer(working_bank.bank.clone_without_scheduler());
+    let resp = cert_receiver
```
akhi3030: should collect metrics on how long we spend waiting here.

akhi3030:
I am still trying to add a test case for testing the consensus rewards service.  I need to figure out how to generate a custom leader schedule for it.  Other than this one missing test case, the rest of the PR should be reviewable.

PR #583, opened 2025-11-04
migration: do not clone record receiver, get it from PoHService

AshwinSekar:
One more before we upstream BCL (I swear).

Andrew gave feedback that `RecordReceiver` is not really safe to clone, and instead sending it over a channel when `PohService` shuts down is preferred. Also gives us a strong guarantee that we're not building blocks with both.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -33,7 +33,6 @@ use {
     },
     solana_version::version,
     solana_votor::{common::block_timeout, event::LeaderWindowInfo, votor::LeaderWindowNotifier},
-    solana_votor_messages::migration::MigrationStatus,
     stats::{BlockCreationLoopMetrics, SlotMetrics},
     std::{
         sync::{
@@ -71,7 +70,6 @@ impl BlockCreationLoop {
 
 pub struct BlockCreationLoopConfig {
     pub exit: Arc<AtomicBool>,
-    pub migration_status: Arc<MigrationStatus>,
 
     // Shared state
     pub bank_forks: Arc<RwLock<BankForks>>,
@@ -90,7 +88,7 @@ pub struct BlockCreationLoopConfig {
     pub replay_highest_frozen: Arc<ReplayHighestFrozen>,
 
     // Channel to receive RecordReceiver from PohService
-    pub record_receiver_channel: Receiver<RecordReceiver>,
+    pub record_receiver_receiver: Receiver<RecordReceiver>,
 }
 
 struct LeaderContext {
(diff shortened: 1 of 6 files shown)


core/src/block_creation_loop.rs
```diff
@@ -174,6 +177,19 @@ fn start_loop(config: BlockCreationLoopConfig) {
     // get latest identity pubkey during startup
     let mut my_pubkey = cluster_info.id();
 
+    info!("{my_pubkey}: Block creation loop initialized");
+    // Wait for PohService to be shutdown
+    migration_status.wait_for_migration_or_exit(&exit);
+    info!("{my_pubkey}: Block creation loop starting");
```
bw-solana: maybe move this log below the wait for `RecordReceiver`?

core/src/block_creation_loop.rs
```diff
@@ -174,6 +177,19 @@ fn start_loop(config: BlockCreationLoopConfig) {
     // get latest identity pubkey during startup
     let mut my_pubkey = cluster_info.id();
 
+    info!("{my_pubkey}: Block creation loop initialized");
+    // Wait for PohService to be shutdown
+    migration_status.wait_for_migration_or_exit(&exit);
```
bw-solana: do we even need this anymore? Could we just wait on PoH to send the record receiver?

core/src/validator.rs
```diff
@@ -1413,16 +1413,22 @@ impl Validator {
 
         let replay_highest_frozen = Arc::new(ReplayHighestFrozen::default());
         let leader_window_notifier = Arc::new(LeaderWindowNotifier::default());
+
+        // Pass RecordReceiver from PohService to BlockCreationLoop when shutting down. Gives us a strong guarentee
+        // that both block producers are not running at the same time
+        let (record_receiver_sender, record_receiver_channel) = bounded(1);
```
bw-solana: `record_receiver_channel` -> `record_receiver_receiver` ? maybe a little deranged... but I keep looking at this and thinking it's sending records

poh/src/poh_service.rs
```diff
@@ -169,6 +176,12 @@ impl PohService {
                     return;
                 }
 
+                // Pass the RecordReceiver to BlockCreationLoop,
+                if let Err(e) = record_receiver_sender.send(record_receiver) {
+                    info!("Unable to send record receiver, shutting down {e:}");
```
bw-solana: `error!` ?

core/src/block_creation_loop.rs
AshwinSekar: Yeah that's a fair point, i'll update

PR #584, opened 2025-11-05
fast leader handover: send `UpdateParent` and clear banks

ksn6:
This PR implements Milestone 1 as articulated here: https://github.com/anza-xyz/alpenglow/issues/293#issuecomment-3145962346.

We implement the "happy case" for fast leader handover; i.e., the case where the optimistic parent matches the confirmed parent obtained via the parent ready channel.

For the "sad case," the leader disseminates an `UpdateParent` message. Leaders and non-leaders both purge the bank and replace it with a new one. The block essentially is a no-op.

Follow-up PRs:
- Milestone 2: Allow new transactions to play on top of this new bank
- Local cluster testing

NOTE: this PR will be broken up into a bunch of smaller PRs.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -247,50 +247,30 @@ fn start_loop(config: BlockCreationLoopConfig) {
 
         // Race between parent ready notification and optimistic parent events
         // Parent ready is checked first and has priority if both channels are ready
-        let window_source = if ctx
-            .bank_forks
-            .read()
-            .unwrap()
-            .migration_status()
-            .should_allow_block_markers(ctx.bank_forks.read().unwrap().root())
-        {
-            select_biased! {
-                recv(ctx.leader_window_info_receiver) -> msg => {
-                    // Drain all pending messages and keep the latest one
-                    msg.ok()
-                        .and_then(|window| {
-                            ctx.leader_window_info_receiver
-                                .try_iter()
-                                .last()
-                                .or(Some(window))
-                        })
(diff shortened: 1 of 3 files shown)


ksn6:
Reminder: I still need to configure the timeouts correctly - i.e., start the block timeout only upon viewing a `ParentReady`.

runtime/src/block_component_processor.rs
```diff
@@ -139,6 +144,28 @@ impl BlockComponentProcessor {
...
+
+        let new_bank = Bank::new_from_parent_with_options(
+            parent_bank.clone(),
+            // TODO(ksn): ensure that we verify on shred ingest that 1 <= slot - parent_slot < 4
+            parent_bank.collector_id(),
+            slot,
+            NewBankOptions::default(),
+        );
+
+        bank_forks.write().unwrap().insert(new_bank);
```
AshwinSekar: I'd really like to avoid `confirm_slot` modifying `bank_forks`  if possible, it's really unintuitive to the reader.

Perhaps something like this could work, lmk if i'm off on the assumptions:
- If an `UpdateParent`s shreds are inserted into blockstore the `ParentMeta` field will be updated
- So instead here, we should just return an error for blockstore processor indicating that this bank has been abandoned
- In `process_replay_results` if we see this error, then we clear the bank
- `generate_new_bank_forks` will recreate this bank with the updated parent on the next iteration.

ksn6: Works - this is done.

PR #585, opened 2025-11-05
migration: discover the genesis block via super-oc tracking

AshwinSekar:
Split from #502 

#### Problem
We need to actually discover the genesis block once the migration has started in order to cast our BLS vote.
From the paper:
```
Definition 1 (strong Optimistic Confirmation). OCp(b) is the condition that
the child of b is in slot slot(b) + 1, and it includes the votes of at least p of stake
on b.
```
English: Block `b` whose child in the next slot contains votes for `b` from 82% of stake is super OC

And to figure out the genesis block:
```
Definition 2 (look for genesis). If there exists an ancestor bOC of b such that
slot(bOC) ≥ s −1 and OCp(bOC) = true, then the function try(b) returns the
lowest ancestor b′of b such that slot(b′) < s.
```
English: From the super OC block, genesis is the greatest by slot # ancestor from before the migration slot

#### Summary of Changes
Add tracking in collect vote lockouts and set the genesis block once it's found.

Originally was going to do this in Cluster Info Vote Listener (CIVL) because there was some confusion about including gossip votes or votes which landed in children *not* in the next slot, but think it's clearer and easier to debug if we only include replay votes as the paper intended.

### DIFF
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -414,12 +414,12 @@ impl MigrationStatus {
     ///
     /// Should only be used during `Migration`, and transitions to `ReadyToEnable` if we have already
     /// received a genesis certificate and it matches.
-    pub fn set_genesis_block(&self, genesis_block_arg: Block) {
+    pub fn set_genesis_block(&self, discovered_genesis_block @ (slot, _): Block) {
         let mut phase = self.phase.write().unwrap();
         let MigrationPhase::Migration {
+            migration_slot,
             ref mut genesis_block,
             ref genesis_cert,
-            ..
         } = &mut *phase
         else {
             unreachable!(
@@ -429,13 +429,19 @@ impl MigrationStatus {
         };
         assert!(
             genesis_block.is_none(),
-            "Attempting to overwrite genesis block to {genesis_block_arg:?}. Programmer error"
+            "Attempting to overwrite genesis block to {discovered_genesis_block:?}. Programmer \
+             error"
+        );
+
+        assert!(
+            slot < *migration_slot,
+            "Attempting to set a genesis block that is past the migration start"
         );
         warn!(
-            "{} Setting genesis block {genesis_block_arg:?}",
+            "{} Setting genesis block {discovered_genesis_block:?}",
             self.my_pubkey
         );
-        *genesis_block = Some(genesis_block_arg);
+        *genesis_block = Some(discovered_genesis_block);
 
         let Some(genesis_cert) = genesis_cert else {
             return;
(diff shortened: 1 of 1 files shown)


votor-messages/src/migration.rs
```diff
@@ -139,6 +139,23 @@ impl MigrationPhase {
...
+        match self {
+            MigrationPhase::Migration {
+                migration_slot,
+                genesis_block,
+                ..
+            } => genesis_block.is_none() && slot > *migration_slot,
+            MigrationPhase::PreFeatureActivation
+            | MigrationPhase::ReadyToEnable { .. }
+            | MigrationPhase::AlpenglowEnabled { .. }
+            | MigrationPhase::FullAlpenglowEpoch { .. } => false,
```
bw-solana, line 155: `_` would probably be fine given I don't see this conditional expanding beyond the `Migration` phase, but this is certainly safe :)

core/src/replay_stage.rs
```diff
@@ -3867,8 +3871,42 @@ impl ReplayStage {
...
+                            .expect("Genesis bank must exist, no rooting past migration slot");
+
+                        let genesis_block = (
+                            genesis_slot,
+                            genesis_bank.block_id().expect(
+                                "It is impossible for block id to not be known at this point, as \
+                                 a descendant of this block has reached super oc status",
+                            ),
+                        );
+                        migration_status.set_genesis_block(genesis_block);
```
bw-solana, line 3907: slightly off topic, but I believe we used to have an assert in this function to check if the genesis block (slot number) was in expected range. Might be nice to have that condition somewhere to catch any oversights when we start hammering this in testing

core/src/consensus.rs
```diff
@@ -496,6 +506,9 @@ impl Tower {
             total_stake += voted_stake;
         }
 
+        let parent_is_super_oc = bank_slot == parent_slot + 1
+            && super_oc_stake as f64 / total_stake as f64 > GENESIS_VOTE_THRESHOLD;
```
bw-solana, line 510: feels yucky shoving this detection in `collect_vote_lockouts`, but it does seem like the most convenient place. I can look the other way for now

AshwinSekar: ah yeah definitely, there's a couple of options here:
- CIVL since it already collects votes, however if we want to enforce the "parent/child in adjacent slot" we'll end up having to interact with bank forks which sucks
- collect vote lockouts, we're already iterating through every vote account in order to collect lockouts and update voted stakes, seems like a reasonable place
- entirely new location, we incur the AccountsDB cost of fetching VoteAccounts

collect vote lockouts seems like the least bad option haha

core/src/replay_stage.rs
AshwinSekar: yeah good call added asserts to set_genesis_block / cert and spurced up here https://github.com/anza-xyz/alpenglow/pull/585/commits/cd03b4ccacaec9b4c2bc96ee5e657f4a707306b5

core/src/consensus.rs
```diff
@@ -442,14 +446,20 @@ impl Tower {
...
+                // Add the last vote to update the `heaviest_subtree_fork_choice`
                 latest_validator_votes_for_frozen_banks.check_add_vote(
                     key,
                     last_landed_voted_slot,
                     get_frozen_hash(last_landed_voted_slot),
                     true,
                 );
+
+                // For migration - a block is super OC there are 82% of votes in the direct child
+                // on the very next slot
```
carllin, line 459: here any OC block would be good right, but votes fr the parent in the direct child are easier to track?

AshwinSekar: I believe so but others disagree. 

I think we can save the headache and just do direct child & no gossip. Latency 1 votes on mainnet are extremely high so probably effort is best spent elsewhere.

Side effect is that yes, it becomes super easy to track if we don't count all descendants or gossip :)

PR #586, opened 2025-11-06
migration: plumb `migration_status` through blockstore

ksn6:
We need access to `migration_status` in blockstore for block component parsing.

We will use `migration_status` to ensure that, pre-Alpenglow, validators that (1) have Alpenglow and (2) do not have Alpenglow both mark slots with incoming `BlockComponent::BlockMarker`s as dead.


ksn6:
there are other ways of attaining the same effects of this code here, albeit with simpler code; closing this PR

PR #587, opened 2025-11-06
migration: send the GenesisVote once discovery has succeeded

AshwinSekar:
Split from #502 

#### Problem
Replay stage discovers the genesis block and sets it on `MigrationStatus`. We then need to send our BLS vote on the block.

#### Summary of Changes
Following up from discussion here https://github.com/anza-xyz/alpenglow/pull/580/files#r2488310874 we detect this in `EventHandler` and send the vote

### DIFF
--- a/votor/src/event_handler.rs
+++ b/votor/src/event_handler.rs
@@ -25,7 +25,11 @@ use {
     solana_pubkey::Pubkey,
     solana_runtime::{bank::Bank, bank_forks::SetRootError},
     solana_signer::Signer,
-    solana_votor_messages::{consensus_message::Block, migration::MigrationStatus, vote::Vote},
+    solana_votor_messages::{
+        consensus_message::Block,
+        migration::{MigrationStatus, GENESIS_VOTE_REFRESH},
+        vote::Vote,
+    },
     std::{
         collections::{BTreeMap, BTreeSet},
         sync::{
@@ -124,10 +128,26 @@ impl EventHandler {
 
         // Wait until migration has completed
         info!("{}: Event loop initialized", local_context.my_pubkey);
-        let Some(genesis_block) = migration_status.wait_for_migration_or_exit(&exit) else {
-            // Exited during migration
-            return Ok(());
-        };
+        let mut last_genesis_vote_refresh_time = Instant::now();
+        loop {
+            if exit.load(Ordering::Relaxed) {
+                // Exited during the migration
+                return Ok(());
+            }
+
+            if migration_status.is_alpenglow_enabled() {
+                // Migration successful
+                break;
+            }
+
+            if last_genesis_vote_refresh_time.elapsed() > GENESIS_VOTE_REFRESH
+                && Self::maybe_send_genesis_vote(&migration_status, &mut vctx)?
+            {
+                last_genesis_vote_refresh_time = Instant::now();
+            }
+            std::thread::sleep(Duration::from_millis(10));
+        }
+        let genesis_block = migration_status.genesis_block();
         let root_slot = vctx.sharable_banks.root().slot();
         info!(
             "{}: Event loop starting genesis {genesis_block:?} root {root_slot}",
@@ -195,6 +215,27 @@ impl EventHandler {
         }
 
         Ok(())
+    }
+
+    fn maybe_send_genesis_vote(
+        migration_status: &MigrationStatus,
+        vctx: &mut VotingContext,
+    ) -> Result<bool, EventLoopError> {
+        let Some((slot, block_id)) = migration_status.get_votable_genesis_block() else {
+            return Ok(false);
+        };
+
+        let vote = Vote::new_genesis_vote(slot, block_id);
+        let Some(bls_op) = generate_vote_message(
+            vote,
+            // We set `is_refresh` because we don't want this vote in the `VoteHistory`
+            true, vctx,
+        )?
+        else {
+            return Ok(false);
+        };
+        vctx.bls_sender.send(bls_op).map_err(|_| SendError(()))?;
+        Ok(true)
     }
 
     fn handle_parent_ready_event(
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -156,6 +156,18 @@ impl MigrationPhase {
         }
     }
 
+    /// If we have entered the migration and discovered the genesis block, return the block
+    /// so that we may vote on it.
+    fn get_votable_genesis_block(&self) -> Option<Block> {
+        match self {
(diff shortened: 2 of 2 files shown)


votor/src/event_handler.rs
```diff
@@ -124,10 +128,26 @@ impl EventHandler {
...
+                // Migration successful
+                break;
+            }
+
+            if last_genesis_vote_refresh_time.elapsed() > GENESIS_VOTE_REFRESH
+                && Self::maybe_send_genesis_vote(&migration_status, &mut vctx)?
+            {
+                last_genesis_vote_refresh_time = Instant::now();
+            }
+            std::thread::sleep(Duration::from_millis(10));
```
AshwinSekar, line 148: Think it makes sense to have this in `EventHandler` instead of replay as `EventHandler` is normally responsible for sending bls votes. Only downside here is this thread no longer sleeps and spins.

Could tradeoff clarity and add another condvar / crossbeam to avoid the spin here if preferred.

votor/src/event_handler.rs
```diff
@@ -197,6 +217,27 @@ impl EventHandler {
...
 
+    fn maybe_send_genesis_vote(
+        migration_status: &MigrationStatus,
+        vctx: &mut VotingContext,
+    ) -> Result<bool, EventLoopError> {
+        let Some((slot, block_id)) = migration_status.get_votable_genesis_block() else {
+            return Ok(false);
+        };
+
+        let vote = Vote::new_genesis_vote(slot, block_id);
```
wen-coding, line 228: Could the genesis vote change if you stay in the loop?

votor/src/event_handler.rs
```diff
@@ -197,6 +217,27 @@ impl EventHandler {
...
+        let Some(bls_op) = generate_vote_message(
+            vote,
+            // We set `is_refresh` because we don't want this vote in the `VoteHistory`
+            true, vctx,
```
wen-coding, line 232: nit: we probably should rename `is_refresh` to `do_not_update_vote_history` or something?

votor/src/event_handler.rs
```diff
@@ -124,10 +128,26 @@ impl EventHandler {
 
         // Wait until migration has completed
         info!("{}: Event loop initialized", local_context.my_pubkey);
-        let Some(genesis_block) = migration_status.wait_for_migration_or_exit(&exit) else {
-            // Exited during migration
-            return Ok(());
-        };
+        let mut last_genesis_vote_refresh_time = Instant::now();
+        loop {
```
wen-coding, line 132: nit: can we maybe move the loop to some function?

votor/src/event_handler.rs
```diff
@@ -124,10 +128,26 @@ impl EventHandler {
...
+                return Ok(());
+            }
+
+            if migration_status.is_alpenglow_enabled() {
+                // Migration successful
+                break;
+            }
+
+            if last_genesis_vote_refresh_time.elapsed() > GENESIS_VOTE_REFRESH
+                && Self::maybe_send_genesis_vote(&migration_status, &mut vctx)?
```
wen-coding, line 144: So on InvalidConfig and TransientError we will retry every 10 ms? Is that a bit aggressive?

votor/src/event_handler.rs
```diff
@@ -124,10 +128,26 @@ impl EventHandler {
 
         // Wait until migration has completed
         info!("{}: Event loop initialized", local_context.my_pubkey);
-        let Some(genesis_block) = migration_status.wait_for_migration_or_exit(&exit) else {
-            // Exited during migration
-            return Ok(());
-        };
+        let mut last_genesis_vote_refresh_time = Instant::now();
```
wen-coding, line 131: So the earliest we send out genesis vote is after 400ms?

votor/src/event_handler.rs
AshwinSekar: Nope, there's only one genesis block

votor/src/event_handler.rs
AshwinSekar: yeah can do it in a follow up

votor/src/event_handler.rs
AshwinSekar: Yeah this is a really good point, with this minimal loop we don't handle set-identity.

It seems to get this voting to work via `EventHandler`, we'll want to startup the full loop, and add a new `VotorEvent` for when the genesis block is found. Then we can create a timer event that indicates that we should refresh the genesis vote in case it doesn't succeed 400ms after the original vote. 
The ConsensusPoolService will notice when the migration succeeds in order to send the kick off parent ready and we can proceed as normal.

I think this is probably best if we want EventHandler to handle the migration vote, rather than trying to special case a minimal event loop here. Other alternative is to keep this vote happening in replay. lmk what you think.

votor/src/event_handler.rs
AshwinSekar: yeah we have to freeze at least 2 blocks along with all the normal startup things before we can identify a genesis block

wen-coding: Sounds good, maybe add a comment here saying this wait of 400ms is intentional.

votor/src/event_handler.rs
wen-coding: I like option 1, replay is already very complicated. But how much work is it?

wen-coding: Also, maybe it's not end of the world if we don't handle set-identity here?

AshwinSekar: I think it's very important that we handle set-identity here

If we don't handle set-identity then this could happen:
- Helius upgrades to v4.1 two weeks before the migration with a dummy identity
- Once caught up they set-identity to their staked identity
- The EventHandler spins for 2 weeks with the dummy identity
- When the migration starts the Helius is unable to send their GenesisVote because we have the wrong identity

Almost everyone in the super majority uses this workflow to upgrade software (you can check the vote credits from an epoch where a software upgrade happens), so if we don't capture the set-identity i'd expect the migration to fail.

AshwinSekar: I'd argue that modifying the event loop is probably more complexity than adding a function call to the replay loop. 

We'll have to modify replay to then send over some sort of `VotorEvent::GenesisBlock` event to the event loop - seems simpler if replay just sent the vote rather than going through an intermediate channel? 

Conceptually this also means the event handler thread is fully started rather than just blocking on the condvar wakeup.

wen-coding: Sure, for Migration I think we can do that, since this logic will not stay in replay forever.

PR #588, opened 2025-11-06
migration: update tick height logic

AshwinSekar:
Split from #502 

#### Problem
Setting alpenglow ticks is still using the old logic of feature flag + handling blocks that "straddle" the boundary.

#### Summary of Changes

First commit - Initialize `MigrationStatus` in `BankForks` initialization, this is necessary for multiple reasons.
Imagine we crashed and started up right after the migration  we have
`PohBlock (snapshot) -> PohBlock -> ... -> AGBlock -> AGBlock`:
- On startup we process blockstore, when we get to the AG blocks we won't know how to process them. We need a `MigrationStatus` to figure out the expected ticks in each block.
- We'll erroneously think that we're in migration after startup as we use the snapshot block to construct `MigrationStatus`

Instead we initialize a `MigrationStatus` with the `BankForks` root (snapshot or genesis) and then:
- (future PR) update `MigrationStatus` while replaying startup blocks - ifwe see the first AGBlock with valid genesis certificate in the header we enable alpenglow
- Pass it to `load_frozen_forks` so that we know how many ticks per block we need for these startup blocks
- Also useful for Karthik's work - we need to know whether the block is Alpenglow so we can check header/footer

This way after startup we'll have an accurate `MigrationStatus`

Second commit - Update tick verification and setting from ff to `migration_status`. Also we don't need the complexity of half poh / half alpenglow blocks anymore so greatly simplifies the logic here.

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -60,6 +60,7 @@ use {
     solana_transaction_error::{TransactionError, TransactionResult as Result},
     solana_transaction_status::token_balances::TransactionTokenBalancesSet,
     solana_vote::vote_account::VoteAccountsHashMap,
+    solana_votor_messages::migration::MigrationStatus,
     std::{
         borrow::Cow,
         collections::{HashMap, HashSet},
@@ -1828,7 +1829,7 @@ fn process_next_slots(
                     .unwrap(),
                 *next_slot,
             );
-            set_alpenglow_ticks(&next_bank);
+            set_alpenglow_ticks(&next_bank, migration_status);
             trace!(
                 "New bank for slot {}, parent slot is {}",
                 next_slot,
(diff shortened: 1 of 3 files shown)


runtime/src/bank_forks.rs
```diff
@@ -145,6 +151,8 @@ impl BankForks {
             descendants.entry(parent).or_default().insert(root_slot);
         }
 
+        let migration_status = Arc::new(Self::initialize_migration_status(&root_bank));
```
AshwinSekar, line 154: This will initialize the `migration_status` on startup with either the snapshot or genesis block.
In the next PR when we process our startup blocks from `Blockstore`, i'll potentially enable alpenglow if we run into a valid Alpenglow block.

The idea is that once the validator is done with process blockstore the `migration_status` is correctly initialized. Currently we only initialize `migration_status` using the `BankForks` root, so we have incorrect behavior regarding the startup blocks.

ledger/src/blockstore_processor.rs
```diff
@@ -1939,6 +1882,7 @@ fn load_frozen_forks(
     timing: &mut ExecuteTimings,
     snapshot_controller: Option<&SnapshotController>,
 ) -> result::Result<(u64, usize), BlockstoreProcessorError> {
+    let migration_status = bank_forks.read().unwrap().migration_status();
```
AshwinSekar, line 1885: this is the fn that loads all the blockstore banks on startup. Basically the reason for all this refactor to move up MigrationStatus into BankForks initialization.

In this PR we just read migration status to figure out ticks - in a future PR we'll modify MigrationStatus if we see a valid AG block here

bw-solana (approved):
lgtm

PR #589, opened 2025-11-06
migration: Send the GenesisVote when we discover the genesis block

AshwinSekar:
Split from #502 

#### Problem
Once compute bank stats discovers the genesis block we need to send out the GenesisVote for it.

#### Summary of Changes
In replay if alpenglow is yet to be enabled and we have discovered a genesis block, send the vote.
Refactored voting_utils to be accessible from replay by passing in arguments rather than the VotingContext

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -1494,7 +1494,7 @@ impl ReplayStage {
     }
 
     /// If we have an eligble genesis block, send out a genesis vote
-    /// Returns false if no eligble block was found
+    /// Returns false if no eligible block was found
     fn maybe_send_genesis_vote(
         migration_status: &MigrationStatus,
         bank_forks: &RwLock<BankForks>,
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -407,11 +407,7 @@ impl MigrationStatus {
     pub fn eligible_genesis_block(&self) -> Option<Block> {
         let phase = self.phase.read().unwrap();
         let MigrationPhase::Migration { genesis_block, .. } = &*phase else {
-            unreachable!(
-                "{}: Programmer error, attempting to find eligble genesis block while not in \
-                 migration",
-                self.my_pubkey
-            );
+            return None;
         };
         *genesis_block
     }


core/src/replay_stage.rs
```diff
@@ -1467,6 +1488,59 @@ impl ReplayStage {
         );
     }
 
+    /// If we have an eligble genesis block, send out a genesis vote
+    /// Returns false if no eligble block was found
```
carllin: nit: eligible

core/src/replay_stage.rs
```diff
@@ -1467,6 +1488,59 @@ impl ReplayStage {
...
+    fn maybe_send_genesis_vote(
+        migration_status: &MigrationStatus,
+        bank_forks: &RwLock<BankForks>,
+        vote_account: Pubkey,
+        identity_keypair: &Arc<Keypair>,
+        authorized_voter_keypairs: &Arc<std::sync::RwLock<Vec<Arc<Keypair>>>>,
+        own_vote_sender: &Sender<ConsensusMessage>,
+        bls_sender: &Sender<BLSOp>,
+    ) -> bool {
+        let Some((slot, block_id)) = migration_status.eligible_genesis_block() else {
```
carllin, line 1507: I know we check that we're in migration before we call this function, but can the certificate pool get a certificate and switch to alpenglow enabled in between that check and here? 

Might be easier to reason about to remove the panic from `eligible_genesis_block()` and just return None if we're not in that migration mode

core/src/replay_stage.rs
```diff
@@ -1467,6 +1488,59 @@ impl ReplayStage {
...
+                    identity_keypair.pubkey()
+                );
+                // If sending fails that means the channel is disconnected and we are shutting down
+                let _ = own_vote_sender.send(message.clone());
+                let _ = bls_sender.send(BLSOp::PushVote {
+                    message: Arc::new(message),
+                    slot,
+                    saved_vote_history:
+                        solana_votor::vote_history_storage::SavedVoteHistoryVersions::Current(
+                            SavedVoteHistory::default(),
```
carllin, line 1535: saving the default vote history won't overwrite anything important because there's no alpenglow vote history yet?

On restart, if we detect an empty vote history, that won't prevent us from resending the genesis vote right?

carllin: Another thing, is this check on startup sufficient https://github.com/AshwinSekar/alpenglow/blob/f4abd1fb6aafda6248cb3c0431ee52efe6b371b3/core/src/validator.rs#L1029-L1047 to switch from using tower history vs alpenglow vote history on startup, or do should we instead check for successful migration?

AshwinSekar: this is next on the list, basically anything that uses the feature flag will be switched to MigrationStatus. Likely I'll modify it so that EventHandler will instantiate the VoteHistory since we shouldn't need it before then.

> saving the default vote history won't overwrite anything important because there's no alpenglow vote history yet?

Yep exactly, also the default one uses Hash::default() for the filename so shouldn't clobber in case we get into some weird scenario where we have a VoteHistory

> On restart, if we detect an empty vote history, that won't prevent us from resending the genesis vote right?

Nope, the point here is that we should always try to send the genesis vote after we freeze and detect the genesis block. Nothing reads VoteHistory prior to event handler turning on

core/src/replay_stage.rs
AshwinSekar: brilliant, you're correct I totally missed that
fixed here https://github.com/anza-xyz/alpenglow/pull/589/commits/3e33e96987ccad867ef513cd30a68ff04d853923

carllin: nice, looks good

What do you think about moving the `unreachable!` type panics in functins not associated with state transitions in migration.rs to the caller and having them handle it?

I think those would be `alpenglow_rooted_new_epoch` and `genesis_block`

AshwinSekar: yeah that should work, i'll send a follow up PR cleaning the unreachables

PR #590, opened 2025-11-07
fast leader handover: parse block header on shred ingest

ksn6:
We need to create a `ParentMeta` column in blockstore to start tracking parent-related items. Eventually, `UpdateParent` markers will update this column as well. This PR parses block headers on shred ingest.

### DIFF
--- a/ledger/src/blockstore/column.rs
+++ b/ledger/src/blockstore/column.rs
@@ -1072,38 +1072,33 @@ impl TypedColumn for columns::AlternateMerkleRootMeta {
 
 impl Column for columns::ParentMeta {
     type Index = (Slot, BlockLocation);
-    // Key size: Slot (8 bytes) + discriminator (1 byte) + optional Hash (32 bytes)
-    // Maximum size when BlockLocation::Alternate
-    type Key = [u8; std::mem::size_of::<Slot>() + 1 + HASH_BYTES];
+    // Key size: Slot (8 bytes) + Hash (32 bytes)
+    // When BlockLocation::Original, the hash is Hash::default().
+    type Key = [u8; std::mem::size_of::<Slot>() + HASH_BYTES];
 
     #[inline]
     fn key((slot, location): &Self::Index) -> Self::Key {
-        let mut key = [0u8; std::mem::size_of::<Slot>() + 1 + HASH_BYTES];
+        let mut key = [0u8; std::mem::size_of::<Slot>() + HASH_BYTES];
         key[..8].copy_from_slice(&slot.to_le_bytes());
 
-        match location {
-            BlockLocation::Original => {
-                key[8] = 0; // discriminator for Original
-                            // Rest of the key remains zeros
-            }
-            BlockLocation::Alternate { block_id } => {
-                key[8] = 1; // discriminator for Alternate
-                key[9..41].copy_from_slice(&block_id.to_bytes());
-            }
-        }
+        let hash_bytes = match location {
+            BlockLocation::Original => &Hash::default().to_bytes(),
+            BlockLocation::Alternate { block_id } => &block_id.to_bytes(),
+        };
+
+        key[8..40].copy_from_slice(hash_bytes);
+
         key
     }
 
     fn index(key: &[u8]) -> Self::Index {
         let slot = Slot::from_le_bytes(key[0..8].try_into().unwrap());
-        let location = match key[8] {
-            0 => BlockLocation::Original,
-            1 => {
-                let block_id = Hash::new_from_array(key[9..41].try_into().unwrap());
-                BlockLocation::Alternate { block_id }
-            }
-            _ => panic!("Invalid BlockLocation discriminator"),
+        let hash = Hash::new_from_array(key[8..40].try_into().unwrap());
+        let location = match hash == Hash::default() {
+            true => BlockLocation::Original,
+            false => BlockLocation::Alternate { block_id: hash },
         };
+
         (slot, location)
     }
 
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore_meta.rs
```diff
@@ -979,6 +979,13 @@ impl OptimisticSlotMetaVersioned {
     }
 }
 
+#[derive(Clone, Copy, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ParentMeta {
+    pub parent_slot: Slot,
+    pub parent_block_id: Hash,
+    pub replay_fec_set_index: u32,
```
AshwinSekar, line 986: do we want to add an explicit field on whether we have seen an `UpdateParent` or not? Or I guess we could check by seeing if `replay_fec_set_index` is nonzero?

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1928,70 @@ impl Blockstore {
...
+                    VersionedBlockHeader::V1(update) | VersionedBlockHeader::Current(update) => {
+                        Some((update.parent_slot, update.parent_block_id))
+                    }
+                }
+            }
+            _ => None,
+        }
+    }
+
+    fn check_for_block_header(
```
AshwinSekar: I guess yeah we don't really need the `has_header` checks in `block_component_processor`? 
Or would we expect there to be a second header in a block with `UpdateParent` present?

core/src/replay_stage.rs
```diff
@@ -3697,6 +3742,29 @@ impl ReplayStage {
                     ("hash", bank.hash().to_string(), String),
                 );
 
+                // Verify parent meta consistency based on migration status
+                if !is_leader_block {
+                    if let Err(err) =
+                        Self::parent_meta_checks(bank_slot, blockstore, migration_status, bank)
```
AshwinSekar: I wouldn't do this here, instead in a future PR we'll need to modify `generate_new_bank_forks` to use the `ParentMeta` & `NextSlots` to create the bank from parent when `MigrationStatus` is active.

Thus if we create the bank we already know it's consistent.

ledger/src/blockstore/column.rs
```diff
@@ -278,6 +278,9 @@ pub mod columns {
     /// * index type: `u64` (see [`SlotColumn`])
     /// * value type: [`blockstore_meta::SlotCertificates`]
     pub struct SlotCertificates;
+
+    #[derive(Debug)]
```
AshwinSekar, line 282: Add doc comment

ledger/src/blockstore.rs
```diff
@@ -1896,6 +1928,70 @@ impl Blockstore {
...
+        // Try to parse BlockHeader from the payload
+        let (parent_slot, parent_block_id) = Self::parse_block_header_from_data_payload(payload)?;
+
+        let parent_meta = ParentMeta {
+            parent_slot,
+            parent_block_id,
+            replay_fec_set_index: 0,
+        };
+
+        parent_meta_working_set
```
AshwinSekar, line 1986: I would call this function something like `maybe_insert_parent_meta` and guard the caller by `shred_index == 0` and then remove the return value.

ledger/src/blockstore_meta.rs
ksn6: yep, the idea is to check whether `replay_fec_set_index == 0`

core/src/replay_stage.rs
ksn6: yep, makes sense - tossed this in while we were still using shred 0 to infer the parent

ledger/src/blockstore.rs
ksn6: we should probably keep around `BlockComponentProcessor`, since `BlockComponentProcessor` enforces that there's exactly one header that appears at the beginning of a block; we don't enforce this in blockstore. it's also tricky to do here, since we don't have access to migration status.

no, we should only ever have a single block header per block - `UpdateParent` is essentially the "second header" upon a fast leader handover transition

ledger/src/blockstore/column.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore/column.rs
```diff
@@ -1059,3 +1069,57 @@ impl ColumnName for columns::AlternateMerkleRootMeta {
...
+
+impl Column for columns::ParentMeta {
+    type Index = (Slot, BlockLocation);
+    // Key size: Slot (8 bytes) + discriminator (1 byte) + optional Hash (32 bytes)
+    // Maximum size when BlockLocation::Alternate
+    type Key = [u8; std::mem::size_of::<Slot>() + 1 + HASH_BYTES];
+
+    #[inline]
+    fn key((slot, location): &Self::Index) -> Self::Key {
+        let mut key = [0u8; std::mem::size_of::<Slot>() + 1 + HASH_BYTES];
```
AshwinSekar: Probably doesn't matter but we could use `Hash::default()` to represent `BlockLocation::Original` and save a byte

ledger/src/blockstore_processor.rs
```diff
@@ -845,6 +845,15 @@ pub enum BlockstoreProcessorError {
```
AshwinSekar, line 845: revert this file

ksn6: done

ledger/src/blockstore/column.rs
ksn6: nice let's do it

AshwinSekar (approved):

PR #592, opened 2025-11-08
local-cluster: add migration tests and VoteStateV4 hack

AshwinSekar:
Split from #502 

#### Problem
Enough migration PRs have landed such that basic migration tests can be enabled.

#### Summary of Changes
Add local-cluster tests for the migration - i'm sorry for how hacky this is 😭 :
- Allow Vote program to toggle which VoteState version it processes
- Use a test keypair for the alpenglow feature flag
- Reduce migration offset from 5000 -> 32 for tests
- Add a new local cluster type "PreMigration" 
   - disables alpenglow feature and genesis certificate BUT use VoteStateV4 w/ bls pubkeys
- Add 1, 2, and 4 node test, send feature activation tx -> run Tower for an epoch -> verify migration succeeded -> check roots are being made

### DIFF
--- a/local-cluster/src/local_cluster.rs
+++ b/local-cluster/src/local_cluster.rs
@@ -211,16 +211,12 @@ impl LocalCluster {
     }
 
     pub fn new(config: &mut ClusterConfig, socket_addr_space: SocketAddrSpace) -> Self {
-        *solana_vote_program::vote_state::TEMP_HARDCODED_TARGET_VERSION
-            .lock()
-            .unwrap() = VoteStateTargetVersion::V3;
+        *vote_state::TEMP_HARDCODED_TARGET_VERSION.lock().unwrap() = VoteStateTargetVersion::V3;
         Self::init(config, socket_addr_space, AlpenglowMode::Disabled)
     }
 
     pub fn new_alpenglow(config: &mut ClusterConfig, socket_addr_space: SocketAddrSpace) -> Self {
-        *solana_vote_program::vote_state::TEMP_HARDCODED_TARGET_VERSION
-            .lock()
-            .unwrap() = VoteStateTargetVersion::V4;
+        *vote_state::TEMP_HARDCODED_TARGET_VERSION.lock().unwrap() = VoteStateTargetVersion::V4;
         Self::init(config, socket_addr_space, AlpenglowMode::Enabled)
     }
 
@@ -228,9 +224,7 @@ impl LocalCluster {
         config: &mut ClusterConfig,
         socket_addr_space: SocketAddrSpace,
     ) -> Self {
-        *solana_vote_program::vote_state::TEMP_HARDCODED_TARGET_VERSION
-            .lock()
-            .unwrap() = VoteStateTargetVersion::V4;
+        *vote_state::TEMP_HARDCODED_TARGET_VERSION.lock().unwrap() = VoteStateTargetVersion::V4;
         Self::init(config, socket_addr_space, AlpenglowMode::PreMigration)
     }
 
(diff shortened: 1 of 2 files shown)


local-cluster/src/local_cluster.rs
```diff
@@ -200,17 +211,33 @@ impl LocalCluster {
     }
 
     pub fn new(config: &mut ClusterConfig, socket_addr_space: SocketAddrSpace) -> Self {
-        Self::init(config, socket_addr_space, false)
+        *solana_vote_program::vote_state::TEMP_HARDCODED_TARGET_VERSION
```
bw-solana: can drop `solana_vote_program` prefix for these

local-cluster/tests/local_cluster.rs
```diff
@@ -6222,6 +6223,140 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+    let vote_listener_socket = solana_net_utils::bind_to_localhost().unwrap();
+    let vote_listener_addr = vote_listener_socket.try_clone().unwrap();
+    let mut validator_config = ValidatorConfig::default_for_test();
+    validator_config.voting_service_test_override = Some(VotingServiceOverride {
+        additional_listeners: vec![vote_listener_addr.local_addr().unwrap()],
+        alpenglow_port_override: AlpenglowPortOverride::default(),
+    });
+    // Since we don't skip warmup slots and have very small epochs + tight MIGRATION_SLOT_OFFSET
+    // we can't afford any forking due to rolling start
+    validator_config.wait_for_supermajority = Some(0);
```
bw-solana, line 6239: even this can be a bit dicey because the sync happens before the validators are fully spun up and ready to go.

not sure what the best way to reduce flakiness is. I tried using `no_wait_for_vote_to_start_leader` in #8673, but not sure that would necessarily work here

local-cluster/tests/local_cluster.rs
```diff
@@ -6222,6 +6223,140 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
...
+        .map(|i| (Arc::new(keypair_from_seed(&[i as u8; 32]).unwrap()), true))
+        .collect::<Vec<_>>();
+    let node_stakes = vec![DEFAULT_NODE_STAKE; num_nodes];
+
+    let mut cluster_config = ClusterConfig {
+        validator_configs: make_identical_validator_configs(&validator_config, num_nodes),
+        validator_keys: Some(validator_keys),
+        node_stakes: node_stakes.clone(),
+        slots_per_epoch: 2 * MINIMUM_SLOTS_PER_EPOCH,
+        stakers_slot_offset: 2 * MINIMUM_SLOTS_PER_EPOCH,
```
bw-solana: might be nice to explain these choices a little more. And maybe assert on these values relative to `MIGRATION_SLOT_OFFSET` so the relationship is obvious

local-cluster/src/local_cluster.rs
```diff
@@ -327,8 +360,17 @@ impl LocalCluster {
             &keys_in_genesis,
             stakes_in_genesis,
             config.cluster_type,
-            is_alpenglow,
+            create_v4_accounts,
```
bw-solana: this is still controlling more than just v4 accounts, right? It dictates if we activate the alpenglow feature.

Might be better to just leave it as `is_alpenglow` or `enable_alpenglow_feature_and_accounts` or something

local-cluster/tests/local_cluster.rs
AshwinSekar: Yeah this isn't perfect unfortunately - if we start to see this test flaky I can put in more effort to addressing this problem.
I know for a fact that not using WFSM will cause problems (observed initially here https://github.com/anza-xyz/agave/pull/8374) which is why I added the ability to specify it in the first place.

bw-solana (approved):

PR #593, opened 2025-11-08
fast leader handover: only notify frozen slots with valid banks

ksn6:
In replay, we currently hard unwrap to access banks for fast leader handover optimistic parent notification. These banks may not exist if the latest rooted slot is past the slot we're analyzing.

Accordingly, we should only process banks that exist in bank forks.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -33,6 +33,7 @@ use {
         window_service::DuplicateSlotReceiver,
     },
     crossbeam_channel::{Receiver, RecvTimeoutError, Sender},
+    itertools::Itertools,
     rayon::{
         iter::{IntoParallelIterator, ParallelIterator},
         ThreadPool,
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -869,16 +869,17 @@ impl ReplayStage {
...
+                        .filter_map(|slot| bank_forks_r.get(*slot));
+                    for bank in fast_leader_handover_notifies {
                         Self::maybe_notify_of_optimistic_parent(
                             &bank,
                             &my_pubkey,
                             &leader_schedule_cache,
                             &optimistic_parent_sender,
                         );
                     }
-                    drop(bank_forks_r);
```
AshwinSekar, line 881: still want to drop no? guess we can move this up to before the for loop

ksn6: oops yep

AshwinSekar (dismissed):

AshwinSekar (approved):

PR #594, opened 2025-11-10
feat: perform all `BlockComponent` checks in blockstore processor

ksn6:
Right now, we need to complete replay to verify that `BlockComponent`s in a block satisfy a number of invariants (e.g., exactly one header, exactly one footer, etc.).

@AshwinSekar found a clever way to avoid this: https://github.com/anza-xyz/alpenglow/pull/575#discussion_r2508438828.

Implement the above.

### DIFF
--- a/ledger/src/blockstore_processor.rs
+++ b/ledger/src/blockstore_processor.rs
@@ -1547,8 +1547,9 @@ pub fn confirm_slot(
             BlockComponent::EntryBatch(entries) => {
                 let slot_full = slot_full && ix == last_entry_batch_index.unwrap();
 
-                // Skip block component validation for genesis block
-                if bank.parent().is_some() {
+                // Skip block component validation for genesis block. Slot 0 is handled specially,
+                // since it won't have the required block markers (e.g., the header and the footer).
+                if bank.slot() != 0 {
                     processor
                         .on_entry_batch(migration_status, is_final)
                         .inspect_err(|err| {


runtime/src/block_component_processor.rs
```diff
@@ -53,12 +42,41 @@ impl BlockComponentProcessor {
...
     pub fn on_marker(
         &mut self,
         bank: Arc<Bank>,
         parent_bank: Arc<Bank>,
         marker: &VersionedBlockMarker,
+        migration_status: &MigrationStatus,
+        is_final: bool,
     ) -> Result<(), BlockComponentProcessorError> {
+        // Pre-migration: blocks with block components should be marked as dead
+        if !migration_status.is_alpenglow_enabled() {
```
AshwinSekar, line 75: we need to be able to support replaying across the migration boundary so i would use `should_have_alpenglow_ticks(bank.slot())` (or create a separate helper `should_have_block_markers`)

ksn6: done

ksn6: reverting, as discussed offline

ledger/src/blockstore_processor.rs
```diff
@@ -1541,11 +1541,21 @@ pub fn confirm_slot(
         .enumerate()
     {
         let num_shreds = completed_range.end - completed_range.start;
+        let is_final = slot_full && ix == completed_ranges.len() - 1;
 
         match component {
             BlockComponent::EntryBatch(entries) => {
                 let slot_full = slot_full && ix == last_entry_batch_index.unwrap();
 
+                // Skip block component validation for genesis block
```
AshwinSekar: Hmm I thought that the genesis block doesn't have any shreds associated with it?
Been a while since i traced through bank0 logic, what did you observe if we omit this check?

ksn6: If we omit this check, we fail here:

https://github.com/anza-xyz/alpenglow/blob/master/local-cluster/src/local_cluster.rs#L431-L448

with this error:

```assume successful validator start: failed to process blockstore from genesis: failed to replay bank 0, did you forget to provide a snapshot```

AshwinSekar: ah figured it out https://github.com/anza-xyz/alpenglow/blob/7437655eecf62cdb77dc354fc0291fee4a08da19/ledger/src/blockstore.rs#L5349 we create a block with 1 entry batch from genesis.

Do you mind making this check `bank.slot() != 0` and updating the comment to say that block 0 is handled specially because it will not have the required block markers?

Either that or we can update the genesis shreds to be correct (e.g. insert a header and footer).

ksn6: nice! done

PR #595, opened 2025-11-10
feat: update clock sysvar in `BlockComponentProcessor`

ksn6:
Now that we can process `BlockComponent`s during replay, update the clock sysvar when we observe the block footer.

Rather than constructing the footer purely in poh recorder, we split construction across block creation loop and poh recorder to simplify the introduction of rewards into the footer in a future PR.

This PR exposes the Alpenglow clock to `BlockComponentProcessor`, which we'll next modify to enforce the Alpenglow clock bounds outlined in https://github.com/solana-foundation/solana-improvement-documents/pull/363.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -2026,11 +2026,11 @@ impl Bank {
             .unwrap_or_default()
     }
 
-    pub fn update_clock_from_footer(&self, unix_timestamp: UnixTimestamp) {
+    pub fn update_clock_from_footer(&self, unix_timestamp_nanos: i64) {
         let mut epoch_start_timestamp =
             // On epoch boundaries, update epoch_start_timestamp
             if self.parent().is_some() && self.parent().unwrap().epoch() != self.epoch() {
-                unix_timestamp
+                unix_timestamp_nanos / 1_000_000_000
             } else {
                 self.clock().epoch_start_timestamp
             };
@@ -2047,7 +2047,7 @@ impl Bank {
             epoch_start_timestamp,
             epoch: self.epoch_schedule().get_epoch(self.slot),
             leader_schedule_epoch: self.epoch_schedule().get_leader_schedule_epoch(self.slot),
-            unix_timestamp: unix_timestamp / 1_000_000_000,
+            unix_timestamp: unix_timestamp_nanos / 1_000_000_000,
         };
         self.update_sysvar_account(&sysvar::clock::id(), |account| {
             create_account(
--- a/runtime/src/block_component_processor.rs
+++ b/runtime/src/block_component_processor.rs
@@ -1,6 +1,5 @@
 use {
     crate::bank::Bank,
-    solana_clock::UnixTimestamp,
     solana_entry::block_component::{
         BlockFooterV1, BlockMarkerV1, VersionedBlockFooter, VersionedBlockHeader,
         VersionedBlockMarker,
@@ -136,7 +135,7 @@ impl BlockComponentProcessor {
 
     pub fn update_bank_with_footer(bank: Arc<Bank>, footer: &BlockFooterV1) {
         // Update clock sysvar
-        bank.update_clock_from_footer(footer.block_producer_time_nanos as UnixTimestamp);
+        bank.update_clock_from_footer(footer.block_producer_time_nanos as i64);
 
         // TODO: rewards
     }
@@ -525,5 +524,50 @@ mod tests {
         // Process entry batch with header but not full - should succeed even without footer
         let result = processor.on_entry_batch(&migration_status, false);
         assert!(result.is_ok());
+    }
+
+    #[test]
+    fn test_footer_sets_epoch_start_timestamp_on_epoch_change() {
+        let mut processor = BlockComponentProcessor {
+            has_header: true,
+            ..Default::default()
+        };
+
+        // Create genesis bank
+        let genesis_config_info = create_genesis_config(10_000);
+        let genesis_bank = Arc::new(Bank::new_for_tests(&genesis_config_info.genesis_config));
+
+        // Get epoch schedule to find first slot of next epoch
+        let epoch_schedule = genesis_bank.epoch_schedule();
+        let first_slot_in_epoch_1 = epoch_schedule.get_first_slot_in_epoch(1);
+
+        // Create parent bank at last slot of epoch 0
+        let mut parent = genesis_bank.clone();
+        for slot in 1..first_slot_in_epoch_1 {
+            parent = create_child_bank(&parent, slot);
+        }
+
+        // Create bank at first slot of epoch 1
+        let bank = create_child_bank(&parent, first_slot_in_epoch_1);
+
+        // Verify we're in epoch 1
+        assert_eq!(bank.epoch(), 1);
+
+        // Process footer with a specific timestamp
+        let footer_time_nanos = 1_500_000_000_000_000_000;
+        let expected_time_secs = 1_500_000_000;
+        let footer = VersionedBlockFooter::V1(BlockFooterV1 {
+            bank_hash: Hash::new_unique(),
+            block_producer_time_nanos: footer_time_nanos,
+            block_user_agent: vec![],
+        });
+
+        processor.on_footer(bank.clone(), parent, &footer).unwrap();
+
+        // Verify clock sysvar was updated
+        assert_eq!(bank.clock().unix_timestamp, expected_time_secs);
+
+        // Verify epoch_start_timestamp was set correctly for the new epoch
+        assert_eq!(bank.clock().epoch_start_timestamp, expected_time_secs);
     }
 }


runtime/src/bank.rs
```diff
@@ -2021,6 +2021,34 @@ impl Bank {
             .unwrap_or_default()
     }
 
+    pub fn set_clock(&self, parent_epoch: Option<Epoch>, unix_timestamp: UnixTimestamp) {
```
wen-coding: this name is very similar to update_clock, should we call it `set_clock_sysvar_in_alpenglow` or something like that?

runtime/src/bank.rs
```diff
@@ -2021,6 +2021,34 @@ impl Bank {
...
+        let mut epoch_start_timestamp =
+            // On epoch boundaries, update epoch_start_timestamp
+            if parent_epoch.is_some() && parent_epoch.unwrap() != self.epoch() {
+                unix_timestamp
```
wen-coding: Hmm, do you need to consider how many slots you are into this Epoch? We could theoretically skip the first 100 slots and start with slot 101 here?

runtime/src/bank.rs
```diff
@@ -2021,6 +2021,34 @@ impl Bank {
...
+            } else {
+                self.clock().epoch_start_timestamp
+            };
+
+        if self.slot == 0 {
+            epoch_start_timestamp = self.unix_timestamp_from_genesis();
```
wen-coding, line 2039: This can go into the same section as line 2030?

runtime/src/block_component_processor.rs
```diff
@@ -73,14 +74,24 @@ impl BlockComponentProcessor {
...
     ) -> Result<(), BlockComponentProcessorError> {
         if self.has_footer {
             return Err(BlockComponentProcessorError::MultipleBlockFooters);
         }
 
+        // Update the clock sysvar
+        let parent_epoch = parent_bank.epoch();
+        let block_producer_time_nanos = match footer {
+            VersionedBlockFooter::V1(footer) | VersionedBlockFooter::Current(footer) => {
+                footer.block_producer_time_nanos
```
wen-coding: Is it possible that the block footer hasn't specified time?

runtime/src/block_component_processor.rs
```diff
@@ -73,14 +74,24 @@ impl BlockComponentProcessor {
...
+            }
+        } as UnixTimestamp;
+
+        bank.set_clock(Some(parent_epoch), block_producer_time_nanos);
```
wen-coding: Where do you check that the time we are checking in isn't far away from what the parent specified?

runtime/src/bank.rs
```diff
@@ -2021,6 +2021,34 @@ impl Bank {
             .unwrap_or_default()
     }
 
+    pub fn set_clock(&self, parent_epoch: Option<Epoch>, unix_timestamp: UnixTimestamp) {
+        let mut epoch_start_timestamp =
+            // On epoch boundaries, update epoch_start_timestamp
+            if parent_epoch.is_some() && parent_epoch.unwrap() != self.epoch() {
```
AshwinSekar: can use `self.parent().epoch()` to avoid having to pass this around

runtime/src/block_component_processor.rs
AshwinSekar: yeah shouldn't we validate that `block_producer_time_nanos > parent.clock()` and that
`parent.clock() + (bank.slot() - parent.slot()) * 800ms >= block_producer_time_nanos`

runtime/src/bank.rs
ksn6: done

runtime/src/bank.rs
ksn6: this shouldn't matter; the function here is only invoked when replaying block components / setting the clock from the footer

basically, the clock sysvar should be set consistently across all validators, which means we should be all set here

runtime/src/bank.rs
ksn6: I prefer the separation, since the previous block addresses epoch boundary things while this block addresses the one-off genesis case.

runtime/src/block_component_processor.rs
ksn6: yep, we'll do that in an upcoming PR - we'll be making a few small PRs rather than do this in a single shot

e.g., we additionally need to ensure that when a leader produces a clock value, they don't just take the "current clock value," but rather ensure that the leader clock satisfies the above inequality

for now, we're just setting the clock sysvar

runtime/src/block_component_processor.rs
ksn6: yep, see https://github.com/anza-xyz/alpenglow/pull/595#discussion_r2512116327

the consensus checks that we add will address this

runtime/src/bank.rs
ksn6: done

core/src/block_creation_loop.rs
```diff
@@ -404,8 +406,18 @@ fn record_and_complete_block(
     // will properly increment the tick_height to max_tick_height.
     bank.set_tick_height(max_tick_height - 1);
     // Write the single tick for this slot
+
+    // Produce the footer
+    // - fill in the bank hash after poh recorder ticks, and the bank freezes
+    // - fill in the block producer time once we have working_bank in poh_recorder
+    let footer = BlockFooterV1 {
+        bank_hash: Hash::default(),
+        block_producer_time_nanos: 0,
```
AshwinSekar: seems fine to fill this in here right? then we can keep the time calculation in block creation loop rather than PoHRecorder?

runtime/src/block_component_processor.rs
```diff
@@ -125,6 +127,13 @@ impl BlockComponentProcessor {
         self.has_header = true;
         Ok(())
     }
+
+    pub fn update_bank_with_footer(bank: Arc<Bank>, footer: &BlockFooterV1) {
```
AshwinSekar, line 136: should this also be called above in `on_footer`?

poh/src/poh_recorder.rs
```diff
@@ -977,34 +985,21 @@ impl PohRecorder {
...
-            block_producer_time_nanos: self.working_bank_block_producer_time_nanos(),
-            block_user_agent: format!("agave/{}", version!()).into_bytes(),
-        };
-
-        let footer = VersionedBlockFooter::Current(footer);
-        let footer = BlockMarkerV1::BlockFooter(footer);
-
-        VersionedBlockMarker::Current(footer)
-    }
+        BlockComponentProcessor::update_bank_with_footer(
```
AshwinSekar: dunno if it's possible to organize it this way but it would be really nice if we could just call `on_marker` (or if you're planning to put processing in a separate fn something like `process_marker`) here, so that there's no possibility of divergence with the non-leader code

runtime/src/bank.rs
```diff
@@ -2021,6 +2021,34 @@ impl Bank {
...
+        };
+        self.update_sysvar_account(&sysvar::clock::id(), |account| {
+            create_account(
+                &clock,
+                self.inherit_specially_retained_account_fields(account),
+            )
+        });
+    }
+
     fn update_clock(&self, parent_epoch: Option<Epoch>) {
```
AshwinSekar, line 2060: we'll have to ensure that this isn't called. if i'm reading the code correctly `set_alpenglow_clock_sysvar` will immediately be clobbered by `update_clock` in the next bank

core/src/block_creation_loop.rs
ksn6: yep, works - was having trouble reading from poh recorder here before, but realized that we already had a write lock 🤦

runtime/src/block_component_processor.rs
ksn6: oops, yep

poh/src/poh_recorder.rs
ksn6: that's tricky atm, since `on_footer` runs checks which depend on all block components being run through the processor sequentially

agreed that this is a desirable property, though - will think about how to achieve this; probably in a later PR, though

runtime/src/bank.rs
ksn6: nice catch - I removed this from `_new_from_parent` in `runtime/src/bank.rs`

we probably still want the `new_from_genesis` invocation + use-cases in tests

ksn6: actually, removing the `update_clock()` from `_new_from_parent` causes a bunch of tests in `solana-bpf-loader-program-tests` to fail:

https://buildkite.com/anza/alpenglow/builds/2578#019a7114-93db-418f-9037-5854d2f46d01

but this works:

```rust
    fn _new_from_parent(
        parent: Arc<Bank>,
        collector_id: &Pubkey,
        slot: Slot,
        reward_calc_tracer: Option<impl RewardCalcTracer>,
        new_bank_options: NewBankOptions,
    ) -> Self {
        ...
        // Update sysvars before processing transactions
        let (_, update_sysvars_time_us) = measure_us!({
            new.update_slot_hashes();
            new.update_stake_history(Some(parent.epoch()));

            // New logic
            let parent_timestamp = new.parent().unwrap().clock().unix_timestamp;
            new.set_alpenglow_clock_sysvar(parent_timestamp);

            // Old logic
            // new.update_clock(Some(parent.epoch()));

            new.update_last_restart_slot()
        });
        ...
    }
```

but we can't just replace the old logic with the new, since that'd result in bank mismatches - instead, we'll want to condition on `migration_status`

but this requires passing `migration_status` in to `_new_from_parent` 😵‍💫 

thoughts on how to proceed?

AshwinSekar: you can check if the bank has an alpenglow certificate (there's a helper fn), if it does then don't call update_clock

that should work in this case without migration status

runtime/src/bank.rs
```diff
@@ -2021,6 +2020,34 @@ impl Bank {
             .unwrap_or_default()
     }
 
+    pub fn set_alpenglow_clock_sysvar(&self, unix_timestamp: UnixTimestamp) {
```
AshwinSekar: this makes it sound like it's a separate sysvar for alpenglow - maybe `update_clock_from_footer` ?

ksn6: done

runtime/src/bank.rs
ksn6: nice - done

runtime/src/bank.rs
```diff
@@ -1391,7 +1391,15 @@ impl Bank {
         let (_, update_sysvars_time_us) = measure_us!({
             new.update_slot_hashes();
             new.update_stake_history(Some(parent.epoch()));
-            new.update_clock(Some(parent.epoch()));
+
+            // If Alpenglow is enabled, update the clock from the footer.
+            if new.get_alpenglow_genesis_certificate().is_some() {
+                let parent_timestamp = new.parent().unwrap().clock().unix_timestamp;
+                new.update_clock_from_footer(parent_timestamp);
```
AshwinSekar: we actually don't want this right? since we're then updating the clock sysvar twice per bank?

core/src/block_creation_loop.rs
```diff
@@ -404,8 +407,30 @@ fn record_and_complete_block(
     // will properly increment the tick_height to max_tick_height.
     bank.set_tick_height(max_tick_height - 1);
     // Write the single tick for this slot
+
+    // Determine the footer's timestamp
+    let working_bank = w_poh_recorder.working_bank().unwrap();
+    let block_producer_time_nanos = working_bank
```
AshwinSekar: nit: put this in a `fn produce_footer(working_bank)`

poh/src/poh_recorder.rs
```diff
@@ -605,7 +604,7 @@ impl PohRecorder {
             for (entry, tick_height) in &self.tick_cache[..entry_count] {
                 working_bank.bank.register_tick(&entry.hash);
 
-                if is_alpentick {
+                if let Some(footer) = footer.as_ref() {
                     // Wait for the bank to be frozen with timeout
```
AshwinSekar: nit: we put the logic inside this if into a separate function?

runtime/src/bank.rs
ksn6: ah, yep fair point - the clock stays constant from the parent clock until the end of the block

core/src/block_creation_loop.rs

poh/src/poh_recorder.rs
ksn6: sure, done

runtime/src/bank.rs
```diff
@@ -2021,6 +2026,34 @@ impl Bank {
...
+        }
+
+        let clock = sysvar::clock::Clock {
+            slot: self.slot,
+            epoch_start_timestamp,
+            epoch: self.epoch_schedule().get_epoch(self.slot),
+            leader_schedule_epoch: self.epoch_schedule().get_leader_schedule_epoch(self.slot),
+            unix_timestamp,
+        };
+        self.update_sysvar_account(&sysvar::clock::id(), |account| {
```
AshwinSekar, line 2052: did we additionally want to update the off-curve account for the nano second time? or punting that for a later PR?

runtime/src/block_component_processor.rs
```diff
@@ -125,6 +133,13 @@ impl BlockComponentProcessor {
...
+        // Update clock sysvar
+        bank.update_clock_from_footer(footer.block_producer_time_nanos as UnixTimestamp);
```
AshwinSekar: we need to convert this back to seconds right?

poh/src/poh_recorder.rs
AshwinSekar: yeah makes sense, atm header doesn't affect any state change so I guess not really a point in processing it on the leader side.

I suppose if we wanted we could use the same idea of creating the header in block creation loop and processing it after creating the new bank and then letting broadcast fill in the parent block id

runtime/src/bank.rs
ksn6: we'll do that in the next PR with the clock checks

runtime/src/block_component_processor.rs
ksn6: yep - adding tests for this now as well

runtime/src/bank.rs
AshwinSekar: This one is in seconds too right? so I think you need to divide by 1 bil here as well?

runtime/src/bank.rs
```diff
@@ -2021,7 +2026,7 @@ impl Bank {
             .unwrap_or_default()
     }
 
-    pub fn set_alpenglow_clock_sysvar(&self, unix_timestamp: UnixTimestamp) {
+    pub fn update_clock_from_footer(&self, unix_timestamp: UnixTimestamp) {
```
AshwinSekar: might be clearer to use a different type for `UnixTimestamp` as the doc comment on this type specifies seconds.

runtime/src/bank.rs
ksn6: good catch - yes

runtime/src/bank.rs
ksn6: done

AshwinSekar (approved):

PR #596, opened 2025-11-10
feat: skew leader clock in `poh-recorder`

ksn6:
The leader shouldn't populate the block footer clock with just the raw clock value.

Rather, the leader should populate the clock with bounds, as specified in https://github.com/solana-foundation/solana-improvement-documents/pull/363.

### DIFF
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -20,7 +20,7 @@ use {
     arc_swap::ArcSwapOption,
     crossbeam_channel::{bounded, unbounded, Receiver, SendError, Sender, TrySendError},
     log::*,
-    solana_clock::{Slot, NUM_CONSECUTIVE_LEADER_SLOTS},
+    solana_clock::{Slot, DEFAULT_MS_PER_SLOT, NUM_CONSECUTIVE_LEADER_SLOTS},
     solana_entry::{
         block_component::{
             BlockFooterV1, BlockMarkerV1, VersionedBlockFooter, VersionedBlockMarker,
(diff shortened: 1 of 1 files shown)


poh/src/poh_recorder.rs
```diff
@@ -977,13 +977,44 @@ impl PohRecorder {
...
+        let min_working_bank_time = parent_time.saturating_add(1);
+        let max_working_bank_time =
+            parent_time.saturating_add(2 * diff_slots * default_ns_per_slot);
+
+        working_bank_time
+            .max(min_working_bank_time)
+            .min(max_working_bank_time)
+    }
+
     pub fn working_bank_block_producer_time_nanos(&self) -> u64 {
```
AshwinSekar, line 998: this value also needs to go into the clock sysvar right? seems we should do it in bank.freeze

ksn6: this should already happen, no?

- when the leader sends out the footer, broadcast populates blockstore

- and then on the leader's replay, block component processor runs

- once https://github.com/anza-xyz/alpenglow/pull/595 lands, the clock sysvar gets populated

AshwinSekar: 2 things:

We only send out the footer once the bank is frozen. Once a bank is frozen no more account updates should be made (the state of the bank is "frozen").

There is no "leader replay", replay just checks to see if the bank is full and calls freeze. The blockstore processor call line is only used for non leader banks
https://github.com/anza-xyz/alpenglow/blob/4bfe9ed011926f73f6cda04b00ceb07cfce881a6/core/src/replay_stage.rs#L3464-L3466
For leader banks transactions are recorded into the bank directly by banking stage https://github.com/anza-xyz/alpenglow/blob/4bfe9ed011926f73f6cda04b00ceb07cfce881a6/core/src/banking_stage/consumer.rs#L174
And ticks are registered directly via poh recorder.

This is why a bank can become frozen before it has been done shredding

AshwinSekar: So i think when we freeze our bank we should create and update the clock, as well as apply the state changes associated with the rewards.
Then when we're producing our footer we can grab these details for shredding.

Basically need to ensure that the leader and replayer will end up with the same bank hash in the end.

PR #597, opened 2025-11-11
feat: alpenglow clock checks

ksn6:
We enforce Alpenglow clock checks, outlined in detail here: https://github.com/solana-foundation/solana-improvement-documents/pull/363.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -21,9 +21,7 @@ use {
     },
     solana_measure::measure::Measure,
     solana_poh::{
-        poh_recorder::{
-            PohRecorder, PohRecorderError, WorkingBank, GRACE_TICKS_FACTOR, MAX_GRACE_SLOTS,
-        },
+        poh_recorder::{PohRecorder, PohRecorderError, GRACE_TICKS_FACTOR, MAX_GRACE_SLOTS},
         record_channels::RecordReceiver,
     },
     solana_pubkey::Pubkey,
@@ -371,29 +369,25 @@ fn skew_block_producer_time_nanos(
 
 /// Produces a block footer with the current timestamp and version information.
 /// The bank_hash field is left as default and will be filled in after the bank freezes.
-fn produce_block_footer(working_bank: &WorkingBank) -> BlockFooterV1 {
+fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
     let mut block_producer_time_nanos = SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .expect("Misconfigured system clock; couldn't measure block producer time.")
         .as_nanos() as i64;
 
-    let working_bank_slot = working_bank.bank.slot();
+    let slot = bank.slot();
 
-    if let Some(parent_bank) = working_bank.bank.parent() {
+    if let Some(parent_bank) = bank.parent() {
         // Get parent time from alpenglow clock (nanoseconds) or fall back to clock sysvar (seconds -> nanoseconds)
-        let parent_time_nanos = working_bank.bank.get_nanosecond_clock().unwrap_or_else(|| {
-            working_bank
-                .bank
-                .clock()
-                .unix_timestamp
-                .saturating_mul(1_000_000_000)
-        });
+        let parent_time_nanos = bank
+            .get_nanosecond_clock()
+            .unwrap_or_else(|| bank.clock().unix_timestamp.saturating_mul(1_000_000_000));
         let parent_slot = parent_bank.slot();
 
         block_producer_time_nanos = skew_block_producer_time_nanos(
             parent_slot,
             parent_time_nanos,
-            working_bank_slot,
+            slot,
             block_producer_time_nanos,
         );
     }
@@ -468,7 +462,7 @@ fn record_and_complete_block(
 
     // Produce the footer with the current timestamp
     let working_bank = w_poh_recorder.working_bank().unwrap();
-    let footer = produce_block_footer(working_bank);
+    let footer = produce_block_footer(working_bank.bank.clone_without_scheduler());
 
     BlockComponentProcessor::update_bank_with_footer(
         working_bank.bank.clone_without_scheduler(),


core/src/block_creation_loop.rs
```diff
@@ -347,17 +349,58 @@ fn produce_window(
...
+        .min(max_working_bank_time)
+}
+
 /// Produces a block footer with the current timestamp and version information.
 /// The bank_hash field is left as default and will be filled in after the bank freezes.
-fn produce_block_footer(block_producer_start_time: SystemTime) -> BlockFooterV1 {
-    let block_producer_time_nanos = block_producer_start_time
+fn produce_block_footer(working_bank: &WorkingBank) -> BlockFooterV1 {
+    let mut block_producer_time_nanos = working_bank
+        .start
```
AshwinSekar: should we be using the current time? this will report the time from the start of this slot `s` but then set the value for use in slot `s + 1`

runtime/src/bank.rs
```diff
@@ -215,6 +216,14 @@ pub const SECONDS_PER_YEAR: f64 = 365.25 * 24.0 * 60.0 * 60.0;
 
 pub const MAX_LEADER_SCHEDULE_STAKES: Epoch = 5;
 
+/// The off-curve account where we store the Alpenglow clock. The clock sysvar has seconds
+/// resolution while the Alpenglow clock has nanosecond resolution.
+pub static ALPENGLOW_CLOCK_ACCOUNT: LazyLock<Pubkey> = LazyLock::new(|| {
```
AshwinSekar: nit would remove the `ALPENGLOW` prefix here and elsewhere for something else, maybe `NANOSECOND_CLOCK_ACCOUNT`?

AshwinSekar: Also I think it's best we remove the `pub` modifier here, i'd rather this not be public api (yet)

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
...
+    }
+
+    /// Given the parent slot, parent time, and working bank slot, calculate the lower and upper
+    /// bounds for the block producer time. We return (lower_bound, upper_bound), where both bounds
+    /// are inclusive. I.e., the working bank time is valid if
+    /// lower_bound <= working_bank_time <= upper_bound.
+    pub fn alpenglow_time_bounds(
+        parent_slot: Slot,
+        parent_time_nanos: i64,
+        working_bank_slot: Slot,
```
AshwinSekar: nit: working_bank is typically used in leader context, would update this to just be slot

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
         Ok(())
     }
 
+    fn enforce_alpenglow_clock_bounds(
```
AshwinSekar: nit: same on removing the alpenglow prefix in favor of nanosecond

runtime/src/bank.rs
```diff
@@ -2052,6 +2061,24 @@ impl Bank {
...
+        self.store_account_and_update_capitalization(
+            &ALPENGLOW_CLOCK_ACCOUNT,
+            &alpenclock_account_data,
+        );
+    }
+
+    pub fn get_alpenglow_clock(&self) -> Option<i64> {
+        self.get_account(&ALPENGLOW_CLOCK_ACCOUNT).map(|acct| {
+            acct.deserialize_data()
+                .expect("Couldn't deserialize Alpenglow clock")
```
wen-coding: Would we every get invalid data here?

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
...
+        bank: Arc<Bank>,
+        parent_bank: Arc<Bank>,
+        footer: &BlockFooterV1,
```
wen-coding, line 143: dunno how you handle BlockFooter versioning, but I sort of expected BlockFooterCurrent here and footer.block_producer_time_nanos would work in later vesions.

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
...
+    ) -> Result<(), BlockComponentProcessorError> {
+        // Get parent time from alpenglow clock account (nanosecond precision)
+        // If alpenglow clock hasn't been populated, don't enforce the bounds.
```
wen-coding: I have a dumb question, so if someone didn't populate clock in between, will our clock information be lost?

e.g. I'm the leader for block 12~16, I want to dial the clock forward. Can I do this:
1. do not populate clock sysvar for block 12
2. set clock sysvar to whatever I want in block 13, since its parent (block 12) has no clock sysvar, can I ignore whatever the value is in block 11 and just stamp my own time here?

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
...
+        match is_valid {
+            true => Ok(()),
+            false => Err(BlockComponentProcessorError::AlpenglowClockOutOfBounds),
+        }
+    }
+
+    /// Given the parent slot, parent time, and working bank slot, calculate the lower and upper
+    /// bounds for the block producer time. We return (lower_bound, upper_bound), where both bounds
+    /// are inclusive. I.e., the working bank time is valid if
+    /// lower_bound <= working_bank_time <= upper_bound.
```
wen-coding, line 171: nit: maybe refer to SIMD for actual bounds calculation here.

runtime/src/bank.rs
```diff
@@ -2052,6 +2061,24 @@ impl Bank {
                 self.inherit_specially_retained_account_fields(account),
             )
         });
+
+        // Update Alpenglow clock
```
wen-coding, line 2065: Yeah, I do agree with Ashwin about using clock nano sysvar or something like that vs Alpenglow clock. This thing might out-live Alpenglow

runtime/src/block_component_processor.rs
```diff
@@ -133,6 +137,52 @@ impl BlockComponentProcessor {
...
+        parent_bank: Arc<Bank>,
+        footer: &BlockFooterV1,
+    ) -> Result<(), BlockComponentProcessorError> {
+        // Get parent time from alpenglow clock account (nanosecond precision)
+        // If alpenglow clock hasn't been populated, don't enforce the bounds.
+        let Some(parent_time_nanos) = parent_bank.get_alpenglow_clock() else {
+            return Ok(());
+        };
+
+        let parent_slot = parent_bank.slot();
```
wen-coding, line 152: nit: this is done at the end of block so there is no possibility we will do a UpdateParent later right? Might be good to explain that somewhere I guess.

runtime/src/block_component_processor.rs
```diff
@@ -222,9 +272,13 @@ mod tests {
         let parent = create_test_bank();
         let bank = create_child_bank(&parent, 1);
 
+        // Calculate valid timestamp based on parent's time
+        let parent_time_nanos = parent.clock().unix_timestamp.saturating_mul(1_000_000_000);
```
wen-coding, line 280: nit: can you explain what you would do if parent has both clock and nano_clock sysvar but the two differs?

Also, after Alpenglow do we intend to keep populating the clock sysvar?

core/src/block_creation_loop.rs
ksn6: sure, can change this - the precise definition of "what time a block is produced" isn't particularly well-defined in the clock SIMD; we can probably roll with either the current time or the current approach - no strong opinions here

runtime/src/bank.rs
ksn6: done

runtime/src/block_component_processor.rs
ksn6: done

runtime/src/block_component_processor.rs
ksn6: done

runtime/src/bank.rs
ksn6: that can't happen, because as of now, only `update_clock_from_footer` writes to this account, and we only ever write an `i64` there, which should be deserializable

if the account simply doesn't exist, then this function returns `None`

runtime/src/block_component_processor.rs
ksn6: a lot of the footer versioning code is going to change once we rewrite block component in wincode

core/src/block_creation_loop.rs
```diff
@@ -347,17 +349,58 @@ fn produce_window(
...
         .duration_since(UNIX_EPOCH)
         .expect("Misconfigured system clock; couldn't measure block producer time.")
-        .as_nanos() as u64;
+        .as_nanos() as i64;
+
+    let working_bank_slot = working_bank.bank.slot();
+
+    if let Some(parent_bank) = working_bank.bank.parent() {
+        // Get parent time from alpenglow clock (nanoseconds) or fall back to clock sysvar (seconds -> nanoseconds)
+        let parent_time_nanos = parent_bank.get_alpenglow_clock().unwrap_or_else(|| {
```
AshwinSekar: I'm wondering if we'll have issues around the migration with this parent logic.
Can something like this happen?

Suppose the Alpenglow genesis block `G` (PoH block) has `100` in the clock sysvar, then in the first alpenglow block `G + 1`:
- lets say `update_clock` sets the clock sysvar to `101` based on the TowerBFT votes
- `block_producer_time_nanos` is skewed to `100.800`
- We attach `block_producer_time_nanos` in the footer and now the clock sysvar is updated to `100`

So if we query the clock sysvar:
- Block `G` -> `100`
- Block `G + 1` -> `101`
- Block `G + 2` -> `100` (technically `100.800` but truncated by dividing 1 bil)

AshwinSekar: Perhaps we should be using `working_bank` here to account for `update_clock` running on the first alpenglow block?

core/src/block_creation_loop.rs
AshwinSekar: yeah I think current time probably makes more sense with what the consumer would expect

runtime/src/bank.rs
wen-coding: Can we add a comment why we would never get invalid data here? Naive assumption just reading this code would appear that attackers writing invalid data here can make us panic.

runtime/src/block_component_processor.rs
ksn6: our clock information won't be lost - to clarify, "if alpenglow clock hasn't been populated" doesn't refer to a leader not setting the alpenglow clock account. rather, this phrase refers to "we aren't in alpenglow mode yet."

the way this currently works is - once we're in alpenglow mode, a leader sets the value of the clock in the off-curve account `NANOSECOND_CLOCK_ACCOUNT`

at that moment, everybody updates this account locally - if that doesn't happen, bank hash mismatches occur

this account then exists in perpetuity

moving forward, if a leader doesn't update this account / update the clock value correctly, their block gets marked dead either due to a bank hash mismatch or due to a clock bounds check failure

runtime/src/block_component_processor.rs
ksn6: done

runtime/src/bank.rs
ksn6: done

runtime/src/block_component_processor.rs
ksn6: not sure I follow - the decision to have the clock within the footer is somewhat arbitrary; it isn't related to fast leader handover - e.g., we could've totally placed the timestamp + associated checks in the header

wen-coding: I mean, you are using parent_bank.slot() here, if we can somehow still change parent via UpdateParent, then this is bad, the values are not correct any more. I guess it's not a problem because it's in block footer, but maybe add a comment.

runtime/src/block_component_processor.rs
wen-coding: Okay, let's update the comment maybe?

runtime/src/block_component_processor.rs
ksn6: the clock sysvar and nanosecond clock will always be different, since the clock sysvar measures time in seconds while the nanosecond clock measures time in nanoseconds

moving forward, setting the clock sysvar only ever happens through `update_clock_from_footer`; if a leader somehow tries to mess with the clock sysvar while properly advancing the nanosecond clock, this will result in a bank mismatch + the block being marked as dead, since non-leaders will end up with a different clock sysvar value

yep, post-Alpenglow, we'll want to continue populating the clock sysvar, since this is a field that SPL programs can access

runtime/src/bank.rs
ksn6: this is an off-curve account, so it's impossible for an attacker to write invalid data here via a txn (i.e., there exists no keypair whose public key is the clock account)

the only way to affect this field adversely is via the block footer's timestamp field, and the clock checks explicitly conduct checks to ensure that attacks aren't possible

lmk if this explanation works - can add a comment indicating this

runtime/src/block_component_processor.rs
ksn6: the checks are correct irrespective of a parent change, right?
 
the access to `parent_slot` here is correct, since we will have already updated bank forks prior to hitting this line of code in the block component processor

even if this code was in the header, we'd still be correct despite a potential update parent being present in the code

runtime/src/block_component_processor.rs
ksn6: done

core/src/block_creation_loop.rs

core/src/block_creation_loop.rs
```diff
@@ -425,7 +468,7 @@ fn record_and_complete_block(
 
     // Produce the footer with the current timestamp
     let working_bank = w_poh_recorder.working_bank().unwrap();
-    let footer = produce_block_footer(*working_bank.start);
+    let footer = produce_block_footer(working_bank);
```
AshwinSekar: can just use `bank` now

wen-coding (approved):

core/src/block_creation_loop.rs
```diff
@@ -468,7 +462,7 @@ fn record_and_complete_block(
 
     // Produce the footer with the current timestamp
     let working_bank = w_poh_recorder.working_bank().unwrap();
-    let footer = produce_block_footer(working_bank);
+    let footer = produce_block_footer(working_bank.bank.clone_without_scheduler());
```
AshwinSekar, line 465: oh i meant you already have `bank` above, so we don't need to grab the `working_bank()` anymore since they are the same thing, so you could use `bank` here and below in `update_bank_with_footer`

ksn6: oh true - will fix in a separate pr

PR #598, opened 2025-11-12
nit: minor refactor to remove a `mut` and a `unwrap`

akhi3030:
Doesn't save any lines of code but the new code feels more "rusty".

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -2028,15 +2028,12 @@ impl Bank {
 
     pub fn update_clock_from_footer(&self, unix_timestamp_nanos: i64) {
         // On epoch boundaries, update epoch_start_timestamp
-        let epoch_start_timestamp = if self.slot == 0 {
-            self.unix_timestamp_from_genesis()
-        } else {
-            match self.parent() {
-                Some(parent) if parent.epoch() != self.epoch() => {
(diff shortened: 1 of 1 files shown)


runtime/src/bank.rs
```diff
@@ -2027,17 +2027,17 @@ impl Bank {
...
+        let epoch_start_timestamp = if self.slot == 0 {
+            self.unix_timestamp_from_genesis()
+        } else {
+            match self.parent() {
+                Some(parent) if parent.epoch() != self.epoch() => {
+                    unix_timestamp_nanos / 1_000_000_000
+                }
+                _ => self.clock().epoch_start_timestamp,
+            }
+        };
```
ksn6, line 2037: thanks for this PR! how about:

```suggestion
        // On epoch boundaries, update epoch_start_timestamp
        let epoch_start_timestamp = match (self.slot, self.parent()) {
            (0, _) => self.unix_timestamp_from_genesis(),
            (_, Some(parent)) if parent.epoch() != self.epoch() => {
                unix_timestamp_nanos / 1_000_000_000
            }
            _ => self.clock().epoch_start_timestamp,
        };
```

akhi3030: love it!

PR #599, opened 2025-11-12
block_component: Add GenesisCertificate block marker

AshwinSekar:
#### Problem
We need some method to indicate the first Alpenglow block post migration

#### Summary of Changes
Add a block marker that will only be disseminated on the first alpenglow block (direct child of genesis).
This block marker holds the `GenesisCertificate`.

Future PR will populate this in BCL and process it in replay to update the genesis account.

Full details of migration in SIMD https://github.com/solana-foundation/solana-improvement-documents/pull/384/files#diff-9a9498ea503301d23f7af15f620fcb53b060f666419c9214a3fe72ff465affd9R118

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -82,14 +82,31 @@
 /// │ Parent Block ID             (32 bytes)  │
 /// └─────────────────────────────────────────┘
 /// ```
+///
+/// ### GenesisCertificate Layout
+/// ```text
+/// ┌─────────────────────────────────────────┐
+/// │ Genesis Slot                  (8 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Genesis Block ID             (32 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ BLS Signature               (192 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Bitmap length (max 512)       (8 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Bitmap                (up to 512 bytes) │
+/// └─────────────────────────────────────────┘
+/// ```
 use {
     crate::entry::Entry,
     serde::{
         de::{self, Visitor},
         Deserialize, Deserializer, Serialize, Serializer,
     },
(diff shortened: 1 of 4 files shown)


entry/src/block_component.rs
```diff
@@ -82,14 +82,31 @@
...
+/// ### GenesisCertificate Layout
+/// ```text
+/// ┌─────────────────────────────────────────┐
+/// │ Genesis Slot                  (8 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Genesis Block ID             (32 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ BLS Signature               (192 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Bitmap length (max 512)       (8 bytes) │
```
AshwinSekar, line 95: could make this 1 byte, but i don't think it's worth the impl complexity for a one off marker

entry/src/block_component.rs
```diff
@@ -187,6 +204,7 @@ impl From<bincode::Error> for BlockComponentError {
 /// └─────────────────────────────────────────┘
 /// ```
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[allow(clippy::large_enum_variant)]
```
AshwinSekar, line 207: don't think this really matters, we should be getting big `EntryBatch`s when it counts. Happy to box if y'all think otherwise

carllin: Each shred payload is essentially the size of this certificate right?

carllin (approved):

entry/src/block_component.rs
ksn6: fyi, in the SIMD, we can make this one byte since we'll have the ability to specify the number of bytes used to express a vec's size in wincode.

entry/src/block_component.rs
ksn6: nah, would avoid the box - the footer is about to get large 😅

ksn6 (approved):

PR #600, opened 2025-11-13
bcl: pack the GenesisCertificate into the first alpenglow block(s)

AshwinSekar:
#### Problem
The direct child(ren) of the alpenglow genesis block must contain the Alpenglow Genesis certificate. 
This is critical in order to update the off curve account for use in restarting post migration.

#### Summary of Changes
Send the genesis certificate block marker for these blocks. Next PR will plug in the processing of it during replay.

### DIFF
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -35,7 +35,6 @@ use {
     },
     solana_version::version,
     solana_votor::{common::block_timeout, event::LeaderWindowInfo},
-    solana_votor_messages::consensus_message::Certificate,
     stats::{BlockCreationLoopMetrics, SlotMetrics},
     std::{
         sync::{
@@ -117,7 +116,7 @@ struct LeaderContext {
     slot_metrics: SlotMetrics,
 
     // Migration information
-    genesis_cert: Arc<Certificate>,
+    genesis_cert: GenesisCertificate,
 }
 
 #[derive(Default)]
@@ -190,6 +189,8 @@ fn start_loop(config: BlockCreationLoopConfig) {
         .migration_status()
         .genesis_certificate()
         .expect("Migration complete, genesis certificate must exist");
+    let genesis_cert = GenesisCertificate::try_from((*genesis_cert).clone())
+        .expect("Genesis certificate must be valid");
 
     info!("{my_pubkey}: Block creation loop starting");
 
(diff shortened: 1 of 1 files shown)


core/src/block_creation_loop.rs
```diff
@@ -112,6 +115,9 @@ struct LeaderContext {
     // Metrics
     metrics: BlockCreationLoopMetrics,
     slot_metrics: SlotMetrics,
+
+    // Migration information
+    genesis_cert: Arc<Certificate>,
```
ksn6: should probably make this a `GenesisCertificate` rather than a `Certificate`

core/src/block_creation_loop.rs
```diff
@@ -178,6 +184,13 @@ fn start_loop(config: BlockCreationLoopConfig) {
         }
     };
 
+    let genesis_cert = bank_forks
+        .read()
+        .unwrap()
+        .migration_status()
+        .genesis_certificate()
+        .expect("Migration complete, genesis certificate must exist");
+
```
ksn6, line 194: why is migration guaranteed to be complete here?

core/src/block_creation_loop.rs
AshwinSekar: you end up cloning twice if you do the conversion here. Any reason to convert here?

core/src/block_creation_loop.rs
AshwinSekar: PohService has to have shut down and sent the record receiver above which means the migration is complete

AshwinSekar:
actually nvm i think this is better in broadcast - we want the genesis marker before the header for the first alpenglow block

core/src/block_creation_loop.rs
bw-solana: I think the advantage is type safety guarantees (no dumbass can try to stick wrong cert in here).

When you say you clone twice, can you specify how often that happens? Like is it just when we're actually creating the cert?

core/src/block_creation_loop.rs
```diff
@@ -640,6 +654,21 @@ fn create_and_insert_leader_bank(slot: Slot, parent_bank: Arc<Bank>, ctx: &mut L
...
+    if parent_slot == ctx.genesis_cert.cert_type.slot() {
+        let genesis_marker = VersionedBlockMarker::Current(BlockMarkerV1::GenesisCertificate(
+            GenesisCertificate::try_from((*ctx.genesis_cert).clone())
+                .expect("Genesis certificate must be valid"),
+        ));
+        ctx.poh_recorder
+            .write()
+            .unwrap()
+            .send_marker(genesis_marker)
+            .expect("Max tick height cannot have been reached");
```
bw-solana, line 668: is this technically racy but we have ~350ms? Or is there some strong ordering guarantee?

core/src/block_creation_loop.rs
AshwinSekar: yeah we clone when doing the conversion as `migration_status` only exposes an `Arc`.
Then we clone when actually sending the marker.

> I think the advantage is type safety guarantees

This is a fair point though, i'll put up a PR to change `Arc<Certificate>` to `Arc<GenesisCertificate>` now that the block marker is available.

core/src/block_creation_loop.rs
AshwinSekar: At this point we're single threaded on the `poh_recorder`. There's no way for the `working_bank` to have changed state between line 657 (`set_bank`) and now

core/src/block_creation_loop.rs
AshwinSekar: a14725eb53a86a55b83eb9cdf4e3c43c1d635175 does the conversion up front

Unfortunately can't change `MigrationStatus` to `GenesisCertificate` as that would require `solana-votor-messages` to depend on `solana-entry`.

PR #601, opened 2025-11-13
fast leader handover: parse `UpdateParent` on shred ingest

ksn6:
### Summary
Implements online detection and storage of `UpdateParent` `BlockMarker`s during shred ingestion. `UpdateParent` markers enable Alpenglow's fast leader handover by allowing blocks to signal parent block updates.

To do this efficiently, we first note that `BlockComponent`'s will be FEC set aligned. We then observe that an `UpdateParent` marker is a specific type of `BlockComponent::VersionedBlockMarker`.

Accordingly, we detect `UpdateParent` markers at FEC set boundaries as follows:

- **Case A**: When receiving a shred with `DATA_COMPLETE_SHRED = true`, signifying the last shred of a FEC set boundary, we check the 0th shred of the *next* FEC set for an `UpdateParent` marker

- **Case B**: When receiving a shred with `DATA_COMPLETE_SHRED = false`, we check whether the previous shred had `DATA_COMPLETE_SHRED = true`; if so, we examine the current shred for an `UpdateParent` marker

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -331,7 +331,7 @@ struct ParentMetaWorkingSetEntry {
     parent_meta: Option<WorkingEntry<ParentMeta>>,
     /// Whether we've already loaded this key from the blockstore. Once true, we use
     /// the cached `parent_meta` value instead of hitting the blockstore again.
-    hit_blockstore: bool,
+    fetched_from_blockstore: bool,
 }
 
 struct ShredInsertionTracker<'a> {
@@ -2273,10 +2273,9 @@ impl Blockstore {
                 return Ok(());
             }
         }
-
         // If we haven't hit blockstore for this key yet...
-        if !entry.hit_blockstore {
-            entry.hit_blockstore = true;
+        else if !entry.fetched_from_blockstore {
+            entry.fetched_from_blockstore = true;
 
             if let Some(parent_meta) = self.parent_meta_cf.get((slot, location))? {
                 entry.parent_meta = Some(WorkingEntry::Clean(parent_meta));


ledger/src/blockstore.rs
```diff
@@ -2044,6 +2189,15 @@ impl Blockstore {
             self.maybe_insert_parent_meta(&shred, slot, location, parent_meta_working_set);
         }
 
+        // Check for update parent
+        self.maybe_insert_update_parent(
```
AshwinSekar: we'll want to move this and `maybe_insert_parent_meta` to after insertion (see where we update the  `MerkleRootMeta`). If for whatever reason the shred is not actually inserted then we don't want to insert/update the `ParentMeta`

ledger/src/blockstore.rs
```diff
@@ -1395,11 +1396,47 @@ impl Blockstore {
...
             }
 
-            self.parent_meta_cf.put_in_batch(
-                &mut shred_insertion_tracker.write_batch,
-                (slot, location),
-                working_parent_meta.as_ref(),
-            )?;
+            let key = (slot, location);
+
+            let WorkingEntry::Dirty(new_parent_meta) = working_parent_meta else {
```
AshwinSekar: This logic is better placed when we actually update the parent meta rather than in this function where we are committing batches. You can grab the current `ParentMeta` before you call the `maybe` functions.

See the `MerkleRootMeta` logic for an example:
```rust
if let HashMapEntry::Vacant(entry) = merkle_root_metas.entry((location, erasure_set)) {
            if let Some(meta) = self
                .merkle_root_meta_from_location(erasure_set, location)
                .unwrap()
            {
                entry.insert(WorkingEntry::Clean(meta));
            }
        }
```

ledger/src/blockstore.rs
```diff
@@ -1988,6 +2025,114 @@ impl Blockstore {
...
+        let fec_set_index = current_shred.fec_set_index();
+        let data_complete = current_shred.data_complete();
+
+        // Determine which shred to check for UpdateParent and which FEC set it belongs to
+        //
+        // NOTE: this only works when SIMD-0317 and SIMD-0366 have been activated!
+        let (shred_bytes, target_fec_set_index) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + DATA_SHREDS_PER_FEC_BLOCK as u32;
```
AshwinSekar: This is also just the next shred right? You can do `current_index + 1`?

ledger/src/blockstore.rs
```diff
@@ -1988,6 +2025,114 @@ impl Blockstore {
...
+        just_inserted_shreds: &HashMap<(BlockLocation, ShredId), Cow<'_, Shred>>,
+        parent_meta_working_set: &mut HashMap<(BlockLocation, u64), WorkingEntry<ParentMeta>>,
+    ) {
+        let current_index = current_shred.index();
+        let fec_set_index = current_shred.fec_set_index();
+        let data_complete = current_shred.data_complete();
+
+        // Determine which shred to check for UpdateParent and which FEC set it belongs to
+        //
+        // NOTE: this only works when SIMD-0317 and SIMD-0366 have been activated!
```
AshwinSekar, line 2004: note to future self: make sure we activate these SIMDs first before alpenglow and update this comment

ledger/src/blockstore.rs
```diff
@@ -1988,6 +2025,114 @@ impl Blockstore {
...
+        let (shred_bytes, target_fec_set_index) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + DATA_SHREDS_PER_FEC_BLOCK as u32;
+            let shred_id = ShredId::new(slot, next_fec_set_index, ShredType::Data);
+
+            match self.get_shred_from_just_inserted_or_db(just_inserted_shreds, shred_id, location)
+            {
+                Some(payload) => (Some(payload), next_fec_set_index),
+                None => (None, next_fec_set_index),
```
AshwinSekar: I think it's cleaner to `let (Some) = ... else { return; }` rather than having all these `None`s to reduce the indentation in this fn

ledger/src/blockstore.rs
```diff
@@ -1988,6 +2025,114 @@ impl Blockstore {
...
+        } else {
+            (None, fec_set_index)
+        };
+
+        // Process the shred bytes if we have them
+        let Some(shred_bytes) = shred_bytes else {
+            return;
+        };
+
+        let Ok(payload) = shred::layout::get_data(&shred_bytes) else {
```
AshwinSekar:
```suggestion
        let Ok(payload) = shred::wire::get_data(&shred_bytes) else {
```

ledger/src/blockstore.rs
AshwinSekar: Actually I think it's even better if we don't enforce this here - if `ParentMeta` is already populated from `UpdateParent` then don't bother calling `maybe_insert_update_parent`. Instead the `block_component_processor` will mark dead if we see multiple update parents - saves us a bunch of blockstore lookups.

ledger/src/blockstore.rs
```diff
@@ -1988,6 +2025,114 @@ impl Blockstore {
             .or_insert(WorkingEntry::Dirty(parent_meta));
     }
 
+    /// Parse UpdateParent from data shred payload
+    fn parse_update_parent_from_data_payload(data: &[u8]) -> Option<(Slot, Hash)> {
```
AshwinSekar: I think this and the parse header fn are better placed in the `block_component.rs`. `blockstore.rs` is already humungous so avoid putting anything that doesn't need to be here.

ledger/src/blockstore_processor.rs
```diff
@@ -845,6 +845,15 @@ pub enum BlockstoreProcessorError {
 
     #[error("block component processor error: {0}")]
     BlockComponentProcessor(#[from] BlockComponentProcessorError),
+
```
AshwinSekar: revert this file right?

ledger/src/blockstore.rs
AshwinSekar: Then the code below becomes simpler as well, we can just take the one with higher `replay_fec_set_index`

ledger/src/blockstore.rs
ksn6: makes sense - done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore_processor.rs
ksn6: yep, done - I branched from a previous commit; these changes ended up sticking around

ledger/src/blockstore.rs
ksn6: done

entry/src/block_component.rs
```diff
@@ -618,6 +618,49 @@ impl BlockComponent {
             }
         }
     }
+
+    /// Parse BlockHeader from bytes
+    pub fn parse_block_header_from_data_payload(data: &[u8]) -> Option<(Slot, Hash)> {
+        // Try to deserialize as BlockComponent
+        let (component, _) = BlockComponent::from_bytes(data).ok()?;
```
wen-coding, line 625: Is it okay to discard all errors silently here? Should we log or add to stats?

entry/src/block_component.rs
```diff
@@ -618,6 +618,49 @@ impl BlockComponent {
...
+
+        // If we were able to parse the component, it must be a block marker
+        let marker = component.as_marker()?;
+
+        // Check if it's a BlockMarker with BlockHeader
+        match marker {
+            VersionedBlockMarker::V1(BlockMarkerV1::BlockHeader(header))
+            | VersionedBlockMarker::Current(BlockMarkerV1::BlockHeader(header)) => {
```
wen-coding, line 633: nit: `VersionBlockMarker::Current(...V1)` looks kind of weird, I would imagine something like `VersionBlockMarker::Current(...Current....)`

entry/src/block_component.rs
```diff
@@ -618,6 +618,49 @@ impl BlockComponent {
...
+
+    /// Parse UpdateParent from bytes
+    pub fn parse_update_parent_from_data_payload(data: &[u8]) -> Option<(Slot, Hash)> {
+        // Try to deserialize as BlockComponent
+        let (component, _) = BlockComponent::from_bytes(data).ok()?;
+
+        // Check if it's a BlockMarker with UpdateParent
+        match component.as_marker()? {
+            VersionedBlockMarker::V1(BlockMarkerV1::UpdateParent(versioned_update))
+            | VersionedBlockMarker::Current(BlockMarkerV1::UpdateParent(versioned_update)) => {
```
wen-coding, line 653: nit: Can VersionedBlockMarker hide all the details inside a function and this function only needs to call
`VersionedBlockMarker::get_update_parent_v1(...)`?

entry/src/block_component.rs
ksn6: probably a good idea to address this after rewriting `BlockComponent` in wincode

entry/src/block_component.rs
ksn6: same here - probably a good idea to address this after rewriting `BlockComponent` in wincode

ledger/src/blockstore.rs
ksn6: nice - done

entry/src/block_component.rs
ksn6: this should be okay - if this fails and is important, the slot gets marked as dead, at which point we'll end up logging

ledger/src/blockstore.rs
```diff
@@ -1988,6 +1963,86 @@ impl Blockstore {
...
+                })
+                .and_then(|flags| {
+                    flags
+                        .contains(ShredFlags::DATA_COMPLETE_SHRED)
+                        .then(|| Cow::Borrowed(current_shred.payload()))
+                });
+
+            (shred_bytes, fec_set_index)
+        } else {
+            (None, fec_set_index)
```
AshwinSekar: just return here

ledger/src/blockstore.rs
```diff
@@ -1988,6 +1963,86 @@ impl Blockstore {
...
+        //
+        // NOTE: this only works when SIMD-0317 and SIMD-0366 have been activated!
+        let (shred_bytes, target_fec_set_index) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + DATA_SHREDS_PER_FEC_BLOCK as u32;
+            let shred_id = ShredId::new(slot, next_fec_set_index, ShredType::Data);
+            let payload =
+                self.get_shred_from_just_inserted_or_db(just_inserted_shreds, shred_id, location);
+            (payload, next_fec_set_index)
```
AshwinSekar: naming is confusing you call some of them `payload` and some `shred_bytes` when referring to the same thing

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
```diff
@@ -1988,6 +1963,86 @@ impl Blockstore {
...
+        let (shred_bytes, target_fec_set_index) = if data_complete {
+            // Case (a): Current shred has DATA_COMPLETE=true (end of FEC set)
+            // Check the 0th shred in the NEXT FEC set for UpdateParent
+            let next_fec_set_index = fec_set_index + DATA_SHREDS_PER_FEC_BLOCK as u32;
+            let shred_id = ShredId::new(slot, next_fec_set_index, ShredType::Data);
+            let shred_bytes =
+                self.get_shred_from_just_inserted_or_db(just_inserted_shreds, shred_id, location);
+            (shred_bytes, next_fec_set_index)
+        } else if current_index % DATA_SHREDS_PER_FEC_BLOCK as u32 == 0 && current_index > 0 {
+            // Case (b): Current shred has DATA_COMPLETE=false
```
AshwinSekar:
```suggestion
            // Case (b): Current shred is the 0th shred of the FEC set
```

ledger/src/blockstore.rs
```diff
@@ -2159,9 +2218,47 @@ impl Blockstore {
...
 
+    fn maybe_update_parent_meta(
+        &self,
+        shred: &Shred,
+        location: BlockLocation,
+        slot: u64,
+        just_inserted_shreds: &mut HashMap<(BlockLocation, ShredId), Cow<'_, Shred>>,
+        parent_meta_working_set: &mut HashMap<(BlockLocation, u64), WorkingEntry<ParentMeta>>,
+    ) {
+        match parent_meta_working_set.get(&(location, slot)) {
```
AshwinSekar: Sorry missed this on the last review, you need to be able to fall back to the column if this is missing.
Otherwise you could end up in a scenario where something like this happens:
- In one shred insertion batch you receive an `UpdateParent` and populate the `parent_meta_working_set`
- At the end of the batch you commit the `parent_meta_working_set` to blockstore
- Next shred insertion batch you receive a header shred
- `parent_meta_working_set` is cleared between insertion batches so now you think we don't have a `ParentMeta` yet and overwrite the one populated by the `UpdateParent`

AshwinSekar: see https://github.com/anza-xyz/alpenglow/pull/601#discussion_r2525649682 for how we populate it for the `MerkleRootMeta` before usage in the merkle root checks. I imagine you could work in something similar here

AshwinSekar: The difference with the other trackers is that they are used on (almost) every single shred while `ParentMeta` is only useful when we're processing a `shred_index % 32 == 0 || shred.is_data_complete()`. 

To avoid hitting blockstore as often I would gate all this logic behind these easy checks first.

ledger/src/blockstore.rs
ksn6: done

entry/src/block_component.rs
wen-coding: But we might not be able to get all logs.... Can we add some metrics here?

ledger/src/blockstore.rs
ksn6: nice! yep, this makes sense - done

entry/src/block_component.rs
ksn6: do metrics get logged on dead slots?

separately - rather than here, it's probably worth adding metrics to `BlockComponentProcessor`; can do this in a follow-up PR

ledger/src/blockstore.rs
```diff
@@ -2157,6 +2218,57 @@ impl Blockstore {
...
+        &self,
+        shred: &Shred,
+        location: BlockLocation,
+        slot: u64,
+        just_inserted_shreds: &mut HashMap<(BlockLocation, ShredId), Cow<'_, Shred>>,
+        parent_meta_working_set: &mut HashMap<(BlockLocation, u64), WorkingEntry<ParentMeta>>,
+    ) -> Result<()> {
+        let key = (location, slot);
+
+        let parent_meta = parent_meta_working_set
```
AshwinSekar: This is not quite right - if we get a blockstore hit we need to update `parent_meta_working_set` (as clean) otherwise we're calling blockstore for every shred that makes it here in the insertion batch.

AshwinSekar: See https://github.com/anza-xyz/alpenglow/pull/601#discussion_r2525649682 for an example of how we handle it for `MerkleRootMeta`.

Unlike the other trackers the `parent_meta` is only updated rarely - so I would even augment the `parent_meta_working_set` map with a bool to indicate whether we need to try fetching from blockstore. 
- If this is the first time we're seeing a shred from this slot in this insertion batch and `parent_meta_working_set` doesn't have it, try fetching from blockstore
- For further calls in this insertion batch no need to fetch from blockstore again, it wouldn't have changed.

ksn6: hmm, gotcha - updated the logic to reflect the above; lmk what you think

ledger/src/blockstore.rs
```diff
@@ -2157,6 +2246,58 @@ impl Blockstore {
...
+                }
+            };
+
+            if populated_from_update_parent {
+                return Ok(());
+            }
+        }
+
+        // If we haven't hit blockstore for this key yet...
+        if !entry.hit_blockstore {
```
AshwinSekar: think it's clearer if you make this an `else if` as a continuation of the above `if`.

The concern I had with reading this code is whether we could already have `entry.parent_meta = Some(Dirty(...))` at this point and overwrite it with the stale value from blockstore.
I *think* that's not possible, because `entry.hit_blockstore` would have already been set to true? But if we make it an `else if` then it becomes obvious.

ledger/src/blockstore.rs
```diff
@@ -310,6 +308,32 @@ pub struct SlotMetaWorkingSetEntry {
...
+/// - `{ parent_meta: Some(Clean(meta)), hit_blockstore: true }` - Loaded from blockstore
+/// - `{ parent_meta: Some(Dirty(meta)), hit_blockstore: true }` - Modified in this batch
+#[derive(Default)]
+struct ParentMetaWorkingSetEntry {
+    /// The parent meta value: None if blockstore was empty, Some(Clean) if loaded
+    /// from blockstore unchanged, or Some(Dirty) if inserted/modified in this batch
+    parent_meta: Option<WorkingEntry<ParentMeta>>,
+    /// Whether we've already loaded this key from the blockstore. Once true, we use
+    /// the cached `parent_meta` value instead of hitting the blockstore again.
+    hit_blockstore: bool,
```
AshwinSekar: nit: maybe `checked_blockstore` or `fetched_from_blockstore`

AshwinSekar (commented):
Looks really good just some minor comments

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: sure, `else if` works here

to elucidate the logic:

if we think about the "lifecycle" of an `entry`, every `entry` must go through the `if !entry.fetched_from_blockstore { ... }` logic the first time around, and thus will end up having `entry.fetched_from_blockstore = true;`.

given the above statement, if `entry.parent_meta = Some(Dirty(...))` at the beginning of `maybe_update_parent_meta`, then either `self.maybe_insert_update_parent` or `self.maybe_insert_parent_meta` must have been invoked.

but this implies that `entry.fetched_from_blockstore = true`, so a bad override doesn't occur.

PR #602, opened 2025-11-14
wip: validated block footer

akhi3030:
Instead of using `BlockFooterV1` in code and relying on the caller to ensure that validation checks have been performed, introduce a new type to give us that guarantee.

### DIFF
--- a/runtime/src/validated_block_footer.rs
+++ b/runtime/src/validated_block_footer.rs
@@ -1,0 +1,126 @@
+use {
+    crate::{bank::Bank, block_component_processor::BlockComponentProcessor},
+    solana_entry::block_component::BlockFooterV1,
+    solana_hash::Hash,
+    thiserror::Error,
+};
+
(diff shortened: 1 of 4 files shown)


runtime/src/validated_block_footer.rs
```diff
@@ -0,0 +1,126 @@
...
+        bank: &Bank,
+        footer: BlockFooterV1,
+    ) -> Result<Self, ValidatedBlockFooterError> {
+        let BlockFooterV1 {
+            bank_hash,
+            block_producer_time_nanos,
+            block_user_agent,
+        } = footer;
+        let block_producer_time_ns = validate_clock(parent_bank, bank, block_producer_time_nanos)?;
+        if bank_hash != bank.hash() {
```
akhi3030, line 53: feedback from Ashwin.  This check cannot be performed yet because the bank is not frozen yet.

PR #603, opened 2025-11-14
Add Finalization Certificate to Block Footer.

wen-coding:
#### Summary of Changes
Add Finalization Certificate to Block Footer, there will be a follow up PR to actually populate the Certificate.


entry/src/block_component.rs
```diff
@@ -1010,6 +1062,25 @@ impl BlockFooterV1 {
         buffer.push(capped_len as u8);
         buffer.extend_from_slice(&self.block_user_agent[..capped_len]);
 
+        // Serialize final certificate if present, add a byte of 1 if present, otherwise a byte of 0
```
AshwinSekar, line 1112: How do you envision the leader to populate the finalization certificate? Specifically when would they not include it?

Would it be easier to just say the leader always includes their latest finalization?

entry/src/block_component.rs
```diff
@@ -287,13 +289,61 @@ pub enum VersionedBlockFooter {
 /// │ User Agent Length            (1 byte)   │
 /// ├─────────────────────────────────────────┤
 /// │ User Agent Bytes          (0-255 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Final cert Length             (2 bytes) │
```
AshwinSekar: Is the comment correct? I don't think you actually serialize the length right?

Also there's a comment at the top of this file we should update that one as well

entry/src/block_component.rs
```diff
@@ -287,13 +289,61 @@ pub enum VersionedBlockFooter {
...
+/// ├─────────────────────────────────────────┤
+/// │ Final cert Data           (variable)    │
```
AshwinSekar: Maybe would be easier to spell out the fields here? Or have another comment that shows the serialization format of `FinalCertificate`

AshwinSekar (commented):
Overall direction looks good just a couple of comments.

I think also we should get the SIMD out for this & adding the rewards aggregate to footer early to avoid any delays when upstreaming.

entry/src/block_component.rs
wen-coding: I would imagine if the leader just started up then it probably has no cert or a very old cert which is not useful. We can require the leader to supply an old cert just in case, but I don't know how useful it is.

Also, if you start from genesis, then block 1 has no cert.

entry/src/block_component.rs
wen-coding: Waiting for Karthik to okay the validity of the fields here, then I would put it in SIMD and link the SIMD in here. It's a bit long for comments.

wen-coding:
> Overall direction looks good just a couple of comments.
> 
> I think also we should get the SIMD out for this & adding the rewards aggregate to footer early to avoid any delays when upstreaming.

https://github.com/solana-foundation/solana-improvement-documents/pull/402

Almost ready but waiting for check of field validity after wincode here.

entry/src/block_component.rs
wen-coding: Oops, changed. I added length serialization later: line 1070~1086
```
        // Serialize final certificate if present, add a byte of 1 if present, otherwise a byte of 0
        if let Some(final_cert) = &self.final_cert {
            let cert_size: u16 = final_cert
                .serialized_size()
                .try_into()
                .map_err(|_| BlockComponentError::DataLengthOverflow)?;
            let cert_size_bytes = cert_size.to_le_bytes();
            if cert_size_bytes.len() != Self::FINAL_CERT_LENGTH_SIZE {
                return Err(BlockComponentError::DataLengthOverflow);
            }
            buffer.extend_from_slice(&cert_size_bytes);
            ....
        } else {
            // Push 2 bytes of 0 to indicate absence of final certificate
            buffer.extend_from_slice(&[0; Self::FINAL_CERT_LENGTH_SIZE]);
        }

```

entry/src/block_component.rs
```diff
@@ -62,6 +62,10 @@
 /// │ User Agent Length            (1 byte)   │
 /// ├─────────────────────────────────────────┤
 /// │ User Agent Bytes          (0-255 bytes) │
+/// ├─────────────────────────────────────────┤
+/// │ Final cert Length             (2 bytes) │
```
ksn6: See https://github.com/solana-foundation/solana-improvement-documents/pull/402/files#r2536008088

entry/src/block_component.rs
ksn6: See https://github.com/solana-foundation/solana-improvement-documents/pull/402/files#r2536008088

entry/src/block_component.rs
```diff
@@ -287,13 +293,61 @@ pub enum VersionedBlockFooter {
...
 /// ```
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub struct BlockFooterV1 {
     pub bank_hash: Hash,
     pub block_producer_time_nanos: u64,
     pub block_user_agent: Vec<u8>,
+    pub final_cert: Option<FinalCertificate>,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
```
ksn6, line 310: fyi: see https://github.com/solana-foundation/solana-improvement-documents/pull/402/files#r2536019237.

this being said, no worries re. writing custom serde code in this PR, since we'll be replacing that with `wincode`

entry/src/block_component.rs
wen-coding: Changed.

entry/src/block_component.rs
wen-coding: Changed.

entry/src/block_component.rs
wen-coding: Yeah, let's review the data structure here and change to wincode later.

ksn6:
@wen-coding - it's likely worth building on top of https://github.com/anza-xyz/alpenglow/pull/616 rather than using `block_component.rs`, which is going away shortly.

PR #605, opened 2025-11-19
fast leader handover: error on bad `UpdateParent` during shred ingest

ksn6:
When disseminating an `UpdateParent` component, we need to ensure that the parent change passes validity checks with respect to the `BlockHeader`.

We explicitly cover the four cases:
- `UpdateParent` observed before `BlockHeader` and...
  - ... `UpdateParent` and `BlockHeader` are in the same shred insertion batch
  - ... `UpdateParent` and `BlockHeader` are in different shred insertion batches

- `BlockHeader` observed before `UpdateParent` and...
  - ... `UpdateParent` and `BlockHeader` are in the same shred insertion batch
  - ... `UpdateParent` and `BlockHeader` are in different shred insertion batches

These checks are efficient, since we can detect the presence of a `BlockHeader` in a shred's payload without having to touch blockstore.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2247,16 +2247,16 @@ impl Blockstore {
     /// - `Err` if validation fails
     fn should_write_parent_meta(slot: Slot, new: &ParentMeta, prev: &ParentMeta) -> Result<bool> {
         // Determine which is the UpdateParent and which is the BlockHeader
-        let (update_parent_meta, block_header_meta) = match (
+        let (update_parent_meta, block_header_meta, should_write) = match (
             new.populated_from_block_header(),
             prev.populated_from_block_header(),
         ) {
             // New is BlockHeader, prev is UpdateParent
             // UpdateParent takes precedence, so don't overwrite
-            (true, false) => (prev, new),
+            (true, false) => (prev, new, false),
             // New is UpdateParent, prev is BlockHeader
             // Validate and allow UpdateParent to replace BlockHeader
-            (false, true) => (new, prev),
+            (false, true) => (new, prev, true),
             // Both are the same type - ensure they match
             _ => match new == prev {
                 true => return Ok(false),
@@ -2265,15 +2265,17 @@ impl Blockstore {
         };
 
         // Validate that the UpdateParent is compatible with the BlockHeader
-        match () {
-            _ if update_parent_meta.block() == block_header_meta.block() => {
-                Err(BlockstoreError::UpdateParentMatchesBlockHeader(slot))
-            }
-            _ if update_parent_meta.parent_slot > block_header_meta.parent_slot => Err(
-                BlockstoreError::UpdateParentSlotGreaterThanBlockHeader(slot),
-            ),
-            _ => Ok(true),
+        if update_parent_meta.block() == block_header_meta.block() {
+            return Err(BlockstoreError::UpdateParentMatchesBlockHeader(slot));
         }
+
+        if update_parent_meta.parent_slot > block_header_meta.parent_slot {
+            return Err(BlockstoreError::UpdateParentSlotGreaterThanBlockHeader(
+                slot,
+            ));
+        }
+
+        Ok(should_write)
     }
 
     /// Fetches or retrieves the existing ParentMeta for a given slot and location.


ledger/src/blockstore.rs
```diff
@@ -2263,13 +2284,13 @@ impl Blockstore {
 
         // If the working set ParentMeta exists and has an UpdateParent, then return early
         if let Some(parent_meta) = entry.parent_meta.as_ref() {
```
AshwinSekar: I think it's clearer if we encapsulated this in a function that didn't run the validity checks but just did the following:
- Populate `entry.parent_meta`, fetching from blockstore if necessary
- Return this previous `parent_meta`

Then below depending on whether `shred.index() == 0` we try to parse the current shred as a `Header` or an `UpdateParent` and return the newly constructed `ParentMeta`.

Then finally we validate this new `ParentMeta` against previous `ParentMeta` and decide whether to update `entry.parent_meta`

ledger/src/blockstore.rs
```diff
@@ -1954,37 +1954,35 @@ impl Blockstore {
         None
     }
 
-    fn maybe_insert_parent_meta(
-        &self,
-        current_shred: &Shred,
-        entry: &mut ParentMetaWorkingSetEntry,
-    ) {
+    fn maybe_insert_parent_meta(&self, current_shred: &Shred) -> Option<ParentMeta> {
```
AshwinSekar: Since these functions no longer insert, we should update the naming to reflect that

ledger/src/blockstore.rs
AshwinSekar: Right now we mix the construction with the validation & fetching from blockstore code which makes it harder to reason about

ledger/src/blockstore.rs
```diff
@@ -1954,37 +1954,35 @@ impl Blockstore {
...
-        else {
-            return;
-        };
+        let (parent_slot, parent_block_id) =
+            BlockComponent::parse_block_header_from_data_payload(payload)?;
 
         let parent_meta = ParentMeta {
             parent_slot,
             parent_block_id,
             replay_fec_set_index: 0,
```
AshwinSekar, line 1982: For a future PR, I'm wondering if we should just start at 32 (and for UpdateParent the next FEC set). Since all validation is done in shred ingest there's not really a point to deserialize and process these block components in replay again?

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: yeah makes sense - turns out that (in the next PR) we'll also need to cache the fetches as well once we populate `ParentMeta` children; this helps make that code cleaner as well

ledger/src/blockstore.rs
ksn6: validation isn't fully done in shred ingest - we still need to ensure that block components aren't duplicated + that headers are in fact present

agreed that it's worth thinking through though

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,64 @@ impl Blockstore {
         Ok(())
     }
 
+    /// Validates compatibility between an UpdateParent and a BlockHeader ParentMeta.
+    ///
+    /// This function determines which ParentMeta is which (UpdateParent vs BlockHeader) and
+    /// validates that they are compatible. Works regardless of parameter order.
+    fn update_parent_block_header_checks(slot: Slot, x: &ParentMeta, y: &ParentMeta) -> Result<()> {
```
AshwinSekar: maybe better named as `check_parent_meta_consistency` since you allow block components of the same type. And update doc comment

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,64 @@ impl Blockstore {
...
+        // Determine which is the UpdateParent and which is the BlockHeader
+        let (update_parent_meta, block_header_meta) = match (
+            x.populated_from_block_header(),
+            y.populated_from_block_header(),
+        ) {
+            (true, false) => (y, x),
+            (false, true) => (x, y),
+            // If both block components are of the same type, ensure that they match
+            _ => match x == y {
+                true => return Ok(()),
```
AshwinSekar: Think return type could be `Result<bool>` indicating whether we need to write this new parent meta and then you could return false here and if the newer one is a header while the older one is an update parent

ledger/src/blockstore.rs
```diff
@@ -2261,42 +2305,47 @@ impl Blockstore {
...
-                    parent_meta.populated_from_update_parent()
-                }
-            };
+        // Fetch or retrieve existing ParentMeta (from working set or blockstore)
+        let previous_parent_meta = self.get_or_fetch_parent_meta(slot, location, entry)?;
 
-            if populated_from_update_parent {
+        // If we already have an UpdateParent, only validate block headers against it. We don't need
+        // to attempt parsing another UpdateParent since UpdateParent takes precedence.
+        if let Some(prev) = &previous_parent_meta {
```
AshwinSekar: if you do https://github.com/anza-xyz/alpenglow/pull/605/files#r2543506897 then you can drop this block and instead guard 2328 by previous_parent_meta not existing or being a header.

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: nice - done

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,77 @@ impl Blockstore {
         Ok(())
     }
 
+    /// Validates compatibility between an UpdateParent and a BlockHeader ParentMeta.
```
AshwinSekar: update doc comment

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,77 @@ impl Blockstore {
...
+        prev: &ParentMeta,
+    ) -> Result<bool> {
+        // Determine which is the UpdateParent and which is the BlockHeader
+        let (update_parent_meta, block_header_meta) = match (
+            new.populated_from_block_header(),
+            prev.populated_from_block_header(),
+        ) {
+            // New is BlockHeader, prev is UpdateParent
+            // UpdateParent takes precedence, so don't overwrite
+            (true, false) => return Ok(false),
```
AshwinSekar: we still want to validate below right?

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,77 @@ impl Blockstore {
...
+            (false, true) => (new, prev),
+            // Both are the same type - ensure they match
+            _ => match new == prev {
+                true => return Ok(false),
+                false => return Err(BlockstoreError::BlockComponentMismatch(slot)),
+            },
+        };
+
+        // Validate that the UpdateParent is compatible with the BlockHeader
+        match () {
```
AshwinSekar: hallucination or intentional?
clearer if you just use `if`s

ledger/src/blockstore.rs
```diff
@@ -2261,42 +2318,41 @@ impl Blockstore {
...
             return Ok(());
-        }
+        };
 
-        // If there isn't any parent_meta, then attempt to detect a block header.
-        if entry.parent_meta.is_none() && shred.index() == 0 {
-            self.maybe_insert_parent_meta(shred, entry);
+        // Validate new ParentMeta against previous (if both exist)
+        if let Some(prev) = &previous_parent_meta {
+            if !Self::check_parent_meta_consistency(slot, &new_parent_meta, prev)? {
```
AshwinSekar: nit: maybe prefix with should_write or something now. This invocation makes it look like we're cool with them not being consistent.

ledger/src/blockstore.rs
ksn6: intentional - I like this a bit more, since things line up more cleanly; never really liked `if`, `else-if` chains - going to try avoid using these in the future for small blocks like this

ledger/src/blockstore.rs
ksn6: yep - done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
ksn6: done

ledger/src/blockstore.rs
AshwinSekar: I think this is confusing, please change it to

```rust
if update_parent_meta.block() == block_header_meta.block() {
  return Err(BlockstoreError::UpdateParentMatchesBlockHeader(slot));
}
if update_parent_meta.parent_slot > block_header_meta.parent_slot {
  return Err(BlockstoreError::UpdateParentSlotGreaterThanBlockHeader(slot));
)
Ok(true)
```

ledger/src/blockstore.rs
```diff
@@ -2250,6 +2236,73 @@ impl Blockstore {
...
+    /// - `Err` if validation fails
+    fn should_write_parent_meta(slot: Slot, new: &ParentMeta, prev: &ParentMeta) -> Result<bool> {
+        // Determine which is the UpdateParent and which is the BlockHeader
+        let (update_parent_meta, block_header_meta) = match (
+            new.populated_from_block_header(),
+            prev.populated_from_block_header(),
+        ) {
+            // New is BlockHeader, prev is UpdateParent
+            // UpdateParent takes precedence, so don't overwrite
+            (true, false) => (prev, new),
```
AshwinSekar: This overwrites the `UpdateParent` with the block header. You need to pass on this information to below

ksn6: done

ledger/src/blockstore.rs
ksn6: done

AshwinSekar (approved):

PR #607, opened 2025-11-19
wip: pay rewards

akhi3030:
#### Summary of Changes

Fixes #

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -162,7 +162,6 @@ use {
     solana_transaction_context::{transaction_accounts::TransactionAccount, TransactionReturnData},
     solana_transaction_error::{TransactionError, TransactionResult as Result},
     solana_vote::vote_account::{VoteAccount, VoteAccountsHashMap},
-    solana_vote_interface::state::VoteStateV3,
     solana_votor_messages::{
         consensus_message::Certificate, migration::GENESIS_CERTIFICATE_ACCOUNT,
     },
@@ -209,6 +208,7 @@ mod fee_distribution;
 mod metrics;
 pub(crate) mod partitioned_epoch_rewards;
 mod recent_blockhashes_account;
+mod reward_pda;
 mod serde_snapshot;
 mod sysvar_cache;
 pub(crate) mod tests;
@@ -2141,17 +2141,17 @@ impl Bank {
         });
     }
 
-    pub(crate) fn pay_vote_reward(&self, voter: Pubkey, reward: u64) {
-        let mut accounts = Arc::unwrap_or_clone(self.vote_accounts());
-        let (_stake, account) = accounts.remove(&voter).unwrap();
-        let data = account.account().data_clone();
-        let mut vote_state: VoteStateV3 = bincode::deserialize(&data).unwrap();
-        vote_state.increment_credits(self.epoch(), reward);
-        let shared_account =
-            AccountSharedData::new_data(account.lamports(), &vote_state, account.owner()).unwrap();
-        // XXX: is the stake getting set properly?
-        self.store_account(&voter, &shared_account);
-    }
+    // pub(crate) fn pay_vote_reward(&self, voter: Pubkey, reward: u64) {
+    //     let mut accounts = Arc::unwrap_or_clone(self.vote_accounts());
+    //     let (_stake, account) = accounts.remove(&voter).unwrap();
+    //     let data = account.account().data_clone();
+    //     let mut vote_state: VoteStateV3 = bincode::deserialize(&data).unwrap();
+    //     vote_state.increment_credits(self.epoch(), reward);
+    //     let shared_account =
+    //         AccountSharedData::new_data(account.lamports(), &vote_state, account.owner()).unwrap();
+    //     // XXX: is the stake getting set properly?
+    //     self.store_account(&voter, &shared_account);
+    // }
 
     pub fn update_last_restart_slot(&self) {
         let feature_flag = self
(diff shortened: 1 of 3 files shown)


core/src/block_creation_loop.rs
```diff
@@ -424,9 +425,17 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
 fn record_and_complete_block(
     poh_recorder: &RwLock<PohRecorder>,
     record_receiver: &mut RecordReceiver,
+    consensus_rewards: Arc<PLRwLock<ConsensusRewards>>,
     block_timer: Instant,
     block_timeout: Duration,
 ) -> Result<(), PohRecorderError> {
+    // Taking a read lock on consensus_rewards can contend with the write lock in bls_sigverifier.
+    // We are ready to produce the block footer now, while we gather other bits of data, we can block on the consensus_rewards lock in a separate thread to minimise contention.
+    let handle = std::thread::spawn(move || {
```
AshwinSekar, line 434: instead of spawning a thread I would suggest the following:
- Add a `consensus_rewards_service` that waits for votes to be sent from the bls sigverifier
- The `wants_vote` filtering can be done in `consensus_rewards_service`
- Here in bcl we can notify the rewards service that we would like the reward aggregates for `slot`
- When notified the rewards service can use its dedicated thread to build the aggregates
- Then in bcl after the block has finished, we can receive the finished aggregates from the rewards service for insertion into the footer.

Just a rough sketch, feel free to modify - I think this sort of design avoids the locking in the hot path and necessity to spin up ad hoc threads.

runtime/src/bank.rs
```diff
@@ -2140,6 +2141,18 @@ impl Bank {
         });
     }
 
+    pub(crate) fn pay_vote_reward(&self, voter: Pubkey, reward: u64) {
+        let mut accounts = Arc::unwrap_or_clone(self.vote_accounts());
+        let (_stake, account) = accounts.remove(&voter).unwrap();
+        let data = account.account().data_clone();
+        let mut vote_state: VoteStateV3 = bincode::deserialize(&data).unwrap();
```
AshwinSekar: We want to be targetting `VoteStateV4` here

runtime/src/bank.rs
```diff
@@ -2140,6 +2141,18 @@ impl Bank {
...
+        vote_state.increment_credits(self.epoch(), reward);
+        let shared_account =
+            AccountSharedData::new_data(account.lamports(), &vote_state, account.owner()).unwrap();
+        // XXX: is the stake getting set properly?
+        self.store_account(&voter, &shared_account);
```
AshwinSekar: Generally prefer `store_accounts` when storing multiple accounts as AccountsDB is able to use its thread pool to do some of the stuff in parallel.

runtime/src/block_component_processor.rs
```diff
@@ -189,11 +239,37 @@ impl BlockComponentProcessor {
...
+                epoch_schedule.slots_per_epoch,
+                total_supply_lamports,
+                bank.slots_per_year(),
+                yearly_inflation,
+            );
+            let validator_reward_lamports = reward_lamports / 2;
+            // XXX: pay the leader rewards
+            let _leader_reward_lamports = reward_lamports - validator_reward_lamports;
+
+            bank.pay_vote_reward(validator, validator_reward_lamports);
```
AshwinSekar, line 271: This isn't quite correct as credits != lamports.

Here's how inflationary rewards are calculated. Let `V` be the validator set:
Let 
```
points(v) = credits(v) * stake(v) for any validator v
```

If the total inflation this epoch is `T`
```
rewards(v) = T * points(v) / Sum_{u \in V}(points(u))
```

You can see the full details by poking around in the caller of 
https://github.com/anza-xyz/alpenglow/blob/d6f08a5b9126aec839e462a76a0cd72cdda649ea/runtime/src/inflation_rewards/points.rs#L106

AshwinSekar: We no longer want the sum of points in the denominator as in Alpenglow rewards are not based on relative performance.

When alpenglow is active we'll want to replace that with `432000 (slots per epoch) * total stake` and then I believe you can just add 1 credit here.

runtime/src/bank.rs
```diff
@@ -2140,6 +2141,18 @@ impl Bank {
         });
     }
 
+    pub(crate) fn pay_vote_reward(&self, voter: Pubkey, reward: u64) {
+        let mut accounts = Arc::unwrap_or_clone(self.vote_accounts());
```
AshwinSekar: Might be able to avoid some of these clones with something like this
```rust
let Some((_, vote_account)) = self.vote_accounts().get(&voter) else {
  return;
};
let mut vote_state: VoteStateV4 = vote_account.account().deserialize_data();
```

AshwinSekar: We could even avoid the `deserialize` and calculate the offset into epoch credits to update if necessary.

runtime/src/block_component_processor.rs
AshwinSekar: Sorry I realized that the leader also needs to be paid out for the same sol value. This is difficult because the leader's point calculation will multiply by the leader's stake rather than the voter's stake.

I'll need to think more about how to solve the leader's side of rewards.

core/src/block_creation_loop.rs
akhi3030: Moving the feedback to https://github.com/anza-xyz/alpenglow/pull/582 as that is where the rewards generation work is being done.

PR #608, opened 2025-11-19
blockstore_processor: advance migration during startup

AshwinSekar:
#### Problem
We currently do not advance `MigrationStatus` during startup at all, it's just set at the bank forks root.
This causes problems if we restart after the migration has succeeded or during the migrationary period before we've advanced the root. 
The old root could indicate that we are still in the migration while blockstore has already been cleaned up and we have Alpenglow blocks present.

Additionally we do not process the `GenesisCertificate` marker in the first Alpenglow block. This is key to inform startup and sometimes lagging steady state that the migration was successful. There's no need to try to do genesis discovery if this marker is processed, we can just go straight into Alpenglow.

#### Summary of Changes
Allow processing of the `GenesisCertificate` marker:
- Immediately move us to `ReadyToEnable` as the processing of the `GenesisCertificate` via block means we have the genesis block frozen
- Allow header and genesis certificate markers to be processed when we're in the migrationary period, for such cases where we missed the migration's completion and are receiving the first alpenglow block.

Advance `MigrationStatus` during startup in `load_frozen_forks`:
- If we root the feature flag activation, transition from `PreFeatureActivation` to `Migration`
- Don't root startup blocks that are in the `Migration`
- If we process a `GenesisCertificate` marker go to `ReadyToEnable`
- If we are `ReadyToEnable`, enable alpenglow immediately and retry processing the current block onwards as Alpenglow blocks instead
- Do not do genesis discovery - if we don't process a `GenesisCertificate` then we rely on the standard `compute_bank_stats` pathway in replay to find the genesis block post startup

Add a test where a node restarts post migration from a root before the migration and ensure it can catch up.
Add a test where a node misses the migration entirely and ensure that it can catch up.

Note: The certificate in the `GenesisCertificate` marker is not validated yet. Future PR will plug in BLS verification.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -922,6 +922,9 @@ impl ReplayStage {
 
                 let forks_root = bank_forks.read().unwrap().root();
                 let start_leader_time = if !migration_status.is_alpenglow_enabled() {
+                    // Will address this in https://github.com/anza-xyz/alpenglow/issues/619
+                    debug_assert!(!votor_event_receiver.is_full());
+
                     // Process cluster-agreed versions of duplicate slots for which we potentially
                     // have the wrong version. Our version was dead or pruned.
                     // Signalled by ancestor_hashes_service.


ledger/src/blockstore_processor.rs
```diff
@@ -2043,6 +2091,31 @@ fn load_frozen_forks(
...
                 }
+
+                // If this block was the first alpenglow block and advanced the migration phase, we can enable alpenglow.
+                //
+                // Note: since this code is all startup code we don't have to worry about shutting down `PohService` or any
+                // in flight activity of `ReplayStage`. This bank must have failed to freeze as it is an Alpenglow block
+                // being verified as a TowerBFT one.
+                //
+                // We are safe to cleanly transition to alpenglow here
+                if migration_status.is_ready_to_enable() {
```
AshwinSekar, line 2108: This is basically analogous to when we enable alpenglow during steady state in replay.
Difference here is:
- We reach `ReadyToEnable` by trying to process the first alpenglow block as a TowerBFT block and failing. While processing we observed the `GenesisCertificate` marker so we know that the migration happened, and which block is the genesis.
- We don't have to purge the blocks > genesis, instead we just reprocesses them as Alpenglow blocks (ticks adjusted and markers allowed).
- We have to reset the dead status and retry this first alpenglow block since we just failed to process it as a TowerBFT block

ledger/src/blockstore_processor.rs
```diff
@@ -2105,7 +2178,11 @@ fn load_frozen_forks(
                 } else {
                     None
                 }
-            };
+            }.filter(|new_root_bank| {
+                // In the case that we've restarted while the migrationary period is going on but before alpenglow
+                // is enabled, don't root blocks past the migration slot
+                migration_status.should_root_during_startup(new_root_bank.slot())
```
AshwinSekar, line 2194: startup equivalent of https://github.com/anza-xyz/alpenglow/blob/d6f08a5b9126aec839e462a76a0cd72cdda649ea/core/src/replay_stage.rs#L2814-L2816

ledger/src/blockstore_processor.rs
```diff
@@ -2129,6 +2206,19 @@ fn load_frozen_forks(
                 all_banks.retain(|_, bank| bank.ancestors.contains_key(&root));
                 m.stop();
                 root_retain_us += m.as_us();
+
+                // If this root bank activated the feature flag, update migration status
+                if migration_status.is_pre_feature_activation() {
```
AshwinSekar, line 2221: startup equivalent of https://github.com/anza-xyz/alpenglow/blob/d6f08a5b9126aec839e462a76a0cd72cdda649ea/core/src/replay_stage.rs#L2843-L2844

AshwinSekar (commented):
Startup replay is slightly different than steady state replay so must be handled slightly differently.

Almost done with migration thank y'all for reviewing. Remaining work:
- BLS verify the certificate in `GenesisCertificate` block marker
- Plug in Alpenglow finalization cert -> TowerBFT repair for nodes that were completely partitioned off during the migration.

runtime/src/block_component_processor.rs
```diff
@@ -71,36 +78,101 @@ impl BlockComponentProcessor {
...
+        if (bank.parent_slot(), parent_block_id) != (genesis_cert.slot, genesis_cert.block_id) {
+            return Err(BlockComponentProcessorError::GenesisCertificateOnNonChild);
+        }
+
+        if bank.get_alpenglow_genesis_certificate().is_some() {
+            return Err(BlockComponentProcessorError::GenesisCertificateAlreadyPopulated);
+        }
+
+        let genesis_cert = Certificate::from(genesis_cert);
+        // TODO(ashwin): verify genesis cert using bls sigverify and bank
```
AshwinSekar, line 159: Created https://github.com/anza-xyz/alpenglow/issues/609 so I don't forget

local-cluster/tests/local_cluster.rs
```diff
@@ -6341,24 +6355,179 @@ fn test_alpenglow_migration(num_nodes: usize) {
...
+    // - Node 3 now rejoins the network
+    // - It is able to use TowerBFT eager repair to fetch blocks, eventually it will see the
+    //   first Alpenglow block, attempt to process it as a TowerBFT block and switch to Alpenglow.
+    info!("Restarting node to a pre migration state");
+    cluster.restart_node(&node_pubkey, exit_info, SocketAddrSpace::Unspecified);
+    cluster.check_for_new_roots(8, test_name, SocketAddrSpace::Unspecified);
+}
+
+#[test]
+// This test requires alpenglow repair
```
AshwinSekar, line 6430: leaving this test here as a stretch goal that we can revisit once we have alpenglow repair.

It's a pretty gnarly situation:
- Our node observes the start of the migration and views 5 TowerBFT blocks after the migration slot
- Our node partitions/goes offline
- The remaining nodes finish the migration, clean everything past genesis and root a block in Alpenglow
- Since they root a block they stop broadcasting the genesis certificate
- Our node rejoins, it can repair the missing blocks but it has 5 TowerBFT blocks in place of the alpenglow ones
- Once alpenglow repair is implemented it'll repair the 5 Alpenglow blocks into the alternate blockstore column

Need some signal to decide to try to replay the alpenglow blocks from the alternate column -> views the GenesisCertificate block marker -> enables alpenglow.
In practice it might be simpler to just tell the operator to wait for a new snapshot and restart if they end up in this state.

AshwinSekar: another alternative is for nodes to continue broadcasting the genesis cert A2A every 10 seconds for a couple hours or until the end of the epoch.

This is much easier, and if someone's node is partitioned for so long that it still misses the cert we can instruct operators to restart with fresh snapshot.

bw-solana: My knee jerk reaction is that the 1 node that happens to hit this case can just manually purge the shreds for the Tower blocks after Alpenglow block and restart 😄

core/src/block_creation_loop.rs
```diff
@@ -657,15 +657,30 @@ fn create_and_insert_leader_bank(slot: Slot, parent_bank: Arc<Bank>, ctx: &mut L
...
+        let bank = poh_recorder.bank().expect("Bank cannot have been cleared");
+        let processor = bank.block_component_processor.read().unwrap();
+        processor
+            .on_genesis_certificate(
+                bank.clone(),
+                ctx.genesis_cert.clone(),
+                &ctx.bank_forks.read().unwrap().migration_status(),
+            )
+            .expect("Recording genesis certificate should not fail");
     }
```
bw-solana: good candidate for a function like `maybe_include_genesis_certificate`

core/src/replay_stage.rs
```diff
@@ -922,7 +922,6 @@ impl ReplayStage {
 
                 let forks_root = bank_forks.read().unwrap().root();
                 let start_leader_time = if !migration_status.is_alpenglow_enabled() {
-                    debug_assert!(votor_event_receiver.is_empty());
```
bw-solana, line 925: is it okay if the receiver is non-empty? Or just getting rid of this now that we've validated it holds true?

local-cluster/tests/local_cluster.rs
AshwinSekar: agreed, this is a pretty wild scenario. Ideally the super OC block will happen in 2 blocks anyway, for the node to partition off in this exact time window is unlikely.

I won 't address this case for now and we can revisit when we do some large scale testing.

ledger/src/blockstore_processor.rs
```diff
@@ -1856,6 +1856,54 @@ fn process_bank_0(
...
 }
 
+/// Clean up a failed slot and restart processing from the given genesis slot
+fn cleanup_and_populate_pending_from_alpenglow_genesis(
+    first_alpenglow_bank: &BankWithScheduler,
+    genesis_slot: Slot,
+    bank_forks: &RwLock<BankForks>,
+    blockstore: &Blockstore,
+    leader_schedule_cache: &LeaderScheduleCache,
+    pending_slots: &mut Vec<(SlotMeta, Bank, Hash)>,
```
bw-solana, line 1870: might be good to clarify what `pending_slots` means

ledger/src/blockstore_processor.rs
```diff
@@ -1856,6 +1856,54 @@ fn process_bank_0(
...
+    root_bank.prune_program_cache_by_deployment_slot(first_alpenglow_bank.slot());
+    blockstore
+        .remove_dead_slot(first_alpenglow_bank.slot())
+        .unwrap();
+
+    let genesis_slot_meta = blockstore
+        .meta(genesis_slot)
+        .map_err(|err| {
+            warn!("Failed to load meta for slot {genesis_slot}: {err:?}");
+            BlockstoreProcessorError::FailedToLoadMeta
```
bw-solana: is this recoverable?

ledger/src/blockstore_processor.rs
```diff
@@ -1856,6 +1856,54 @@ fn process_bank_0(
...
+        })?
+        .unwrap();
+
+    warn!(
+        "{}: load_frozen_forks() restart processing from {genesis_slot} treating further blocks \
+         as Alpenglow banks",
+        migration_status.my_pubkey()
+    );
+    pending_slots.clear();
```
bw-solana, line 1898: should we declare that this has to be passed in empty?

ledger/src/blockstore_processor.rs
```diff
@@ -1856,6 +1856,54 @@ fn process_bank_0(
...
+    process_next_slots(
```
bw-solana, line 1900: comment explaining what we're doing here could be nice (queueing up slots descendant from genesis slot to replay as Alpenglow blocks)

ledger/src/blockstore_processor.rs
```diff
@@ -1959,8 +2007,8 @@ fn load_frozen_forks(
     let mut root = bank_forks.read().unwrap().root();
     let max_root = std::cmp::max(root, blockstore_max_root);
     info!(
-        "load_frozen_forks() latest root from blockstore: {blockstore_max_root}, max_root: \
-         {max_root}",
+        "load_frozen_forks() bank forks root {root}, latest root from blockstore: \
```
bw-solana: nit `"load_frozen_forks() bank forks root: {root}, latest root from blockstore: \`

ledger/src/blockstore_processor.rs
```diff
@@ -2043,6 +2091,31 @@ fn load_frozen_forks(
                 if opts.abort_on_invalid_block {
                     Err(error)?
                 }
+
+                // If this block was the first alpenglow block and advanced the migration phase, we can enable alpenglow.
+                //
+                // Note: since this code is all startup code we don't have to worry about shutting down `PohService` or any
+                // in flight activity of `ReplayStage`. This bank must have failed to freeze as it is an Alpenglow block
+                // being verified as a TowerBFT one.
```
bw-solana, line 2105: should we have a debug assert on the exact error we expect (I'm assuming too few ticks)?

local-cluster/tests/local_cluster.rs
```diff
@@ -6341,24 +6355,179 @@ fn test_alpenglow_migration(num_nodes: usize) {
...
-    test_alpenglow_migration(1)
+    test_alpenglow_migration(1, "test_alpenglow_migration_1", &[4]);
 }
 
 #[test]
 #[serial]
 fn test_alpenglow_migration_2() {
-    test_alpenglow_migration(2)
+    test_alpenglow_migration(2, "test_alpenglow_migration_2", &[4, 4]);
 }
```
bw-solana: what are these tests buying us that we don't get with `test_alpenglow_migration_4`?

local-cluster/tests/local_cluster.rs
```diff
@@ -6341,24 +6355,179 @@ fn test_alpenglow_migration(num_nodes: usize) {
...
+#[test]
+#[serial]
+fn test_alpenglow_missed_migration_entirely() {
+    let test_name = "test_alpenglow_missed_migration_entirely";
+
+    // Start a 3 node cluster and have it go through the migration and root some slots
+    // Critical that the third node is not in the leader schedule, as since
+    // we clear blockstore later, we could end up producing duplicate blocks
+    let (mut cluster, validator_keys, migration_slot) =
+        test_alpenglow_migration(3, test_name, &[4, 4, 0]);
```
bw-solana, line 6400: what are the stakes in this case? I was assuming even stake, but can't super OC with only 2/3

runtime/src/bank.rs
```diff
@@ -2826,16 +2826,22 @@ impl Bank {
...
             acct.deserialize_data()
                 .expect("Programmer error deserializing genesis certificate")
         })
     }
 
+    /// For use in the first Alpenglow block, set the genesis certificate.
+    pub fn set_alpenglow_genesis_certificate(&self, cert: &Certificate) {
+        let cert_size = bincode::serialized_size(cert).unwrap();
+        let lamports = Rent::default().minimum_balance(cert_size as usize);
+        let cert_acct = AccountSharedData::new_data(lamports, cert, &system_program::ID).unwrap();
```
bw-solana, line 2840: okay John Law

runtime/src/block_component_processor.rs
```diff
@@ -71,36 +78,101 @@ impl BlockComponentProcessor {
...
-        // Here onwards, Alpenglow is enabled
+        let slot = bank.slot();
         let marker = match marker {
             VersionedBlockMarker::V1(marker) | VersionedBlockMarker::Current(marker) => marker,
         };
 
-        match marker {
-            BlockMarkerV1::BlockFooter(footer) => self.on_footer(bank, parent_bank, footer),
-            BlockMarkerV1::BlockHeader(header) => self.on_header(header),
+        match (marker, migration_status.should_allow_block_markers(slot)) {
```
bw-solana: this is confusing, and I think it's because `fn should_allow_block_markers` doesn't really do what it says anymore. It's more like `should_allow_all_block_markers`. Because even if it is `false`, we still allow certain block markers is we're in migration.

Maybe something like this would be clearer:
```rust
    let markers_fully_enabled = migration_status.should_allow_block_markers(slot);
    let in_migration = migration_status.is_in_migration();

    match marker {
        // Header and genesis cert can be processed either:
        //  - once migration is fully enabled, or
        //  - while we're still in the migration phase (to let us advance it).
        BlockMarkerV1::BlockHeader(header)
            if markers_fully_enabled || in_migration =>
        {
            self.on_header(&header)
        }

        BlockMarkerV1::GenesisCertificate(genesis_cert)
            if markers_fully_enabled || in_migration =>
        {
            self.on_genesis_certificate(bank, genesis_cert, migration_status)
        }

        // Everything else is only valid once migration is complete.
        BlockMarkerV1::BlockFooter(footer) if markers_fully_enabled => {
            self.on_footer(bank, parent_bank, &footer)
        }

        // We process UpdateParent on shred ingest; once migration is complete
        // we just treat this as a no-op marker.
        BlockMarkerV1::UpdateParent(_) if markers_fully_enabled => {
            Ok(())
        }

        // Any other combination means we saw a marker too early.
        _ => Err(BlockComponentProcessorError::BlockComponentPreMigration),
    }?;
```

runtime/src/block_component_processor.rs
```diff
@@ -71,36 +78,101 @@ impl BlockComponentProcessor {
         &mut self,
         bank: Arc<Bank>,
         parent_bank: Arc<Bank>,
-        marker: &VersionedBlockMarker,
+        marker: VersionedBlockMarker,
         migration_status: &MigrationStatus,
         is_final: bool,
```
bw-solana, line 96: outside scope of this PR, but we should really document what `is_final` means

votor-messages/src/migration.rs
```diff
@@ -187,6 +187,20 @@ impl MigrationPhase {
         }
     }
 
+    /// Should we root this slot when loading frozen slots during startup?
+    /// Similar to `should_report_commitment_or_root`, but we also continue root post migration.
+    /// This is only relevant if we restart during the migration period before it completes, we don't
+    /// want to root any slots >= migraiton_slot
+    fn should_root_during_startup(&self, slot: Slot) -> bool {
+        match self {
+            MigrationPhase::PreFeatureActivation => true,
```
bw-solana: could combine with cases below?

AshwinSekar: done [d18b79f](https://github.com/anza-xyz/alpenglow/pull/608/commits/d18b79f123a18efcd20b1e3765d82fb35e8ac5f4)

runtime/src/block_component_processor.rs
AshwinSekar: added some documentation here [2d9140a](https://github.com/anza-xyz/alpenglow/pull/608/commits/2d9140a94a1ba6290ae3e2cd48c1cc792d514335)

runtime/src/block_component_processor.rs
AshwinSekar: yeah good call, i've reworked it here [e5beb68](https://github.com/anza-xyz/alpenglow/pull/608/commits/e5beb685a2b56c12b1a62c0fdb0d6125e73071e3)

local-cluster/tests/local_cluster.rs
AshwinSekar: not much really, i've removed them here [a0dd71f](https://github.com/anza-xyz/alpenglow/pull/608/commits/a0dd71f4423aad2ba655d4b329f6ebc459aff78a)

local-cluster/tests/local_cluster.rs
AshwinSekar: it is even stake. I kind of hacked this situation as the third node is initially part of the migration, then we clear it's blockstore to make it look like it missed the migration entirely.

If you prefer, I can replicate the situation more accurately by killing the third node immediately on startup, adjusting the stakes so that the remaining 2 nodes can migrate and then restarting the third node.

ledger/src/blockstore_processor.rs
AshwinSekar: added and yeah it will be too few ticks in this case

ledger/src/blockstore_processor.rs
AshwinSekar: done addressed your comments in this file here https://github.com/anza-xyz/alpenglow/pull/608/commits/04aa337146980e3c2bbad0b63d5df01cd65bf0d1

ledger/src/blockstore_processor.rs
AshwinSekar: it's not recoverable, this error is propagated all the way up and cause `Validator::new` to return an error.

ledger/src/blockstore_processor.rs
AshwinSekar: is the suggestion that the caller does the clear instead? I'm not opposed but I thought it fits here as we are removing `first_alpenglow_bank` from other runtime things, so naturally we'd want to clean up its influence on `pending_slots` as well

core/src/block_creation_loop.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/608/commits/e1c570c0b5177a70c42d0a9aad65e289b6f603c4

core/src/replay_stage.rs
AshwinSekar: ah yeah so turns out the receiver will not be empty if our node is slow to observe the completion of the migration.
If the rest of the cluster starts sending votes / certs, the certificate pool will emit events (`Finalized`, `ParentReady` etc.).

An argument could be made that it's fine to not emit these events - if the cluster was able to finalize or otherwise parent ready blocks in our absence then they could continue to do so. I haven't fully thought it through but I think this should work.

This is a bounded channel so having it fill up is not ideal. If you're cool with it I can think this through and address in a follow up.

bw-solana: I guess the risk is that:

1. Pool can be generating events (observing consensus messages)
2. This node is slow to observe the migration for whatever reason (e.g. replay got bogged down)
3. Event handler is not processing events because it's waiting on the condvar
4. The bounded channel fills up and we stall the sending thread

Seems pretty bad depending on what messages get dropped..

At a minimum we should file an issue to make sure we don't forget about this. It might be nice to `try_send` and assert on full for now so things blow up in our face during testing, but not sure how much code churn that will cause... maybe we could change this debug_assert to something like `debug_assert!(!votor_event_receiver.is_full());`

ledger/src/blockstore_processor.rs
bw-solana: I'm following better now and agree with you that clearing here feels right

local-cluster/tests/local_cluster.rs
bw-solana: I think the more representative test would be better, but doesn't need to gate this PR

core/src/replay_stage.rs
AshwinSekar: add the not full check https://github.com/anza-xyz/alpenglow/pull/608/commits/9b0a86a218be5ba20a6b1e9d3184fe89edd98047 and created an issue #619

local-cluster/tests/local_cluster.rs
AshwinSekar: sounds good I filed https://github.com/anza-xyz/alpenglow/issues/620

bw-solana (approved):

runtime/src/block_component_processor.rs
```diff
@@ -67,40 +78,116 @@ impl BlockComponentProcessor {
...
+            "{}: Alpenglow genesis marker processed during replay of {}. Transitioning Alpenglow \
+             to ReadyToEnable",
+            migration_status.my_pubkey(),
+            bank.slot()
+        );
+        migration_status.set_genesis_block(
+            genesis_cert
+                .cert_type
+                .to_block()
+                .expect("Genesis cert must correspond to a block"),
```
carllin, line 183: this is only safe to do here because we have the block at this point?

but it's not safe to do here in the consensus pool? https://github.com/anza-xyz/alpenglow/blob/64c51e401b8193612ae170da8cbd9e1ed9cc4ca1/votor/src/consensus_pool.rs#L346

runtime/src/block_component_processor.rs
```diff
@@ -67,40 +78,116 @@ impl BlockComponentProcessor {
...
-        }
-
-        // Here onwards, Alpenglow is enabled
+        let slot = bank.slot();
         let marker = match marker {
             VersionedBlockMarker::V1(marker) | VersionedBlockMarker::Current(marker) => marker,
         };
 
+        let markers_fully_enabled = migration_status.should_allow_block_markers(slot);
+        let in_migration = migration_status.is_in_migration();
```
carllin, line 104: Can this be simplified to just checking that the slot is greater than the migration start slot?

votor-messages/src/migration.rs
```diff
@@ -575,6 +605,34 @@ impl MigrationStatus {
...
+        *self.phase.write().unwrap() = MigrationPhase::AlpenglowEnabled { genesis_cert };
+        let (is_alpenglow_enabled, _condvar) = &self.migration_wait;
+        *is_alpenglow_enabled.lock().unwrap() = true;
+        // No need to condvar as we're in startup and no one is waiting for us.
+        warn!(
+            "{}: Alpenglow enabled during startup! Genesis slot {genesis_slot}",
+            self.my_pubkey()
+        );
+        genesis_slot
+    }
```
carllin, line 634: Would it be possible to just have this plumb the necessary signals beforehand to avoid blocking and then call `enable_alpenglow()` directly? 

Might make both paths easier to reason about

runtime/src/block_component_processor.rs
AshwinSekar: that's basically what `in_migration` tells you. And we only want the header and genesis certificate to be allowed in that case

PR #613, opened 2025-11-26
blockstore: compute and populate DoubleMerkleRootMeta

AshwinSekar:
Adds an unoptimized double merkle root computation to get the ball rolling.

Computes and stores the double merkle root and associated proofs once the block is full.
This places all the blockstore lookups in the hot path (computation will run once block has been frozen).

Future optimizations:
1. Join nodes as each FEC set comes in, no need to wait until the end and perform up to 1000 blockstore lookups for FEC set roots, this can be done in line when populating new MerkleRootMetas
2. Generate proofs only when the first repair request comes in. The majority of blocks should not be repaired - we can defer this out of the hot path.

Next PR will actually call this fn and set block id in either replay or broadcast

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -910,7 +910,7 @@ impl Blockstore {
 
         let last_index = slot_meta.last_index.expect("Slot is full");
 
-        // This function is only used post Alpenglow, so implicitely gated by SIMD-0317 as that is a prereq
+        // This function is only used post Alpenglow, so implicitly gated by SIMD-0317 as that is a prereq
         let fec_set_count = (last_index / (DATA_SHREDS_PER_FEC_BLOCK as u64) + 1) as usize;
 
         let parent_meta = self
@@ -923,7 +923,7 @@ impl Blockstore {
         let fec_set_indices =
             (0..fec_set_count).map(|i| (slot, (i * DATA_SHREDS_PER_FEC_BLOCK) as u32));
         let keys = self.merkle_root_meta_cf.multi_get_keys(fec_set_indices);
-        let mut merkle_tree_leaves = self
+        let merkle_tree_leaves: Vec<_> = self
             .merkle_root_meta_cf
             .multi_get_bytes(&keys)
             .map(|get_result| {
@@ -936,14 +936,12 @@ impl Blockstore {
                     .expect("Legacy shreds no longer exist, merkle root must be present");
                 Ok(merkle_root)
             })
-            .collect::<Vec<_>>();
-
-        // Add parent info as the last leaf
-        let parent_info_hash = hashv(&[
-            &parent_meta.parent_slot.to_le_bytes(),
-            parent_meta.parent_block_id.as_ref(),
-        ]);
-        merkle_tree_leaves.push(Ok(parent_info_hash));
+            // Add parent info as the last leaf
+            .chain(std::iter::once(Ok(hashv(&[
+                &parent_meta.parent_slot.to_le_bytes(),
+                parent_meta.parent_block_id.as_ref(),
+            ]))))
+            .collect();
 
         // Build the merkle tree
         let merkle_tree = make_merkle_tree(merkle_tree_leaves)
@@ -954,24 +952,23 @@ impl Blockstore {
 
         // Build proofs
         let tree_size = fec_set_count + 1;
-        let mut proofs = Vec::with_capacity(tree_size);
-
-        for leaf_index in 0..tree_size {
-            let proof_iter = make_merkle_proof(leaf_index, tree_size, &merkle_tree);
-            let proof: Vec<u8> = proof_iter
-                .flat_map(|proof| {
-                    proof
-                        .expect("Merkle proof construction cannot have failed")
-                        .as_slice()
-                })
-                .copied()
-                .collect();
-            debug_assert_eq!(
-                proof.len(),
-                get_proof_size(tree_size) as usize * SIZE_OF_MERKLE_PROOF_ENTRY
-            );
-            proofs.push(proof);
-        }
+        let proofs: Vec<Vec<u8>> = (0..tree_size)
+            .map(|leaf_index| {
+                make_merkle_proof(leaf_index, tree_size, &merkle_tree)
+                    .map(|proof_entry| {
+                        proof_entry.expect("Merkle proof construction cannot have failed")
+                    })
+                    .flat_map(|proof_entry| proof_entry.as_slice())
+                    .copied()
+                    .collect()
+            })
+            .inspect(|proof: &Vec<u8>| {
+                debug_assert_eq!(
+                    proof.len(),
+                    get_proof_size(tree_size) as usize * SIZE_OF_MERKLE_PROOF_ENTRY
+                );
+            })
+            .collect();
 
         // Create and store DoubleMerkleMeta
         let double_merkle_meta = DoubleMerkleMeta {


ledger/src/blockstore.rs
```diff
@@ -866,6 +874,108 @@ impl Blockstore {
...
+        for leaf_index in 0..tree_size {
+            let proof_iter = make_merkle_proof(leaf_index, tree_size, &merkle_tree);
+            let proof: Vec<u8> = proof_iter
+                .map(|proof| proof.map(|p| p.as_slice()))
+                .collect::<std::result::Result<Vec<_>, _>>()
+                .map_err(|_| {
+                    BlockstoreError::FailedDoubleMerkleRootConstruction(slot, block_location)
+                })?
+                .into_iter()
+                .flatten()
```
AshwinSekar: opted to just store the proof as flattened to avoid the extra work when serving repair - will have to flatten when sending over the wire and unflatten when receiver verifies anyway

ledger/src/blockstore/blockstore_purge.rs
```diff
@@ -437,6 +441,14 @@ impl Blockstore {
                 .alt_merkle_root_meta_cf
                 .delete_file_in_range(from_slot, to_slot)
                 .is_ok()
+            & self
+                .parent_meta_cf
```
ksn6, line 445: oof, good catch

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,128 @@ impl Blockstore {
...
+                    BlockstoreProcessorError::FailedDoubleMerkleRootConstruction(
+                        slot,
+                        block_location,
+                    )
+                })?
+                .into_iter()
+                .flatten()
+                .copied()
+                .collect();
+            debug_assert!(
```
ksn6: debug_assert_eq!

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,128 @@ impl Blockstore {
         }
     }
 
+    /// Fetches (and populates if needed) the DoubleMerkleMeta for the given block.
+    /// Returns the double_merkle_root
+    ///
+    /// Should only be used on full blocks.
+    pub fn get_or_compute_double_merkle_root(
```
ksn6: Are you planning on invoking this fn within blockstore in a follow-up PR?

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,128 @@ impl Blockstore {
...
+        }
+
+        // Compute double merkle - slot must be full at this point
+        let Some(slot_meta) = self
+            .meta_cf
+            .get(slot)
+            .expect("Blockstore operations must succeed")
+        else {
+            return Err(BlockstoreProcessorError::FailedToLoadMeta);
+        };
```
ksn6: Rather than let-else-return error, we can do (here and elsewhere in the function):

```suggestion
        let slot_meta = self
            .meta_cf
            .get(slot)
            .expect("Blockstore operations must succeed")
            .ok_or(BlockstoreProcessorError::FailedToLoadMeta)?;
```

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,128 @@ impl Blockstore {
...
+
+        if !slot_meta.is_full() {
+            return Err(BlockstoreProcessorError::SlotNotFull(slot, block_location));
+        }
+
+        let Some(last_index) = slot_meta.last_index else {
+            return Err(BlockstoreProcessorError::SlotNotFull(slot, block_location));
+        };
```
ksn6:
```suggestion
        let last_index = slot_meta
            .last_index
            .filter(|_| slot_meta.is_full())
            .ok_or(BlockstoreProcessorError::SlotNotFull(slot, block_location))?;
```

ksn6:
> Next PR will actually call this fn and set block id in either replay or broadcast

We'll need to set the block id for the header in broadcast atm, as that's where we construct the header.

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,128 @@ impl Blockstore {
...
+
+        // Collect merkle roots for each FEC set
+        let mut merkle_tree_leaves = Vec::with_capacity(fec_set_count + 1);
+
+        for i in 0..fec_set_count {
+            let fec_set_index = (i * DATA_SHREDS_PER_FEC_BLOCK) as u32;
+            let erasure_set_id = ErasureSetId::new(slot, fec_set_index);
+
+            let Some(merkle_root) = self
+                .merkle_root_meta_from_location(erasure_set_id, block_location)
```
ksn6: A single invocation to `.multi_get_keys` is probably better here, in case we don't get to the online merkle idea.

AshwinSekar: yeah good call that's definitely better - updated

ledger/src/blockstore.rs
AshwinSekar: yeah good call, this fn can just return an `Option`, any error apart from the slot not being full indicates an inconsistency in blockstore or a bug in the merkle tree impl

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,119 @@ impl Blockstore {
...
+            .get(slot)
+            .expect("Blockstore operations must succeed")?;
+
+        if !slot_meta.is_full() {
+            return None;
+        }
+
+        let last_index = slot_meta.last_index.expect("Slot is full");
+
+        // This function is only used post Alpenglow, so implicitely gated by SIMD-0317 as that is a prereq
```
ksn6: nit: implicitly

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,119 @@ impl Blockstore {
...
+
+        let parent_meta = self
+            .parent_meta_cf
+            .get((slot, block_location))
+            .expect("Blockstore operations must succeed")
+            .expect("Slot cannot be full without parent");
+
+        // Collect merkle roots for each FEC set
+        let fec_set_indices =
+            (0..fec_set_count).map(|i| (slot, (i * DATA_SHREDS_PER_FEC_BLOCK) as u32));
```
ksn6, line 924: Would use `checked_mul` + return `None` on failure, just in case.

Slightly far-fetched - given proposals to remove CU limits and have block limits only be time-based, if a malicious leader constructed a block with a very large number of no-ops, I wonder whether we could have `2^32 / DATA_SHREDS_PER_FEC_SET_BLOCK ~ 135M` FEC sets, which would then cause the cluster to crash here.

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,119 @@ impl Blockstore {
...
+                Ok(merkle_root)
+            })
+            .collect::<Vec<_>>();
+
+        // Add parent info as the last leaf
+        let parent_info_hash = hashv(&[
+            &parent_meta.parent_slot.to_le_bytes(),
+            parent_meta.parent_block_id.as_ref(),
+        ]);
+        merkle_tree_leaves.push(Ok(parent_info_hash));
```
ksn6:
```suggestion
      let merkle_tree_leaves: Vec<_> = self
          .merkle_root_meta_cf
          .multi_get_bytes(&keys)
          .map(|get_result| {
              let bytes = get_result
                  .expect("Blockstore operations must succeed")
                  .expect("Merkle root meta must exist for all fec sets in full slot");
              bincode::deserialize::<MerkleRootMeta>(bytes.as_ref())
                  .expect("Merkle root meta column only contains valid MerkleRootMetas")
                  .merkle_root()
                  .expect("Legacy shreds no longer exist, merkle root must be present")
          })
          // Add parent info as the last leaf
          .chain(std::iter::once(hashv(&[
              &parent_meta.parent_slot.to_le_bytes(),
              parent_meta.parent_block_id.as_ref(),
          ])))
          .collect();
```

ledger/src/blockstore.rs
```diff
@@ -866,6 +874,119 @@ impl Blockstore {
...
+                        .as_slice()
+                })
+                .copied()
+                .collect();
+            debug_assert_eq!(
+                proof.len(),
+                get_proof_size(tree_size) as usize * SIZE_OF_MERKLE_PROOF_ENTRY
+            );
+            proofs.push(proof);
+        }
```
ksn6:
```suggestion
        let tree_size = fec_set_count + 1;
        let proofs: Vec<Vec<u8>> = (0..tree_size)
            .map(|leaf_index| {
                make_merkle_proof(leaf_index, tree_size, &merkle_tree)
                    .map(|hash| hash.expect("Merkle proof construction cannot fail"))
                    .flat_map(|hash| hash.as_ref())
                    .copied()
                    .collect()
            })
            .inspect(|proof| {
                debug_assert_eq!(
                    proof.len(),
                    get_proof_size(tree_size) as usize * SIZE_OF_MERKLE_PROOF_ENTRY
                );
            })
            .collect();
```

ksn6 (approved):
LGTM

ledger/src/blockstore.rs
AshwinSekar: we're actually protected by that type of attack here https://github.com/anza-xyz/alpenglow/blob/079cd89725b6b6b2023ea5875a91979e5375bad5/ledger/src/shred.rs#L801-L803

Before the shred is even inserted into blockstore we verify that the index is less than 32k, so `fec_set_count` cannot be more than ~1000

PR #614, opened 2025-11-28
block component: process one at a time

ksn6:
Per SIMD-0337, there is a a one-to-one mapping between one range in `CompletedRanges` and one `BlockComponent`. I.e., a `CompletedRanges` with `n` ranges resolves to exactly `n` `BlockComponent`s.

We can simplify `BlockComponent` parsing as a result.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -4299,13 +4299,7 @@ impl Blockstore {
         let components =
             self.get_slot_components_in_block(slot, &completed_ranges, Some(&slot_meta))?;
 
-        // TODO(ksn): get rid of this once we remove BlockComponent::from_bytes_multiple
-        if completed_ranges.len() != components.len() {
-            return Err(BlockstoreError::BlockComponentMisalignment(
-                slot,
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore.rs
```diff
@@ -4512,13 +4512,16 @@ impl Blockstore {
     ///   completed_ranges = [..., (s_i..e_i), (s_i+1..e_i+1), ...]
     /// Then, the following statements are true:
     ///   s_i < e_i == s_i+1 < e_i+1
+    /// Note that one CompletedRanges corresponds to one BlockComponent.
     fn get_slot_components_in_block(
         &self,
         slot: Slot,
         completed_ranges: &CompletedRanges,
         slot_meta: Option<&SlotMeta>,
     ) -> Result<Vec<BlockComponent>> {
```
wen-coding, line 4515: We probably should change the return type as well?

ksn6: Had an error in the PR description - fixed this and updated the code; accordingly, the return type here should be correct.

PR #615, opened 2025-11-29
block component: introduce wincode

ksn6:
`block_component.rs` is getting hard to manage. Let's use `wincode` to solve this.

Notes:

- `block_component_v2.rs` now has close to zero custom serde code, and is much simpler to maintain. We have tests to validate that behavior identically matches that of `block_component.rs`.

- Once this lands, we'll entirely remove `block_component.rs`.

- This PR does not use `wincode` to serde `EntryBatch`-es just yet. We'll do this as soon as we upstream to `agave`. I tried cherry-picking the PRs to get `Entry` to implement `SchemaRead` / `SchemaWrite`, but ran into some versioning problems. Could probably get this to work with some effort, but there's no point, since we'll be upstreaming soon anyways.

### DIFF
--- a/entry/src/block_component_v2.rs
+++ b/entry/src/block_component_v2.rs
@@ -97,13 +97,13 @@ use {
     solana_clock::Slot,
     solana_hash::Hash,
     solana_votor_messages::consensus_message::{Certificate, CertificateType},
-    std::{marker::PhantomData, mem::MaybeUninit},
+    std::mem::MaybeUninit,
     wincode::{
         containers::{Pod, Vec as WincodeVec},
-        error::{preallocation_size_limit, write_length_encoding_overflow},
+        error::write_length_encoding_overflow,
         io::{Reader, Writer},
         len::{BincodeLen, SeqLen},
-        ReadResult, SchemaRead, SchemaWrite, WriteResult,
+        ReadResult, SchemaRead, SchemaWrite, TypeMeta, WriteResult,
     },
 };
 
@@ -128,51 +128,17 @@ impl SeqLen for U8Len {
     }
 }
 
-/// Default maximum preallocation size for U16Len (4 MiB).
-/// This is a safety precaution against malicious input causing OOM.
-const DEFAULT_U16_LEN_MAX_SIZE: usize = 4 << 20;
-
-/// 2-byte length prefix (max 65535 elements).
-#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
-pub struct U16Len<const MAX_SIZE: usize = DEFAULT_U16_LEN_MAX_SIZE>;
-
-impl<const MAX_SIZE: usize> SeqLen for U16Len<MAX_SIZE> {
-    fn read<'de, T>(reader: &mut impl Reader<'de>) -> ReadResult<usize> {
-        let len = u16::get(reader)? as usize;
-        let needed = len
-            .checked_mul(std::mem::size_of::<T>())
-            .ok_or_else(|| preallocation_size_limit(usize::MAX, MAX_SIZE))?;
-        if needed > MAX_SIZE {
-            return Err(preallocation_size_limit(needed, MAX_SIZE));
-        }
-        Ok(len)
-    }
-
-    fn write(writer: &mut impl Writer, len: usize) -> WriteResult<()> {
-        let Ok(len): Result<u16, _> = len.try_into() else {
-            return Err(write_length_encoding_overflow("u16::MAX"));
-        };
-        Ok(writer.write(&len.to_le_bytes())?)
-    }
-
-    fn write_bytes_needed(_len: usize) -> WriteResult<usize> {
-        Ok(2)
-    }
-}
-
-/// Wraps a value with a length prefix for TLV-style serialization.
+/// Wraps a value with a u16 length prefix for TLV-style serialization.
+///
+/// The length prefix represents the serialized byte size of the inner value.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct LengthPrefixed<T, L: SeqLen = U16Len> {
+pub struct LengthPrefixed<T> {
     inner: T,
-    _marker: PhantomData<L>,
 }
 
-impl<T, L: SeqLen> LengthPrefixed<T, L> {
+impl<T> LengthPrefixed<T> {
     pub fn new(inner: T) -> Self {
-        Self {
-            inner,
-            _marker: PhantomData,
-        }
+        Self { inner }
     }
 
     pub fn inner(&self) -> &T {
@@ -180,27 +146,45 @@ impl<T, L: SeqLen> LengthPrefixed<T, L> {
     }
 }
 
-impl<T: SchemaWrite<Src = T>, L: SeqLen> SchemaWrite for LengthPrefixed<T, L> {
+impl<T: SchemaWrite<Src = T>> SchemaWrite for LengthPrefixed<T> {
     type Src = Self;
+
+    const TYPE_META: TypeMeta = match T::TYPE_META {
+        TypeMeta::Static { size, zero_copy } => TypeMeta::Static {
+            size: size + std::mem::size_of::<u16>(),
+            zero_copy,
+        },
+        TypeMeta::Dynamic => TypeMeta::Dynamic,
+    };
 
     fn size_of(src: &Self::Src) -> WriteResult<usize> {
         let inner_size = T::size_of(&src.inner)?;
-        let len_size = L::write_bytes_needed(inner_size)?;
-        Ok(len_size + inner_size)
+        Ok(std::mem::size_of::<u16>() + inner_size)
     }
 
     fn write(writer: &mut impl Writer, src: &Self::Src) -> WriteResult<()> {
         let inner_size = T::size_of(&src.inner)?;
-        L::write(writer, inner_size)?;
+        let Ok(len): Result<u16, _> = inner_size.try_into() else {
+            return Err(write_length_encoding_overflow("u16::MAX"));
+        };
+        u16::write(writer, &len)?;
         T::write(writer, &src.inner)
     }
 }
 
-impl<'de, T: SchemaRead<'de, Dst = T>, L: SeqLen> SchemaRead<'de> for LengthPrefixed<T, L> {
+impl<'de, T: SchemaRead<'de, Dst = T>> SchemaRead<'de> for LengthPrefixed<T> {
     type Dst = Self;
 
+    const TYPE_META: TypeMeta = match T::TYPE_META {
+        TypeMeta::Static { size, zero_copy } => TypeMeta::Static {
+            size: size + std::mem::size_of::<u16>(),
+            zero_copy,
+        },
+        TypeMeta::Dynamic => TypeMeta::Dynamic,
+    };
+
     fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
-        let _len = L::read::<Self>(reader)?;
+        let _len = u16::get(reader)?;
         let inner_dst = unsafe { &mut *(&raw mut (*dst.as_mut_ptr()).inner).cast() };
         T::read(reader, inner_dst)?;
         Ok(())
@@ -328,10 +312,10 @@ pub enum VersionedUpdateParent {
 #[derive(Debug, Clone, PartialEq, Eq, SchemaWrite, SchemaRead)]
 #[wincode(tag_encoding = "u8")]
 pub enum BlockMarkerV1 {
-    BlockFooter(LengthPrefixed<VersionedBlockFooter, U16Len>),
-    BlockHeader(LengthPrefixed<VersionedBlockHeader, U16Len>),
-    UpdateParent(LengthPrefixed<VersionedUpdateParent, U16Len>),
-    GenesisCertificate(LengthPrefixed<GenesisCertificate, U16Len>),
+    BlockFooter(LengthPrefixed<VersionedBlockFooter>),
+    BlockHeader(LengthPrefixed<VersionedBlockHeader>),
+    UpdateParent(LengthPrefixed<VersionedUpdateParent>),
+    GenesisCertificate(LengthPrefixed<GenesisCertificate>),
 }
 
 impl BlockMarkerV1 {


entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+
+/// 1-byte length prefix (max 255 elements).
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct U8Len;
+
+impl SeqLen for U8Len {
+    fn read<'de, T>(reader: &mut impl Reader<'de>) -> ReadResult<usize> {
+        let mut buf = [MaybeUninit::uninit(); 1];
+        reader.copy_into_slice(&mut buf)?;
+        Ok(unsafe { buf[0].assume_init() } as usize)
```
cpubot:
```suggestion
        u8::get(reader).map(|len| len as usize)
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    fn write(writer: &mut impl Writer, len: usize) -> WriteResult<()> {
+        if len > u8::MAX as usize {
+            return Err(WriteError::Custom("Length exceeds u8::MAX"));
+        }
+        writer.write(&[len as u8]).map_err(WriteError::Io)
```
cpubot:
```suggestion
        let Ok(len) = len.try_into() else {
            return Err(write_length_encoding_overflow("u8::MAX"));
        };
        Ok(writer.write(&[len])?)
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct U16Len;
+
+impl SeqLen for U16Len {
+    fn read<'de, T>(reader: &mut impl Reader<'de>) -> ReadResult<usize> {
+        let mut len_bytes = [MaybeUninit::uninit(); 2];
+        reader.copy_into_slice(&mut len_bytes)?;
+        // UNSAFE: copy_into_slice initializes all bytes
+        let bytes = unsafe { [len_bytes[0].assume_init(), len_bytes[1].assume_init()] };
+        Ok(u16::from_le_bytes(bytes) as usize)
```
cpubot:
```suggestion
        u16::get(reader).map(|len| len as usize)
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    fn write(writer: &mut impl Writer, len: usize) -> WriteResult<()> {
+        if len > u16::MAX as usize {
+            return Err(WriteError::Custom("Length exceeds u16::MAX"));
+        }
+        writer
+            .write(&(len as u16).to_le_bytes())
+            .map_err(WriteError::Io)
```
cpubot:
```suggestion
        let Ok(len): Result<u16, _> = len.try_into() else {
            return Err(write_length_encoding_overflow("u16::MAX"));
        };
        Ok(writer.write(&len.to_le_bytes())?)
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+        Ok(1)
+    }
+}
+
+/// 2-byte length prefix (max 65535 elements).
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct U16Len;
+
+impl SeqLen for U16Len {
+    fn read<'de, T>(reader: &mut impl Reader<'de>) -> ReadResult<usize> {
```
cpubot: May be worth adding a preallocation size limit in the u16 case [like `BincodeLen` does](https://docs.rs/wincode/latest/src/wincode/len.rs.html#29-49) -- but at your discretion. `u8` version is probably fine without.

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+}
+
+impl<'de, T: SchemaRead<'de, Dst = T>, L: SeqLen> SchemaRead<'de> for LengthPrefixed<T, L> {
+    type Dst = Self;
+
+    fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        let _len = L::read::<Self>(reader)?;
+        let mut inner_dst = MaybeUninit::uninit();
+        T::read(reader, &mut inner_dst)?;
+        dst.write(Self::new(unsafe { inner_dst.assume_init() }));
```
cpubot: We can avoid putting the `inner` on the stack by writing directly to `dst`

```suggestion
        let _len = L::read::<Self>(reader)?;
        let inner_dst = unsafe { &mut *(&raw mut (*dst.as_mut_ptr()).inner).cast() };
        T::read(reader, inner_dst)?;
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    fn write_bytes_needed(_len: usize) -> WriteResult<usize> {
+        Ok(2)
+    }
+}
+
+/// Wraps a value with a length prefix for TLV-style serialization.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LengthPrefixed<T, L: SeqLen = U16Len> {
```
cpubot: It appears `LengthPrefixed` is used for data structs rather than sequences. What is length meant to denote here? The serialized size of the inner variant?

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    #[wincode(tag = 1)]
+    BlockHeader(LengthPrefixed<VersionedBlockHeader, U16Len>),
+    #[wincode(tag = 2)]
+    UpdateParent(LengthPrefixed<VersionedUpdateParent, U16Len>),
+    #[wincode(tag = 3)]
+    GenesisCertificate(LengthPrefixed<GenesisCertificate, U16Len>),
+}
+
+impl BlockMarkerV1 {
+    pub fn new_block_footer(f: VersionedBlockFooter) -> Self {
```
cpubot, line 322: I'm a bit confused as to what `LengthPrefixed` does in this case. Presumably each variant might have a different serialized size, so it seems a bit strange for it to exist solely as a type parameter on `LengthPrefixed` (presumably we need to embed the actual size of the variant in the struct?).

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+}
+
+impl VersionedBlockMarker {
+    pub const fn new(marker: BlockMarkerV1) -> Self {
+        Self::V1(marker)
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
+        wincode::serialize(self)
+            .map_err(|e| BlockComponentError::SerializationFailed(e.to_string()))
```
cpubot: Why not add variants for both `wincode::ReadError` and `wincode::WriteError` in the `BlockComponentError`?

E.g.,
```rs
#[derive(Debug, thiserror::Error)]
pub enum BlockComponentError {
    // ...
    #[error(transparent)]
    SerializationFailed(#[from] wincode::WriteError),
    #[error(transparent)]
    DeserializationFailed(#[from] wincode::ReadError),
}
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+            }
+            Self::UnsupportedVersion { version } => write!(f, "Unsupported version: {version}"),
+            Self::CursorOutOfBounds => write!(f, "Cursor out of bounds"),
+            Self::SerializationFailed(msg) => write!(f, "Serialization failed: {msg}"),
+            Self::DeserializationFailed(msg) => write!(f, "Deserialization failed: {msg}"),
+        }
+    }
+}
+
+impl Error for BlockComponentError {}
```
cpubot: `thiserror` would simplify this

```suggestion
#[derive(Debug, thiserror::Error)]
pub enum BlockComponentError {
    #[error("Insufficient data")]
    InsufficientData,
    #[error("Entry count {count} exceeds max {max}")]
    TooManyEntries { count: usize, max: usize },
    #[error("Entry batch cannot be empty")]
    EmptyEntryBatch,
    #[error("Unknown {variant_type} variant: {id}")]
    UnknownVariant { variant_type: String, id: u8 },
    #[error("Unsupported version: {version}")]
    UnsupportedVersion { version: u16 },
    #[error("Cursor out of bounds")]
    CursorOutOfBounds,
    #[error(transparent)]
    SerializationFailed(#[from] wincode::WriteError),
    #[error(transparent)]
    DeserializationFailed(#[from] wincode::ReadError),
}
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
+        match self {
+            Self::EntryBatch(entries) => bincode::serialize(entries)
+                .map_err(|e| BlockComponentError::SerializationFailed(e.to_string())),
+            Self::BlockMarker(marker) => {
+                let mut buf = 0u64.to_le_bytes().to_vec();
+                buf.extend(marker.to_bytes()?);
+                Ok(buf)
```
cpubot: Can avoid the extra vec allocation of `0u64.to_le_bytes().to_vec()` and avoid a resize / realloc from the `buf.extend(marker.to_bytes()?)`
```suggestion
                let size = wincode::serialized_size(marker)? as usize + size_of::<u64>();
                let mut buffer = Vec::with_capacity(size);
                buffer.extend_from_slice(&0u64.to_le_bytes());
                let mut target = buffer.spare_capacity_mut();
                wincode::serialize_into(&mut target, marker)?;
                Ok(buffer)
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    pub const fn as_marker(&self) -> Option<&VersionedBlockMarker> {
+        match self {
+            Self::BlockMarker(m) => Some(m),
+            _ => None,
+        }
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
+        match self {
+            Self::EntryBatch(entries) => bincode::serialize(entries)
```
cpubot: `Entry` supports wincode as of a few months ago -- may need a resync with agave

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    #[wincode(tag = 1)]
+    V1(BlockMarkerV1),
+}
+
+impl VersionedBlockMarker {
+    pub const fn new(marker: BlockMarkerV1) -> Self {
+        Self::V1(marker)
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
```
cpubot: Not sure we need these `to_bytes` and `from_bytes` functions. Users can just `wincode::serialize` and `wincode::deserialize`.

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    pub const fn as_marker(&self) -> Option<&VersionedBlockMarker> {
+        match self {
+            Self::BlockMarker(m) => Some(m),
+            _ => None,
+        }
+    }
+
+    pub fn to_bytes(&self) -> Result<Vec<u8>, BlockComponentError> {
```
cpubot: Presumably we should just implement `SchemaWrite` for `BlockComponent`

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+                .map_err(|e| BlockComponentError::SerializationFailed(e.to_string())),
+            Self::BlockMarker(marker) => {
+                let mut buf = 0u64.to_le_bytes().to_vec();
+                buf.extend(marker.to_bytes()?);
+                Ok(buf)
+            }
+        }
+    }
+
+    pub fn from_bytes(data: &[u8]) -> Result<(Self, usize), BlockComponentError> {
```
cpubot: Similarly we should just implement `SchemaRead` for `BlockComponent`

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,678 @@
...
+    }
+
+    pub fn infer_is_entry_batch(data: &[u8]) -> Option<bool> {
+        data.get(..8)?
+            .try_into()
+            .ok()
+            .map(|b| u64::from_le_bytes(b) != 0)
+    }
+
+    pub fn serialized_size(&self) -> Result<u64, BlockComponentError> {
```
cpubot: Should get this from implementing `SchemaWrite`

entry/src/block_component_v2.rs
ksn6: doing a full `agave` resync is pretty challenging at this point - we'll definitely get this on the upstream, though

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: > It appears LengthPrefixed is used for data structs rather than sequences.

Correct.

> What is length meant to denote here? The serialized size of the inner variant?

Correct - it's the serialized size of the inner variant.

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: `LengthPrefixed` purely emits the length (in bytes) of the inner type when serialized.

For now, we don't need to use this length anywhere, which is why I'm not embedding the size of the variant in the struct directly.

For a bit more context - this length field is something that Jump requested when creating `BlockMarker` variants; they're interested in TLV serialization for FPGA-related things.

entry/src/block_component_v2.rs
ksn6: (rewriting this in `SchemaWrite`)

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: nice! done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
cpubot: I see. Thanks for clarifying.

I don't think using `SeqLen` is necessary for `LengthPrefixed`, given we aren't using it to protect against sequence preallocations. Additionally, `SeqLen` doesn't have a [`TypeMeta`](https://docs.rs/wincode/latest/wincode/enum.TypeMeta.html) constant, which means that anything using `LengthPrefixed` immediately loses [static](https://docs.rs/wincode/latest/wincode/enum.TypeMeta.html#variant.Static) size optimizations. 

Another observation --  everything using `LengthPrefixed` uses `u16` -- do we actually need that to be generic over any type? If we do, I suggest constraining the type to `SchemaRead` and `SchemaWrite` rather than `SeqLen` so we can implement `TypeMeta` for `LengthPrefixed`.

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+impl<'de> SchemaRead<'de> for BlockComponent {
+    type Dst = Self;
+
+    fn read(reader: &mut impl Reader<'de>, dst: &mut MaybeUninit<Self::Dst>) -> ReadResult<()> {
+        // Read the entry count (first 8 bytes) to determine variant
+        let count_bytes = reader.fill_array::<8>()?;
+        let entry_count = u64::from_le_bytes(*count_bytes);
```
cpubot, line 470:
```suggestion
        let entry_count = u64::get(reader)?;
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+
+        if entry_count == 0 {
+            // This is a BlockMarker - consume the count bytes and read the marker
+            reader.consume(8)?;
```
cpubot, line 474: Can be removed once `u64::get` is used above

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+            let mut marker_dst = MaybeUninit::uninit();
+            <VersionedBlockMarker as SchemaRead>::read(reader, &mut marker_dst)?;
+            dst.write(Self::BlockMarker(unsafe { marker_dst.assume_init() }));
```
cpubot:
```suggestion
            dst.write(Self::BlockMarker(VersionedBlockMarker::get(reader)?));
```

cpubot: You can avoid the `MaybeUninit::uninit()` -> `T::Read` -> `assume_init()` in general with [`get`](https://docs.rs/wincode/latest/wincode/trait.SchemaRead.html#method.get)

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+            }
+        }
+    }
+
+    fn write(writer: &mut impl Writer, src: &Self::Src) -> WriteResult<()> {
+        match src {
+            Self::EntryBatch(entries) => {
+                // TODO(ksn): replace with wincode:: upon upstreaming to Agave.
+                let bytes = bincode::serialize(entries).map_err(|e| {
+                    wincode::WriteError::Custom(Box::leak(e.to_string().into_boxed_str()))
```
cpubot: `Box::leak` is unfortunate -- recommend just using a static "bincode error" for now until this is upstreamed into agave.

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+pub enum VersionedUpdateParent {
+    #[wincode(tag = 1)]
+    V1(UpdateParentV1),
+}
+
+/// TLV-encoded marker variants.
+#[derive(Debug, Clone, PartialEq, Eq, SchemaWrite, SchemaRead)]
+#[wincode(tag_encoding = "u8")]
+pub enum BlockMarkerV1 {
+    #[wincode(tag = 0)]
```
cpubot: The tags here are exactly what would be generated by `wincode` by default -- do we need the explicit tags here?

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+
+/// Attests to genesis block finalization with a BLS aggregate signature.
+#[derive(Clone, PartialEq, Eq, Debug, SchemaWrite, SchemaRead)]
+pub struct GenesisCertificate {
+    pub slot: Slot,
+    #[wincode(with = "Pod<Hash>")]
+    pub block_id: Hash,
+    #[wincode(with = "Pod<BLSSignature>")]
+    pub bls_signature: BLSSignature,
+    #[wincode(with = "WincodeVec<Pod<u8>, BincodeLen>")]
```
cpubot: `Pod` is not necessary for `u8` since wincode `v0.2.0`
```suggestion
    #[wincode(with = "WincodeVec<u8, BincodeLen>")]
```

entry/src/block_component_v2.rs
```diff
@@ -0,0 +1,692 @@
...
+    pub new_parent_block_id: Hash,
+}
+
+/// Block production metadata. User agent is capped at 255 bytes.
+#[derive(Clone, PartialEq, Eq, Debug, SchemaWrite, SchemaRead)]
+pub struct BlockFooterV1 {
+    #[wincode(with = "Pod<Hash>")]
+    pub bank_hash: Hash,
+    pub block_producer_time_nanos: u64,
+    #[wincode(with = "WincodeVec<Pod<u8>, U8Len>")]
```
cpubot:
```suggestion
    #[wincode(with = "WincodeVec<u8, U8Len>")]
```

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: done

entry/src/block_component_v2.rs
ksn6: kept these in for explicitness - though it's fair to say that tag_encoding above should be a hint that the enum ordering is important; will get rid of the tags here

entry/src/block_component_v2.rs
ksn6: is there a way to do this without advancing the reader?

block components work as follows:
- if we have a block marker, the first eight bytes of a block component are [0_u8; 8]
- if we have an entry batch, the first eight bytes of a block component denote the length of the Vec<Entry>, i.e., the beginning of the bincode representation of the Vec<Entry>

if we advance the reader, this isn't great, since we lose the ability to b/wincode deserialize the Vec

entry/src/block_component_v2.rs
ksn6: (see above)

entry/src/block_component_v2.rs
ksn6: Yep, fair - we're using `u16` everywhere for now, so might as well hard-code that in until we need to generalize. Let me know if this works.

entry/src/block_component_v2.rs
cpubot: Gotcha. Yeah, if we want to avoid advancing, the way you have it is the correct way.

entry/src/block_component_v2.rs
cpubot: Looks good!

PR #616, opened 2025-11-29
block component: switch to wincode

ksn6:
PR #615 introduced `block_component_v2.rs`, which:

- Is identical to `block_component.rs` from a functionality standpoint (demonstrated via tests)
- Is written via `wincode`, resulting in far cleaner logic

This PR:

- Moves the contents of `block_component_v2.rs` into `block_component.rs`, removing the old `BlockComponent` implementation
- Updates call-sites (all cosmetic changes)

### DIFF
--- a/entry/src/block_component.rs
+++ b/entry/src/block_component.rs
@@ -144,6 +144,10 @@ impl<T> LengthPrefixed<T> {
     pub fn inner(&self) -> &T {
         &self.inner
     }
+
+    pub fn into_inner(self) -> T {
+        self.inner
+    }
 }
 
 impl<T: SchemaWrite<Src = T>> SchemaWrite for LengthPrefixed<T> {
--- a/runtime/src/block_component_processor.rs
+++ b/runtime/src/block_component_processor.rs
@@ -111,7 +111,7 @@ impl BlockComponentProcessor {
             BlockMarkerV1::GenesisCertificate(genesis_cert)
                 if markers_fully_enabled || in_migration =>
             {
-                self.on_genesis_certificate(bank, genesis_cert.inner().clone(), migration_status)
+                self.on_genesis_certificate(bank, genesis_cert.into_inner(), migration_status)
             }
 
             // Everything else is only valid once migration is complete


entry/src/block_component.rs
```diff
@@ -52,20 +46,16 @@
 /// └─────────────────────────────────────────┘
 /// ```
 ///
-/// ### BlockFooterV1 Layout
+/// ### BlockHeaderV1 Layout
```
wen-coding, line 49: What is the reason we are changing order here? Makes the code review harder.

entry/src/block_component.rs
```diff
@@ -99,62 +93,142 @@
...
     TooManyEntries { count: usize, max: usize },
-    /// EntryBatch is empty when it shouldn't be
     EmptyEntryBatch,
-    /// Unknown variant identifier
     UnknownVariant { variant_type: String, id: u8 },
-    /// Unsupported version number
     UnsupportedVersion { version: u16 },
-    /// Data length conversion failed (e.g., usize to u16)
-    DataLengthOverflow,
-    /// Cursor position exceeded data boundary
```
wen-coding, line 128: Why do we not need all the comments any more?

entry/src/block_component.rs
```diff
@@ -99,62 +93,142 @@
...
     SerializationFailed(String),
-    /// Deserialization failed
     DeserializationFailed(String),
 }
 
 impl fmt::Display for BlockComponentError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::InsufficientData => write!(f, "Insufficient data for deserialization"),
+            Self::InsufficientData => write!(f, "Insufficient data"),
```
wen-coding: Can we use #[error(...)] instead?

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
-pub enum VersionedBlockMarker {
-    V1(BlockMarkerV1),
-    Current(BlockMarkerV1),
+/// Block production metadata. User agent is capped at 255 bytes.
+#[derive(Clone, PartialEq, Eq, Debug, SchemaWrite, SchemaRead)]
+pub struct BlockFooterV1 {
+    #[wincode(with = "Pod<Hash>")]
+    pub bank_hash: Hash,
+    pub block_producer_time_nanos: u64,
+    #[wincode(with = "WincodeVec<Pod<u8>, U8Len>")]
```
wen-coding: Do we have a wincode doc somewhere explaining the macros?

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
-    BlockHeader(VersionedBlockHeader),
-    UpdateParent(VersionedUpdateParent),
-    GenesisCertificate(GenesisCertificate),
+/// Attests to genesis block finalization with a BLS aggregate signature.
+#[derive(Clone, PartialEq, Eq, Debug, SchemaWrite, SchemaRead)]
+pub struct GenesisCertificate {
+    pub slot: Slot,
+    #[wincode(with = "Pod<Hash>")]
+    pub block_id: Hash,
+    #[wincode(with = "Pod<BLSSignature>")]
```
wen-coding, line 236: Out of curiosity, where is BLSSignature defined and how can we add new types?

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
-/// ├─────────────────────────────────────────┤
-/// │ User Agent Bytes          (0-255 bytes) │
-/// └─────────────────────────────────────────┘
-/// ```
-#[derive(Clone, PartialEq, Eq, Debug)]
-pub struct BlockFooterV1 {
-    pub bank_hash: Hash,
-    pub block_producer_time_nanos: u64,
-    pub block_user_agent: Vec<u8>,
+impl From<GenesisCertificate> for Certificate {
```
wen-coding, line 270: Can we some day split different type of block component into their own file, instead of piling them in one file here?

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
+            bitmap: cert.bitmap,
+        }
+    }
 }
 
-// ============================================================================
-// Block Header Types
-// ============================================================================
+#[derive(Debug, Clone, PartialEq, Eq, SchemaWrite, SchemaRead)]
+#[wincode(tag_encoding = "u8")]
```
wen-coding, line 281: Hmm, I wonder whether u8 is enough for block footer, it will probably have the most churn.

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
 
-    /// Check if data looks like an entry batch (non-zero entry count). Returns `None` if we can't
-    /// deduce whether the data is an entry batch.
     pub fn infer_is_entry_batch(data: &[u8]) -> Option<bool> {
-        // Per documentation, the first 8 bytes denote the length of an entry batch, where a length
-        // of zero indicates a block marker.
-        data.get(..8)
-            .and_then(|bytes| bytes.try_into().ok())
-            .map(|bytes| u64::from_le_bytes(bytes) != 0)
+        data.get(..8)?
```
wen-coding: Can we make this 8 a constant so it's clearer why we chose this number?

entry/src/block_component.rs
```diff
@@ -163,440 +237,286 @@ impl fmt::Display for BlockComponentError {
...
-    /// Returns an error if size calculation fails.
-    pub fn serialized_size(&self) -> Result<u64, BlockComponentError> {
-        match self {
-            Self::EntryBatch(entries) => bincode::serialized_size(entries)
-                .map_err(|e| BlockComponentError::SerializationFailed(e.to_string())),
-            Self::BlockMarker(marker) => {
-                Ok(Self::ENTRY_COUNT_SIZE as u64 + marker.serialized_size())
-            }
-        }
+        Self::infer_is_entry_batch(data).map(|result| !result)
```
wen-coding: nit: it's easy to misread `! `here, add a comment explaining what it does maybe?

entry/src/block_component.rs
```diff
@@ -1,16 +1,10 @@
-/// Block components and friends.
+/// Block components using wincode serialization.
```
wen-coding, line 1: Do we have a benchmark somewhere?

also: would be nice to link to wincode doc here.

entry/src/block_component.rs
ksn6: turns out we don't need most of the errors - did a bit of cleanup

entry/src/block_component.rs
ksn6: see above - this should all be fixed

entry/src/block_component.rs
ksn6: crates.io: https://crates.io/crates/wincode

entry/src/block_component.rs
ksn6: This is just an alias for `solana_bls_signatures::Signature` (see the imports).

New types should implement `SchemaRead` and `SchemaWrite`. E.g., see `LengthPrefixed` in `block_component.rs`. Also happy to help with writing new types if you're looking to do so.

entry/src/block_component.rs
ksn6: Not sure I see the benefit - most component definitions are very short + `GenesisCertificate`, the most complex one, goes away post-Alpenglow.

entry/src/block_component.rs
ksn6: It'll probably be a while before we get to 255 versions - we're still on v1. At that point, it won't be to hard to change this via SIMD.

entry/src/block_component.rs
ksn6: Could you clarify? `!` negates a bool in Rust - a slice of bytes is either (1) an entry batch, (2) a block marker, or (3) can't infer because we don't have at least 8 bytes.

entry/src/block_component.rs
ksn6: We'll need to bench once we upstream to `agave`. See here: https://github.com/anza-xyz/alpenglow/pull/615:

> This PR does not use wincode to serde EntryBatch-es just yet. We'll do this as soon as we upstream to agave. I tried cherry-picking the PRs to get Entry to implement SchemaRead / SchemaWrite, but ran into some versioning problems. Could probably get this to work with some effort, but there's no point, since we'll be upstreaming soon anyways.

wincode docs are on crates.io: https://crates.io/crates/wincode (probably don't need to mention this in the code)

wen-coding (approved):

runtime/src/block_component_processor.rs
```diff
@@ -108,16 +106,16 @@ impl BlockComponentProcessor {
...
             // - while we're still in the migration phase (to let us advance it)
             BlockMarkerV1::BlockHeader(header) if markers_fully_enabled || in_migration => {
-                self.on_header(&header)
+                self.on_header(header.inner())
             }
             BlockMarkerV1::GenesisCertificate(genesis_cert)
                 if markers_fully_enabled || in_migration =>
             {
-                self.on_genesis_certificate(bank, genesis_cert, migration_status)
+                self.on_genesis_certificate(bank, genesis_cert.inner().clone(), migration_status)
```
AshwinSekar: do we need the clone?

AshwinSekar: maybe a `to_inner` or from  that consumes it?

wen-coding (approved):

PR #617, opened 2025-11-29
Split cert verification into bls-cert-verify to share between BLS Sigverifier and Runtime.

wen-coding:
#### Problem
We need to verify certs in Block header from runtime.

#### Summary of Changes
- This is the first step to just move cert verification logic to bls-cert-verify
- We will add cert limit checking in next PR.

### DIFF
--- a/bls-cert-verify/src/cert_verify.rs
+++ b/bls-cert-verify/src/cert_verify.rs
@@ -40,6 +40,10 @@ fn aggregate_keys_from_bitmap<F>(
 where
     F: Fn(usize) -> Option<BlsPubkey> + Sync,
 {
+    // This function should return None if any of the following happens:
+    // - A rank in the bitmap does not have a corresponding pubkey.
+    // - A pubkey fails to convert to projective form.
+    // It only returns Some(aggregated_pubkey) if all pubkeys are found and valid.
     let pubkeys: Vec<PubkeyProjective> = bit_vec
         .iter_ones()
         .map(|rank| {


bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+    let pubkeys: Result<Vec<_>, _> = bit_vec
+        .iter_ones()
+        .filter_map(rank_to_pubkey)
+        .map(PubkeyProjective::try_from)
+        .collect();
+    let pubkeys = pubkeys.ok()?;
+    PubkeyProjective::par_aggregate(pubkeys.par_iter()).ok()
+}
+
+pub fn verify_base2_certificate<F>(
```
ksn6, line 60: nit: make this private

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -48,38 +47,13 @@ fn get_key_to_rank_map(bank: &Bank, slot: Slot) -> Option<&Arc<BLSPubkeyToRankMa
...
-        .filter_map(|rank| key_to_rank_map.get_pubkey(rank))
-        .map(|(_, bls_pubkey)| PubkeyProjective::try_from(*bls_pubkey))
-        .collect();
-    let pubkeys = pubkeys.ok()?;
-    PubkeyProjective::par_aggregate(pubkeys.par_iter()).ok()
-}
-
 #[derive(Debug, Error, PartialEq)]
-enum CertVerifyError {
+enum CertVerificationError {
```
ksn6: This is confusing - why do we have a `CertVerificationError` here and a `CertVerifyError` in `cert_verify.rs`?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+            verify_base2_certificate(cert_to_verify, &bit_vec, &rank_to_pubkey)
+        }
+        solana_signer_store::Decoded::Base3(bit_vec1, bit_vec2) => {
+            verify_base3_certificate(cert_to_verify, &bit_vec1, &bit_vec2, &rank_to_pubkey)
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
ksn6, line 140: The tests here look similar to the ones in `bls_sigverifier.rs` - do we need both / can we delete the equivalents in `bls_sigverifier.rs`?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+) -> Option<PubkeyProjective>
+where
+    F: Fn(usize) -> Option<BlsPubkey> + Sync,
+{
+    let pubkeys: Result<Vec<_>, _> = bit_vec
+        .iter_ones()
+        .filter_map(rank_to_pubkey)
+        .map(PubkeyProjective::try_from)
+        .collect();
+    let pubkeys = pubkeys.ok()?;
```
ksn6:
```suggestion
    let pubkeys: Vec<_> = bit_vec
        .iter_ones()
        .filter_map(rank_to_pubkey)
        .filter_map(|pk| PubkeyProjective::try_from(pk).ok())
        .collect();
```

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+        return Err(CertVerifyError::KeyAggregationFailed);
+    };
+
+    if let Ok(true) =
+        aggregate_bls_pubkey.verify_signature(&cert_to_verify.signature, &signed_payload)
+    {
+        Ok(())
+    } else {
+        Err(CertVerifyError::SignatureVerificationFailed)
+    }
```
ksn6, line 81: Lots of if-else / let else statements; can simplify:

```suggestion
    let original_vote = cert_to_verify.cert_type.to_source_vote();
    let signed_payload =
        bincode::serialize(&original_vote).map_err(|_| CertVerifyError::SerializationFailed)?;
    let aggregate_pubkey =
        aggregate_keys_from_bitmap(bit_vec, rank_to_pubkey).ok_or(CertVerifyError::KeyAggregationFailed)?;

    aggregate_pubkey
        .verify_signature(&cert_to_verify.signature, &signed_payload)
        .ok()
        .filter(|&valid| valid)
        .ok_or(CertVerifyError::SignatureVerificationFailed)
```

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    match SignatureProjective::par_verify_distinct_aggregated(
+        &pubkeys_affine,
+        &cert_to_verify.signature,
+        &messages_to_verify,
+    ) {
+        Ok(true) => Ok(()),
+        _ => Err(CertVerifyError::SignatureVerificationFailed),
+    }
+}
```
ksn6:
```suggestion
    let (vote1, vote2) = cert_to_verify
        .cert_type
        .to_source_votes()
        .ok_or(CertVerifyError::Base3EncodingOnUnexpectedCert(cert_to_verify.cert_type))?;

    let signed_payload1 =
        bincode::serialize(&vote1).map_err(|_| CertVerifyError::SerializationFailed)?;
    let signed_payload2 =
        bincode::serialize(&vote2).map_err(|_| CertVerifyError::SerializationFailed)?;

    let agg_pk1 =
        aggregate_keys_from_bitmap(bit_vec1, rank_to_pubkey).ok_or(CertVerifyError::KeyAggregationFailed)?;
    let agg_pk2 =
        aggregate_keys_from_bitmap(bit_vec2, rank_to_pubkey).ok_or(CertVerifyError::KeyAggregationFailed)?;

    let pubkeys: Vec<BlsPubkey> = vec![agg_pk1.into(), agg_pk2.into()];
    let messages: Vec<&[u8]> = vec![&signed_payload1, &signed_payload2];

    SignatureProjective::par_verify_distinct_aggregated(&pubkeys, &cert_to_verify.signature, &messages)
        .ok()
        .filter(|&valid| valid)
        .ok_or(CertVerifyError::SignatureVerificationFailed)
```

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,270 @@
...
+) -> Result<(), CertVerifyError>
+where
+    F: Fn(usize) -> Option<BlsPubkey> + Sync,
+{
+    let decoded_bitmap = match decode(&cert_to_verify.bitmap, max_len) {
+        Ok(decoded) => decoded,
+        Err(e) => {
+            return Err(CertVerifyError::BitmapDecodingFailed(e));
+        }
+    };
```
ksn6:
```suggestion
    let decoded_bitmap =
        decode(&cert_to_verify.bitmap, max_len).map_err(CertVerifyError::BitmapDecodingFailed)?;
```

bls-cert-verify/src/cert_verify.rs
wen-coding: This will be used by cert verify in runtime.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: We need an error type in cert_verify.rs, it's natural to give it the name CertVerifyError, so I just renamed the Error type with the same name in bls_sigverifier.rs.

bls-cert-verify/src/cert_verify.rs
wen-coding: Yeah, thought about it, since we will add cert limit check later, maybe clean up the tests at that point?

bls-cert-verify/src/cert_verify.rs
ksn6: Should `verify_base3_certificate` also be public then? Seems odd to have one be public and another be private.

core/src/bls_sigverify/bls_sigverifier.rs
ksn6: Having two error classes named `CertVerificationError` and `CertVerifyError` is confusing - ideally we'd have the same error name in both places and use qualifiers, or if possible, just use a single error class and use everywhere.

bls-cert-verify/src/cert_verify.rs
ksn6: Could you clarify what you mean by "we will add cert limit check later?" Not sure how this relates to deleting the equivalent tests in `bls_sigverifier.rs`.

bls-cert-verify/src/cert_verify.rs
wen-coding: As far as I know currently no cert in block marker will use base3 encoding, so let's only make it public when needed.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Changed.

bls-cert-verify/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+license = { workspace = true }
+edition = { workspace = true }
+
+[dependencies]
+bincode = { workspace = true }
+bitvec = { workspace = true }
+rayon = { workspace = true }
+serde = { workspace = true }
+solana-bls-signatures = { workspace = true, features = ["parallel"] }
+solana-logger = { workspace = true }
```
samkim-crypto: Is this crate used anywhere in the new crate?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,255 @@
...
+    bit_vec: &BitVec<u8, Lsb0>,
+    rank_to_pubkey: &F,
+) -> Option<PubkeyProjective>
+where
+    F: Fn(usize) -> Option<BlsPubkey> + Sync,
+{
+    let pubkeys: Vec<_> = bit_vec
+        .iter_ones()
+        .filter_map(rank_to_pubkey)
+        .filter_map(|pk| PubkeyProjective::try_from(pk).ok())
```
samkim-crypto: I think there is technically a difference in behavior compared to the original function. In the original [function](https://github.com/anza-xyz/alpenglow/pull/617/files#diff-44998bea37a652beea78de28073041a995c20d87f97e42c57fd7bd85466cbb39L55-L58), if at least one `PubkeyProjective::try_from(*bls_pubkey)` fails, then the entire batch fails. In this new logic, since we use a `filter_map`, the pubkeys that fails `try_from` will simply be ignored and not collected into the aggregate pubkey.

Is this intended behavior? At least for the case of sigverify, it seems like this is not what we want?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,255 @@
...
+    },
+    solana_signer_store::{decode, DecodeError},
+    solana_votor_messages::consensus_message::{Certificate, CertificateType},
+    thiserror::Error,
+};
+
+#[derive(Debug, Error, PartialEq)]
+pub enum CertVerifyError {
+    #[error("Failed to decode bitmap {0:?}")]
+    BitmapDecodingFailed(DecodeError),
```
samkim-crypto, line 21: What do you think about adding a `#[from]` here to eliminate some `map_err` calls?
```
BitmapDecodingFailed(#[from] DecodeError),
```

Then for something like 
```
let decoded_bitmap = decode(&cert_to_verify.bitmap, max_len)
    .map_err(CertVerifyError::BitmapDecodingFailed)?;
```

we can simply do

```
let decoded_bitmap = decode(&cert_to_verify.bitmap, max_len)?;
```

bls-cert-verify/src/cert_verify.rs
```diff
@@ -0,0 +1,255 @@
...
+    bit_vec: &BitVec<u8, Lsb0>,
+    rank_to_pubkey: &F,
+) -> Result<(), CertVerifyError>
+where
+    F: Fn(usize) -> Option<BlsPubkey> + Sync,
+{
+    let original_vote = cert_to_verify.cert_type.to_source_vote();
+
+    let signed_payload =
+        bincode::serialize(&original_vote).map_err(|_| CertVerifyError::SerializationFailed)?;
```
samkim-crypto: This is a style issue, but we can consider adding a dedicated function for serializing votes:

```
fn serialize_vote(vote: &Vote) -> Result<Vec<u8>, CertVerifyError> {
    // If using the improved error type from the previous suggestion:
    bincode::serialize(vote).map_err(|_| CertVerifyError::SerializationFailed)
}
```

Then we can just invoke cleanly
```
// verify_base2_certificate
let signed_payload = serialize_vote(&original_vote)?;
```

and 
```
// verify_base3_certificate
let signed_payload1 = serialize_vote(&vote1)?;
let signed_payload2 = serialize_vote(&vote2)?;
```

bls-cert-verify/src/cert_verify.rs
wen-coding: Kept only the last sentence because I think that's clearer.

bls-cert-verify/src/cert_verify.rs
wen-coding: Kept the last statement, applied others.

bls-cert-verify/Cargo.toml
wen-coding: Removed.

bls-cert-verify/src/cert_verify.rs
wen-coding: Hmm, that's true, also we probably shouldn't use .filter_map on rank_to_pubkey either, if the bitmap contains an error, we should just bail out, changed and added tests.

Thanks for catching that.

bls-cert-verify/src/cert_verify.rs
wen-coding: Tried that, doesn't compile currently:
```
   --> bls-cert-verify/src/cert_verify.rs:18:28
    |
18  |     BitmapDecodingFailed(#[from] DecodeError),
    |                            ^^^^ method cannot be called on `&DecodeError` due to unsatisfied trait bounds
    |
   ::: /Users/wenxu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/solana-signer-store-0.1.0/src/lib.rs:169:1
    |
169 | pub enum DecodeError {
    | -------------------- doesn't satisfy `DecodeError: AsDynError<'_>` or `DecodeError: std::error::Error`
    |
    = note: the following trait bounds were not satisfied:
            `DecodeError: std::error::Error`
            which is required by `DecodeError: AsDynError<'_>`
            `&DecodeError: std::error::Error`
            which is required by `&DecodeError: AsDynError<'_>`
```
This code is using thiserror, can you use thiserror as well?

bls-cert-verify/src/cert_verify.rs
wen-coding: Good point, done.

bls-cert-verify/src/cert_verify.rs
samkim-crypto: Ah sorry, it seems that `thiserror` crate's `#[from]` attribute automatically generates an implementation of the `From` trait and it expects `DecodeError` to implement `std::error::Error`, which `DecodeError` does not. Ignore my previous comment. Sorry for the noise 🙏

bls-cert-verify/src/cert_verify.rs
wen-coding: We will need to change the API significantly to add limit verification. So this PR is mostly about moving the logic around, we can maybe clean up tests in the next PR when the API has been determined and functionality is better split between BLS sigverifier and bls-cert-verify?

PR #618, opened 2025-11-30
fast leader handover: validator configs for local cluster tests

ksn6:
We need to test fast leader handover behavior in local cluster tests.

We expose a flag that, when active for a validator, votes skip for a slot when `slot % 4 == 3` after replay. This will allow us to write local cluster tests for the case in fast leader handover wherea parent switch occurs via `UpdateParent`.


AshwinSekar:
I'd rather not add this to production code and use an A2A spy in local-cluster tests instead.

PR #621, opened 2025-12-03
bank: Use double merkle as the block id for alpenglow blocks

AshwinSekar:
#### Problem
We currently use the chained merkle root as the block id for alpenglow blocks.

#### Summary of Changes
When the migration is complete, use the double merkle root as the block id instead:

Replay:
- After a successful freeze, compute double merkle from blockstore and set on bank
- Future perf improvement: can join nodes as new fec sets are inserted rather than waiting until freeze to do the computation

Startup replay:
- Fetch the double merkle directly from blockstore. Only in rare cases where replay was lagging on our prior run will we need to compute, minimal perf impact.

Broadcast:
- track fec set roots as we shred new batches
- When we finish a slot compute the double merkle , insert into blockstore, and set on bank
- Future perf improvement: if needed some joining can happen in the background while we're waiting for entries
- Future perf improvement: After computing the double merkle, the insertion of the DoubleMerkleMeta into blockstore can happen async

Once all the double merkle stuff is implemented i'll put together an issue summarizing future perf improvements that we can tackle after profiling.

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -3710,11 +3710,6 @@ impl ReplayStage {
                     continue;
                 }
 
-                let bank_progress = progress
-                    .get_mut(&bank.slot())
-                    .expect("Bank fork progress entry missing for completed bank");
-                let replay_stats = bank_progress.replay_stats.clone();
-
                 let r_replay_stats = replay_stats.read().unwrap();
                 let replay_progress = bank_progress.replay_progress.clone();
                 let r_replay_progress = replay_progress.read().unwrap();


votor-messages/src/migration.rs
```diff
@@ -248,6 +248,12 @@ impl MigrationPhase {
         self.should_send_votor_event(slot)
     }
 
+    /// Should this block use the double merkle root as the block id (instead of chained merkle root)?
+    fn should_use_double_merkle_block_id(&self, slot: Slot) -> bool {
+        // Same as votor events, all alpenglow blocks should use double merkle
+        self.should_send_votor_event(slot)
```
bw-solana: poor `fn should_send_votor_event` carrying the weight of the world on his shoulders 😆 

doesn't have to gate this PR, but might be good to create something like this:
```rust
    fn is_alpenglow_block(&self, slot: Slot) -> bool {
        match self {
            MigrationPhase::PreFeatureActivation
            | MigrationPhase::Migration { .. }
            | MigrationPhase::ReadyToEnable { .. } => false,
            MigrationPhase::AlpenglowEnabled { genesis_cert } => {
                slot > genesis_cert.cert_type.slot()
            }
            MigrationPhase::FullAlpenglowEpoch { .. } => true,
        }
    }
```

and then we can reference `is_alpenglow_block` in all these other functions

turbine/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -76,6 +85,57 @@ impl StandardBroadcastRun {
         }
     }
 
+    /// Upon recieval of shreds from a new bank (bank.slot() != self.slot)
```
bw-solana: receipt*

turbine/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -25,7 +30,9 @@ use {
 pub struct StandardBroadcastRun {
     slot: Slot,
     parent: Slot,
+    parent_block_id: Hash,
     chained_merkle_root: Hash,
+    double_merkle_leaves: Vec<Hash>,
```
bw-solana, line 35: Are we going to have a special type for this (not Hash)? I seem to recall a PR from Akhi with a Slice Hash or something

core/src/replay_stage.rs
```diff
@@ -3639,15 +3640,27 @@ impl ReplayStage {
                     }
                 }
 
+                let is_leader_block = bank.collector_id() == my_pubkey;
```
bw-solana, line 3688: seems like this whole thing (3643 - 3691) could be hidden behind some `determine_and_set_block_id` function

runtime/src/bank.rs
```diff
@@ -5660,7 +5660,12 @@ impl Bank {
     }
 
     pub fn set_block_id(&self, block_id: Option<Hash>) {
-        *self.block_id.write().unwrap() = block_id;
+        let mut block_id_w = self.block_id.write().unwrap();
+        if block_id_w.is_some() {
+            debug_assert_eq!(*block_id_w, block_id);
+            return;
+        }
+        *block_id_w = block_id;
```
bw-solana: could just if/else to avoid early return

turbine/src/broadcast_stage/standard_broadcast_run.rs
AshwinSekar: Akhi is exploring changing some of the types to `FecSetRoot` on agave master. it hasn't landed yet, i've added him to this PR to be aware of how we're using the merkle-tree impl so that he can better design the api refactor.

runtime/src/bank.rs
AshwinSekar: yeah good call https://github.com/anza-xyz/alpenglow/pull/621/commits/fd9a7d4a569da8f036aa7558ef2a50da59edb048

core/src/replay_stage.rs
AshwinSekar: yeah fair, factored it out here https://github.com/anza-xyz/alpenglow/pull/621/commits/dc7153e8166a711729dad845a45accc06d7fe49f

votor-messages/src/migration.rs
AshwinSekar: saved his back here https://github.com/anza-xyz/alpenglow/pull/621/commits/1d6c419a8dbacfaac7b034ffcd0b8e90c4f8ef27

core/src/replay_stage.rs
```diff
@@ -3534,6 +3535,51 @@ impl ReplayStage {
...
+            // Unlike CMR there are no checks performed in `get_double_merkle_root`.
+            debug_assert!(block_id.is_some() || is_leader_block);
+            Ok(block_id)
+        } else {
+            // Once SIMD-0317 is activated, this can simply grab the merkle root of the last shred.
+            // Then there is no need for the error checking or to potentially mark the slot as dead below.
+            blockstore.check_last_fec_set_and_get_block_id(
+                bank.slot(),
+                bank.hash(),
+                is_leader_block,
```
AshwinSekar, line 3558: @ksn6 found that we were previously passing in `false` here always 🤦 
This results in warning logs being printed for our leader blocks (block is not actually set as dead because we filter on `is_leader_block` below)  
```
Unable to check the last fec set for slot 16 11111111111111111111111111111111, marking as dead: Err(UnknownLastIndex(16))
```

Since we're modifying stuff here, I bundled the fix into this PR https://github.com/anza-xyz/alpenglow/pull/621/commits/9207c58d3ed795c2c145a877b322ca99d3ddf5d0

votor-messages/src/migration.rs
bw-solana: my hero

bw-solana (approved):

PR #622, opened 2025-12-04
fast leader handover: use `ParentMeta` to update `SlotMeta` `next_slots`

ksn6:
As `UpdateParent` and `BlockHeader` components filter in, we need to update:

- parent information in `SlotMeta`
- the `next_slots` field for `SlotMeta` to enable `generate_new_bank_forks` to work in replay

This is important to allow non-leaders to process `UpdateParent`s during a parent switch in fast leader handover.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -1687,7 +1687,6 @@ impl Blockstore {
                 metrics,
             );
         }
-
         // Handle chaining for the members of the slot_meta_working_set that
         // were inserted into, drop the others.
         self.handle_chaining(
@@ -5408,7 +5407,7 @@ impl Blockstore {
     /// `BlockLocation::Original`
     fn update_chaining_for_parent_metas(
         &self,
-        working_set: &mut HashMap<(BlockLocation, u64), SlotMetaWorkingSetEntry>,
+        working_set: &HashMap<(BlockLocation, u64), SlotMetaWorkingSetEntry>,
         parent_metas: &HashMap<(BlockLocation, u64), ParentMetaWorkingSetEntry>,
         new_chained_slots: &mut HashMap<u64, Rc<RefCell<SlotMeta>>>,
     ) -> Result<()> {
@@ -5427,21 +5426,12 @@ impl Blockstore {
         });
 
         for (slot, new_parent_slot) in update_parent_entries {
-            // Get old parent from slot meta; update parent_slot to new value.
-            // Check new_slot_meta first (handles same-batch case where BlockHeader
-            // already set parent), then fall back to old_slot_meta (from database).
-            let old_parent_slot =
-                if let Some(entry) = working_set.get_mut(&(BlockLocation::Original, slot)) {
-                    let old = entry
-                        .new_slot_meta
-                        .borrow()
-                        .parent_slot
-                        .or_else(|| entry.old_slot_meta.as_ref().and_then(|m| m.parent_slot));
-                    entry.new_slot_meta.borrow_mut().parent_slot = Some(new_parent_slot);
-                    old
-                } else {
-                    None
-                };
+            // Get or create slot meta. When UpdateParent is inserted separately from
+            // BlockHeader, the slot may not be in working_set (filtered by did_insert_occur).
+            let slot_meta =
+                self.find_slot_meta_else_create(working_set, new_chained_slots, slot)?;
+            let old_parent_slot = slot_meta.borrow().parent_slot;
+            slot_meta.borrow_mut().parent_slot = Some(new_parent_slot);
 
             // Remove slot from old parent's next_slots if parent changed
             if let Some(old_parent) = old_parent_slot.filter(|&old| old != new_parent_slot) {
@@ -5454,9 +5444,31 @@ impl Blockstore {
             // Add slot to new parent's next_slots
             let new_parent_meta =
                 self.find_slot_meta_else_create(working_set, new_chained_slots, new_parent_slot)?;
-            let mut new_parent = new_parent_meta.borrow_mut();
-            if !new_parent.next_slots.contains(&slot) {
-                new_parent.next_slots.push(slot);
+            {
+                let mut new_parent = new_parent_meta.borrow_mut();
+                if !new_parent.next_slots.contains(&slot) {
+                    new_parent.next_slots.push(slot);
+                }
+            }
+
+            // Propagate or clear connectivity based on new parent's state.
+            if new_parent_meta.borrow().is_connected() {
+                if !slot_meta.borrow().is_parent_connected() {
+                    slot_meta.borrow_mut().set_parent_connected();
+                    self.traverse_children_mut(
+                        &slot_meta,
+                        working_set,
+                        new_chained_slots,
+                        SlotMeta::set_parent_connected,
+                    )?;
+                }
+            } else if slot_meta.borrow_mut().clear_parent_connected() {
+                self.traverse_children_mut(
+                    &slot_meta,
+                    working_set,
+                    new_chained_slots,
+                    SlotMeta::clear_parent_connected,
+                )?;
             }
         }
 
@@ -13127,6 +13139,7 @@ pub mod tests {
         parent_slot: Slot,
         parent_block_id: Hash,
         shred_index: u32,
+        is_last_in_slot: bool,
     ) -> Vec<Shred> {
         let component = UpdateParentV1 {
             new_parent_slot: parent_slot,
@@ -13142,7 +13155,7 @@ pub mod tests {
             .make_merkle_shreds_from_component(
                 &Keypair::new(),
                 &component,
-                true,
+                is_last_in_slot,
                 Some(Hash::new_unique()),
                 shred_index,
                 shred_index,
@@ -13218,7 +13231,7 @@ pub mod tests {
         let parent_3_id = Hash::new_unique();
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(10, 3, parent_3_id, 32),
+                create_update_parent_shreds(10, 3, parent_3_id, 32, true),
                 None,
                 true,
             )
@@ -13256,7 +13269,7 @@ pub mod tests {
 
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(20, 15, Hash::new_unique(), 32),
+                create_update_parent_shreds(20, 15, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
@@ -13283,7 +13296,7 @@ pub mod tests {
 
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(30, 22, Hash::new_unique(), 32),
+                create_update_parent_shreds(30, 22, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
@@ -13309,7 +13322,7 @@ pub mod tests {
 
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(41, 32, Hash::new_unique(), 32),
+                create_update_parent_shreds(41, 32, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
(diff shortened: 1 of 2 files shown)


ledger/src/blockstore.rs
```diff
@@ -2483,6 +2488,20 @@ impl Blockstore {
             if !Self::should_write_parent_meta(slot, &new_parent_meta, prev)? {
                 return Ok(());
             }
+
+            if prev.populated_from_block_header() && new_parent_meta.populated_from_update_parent()
```
carllin: is this a redundant check from the should_write_parent_meta() check above https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L2400

ledger/src/blockstore.rs
```diff
@@ -1667,6 +1667,9 @@ impl Blockstore {
                 metrics,
             );
         }
+
+        self.synchronize_slot_meta_with_parent_meta(&mut shred_insertion_tracker);
```
carllin: any reason not to wrap this alongside `maybe_update_parent_meta()` in a single function instead?  That function is already handling making the `parent_slot_meta` consistent: https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L2496-L2500

Would be logically nice to have all the weeds of the parent update stuff in one place

ledger/src/blockstore.rs
```diff
@@ -5209,6 +5228,44 @@ impl Blockstore {
...
+        for (location, slot, parent_slot, populated_from_update_parent) in dirty_parent_metas {
+            let slot_meta_wse = self.get_slot_meta_entry(
+                &mut shred_insertion_tracker.slot_meta_working_set,
+                slot,
+                location,
+                parent_slot,
+            );
+
+            // When an UpdateParent shows up, set old_slot_meta to None to force handle_chaining to
+            // run.
```
carllin: chaining only happens when the BlockLocation == original

@AshwinSekar is it possible to have a parent update for a duplicate?

ledger/src/blockstore.rs
```diff
@@ -5209,6 +5228,44 @@ impl Blockstore {
...
+            if populated_from_update_parent {
+                slot_meta_wse.old_slot_meta = None;
```
carllin: My impression is the special-case chaining logic for handling these parent updates is too fragmented, see: https://github.com/anza-xyz/alpenglow/pull/622/files#r2594357652

I think what should happen is:

1. `maybe_update_parent_meta()` I think should just simply update the ParentMeta so that downstream logic knows a ParentUpdate occurred: https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L2449C8-L2449C32. It should avoid handling chaining logic like https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L2494-L2502
2. All chaining logic should be handled in `handle_chaining()`, which I think should be passed an additional argument which is all the dirty ParentMeta's. This makes it clear to readers that the only purpose for the dirty ParentMeta's is to update chaining.
3. `handle_chaining()` can then call either the existing `handle_chaining_for_slot()` for regular orphan slots: https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L5355 or a new `update_chaining_for_slot()` to handle the ParentMeta's. We can factor out some of the logic for `handle_chaining_for_slot()` so that the logic is shared with `update_chaining_for_slot()`

ledger/src/blockstore.rs
ksn6: > chaining only happens when the BlockLocation == original

agreed

> is it possible to have a parent update for a duplicate?

yes, that's possible - iirc, @AshwinSekar stated offline that he'd be handling the duplicate blocks case; can help with this as well - lmk

ledger/src/blockstore.rs
ksn6: nice; this makes things cleaner - let me know what you think

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5401,62 @@ impl Blockstore {
...
+                    None
+                };
+
+            // Remove slot from old parent's next_slots if parent changed
+            if let Some(old_parent) = old_parent_slot.filter(|&old| old != new_parent_slot) {
+                self.find_slot_meta_else_create(working_set, new_chained_slots, old_parent)?
+                    .borrow_mut()
+                    .next_slots
+                    .retain(|&s| s != slot);
+            }
```
carllin: maybe could condense all this logic with somethiing like

```
working_set
    .get_mut(&(BlockLocation::Original, slot))
    .and_then(|entry| {
        entry.new_slot_meta.borrow_mut().parent_slot = Some(new_parent_slot);
        entry.old_slot_meta.as_ref().and_then(|m| m.parent_slot)
    })
    .filter(|&old| old != new_parent_slot)
    .and_then(|old_parent| {
        self.find_slot_meta_else_create(working_set, new_chained_slots, old_parent).ok()
    })
    .map(|meta| meta.borrow_mut().next_slots.retain(|&s| s != slot));
```

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5401,62 @@ impl Blockstore {
...
+            // Add slot to new parent's next_slots
+            let new_parent_meta =
+                self.find_slot_meta_else_create(working_set, new_chained_slots, new_parent_slot)?;
+            let mut new_parent = new_parent_meta.borrow_mut();
+            if !new_parent.next_slots.contains(&slot) {
+                new_parent.next_slots.push(slot);
+            }
+        }
+
+        Ok(())
```
carllin, line 5475: Noting here that this will not propagate the is_connected flag: https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L5413-L5425

Now there's a question of whether this is necessary, I think the `is_connected` logic is outdated and nobody really uses it at this point, i.e. it bubbles up to here https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L1693, but `self.new_shreds_signals` is generally empty


If we want to preserve this logic we would have to ensure this logic runs to handle the case where either the slot is:

1. newly completed
2. switched to a new parent that isn't connected
3. switched from an old parent that wasn't connected but the new parent is connected

tagging @AshwinSekar if I'm missing anything

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5401,62 @@ impl Blockstore {
...
+                }
+                _ => None,
+            }
+        });
+
+        for (slot, new_parent_slot) in update_parent_entries {
+            // Get old parent from slot meta; update parent_slot to new value
+            let old_parent_slot =
+                if let Some(entry) = working_set.get_mut(&(BlockLocation::Original, slot)) {
+                    let old = entry.old_slot_meta.as_ref().and_then(|m| m.parent_slot);
```
carllin: do we have a test where there is an UpdateParent and a BlockHeader parent in the same shred insertion batch, testing both orders where

1. BlockHeader precedes UpdateParent
2. UpdateParent precedes BlockHeader

And then checks all the chaining is correct?

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5401,62 @@ impl Blockstore {
...
+                    .next_slots
+                    .retain(|&s| s != slot);
+            }
+
+            // Add slot to new parent's next_slots
+            let new_parent_meta =
+                self.find_slot_meta_else_create(working_set, new_chained_slots, new_parent_slot)?;
+            let mut new_parent = new_parent_meta.borrow_mut();
+            if !new_parent.next_slots.contains(&slot) {
+                new_parent.next_slots.push(slot);
```
carllin: forgot how painful this logic is haha

this new_parent might not be in the chained cache, so it won't be committed here https://github.com/ksn6/alpenglow/blob/59a53d6cf818f6cc774713668f12bef4925e4e49/ledger/src/blockstore.rs#L5314-L5316

But it will be committed later by the dirty SlotMeta working set?

ledger/src/blockstore.rs
ksn6: had something like this, though - I slightly prefer the existing code for readability

not strongly opinionated on this though

ledger/src/blockstore.rs
ksn6: good catch

ledger/src/blockstore.rs
ksn6: yep, that's right

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5400,81 @@ impl Blockstore {
         Ok(())
     }
 
+    /// Handles chaining updates when an UpdateParent marker overrides a previously
+    /// set parent from a BlockHeader.
+    ///
+    /// TODO(ashwin, ksn): this chaining only occurs for `SlotMeta`s in the column associated with
```
AshwinSekar, line 5406: after our discussion during BP I think this is fine.
We don't need the ability to generate new bank forks out of any column apart from Original

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5400,81 @@ impl Blockstore {
...
+
+            // Remove slot from old parent's next_slots if parent changed
+            if let Some(old_parent) = old_parent_slot.filter(|&old| old != new_parent_slot) {
+                self.find_slot_meta_else_create(working_set, new_chained_slots, old_parent)?
+                    .borrow_mut()
+                    .next_slots
+                    .retain(|&s| s != slot);
+            }
+
+            // Add slot to new parent's next_slots
```
AshwinSekar, line 5444: Maybe a comment somewhere in the function to require this is only used after `working_set` has been filtered so that all members have `did_insert_occur = true`. 

Otherwise we could end up updating the slot meta here from `working_set` with `did_insert_occur = false` and skip committing it to the DB.

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5400,81 @@ impl Blockstore {
...
+                if !new_parent.next_slots.contains(&slot) {
+                    new_parent.next_slots.push(slot);
+                }
+            }
+
+            // Propagate or clear connectivity based on new parent's state.
+            if new_parent_meta.borrow().is_connected() {
+                if !slot_meta.borrow().is_parent_connected() {
+                    slot_meta.borrow_mut().set_parent_connected();
+                    self.traverse_children_mut(
```
carllin, line 5458: hmm you only want to traverse the children if this slot_meta is complete right? https://github.com/ksn6/alpenglow/blob/4ae52765259cbb8bc22f01101c63eef59f5622c5/ledger/src/blockstore.rs#L5387

ledger/src/blockstore.rs
```diff
@@ -5390,6 +5400,81 @@ impl Blockstore {
...
+                self.traverse_children_mut(
+                    &slot_meta,
+                    working_set,
+                    new_chained_slots,
+                    SlotMeta::clear_parent_connected,
+                )?;
+            }
+        }
+
+        Ok(())
```
carllin, line 5475: I think this logic would be much cleaner if we could clear the necessary/unnecessary chaining flags, and just call a function that wraps the same logic as here: https://github.com/ksn6/alpenglow/blob/4ae52765259cbb8bc22f01101c63eef59f5622c5/ledger/src/blockstore.rs#L5386-L5398

That way the chaning logic is the same in both these places

ledger/src/blockstore.rs
ksn6: gotcha - didn't realize this; will fix the test at the end as well

PR #623, opened 2025-12-04
feat: SIMD-0402 - add final certs to block footer

ksn6:

This PR updates the block footer with finalization certificate information as described in https://github.com/solana-foundation/solana-improvement-documents/pull/402. We adapt the implementation from https://github.com/anza-xyz/alpenglow/pull/603 to use `wincode`.

We populate the finalization certificate with `None` in `block_creation_loop.rs` for now. A future PR will populate this value correctly.

### DIFF
--- a/runtime/src/block_component_processor.rs
+++ b/runtime/src/block_component_processor.rs
@@ -627,7 +627,7 @@ mod tests {
             bank_hash: Hash::new_unique(),
             block_producer_time_nanos: 1_000_000_000,
             block_user_agent: vec![],
-            final_cert: Some(FinalCertificate::new_for_tests()),
+            final_cert: None,
         });
 
         // Try to process footer without header - should fail


runtime/src/block_component_processor.rs
```diff
@@ -622,6 +627,7 @@ mod tests {
             bank_hash: Hash::new_unique(),
             block_producer_time_nanos: 1_000_000_000,
             block_user_agent: vec![],
+            final_cert: Some(FinalCertificate::new_for_tests()),
```
wen-coding: nit: can we make `final_cert` None in one of the tests?

entry/src/block_component.rs
```diff
@@ -128,6 +160,28 @@ impl SeqLen for U8Len {
     }
 }
 
+/// 2-byte length prefix (max 65535 elements).
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct U16Len;
```
wen-coding, line 165: nit: can we somehow merge U16Len and U8Len maybe?

ksn6: we could, though that'd require introducing a new trait / `PhantomData` / etc. - would rather keep this simple

can explore this if we needed to add more of these in the future, though arguably at that point, we should throw these types directly into wincode

PR #624, opened 2025-12-05
repair: remove CMR block_id requests

AshwinSekar:
#### Problem
We added some block id repair requests based on the assumption that we were using CMR.
With DMR these requests are not necessary.

#### Summary of Changes
Remove them except for shred for block id.
Next PR will add getFecSetCountWithParent & getFecSetRoot instead to match what we have in the paper.

### DIFF
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -26,6 +26,7 @@ use {
         distributions::{Distribution, WeightedError, WeightedIndex},
         Rng,
     },
+    shred::wire::get_merkle_root,
     solana_clock::Slot,
     solana_cluster_type::ClusterType,
     solana_gossip::{
@@ -171,7 +172,7 @@ impl RequestResponse for ShredRepairType {
             } => {
                 shred_slot == *slot
                     && get_shred_index(shred) == Some(*index)
-                    && shred::wire::get_merkle_root(shred) == Some(*fec_set_merkle_root)
+                    && get_merkle_root(shred) == Some(*fec_set_merkle_root)
             }
         }
     }


core/src/repair/serve_repair.rs
```diff
@@ -108,14 +108,14 @@ pub enum ShredRepairType {
     /// Requesting the missing shred at a particular index
     Shred(Slot, u64),
 
-    // Block Id based variants. Note that unlike the above requests, these require that the responder has
-    // the full block in order to respond.
-    /// Requesting `MAX_ORPHAN_REPAIR_RESPONSES` parent shreds for the specific block ID
-    OrphanForBlockId(Slot, Hash),
```
AshwinSekar, line 114: will be replacing these with the ones in the paper 

Image

core/src/repair/standard_repair_handler.rs
```diff
@@ -55,112 +51,4 @@ impl RepairHandler for StandardRepairHandler {
...
-
-        for packet in packets.take(max_responses) {
-            res.push(packet);
-        }
-        (!res.is_empty()).then_some(res.into())
-    }
-
-    /// Fulfills orphan requests from the turbine column only. This predates block id logic and is left for compatability
-    /// with TowerBFT. This variant will respond to requests even if the block has not been fully ingested.
-    fn run_orphan_turbine_only(
```
AshwinSekar, line 138: this is copied back into `run_orphan` in repair_handler.rs

core/src/repair/serve_repair.rs
```diff
@@ -163,20 +156,22 @@ impl RequestResponse for ShredRepairType {
...
-                shred_slot == *slot && get_shred_index(shred) == Some(*index)
+            ShredRepairType::ShredForBlockId {
+                slot,
+                index,
+                fec_set_merkle_root,
+                ..
+            } => {
+                shred_slot == *slot
+                    && get_shred_index(shred) == Some(*index)
+                    && shred::wire::get_merkle_root(shred) == Some(*fec_set_merkle_root)
```
bw-solana: should probably import `get_merkle_root`

core/src/repair/serve_repair.rs
```diff
@@ -274,7 +269,7 @@ type PingCache = ping_pong::PingCache<REPAIR_PING_TOKEN_SIZE>;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiEnumVisitor, AbiExample),
-    frozen_abi(digest = "4zhY22Ens7hwEGMfq6wLyb9viynpcUYrSUCYnpgAS9rC")
+    frozen_abi(digest = "H4x6ESrXXSWgh9M5Kw3ryDww2okbEUW36V9BJUVZU17V")
```
bw-solana, line 273: I'm assuming this is fine given the added/removed enum variants were never exposed on public cluster

AshwinSekar: yeah exactly

PR #625, opened 2025-12-09
Add cert limit verification.

wen-coding:
#### Problem
We haven't added cert limit verification so far.

#### Summary of Changes
Add optional cert limit the verification logic needs to verify.

### DIFF
--- a/bls-cert-verify/src/cert_verify.rs
+++ b/bls-cert-verify/src/cert_verify.rs
@@ -46,8 +46,6 @@ where
     // It only returns Some(aggregated_pubkey) if all pubkeys are found and valid.
     let results: Option<Vec<_>> = bit_vec
         .iter_ones()
-        .collect::<Vec<_>>()
-        .into_par_iter()
         .map(|rank| {
             let (pubkey, stake) = rank_to_pubkey_and_stake(rank)?;
             let pk_proj = PubkeyProjective::try_from(&pubkey).ok()?;


bls-cert-verify/src/cert_verify.rs
```diff
@@ -31,26 +31,40 @@ pub enum CertVerifyError {
...
+            let pk_proj = PubkeyProjective::try_from(&pubkey).ok()?;
+            Some((pk_proj, stake))
         })
-        .collect::<Option<Vec<_>>>()?;
-    PubkeyProjective::par_aggregate(pubkeys.par_iter()).ok()
+        .collect();
+
+    let results = results?; // fail-fast
+
+    let total_stake = results.iter().map(|(_, stake)| *stake).sum::<u64>();
```
ksn6, line 58:
```suggestion
    let total_stake = results?.iter().map(|(_, stake)| *stake).sum::<u64>();
```

bls-cert-verify/src/cert_verify.rs
```diff
@@ -99,10 +122,20 @@ where
...
-        .ok_or(CertVerifyError::KeyAggregationFailed)?;
+    let (agg_pk1, stake1) =
+        aggregate_keys_and_stakes_from_bitmap(bit_vec1, rank_to_pubkey_and_stake)
+            .ok_or(CertVerifyError::KeyAggregationFailed)?;
+    let (agg_pk2, stake2) =
+        aggregate_keys_and_stakes_from_bitmap(bit_vec2, rank_to_pubkey_and_stake)
+            .ok_or(CertVerifyError::KeyAggregationFailed)?;
+
+    if let Some(required_stake) = required_stake {
+        let total_stake = stake1.saturating_add(stake2);
```
ksn6: `checked_add` + `expect`

bls-cert-verify/src/cert_verify.rs
```diff
@@ -99,10 +122,20 @@ where
...
-    let agg_pk1 = aggregate_keys_from_bitmap(bit_vec1, rank_to_pubkey)
-        .ok_or(CertVerifyError::KeyAggregationFailed)?;
-    let agg_pk2 = aggregate_keys_from_bitmap(bit_vec2, rank_to_pubkey)
-        .ok_or(CertVerifyError::KeyAggregationFailed)?;
+    let (agg_pk1, stake1) =
+        aggregate_keys_and_stakes_from_bitmap(bit_vec1, rank_to_pubkey_and_stake)
+            .ok_or(CertVerifyError::KeyAggregationFailed)?;
+    let (agg_pk2, stake2) =
+        aggregate_keys_and_stakes_from_bitmap(bit_vec2, rank_to_pubkey_and_stake)
+            .ok_or(CertVerifyError::KeyAggregationFailed)?;
```
ksn6, line 117: Does `bit_vec1` always correspond to `notar` votes / `bit_vec2` always correspond to `notar_fallback` votes?

If so, rather than using `bit_vec1` and `bit_vec2`, we should probably use more descriptive names here.

bls-cert-verify/src/cert_verify.rs
ksn6: Also, let's toss in a comment indicating that the validators represented by the bit vecs must be mutually exclusive (because of separation by vote type, e.g.), which is why this addition is okay to do.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -485,15 +497,23 @@ impl BLSSigVerifier {
...
         };
 
-        verify_votor_message_certificate(cert_to_verify, key_to_rank_map.len(), |rank| {
-            key_to_rank_map
-                .get_pubkey(rank)
-                .map(|(_, bls_pubkey)| *bls_pubkey)
-        })?;
+        let (required_stake_percentage, _) =
+            certificate_limits_and_vote_types(&cert_to_verify.cert_type);
+        let required_stake = (required_stake_percentage * total_stake as f64) as u64;
```
ksn6: This isn't correct - `total_stake` ends up losing precision here because of the cast to `u64` -> `f64`, and we end up rounding down rather than rounding up in casting to `u64`.

If we're attempting to verify that `x * 0.6 > y`, e.g., we should instead check that `x * 3 > 5 * y` (may need to use `u128` temps to avoid overflow).

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -1201,6 +1259,60 @@ mod tests {
...
+            .expect("Failed to aggregate votes");
+        let cert = builder.build().expect("Failed to build certificate");
+        let consensus_message = ConsensusMessage::Certificate(cert);
+        let packet_batches = messages_to_batches(&[consensus_message]);
+
+        assert!(verifier.verify_and_send_batches(packet_batches).is_ok());
+        assert_eq!(
+            message_receiver.try_iter().count(),
+            0,
+            "This certificsate should be invalid"
```
ksn6: nit: certificate

bls-cert-verify/src/cert_verify.rs
```diff
@@ -118,21 +151,29 @@ where
 pub fn verify_votor_message_certificate<F>(
     cert_to_verify: &Certificate,
     max_len: usize,
-    rank_to_pubkey: F,
+    required_stake: Option<u64>,
```
ksn6: Why is `required_stake` an `Option<u64>` rather than a `u64`? Looks like this is `Some(...)` everywhere atm.

bls-cert-verify/src/cert_verify.rs
wen-coding: Hmm, I do want to emphasize we are doing fail-fast here.

bls-cert-verify/src/cert_verify.rs
wen-coding: Nope, we could have skip and skip_fallback as well.

bls-cert-verify/src/cert_verify.rs
wen-coding: For rewards it should be None.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Theoretically yes, but we only get a float from `certificate_limits_and_vote_types`, and I don't think our stake would be large enough for the loss of precision to matter.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Done.

bls-cert-verify/src/cert_verify.rs
wen-coding: Done.

bls-cert-verify/src/cert_verify.rs
ksn6: Similar to the rest of the function, where we fail fast as well, isn't this already accomplished when we do `?`:

```rust
let total_stake = results?.iter().map(|(_, stake)| *stake).sum::<u64>();
```

This above code is more idiomatic than doing `let results = results?;` and re-using `results`.

bls-cert-verify/src/cert_verify.rs
ksn6: Is it just these two cases? If so, calling these `vote_bit_vec` and `vote_fallback_bit_vec` would be a bit more descriptive. It would also somewhat self-document the part re. mutual exclusivity of validators.

Right now, it's difficult to intuit exactly what `bit_vec1` and `bit_vec2` are.

ksn6: Also - do we care at all about the case where a malicious node votes incorrectly? E.g., votes both `skip` and `skip_fallback`?

core/src/bls_sigverify/bls_sigverifier.rs
ksn6: Disagreed - this issue isn't theoretical. `f64` stores ~15-16 significant digits, while the top validators currently have ~1.3 * 10^16 lamports worth of stake. This means that `f64` is on the cusp of barely good enough to work with the current precision. This is just with a _single_ top validator; if we start adding together substantial stake as we're doing here, the issue only gets worse.

In general, we should avoid doing financial math with `f64` and stick with ints. Otherwise, we'll end up with hard-to-find bugs in the future. E.g., this is a frequent source bug in Solana programs where accounts end up with dust rather than clean 0's. In our case, `f64`s blur the lines between `<=` and `<` in consensus.

It turns out that avoiding `f64` in `votor` is doable with essentially little-to-no added complexity: https://github.com/anza-xyz/alpenglow/pull/629.

Could we grab the updated `certificate_limits_and_vote_types` from #629 here + update this logic?

bls-cert-verify/src/cert_verify.rs
ksn6: If that's the case, could we make this `u64` instead and then turn this into an `Option` in the rewards PR so there's a bit more context?

bls-cert-verify/src/cert_verify.rs
wen-coding: The cert doesn't really care which one is fallback, it just packs two bitmaps. In the future we can decide to pack notarize and skip together in a cert, that's totally okay. I would rather not limit ourselves here.

Vote both `skip` and `skip_fallback` is detected in consensus pool and literally impossible to encode in base-3 encoding.

bls-cert-verify/src/cert_verify.rs
wen-coding: I believe @akhi3030 is making rewards work now, so why do we change this from Option to u64 and very soon change it back?

bls-cert-verify/src/cert_verify.rs
wen-coding: Both versions are correct and fail fast — I don’t think there’s a functional difference here.

I personally prefer the two-step version because I find it a bit clearer to read, but this is stylistic.
If you strongly prefer the inline `?`, I’m okay with that.

ksn6: I do prefer the `results?` approach inline rather than having a separate line, as the inline approach is more idiomatic Rust.

bls-cert-verify/src/cert_verify.rs
ksn6: For clarity, we shouldn't be splitting things like this across unrelated PRs - it's far more clear if we explicitly make this a `u64` here and then in the rewards PR, it's changed to an `Option`.

wen-coding: @akhi3030 what do you prefer?

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Done.

bls-cert-verify/src/cert_verify.rs
wen-coding: Wait, this doesn't work, because I'm reusing result on next line:
`let pubkeys: Vec<_> = results.into_iter().map(|(pk, _)| pk).collect();`

bls-cert-verify/src/cert_verify.rs
akhi3030: https://github.com/anza-xyz/alpenglow/pull/607/changes#diff-4233078a56eaab34364bc77843352b2bf9c4cac5606081c6de4f94c0313bc018R42.

I am going to be implementing my own verification because the reward certs are not a `Certificate`.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -31,26 +33,40 @@ pub enum CertVerifyError {
...
 {
     // This function should return None if any of the following happens:
     // - A rank in the bitmap does not have a corresponding pubkey.
     // - A pubkey fails to convert to projective form.
     // It only returns Some(aggregated_pubkey) if all pubkeys are found and valid.
-    let pubkeys: Vec<PubkeyProjective> = bit_vec
+    let results: Option<Vec<_>> = bit_vec
         .iter_ones()
+        .collect::<Vec<_>>()
+        .into_par_iter()
```
AshwinSekar: Was there a reason to switch here? is it faster for our use case?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -60,17 +76,31 @@ fn serialize_vote(vote: &Vote) -> Result<Vec<u8>, CertVerifyError> {
 pub fn verify_base2_certificate<F>(
     cert_to_verify: &Certificate,
     bit_vec: &BitVec<u8, Lsb0>,
-    rank_to_pubkey: &F,
+    required_stake: Option<(u64, Fraction)>,
```
AshwinSekar: would it be cleaner if instead we return the `aggregate_stake` and let the caller do any comparisons they need?

keeps `bls-cert-verify` more general purpose for other use case

bls-cert-verify/src/cert_verify.rs
wen-coding: What do you mean? This function is rewritten because now we need to collect pubkey and stake from the bitmaps.

bls-cert-verify/src/cert_verify.rs
wen-coding: Conceptually it's a bit cleaner to verify inside cert_verify, so you verify signature and stake in one step. But I think verifying in the caller also works. Changed.

bls-cert-verify/src/cert_verify.rs
wen-coding: We don't have required_stake argument any more, it's in caller now.

bls-cert-verify/src/cert_verify.rs
ksn6: Similar question here - why are we invoking `.into_par_iter()` here when we weren't doing so before?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -1154,6 +1180,43 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_verify_certificate_base2_not_enough_stake() {
```
ksn6, line 1213: Would be useful to add a test that verifies boundary behavior (the at-threshold stake with perfect equality works as expected).

ksn6 (commented):
LGTM after addressing the outstanding comments

bls-cert-verify/src/cert_verify.rs
wen-coding: Oh that's what you are talking about, yeah I was thinking .into_par_iter() can parallelize collection as well, because we may have 1000+ bits in bitmap. That said, the performance difference is probably small, so I'm fine with reverting this parallelization as well.

core/src/bls_sigverify/bls_sigverifier.rs
wen-coding: Done.

bls-cert-verify/src/cert_verify.rs
wen-coding: Thinking about it again, if we send out the certs right after the threshold is met, it's very likely to only have top 200+ validators, so the performance gain is not obvious. Reverted to serial aggregation.

PR #626, opened 2025-12-09
fast leader handover: migration interactions

ksn6:
We need to enable fast leader handover some short period after Alpenglow migration.

This PR prevents optimistic parents from being sent via replay until said short period has transpired. This prevention forces leaders to have to wait for parent ready events in consensus.

### DIFF
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -45,7 +45,7 @@
 use {
     crate::consensus_message::{Block, Certificate, CertificateType},
     log::*,
-    solana_clock::{Epoch, Slot, NUM_CONSECUTIVE_LEADER_SLOTS},
+    solana_clock::{Epoch, Slot},
     solana_epoch_schedule::EpochSchedule,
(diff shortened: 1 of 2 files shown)


votor-messages/src/migration.rs
```diff
@@ -253,6 +253,12 @@ impl MigrationPhase {
         self.is_alpenglow_block(slot)
     }
 
+    /// Should this block allow the UpdateParent marker, i.e., support fast leader handover?
+    fn should_allow_fast_leader_handover(&self, slot: Slot) -> bool {
+        // Only allow for alpenglow blocks, 2 leader windows after after migration
+        self.is_alpenglow_block(slot.saturating_sub(2 * NUM_CONSECUTIVE_LEADER_SLOTS))
```
AshwinSekar: why can't we enable it immediately, what breaks? 
```suggestion
        self.is_alpenglow_block(slot)
```

ksn6: ran tests locally - things seem fine if we roll with immediately doing fast leader handover; will update this

PR #629, opened 2025-12-10
votor: avoid floats in favor of ints

ksn6:
In `votor`, we have logic that resembles:

```rust
let x: f64 = ...; // some fraction that's a ratio of two u64's
let y: f64 = ...; // some alpenglow threshold that's also expressible as a ratio of two u64's

if x > y {
    ...
}
```

This PR replaces the above logic with purely int-based calculations to avoid loss-of-precision / inopportune rounding issues.

### DIFF
--- a/votor-messages/src/fraction.rs
+++ b/votor-messages/src/fraction.rs
@@ -1,5 +1,8 @@
+//! Fraction type for precise stake threshold comparisons.
+
 use std::num::NonZeroU64;
 
+/// Numerator / denominator, for precise comparisons without floating point.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct Fraction {
     numerator: u64,
@@ -7,6 +10,7 @@ pub struct Fraction {
 }
 
 impl Fraction {
+    /// Creates a new fraction.
     #[inline]
     pub const fn new(numerator: u64, denominator: NonZeroU64) -> Self {
         Self {
@@ -15,10 +19,16 @@ impl Fraction {
         }
     }
 
+    /// Creates a fraction from a percentage (e.g. 60 -> 60/100).
     #[inline]
     pub const fn from_percentage(pct: u64) -> Self {
         // SAFETY: 100 != 0
         Self::new(pct, unsafe { NonZeroU64::new_unchecked(100) })
+    }
+
+    /// Approximates this fraction as an f64.
+    pub fn approx_f64(&self) -> f64 {
+        self.numerator as f64 / self.denominator.get() as f64
     }
 }
 
--- a/votor-messages/src/migration.rs
+++ b/votor-messages/src/migration.rs
@@ -69,14 +69,16 @@ pub const MIGRATION_SLOT_OFFSET: Slot = 5000;
 #[cfg(feature = "dev-context-only-utils")]
 pub const MIGRATION_SLOT_OFFSET: Slot = 32;
 
+use crate::fraction::Fraction;
+
 /// We match Alpenglow's 20 + 20 model, by allowing a maximum of 20% malicious stake during the migration.
-pub const MIGRATION_MALICIOUS_THRESHOLD: u64 = 20;
+pub const MIGRATION_MALICIOUS_THRESHOLD: Fraction = Fraction::from_percentage(20);
 
 /// In order to rollback a block eligible for genesis vote, we need:
 /// `SWITCH_FORK_THRESHOLD` - (1 - `GENESIS_VOTE_THRESHOLD`) = `MIGRATION_MALICIOUS_THRESHOLD` malicious stake.
 ///
 /// Using 38% as the `SWITCH_FORK_THRESHOLD` gives us 82% for `GENESIS_VOTE_THRESHOLD`.
-pub const GENESIS_VOTE_THRESHOLD: u64 = 82;
+pub const GENESIS_VOTE_THRESHOLD: Fraction = Fraction::from_percentage(82);
 
 /// The interval at which we refresh our genesis vote
 pub const GENESIS_VOTE_REFRESH: Duration = Duration::from_millis(400);
@@ -516,7 +518,7 @@ impl MigrationStatus {
             .map(|b| *b != (slot, block_id))
             .unwrap_or(true)
         {
-            let genesis_vote_thresh_f64 = GENESIS_VOTE_THRESHOLD as f64 / 100.0;
+            let genesis_vote_thresh_f64 = GENESIS_VOTE_THRESHOLD.approx_f64();
             panic!(
                 "{}: We wish to cast a genesis vote on {discovered_genesis_block:?}, however we \
                  have received a genesis certificate for ({slot}, {block_id}). This means there \
@@ -567,7 +569,7 @@ impl MigrationStatus {
             return;
         };
         if *genesis_block != (slot, block_id) {
-            let genesis_vote_thresh_f64 = GENESIS_VOTE_THRESHOLD as f64 / 100.0;
+            let genesis_vote_thresh_f64 = GENESIS_VOTE_THRESHOLD.approx_f64();
             panic!(
                 "{}: We cast a genesis vote on {genesis_block:?}, however we have received a \
                  genesis certificate for ({slot}, {block_id}). This means there is significant \
--- a/votor/src/common.rs
+++ b/votor/src/common.rs
@@ -1,7 +1,7 @@
 use {
-    crate::fraction::Fraction,
     solana_votor_messages::{
         consensus_message::CertificateType,
+        fraction::Fraction,
         migration::GENESIS_VOTE_THRESHOLD,
         vote::{Vote, VoteType},
     },
@@ -53,10 +53,7 @@ pub const fn certificate_limits_and_vote_types(
             Fraction::from_percentage(60),
             &[VoteType::Skip, VoteType::SkipFallback],
         ),
-        CertificateType::Genesis(_, _) => (
-            Fraction::from_percentage(GENESIS_VOTE_THRESHOLD),
-            &[VoteType::Genesis],
-        ),
+        CertificateType::Genesis(_, _) => (GENESIS_VOTE_THRESHOLD, &[VoteType::Genesis]),
     }
 }
 
--- a/votor/src/consensus_pool/slot_stake_counters.rs
+++ b/votor/src/consensus_pool/slot_stake_counters.rs
@@ -7,10 +7,9 @@ use {
         },
         consensus_pool::stats::ConsensusPoolStats,
         event::VotorEvent,
-        fraction::Fraction,
     },
     solana_hash::Hash,
-    solana_votor_messages::vote::Vote,
+    solana_votor_messages::{fraction::Fraction, vote::Vote},
     std::{collections::BTreeMap, num::NonZeroU64},
 };
 
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -41,8 +41,7 @@ use {
             MAX_LOCKOUT_HISTORY,
         },
     },
-    solana_votor::fraction::Fraction,
-    solana_votor_messages::migration::GENESIS_VOTE_THRESHOLD,
+    solana_votor_messages::{fraction::Fraction, migration::GENESIS_VOTE_THRESHOLD},
     std::{
         cmp::Ordering,
         collections::{HashMap, HashSet},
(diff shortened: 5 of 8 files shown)


votor/src/common.rs
```diff
@@ -37,19 +37,67 @@ pub const fn conflicting_types(vote_type: VoteType) -> &'static [VoteType] {
...
-        CertificateType::Finalize(_) => (0.6, &[VoteType::Finalize]),
-        CertificateType::Skip(_) => (0.6, &[VoteType::Skip, VoteType::SkipFallback]),
+        CertificateType::FinalizeFast(_, _) => ((4, 5), &[VoteType::Notarize]),
+        CertificateType::Finalize(_) => ((3, 5), &[VoteType::Finalize]),
+        CertificateType::Skip(_) => ((3, 5), &[VoteType::Skip, VoteType::SkipFallback]),
         CertificateType::Genesis(_, _) => (GENESIS_VOTE_THRESHOLD, &[VoteType::Genesis]),
     }
 }
 
+fn bowtie(lhs: (u64, u64), rhs: (u64, u64)) -> Option<(u128, u128)> {
```
akhi3030: this is a great idea!  Some suggestions:

- Did you consider using an external crate e.g. https://github.com/dnsl48/fraction
- As you suggested, it might be a good idea to use a struct so that at instantiation time, we can check if denominator is 0.  As most of the fractions are constants, we could do unwraps there.
- The functions can return a https://doc.rust-lang.org/std/cmp/enum.Ordering.html instead of bool and then we do not need so many of them?

votor-messages/src/migration.rs
```diff
@@ -70,13 +70,13 @@ pub const MIGRATION_SLOT_OFFSET: Slot = 5000;
...
 /// We match Alpenglow's 20 + 20 model, by allowing a maximum of 20% malicious stake during the migration.
-pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
+pub const MIGRATION_MALICIOUS_THRESHOLD: (u64, u64) = (1, 5);
 
 /// In order to rollback a block eligible for genesis vote, we need:
 /// `SWITCH_FORK_THRESHOLD` - (1 - `GENESIS_VOTE_THRESHOLD`) = `MIGRATION_MALICIOUS_THRESHOLD` malicious stake.
 ///
 /// Using 38% as the `SWITCH_FORK_THRESHOLD` gives us 82% for `GENESIS_VOTE_THRESHOLD`.
-pub const GENESIS_VOTE_THRESHOLD: f64 = 82.0 / 100.0;
+pub const GENESIS_VOTE_THRESHOLD: (u64, u64) = (82, 100);
```
wen-coding: I think in the whitepaper and all SIMDs we talk percentage everywhere, and I believe 100 * total_stake will not overflow u64, so can we just use percentage?

ksn6: I considered this alternative initially, though went with fractions because:

- Rational numbers are more general in the event that we wish to have non-whole numbers.

- It's still a bit odd to commit to 100 as a magic denominator. There's no magic with rational numbers, and we don't really benefit from 100 as a magic denominator from a code complexity perspective.

votor/src/common.rs
ksn6: I looked through our current Cargo.toml prior to implementing this; we don't use a fraction crate at the moment. I'd rather not introduce a new dependency for this.

Yep, sounds reasonable - introducing a `Fraction` struct makes this cleaner.

votor/src/consensus_pool/slot_stake_counters.rs
```diff
@@ -94,10 +94,20 @@ impl SlotStakeCounters {
...
+            // UNWRAP: the total stake shouldn't overflow a u64
+            self.skip_total.checked_add(*stake).unwrap(),
+            self.total_stake,
+        );
+
+        let a = notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_ONLY;
+        let b = notarized_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_FOR_NOTARIZE_OR_SKIP;
+        let c = notarized_plus_skip_ratio >= SAFE_TO_NOTAR_MIN_NOTARIZE_AND_SKIP;
+
+        a ||
```
AshwinSekar: ?
can u revert and keep the comments with the logic like before

votor/src/common.rs
```diff
@@ -37,16 +37,65 @@ pub const fn conflicting_types(vote_type: VoteType) -> &'static [VoteType] {
 /// Must be in sync with `vote_to_cert_types`
 pub const fn certificate_limits_and_vote_types(
     cert_type: &CertificateType,
-) -> (f64, &'static [VoteType]) {
+) -> (Fraction, &'static [VoteType]) {
     match cert_type {
-        CertificateType::Notarize(_, _) => (0.6, &[VoteType::Notarize]),
-        CertificateType::NotarizeFallback(_, _) => {
-            (0.6, &[VoteType::Notarize, VoteType::NotarizeFallback])
+        CertificateType::Notarize(_, _) => (Fraction::new(3, 5), &[VoteType::Notarize]),
```
AshwinSekar: can we have a constructor that takes a percentage, `Fraction::new(3, 5)` is much less than clear than something like `Fraction::new(60)

votor/src/common.rs
```diff
@@ -37,16 +37,65 @@ pub const fn conflicting_types(vote_type: VoteType) -> &'static [VoteType] {
...
+            &[VoteType::Genesis],
+        ),
+    }
+}
+
+/// Numerator / denominator, with denominator != 0.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Fraction {
+    pub numerator: u64,
+    pub denominator: u64,
```
akhi3030: Please do not make these `pub`.  That allows one to create a `Fraction` without using a constructor and bypass the checks in the constructor.

Further, this struct should be in its own module otherwise users in this module can access the fields.

votor/src/common.rs
```diff
@@ -37,16 +37,65 @@ pub const fn conflicting_types(vote_type: VoteType) -> &'static [VoteType] {
...
+}
+
+impl Fraction {
+    #[inline]
+    pub const fn new(numerator: u64, denominator: u64) -> Self {
+        debug_assert!(denominator != 0);
```
akhi3030: I think this should either be an `assert` or this function should return an error.

votor/src/common.rs
akhi3030: @bw-solana: do you know if we have a policy on introducing new external dependencies?  I think it might make sense to use the external crate as I would like to use this approach for calculating rewards as well.  And longer term, we could use this approach to remove more uses of floats in the code.

votor/src/consensus_pool/slot_stake_counters.rs
ksn6: sure

votor/src/common.rs
ksn6: Thinking about this further - our `Fraction` implementation is super-tiny (~60 LOC) at the moment, and has an intentionally limited set of functionality. External crates are much more bloated.

I'd rather that we avoid folks doing accidentally arbitrary things with fractions (e.g., adding / subtracting, etc.) since that could potentially lead to hidden "big int" calculations happening. These inadvertent operations are easier to do if we introduce an external crate.

In any case, most reasonable use-cases of fractions involve adding with common denominators, etc., so users should really just be adding the numerators themselves, etc. If there are specific ops on the rewards side that don't fit this, we should consider adding external crates then.

votor-messages/src/migration.rs
ksn6: Majority wins - I'll introduce `Fraction::from_percentage` to address this.

votor/src/consensus_pool/slot_stake_counters.rs
```diff
@@ -94,10 +95,17 @@ impl SlotStakeCounters {
         let notarized_ratio = *stake as f64 / self.total_stake as f64;
         trace!("safe_to_notar {block_id:?} {skip_ratio} {notarized_ratio}");
         // Check if the block fits condition (i) 40% of stake holders voted notarize
+        let notarized_ratio = Fraction::new(*stake, self.total_stake);
+        let notarized_plus_skip_ratio = Fraction::new(
```
akhi3030: it seems like this is not adding the notar ratio?

votor/src/consensus_pool/slot_stake_counters.rs
```diff
@@ -113,11 +121,12 @@ impl SlotStakeCounters {
                 self.notarize_total,
                 self.top_notarized_stake
             );
-            self.skip_total
-                .saturating_add(self.notarize_total.saturating_sub(self.top_notarized_stake))
-                as f64
-                / self.total_stake as f64
-                >= SAFE_TO_SKIP_THRESHOLD
+
+            let num_stake = self
```
akhi3030, line 122: - nit: maybe subtract first and then add to ensure not overflowing?
- maybe we should use `checked_*` and unwrap?  if somehow overflows happen, we will get strange results here.

votor-messages/src/fraction.rs
```diff
@@ -0,0 +1,62 @@
...
+            .unwrap();
+        let rhs = (other.numerator as u128)
+            .checked_mul(self.denominator as u128)
+            .unwrap();
+        lhs.cmp(&rhs)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
akhi3030, line 55: please add a test to test that denominator = 0 panics.

AshwinSekar, line 55: can we also the case that demonstrates the motivation for this, e.g. a test where using f64 > would return the incorrect result but using Fraction returns the correct result?

votor/src/consensus_pool/slot_stake_counters.rs
ksn6: I believe we are - see the checked add below.

votor/src/consensus_pool/slot_stake_counters.rs
ksn6: Let's do these in a separate PR if needed and minimize diffs with the existing math logic.

votor/src/consensus_pool/slot_stake_counters.rs
akhi3030: You are right.  I got confused because the input to the function is `stake`.  All good.

votor/src/common.rs
billf: consider pushing the requirement to callers by using https://doc.rust-lang.org/std/num/type.NonZeroU64.html removes the need for errors or asserts (and the resulting tests). remains const-safe.

ksn6: nice - let's go with the `NonZeroU64` approach

votor-messages/src/fraction.rs
ksn6: not needed anymore given the `NonZeroU64` approach

votor-messages/src/migration.rs
```diff
@@ -70,13 +70,13 @@ pub const MIGRATION_SLOT_OFFSET: Slot = 5000;
 pub const MIGRATION_SLOT_OFFSET: Slot = 32;
 
 /// We match Alpenglow's 20 + 20 model, by allowing a maximum of 20% malicious stake during the migration.
-pub const MIGRATION_MALICIOUS_THRESHOLD: f64 = 20.0 / 100.0;
+pub const MIGRATION_MALICIOUS_THRESHOLD: u64 = 20;
```
akhi3030: why are these not expressed as fractions?

votor/src/fraction.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[inline]
+    pub const fn new(numerator: u64, denominator: NonZeroU64) -> Self {
+        Self {
+            numerator,
+            denominator,
+        }
+    }
+
+    #[inline]
+    pub fn from_tuple((numerator, denominator): (u64, u64)) -> Self {
```
akhi3030: doesn't seem to be used.

votor/src/consensus_pool/slot_stake_counters.rs
```diff
@@ -94,10 +95,15 @@ impl SlotStakeCounters {
         let notarized_ratio = *stake as f64 / self.total_stake as f64;
         trace!("safe_to_notar {block_id:?} {skip_ratio} {notarized_ratio}");
         // Check if the block fits condition (i) 40% of stake holders voted notarize
+        let total_stake = NonZeroU64::new(self.total_stake).unwrap();
```
akhi3030, line 97: this work has highlighted that `self.total_stake` is not guaranteed to be non-zero.  I think we need to do additional work to add checks for that.  Maybe store `self.total_stake` as `NonZero` as well.

ksn6: let's please do this in a separate follow-up PR; I don't think that's essential here + another PR is being held up on this

votor-messages/src/migration.rs
ksn6: yeah, might as well move `Fraction` to `votor-messages`; that's why I kept these as 20

PR #632, opened 2025-12-12
refactor: reduces code duplication in bls_sigverifier

akhi3030:
#### Problem

There is a lot of code duplication in the various tests of bls_sigverifier as they all have to create channels to pass to the setup code.

#### Summary of Changes

Reduces the code duplication by moving the channel creation into the setup code.

This also makes it easier to add new channels to the bls_sigverifier e.g. in https://github.com/anza-xyz/alpenglow/pull/582.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -1151,14 +1151,8 @@ mod tests {
 
     #[test]
     fn test_verify_certificate_base2_just_enough_stake() {
-        let (verified_vote_sender, _) = crossbeam_channel::unbounded();
-        let (message_sender, message_receiver) = crossbeam_channel::unbounded();
-        let (consensus_metrics_sender, _) = crossbeam_channel::unbounded();
-        let (validator_keypairs, mut verifier) = create_keypairs_and_bls_sig_verifier(
-            verified_vote_sender,
-            message_sender,
-            consensus_metrics_sender,
-        );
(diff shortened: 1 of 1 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -581,14 +585,37 @@ mod tests {
...
+    fn create_keypairs_and_bls_sig_verifier() -> (
+        Vec<ValidatorVoteKeypairs>,
+        BLSSigVerifier,
+        VerifiedVoteReceiver,
+        Receiver<ConsensusMessage>,
+    ) {
+        let (verified_votes_sender, verified_votes_receiver) = crossbeam_channel::unbounded();
+        let (message_sender, message_receiver) = crossbeam_channel::unbounded();
+        let (consensus_metrics_sender, consensus_metrics_receiver) = crossbeam_channel::unbounded();
+        std::thread::spawn(move || while consensus_metrics_receiver.recv().is_ok() {});
```
ksn6, line 633: nit: do we need to join this thread anywhere?

If it's a "drain thread," add a comment indicating as such.

ksn6 (commented):
LGTM after the nit

core/src/bls_sigverify/bls_sigverifier.rs
akhi3030: I don't think the thread needs to be joined.  It will exit when the sigverifier is dropped.  I added some comments explaining why we are creating the thread.

AshwinSekar (approved):

PR #637, opened 2025-12-16
feat: introduces skip and notar reward certificates

akhi3030:
#### Problem

We need to introduce new reward certificates for paying vote rewards.

#### Summary of Changes

Introduces the reward certificates without actually populating them.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -41,10 +41,11 @@ use {
             MAX_LOCKOUT_HISTORY,
         },
     },
-    solana_votor_messages::migration::GENESIS_VOTE_THRESHOLD,
+    solana_votor_messages::{fraction::Fraction, migration::GENESIS_VOTE_THRESHOLD},
     std::{
         cmp::Ordering,
         collections::{HashMap, HashSet},
+        num::NonZeroU64,
         ops::{
             Bound::{Included, Unbounded},
             Deref,
@@ -507,7 +508,8 @@ impl Tower {
         }
 
         let parent_is_super_oc = bank_slot == parent_slot + 1
-            && super_oc_stake as f64 / total_stake as f64 > GENESIS_VOTE_THRESHOLD;
+            && Fraction::new(super_oc_stake, NonZeroU64::new(total_stake).unwrap())
+                > GENESIS_VOTE_THRESHOLD;
 
         // TODO: populate_ancestor_voted_stakes only adds zeros. Comment why
         // that is necessary (if so).
--- a/votor-messages/src/fraction.rs
+++ b/votor-messages/src/fraction.rs
@@ -1,0 +1,79 @@
+//! Fraction type for precise stake threshold comparisons.
+
+use std::num::NonZeroU64;
+
+/// Numerator / denominator, for precise comparisons without floating point.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Fraction {
+    numerator: u64,
+    denominator: NonZeroU64,
+}
+
+impl Fraction {
+    /// Creates a new fraction.
+    #[inline]
+    pub const fn new(numerator: u64, denominator: NonZeroU64) -> Self {
+        Self {
+            numerator,
+            denominator,
+        }
+    }
+
+    /// Creates a fraction from a percentage (e.g. 60 -> 60/100).
+    #[inline]
+    pub const fn from_percentage(pct: u64) -> Self {
+        // SAFETY: 100 != 0
+        Self::new(pct, unsafe { NonZeroU64::new_unchecked(100) })
+    }
+
+    /// Approximates this fraction as an f64.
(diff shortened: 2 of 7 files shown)


votor-messages/src/reward_certificate.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl SkipRewardCertificate {
+    /// Creates a new [`SkipRewardCertificate`] for test purposes.
+    #[cfg(feature = "dev-context-only-utils")]
+    pub fn new_for_tests() -> Self {
+        let signature = BLSSignatureCompressed::default();
+        Self {
+            slot: 1234,
+            signature,
+            bitmap: vec![],
```
ksn6: Should probably throw in something non-empty here. Same with `NotarRewardCertificate` below.

votor-messages/src/reward_certificate.rs
```diff
@@ -0,0 +1,64 @@
...
+/// Unlike the skip certificate which can be base-2 or base-3 encoded, this is guaranteed to be base-2 encoded.
+#[derive(Clone, PartialEq, Eq, Debug, SchemaWrite, SchemaRead)]
+pub struct SkipRewardCertificate {
+    /// The slot the certificate is for.
+    pub slot: Slot,
+    /// The signature
+    #[wincode(with = "Pod<BLSSignatureCompressed>")]
+    pub signature: BLSSignatureCompressed,
+    /// The bitmap for validators, see solana-signer-store for encoding format
+    pub bitmap: Vec<u8>,
```
ksn6: This isn't correct, since we'd use 8 bytes to serialize the bitmap length here. Looks like we'll want 2 bytes instead according to SIMD-0428. We'll want to do something like this:

https://github.com/anza-xyz/alpenglow/blob/master/entry/src/block_component.rs#L363-L364

votor-messages/src/reward_certificate.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// The slot the certificate is for.
+    pub slot: Slot,
+    /// The block id the certificate is for.
+    #[wincode(with = "Pod<Hash>")]
+    pub block_id: Hash,
+    /// The signature
+    #[wincode(with = "Pod<BLSSignatureCompressed>")]
+    pub signature: BLSSignatureCompressed,
+    /// The bitmap for validators, see solana-signer-store for encoding format
+    pub bitmap: Vec<u8>,
```
ksn6: See https://github.com/anza-xyz/alpenglow/pull/637/changes#r2624211977

entry/src/block_component.rs
ksn6, line 1: Update the header docstrings to reflect the changes made.

akhi3030: done.

votor-messages/src/reward_certificate.rs
akhi3030: good catch.  Updated.

votor-messages/src/reward_certificate.rs
akhi3030: makes sense.  Done.

votor-messages/src/reward_certificate.rs
akhi3030: Done.

votor-messages/src/reward_certificate.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Error)]
+pub enum RewardCertError {
+    /// Invalid bitmap was supplied.
+    #[error("invalid bitmap was supplied")]
+    InvalidBitmap,
+}
+
+/// 2-byte length prefix (max 65535 elements).
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub struct U16Len;
```
ksn6, line 27: @cpubot - is this the best way of handling this? Is it worth tossing `U16Len` into `wincode` directly / does this already exist in `wincode`?

entry/src/block_component.rs
```diff
@@ -78,6 +78,14 @@
 /// │ Final Cert Present           (1 byte)   │
 /// ├─────────────────────────────────────────┤
 /// │ FinalCertificate (if present, variable) │
+/// ├─────────────────────────────────────────┤
+/// │ Skip reward cert Present     (1 byte)   │
+/// ├─────────────────────────────────────────┤
+/// │ SkipRewardCert (if present, variable)   │
+/// └─────────────────────────────────────────┘
```
ksn6: the ascii art here is slightly wrong lol

you need a ├

votor-messages/src/reward_certificate.rs
cpubot: This is the correct way to do it for now. wincode will have a generalized `FixInt<T>` length type in the future

entry/src/block_component.rs
akhi3030: keen eye =).  fixed.

PR #639, opened 2025-12-19
feat: introduces the `ConsensusRewards` container

akhi3030:
### Problem

We need to introduce a new struct for storing and aggregating votes needed to produce the reward certificates.

- https://github.com/anza-xyz/alpenglow/pull/637 introduced the reward certs

### Summary of Changes

- Introduces the `ConsensusReward` container
- it stores votes relevant for producing reward certs
- when requested to, produces reward certificates
- it receives votes and requests to produce certs over channels and sends the produced certs over another channel
- also introduces a service that manages all the channels
- **Note**: this not wired up to the sigverifier or the block creation loop yet to reduce the size of the PR.

#### Rough design of `ConsensusRewards`:
- Per slot, stores an `Entry` object
- uses `ClusterInfo` and `LeaderScheduleCache` to figure out which votes are actually useful

#### Rough design of `Entry`:
- stores a `PartialCert` for aggregating skip votes
- stores a `PartialCert` per blockid observed for notar votes
- contains code to minimise allocations and reallocations

#### Rough design of `PartialCert`:
- Stores a signature projective and a bitvec to aggregate votes into
- also stores info about max validators and if votes from all validators have been seen, then pre-computes the signature and the bitmap

### DIFF
--- a/votor/src/consensus_rewards.rs
+++ b/votor/src/consensus_rewards.rs
@@ -42,7 +42,7 @@ pub fn wants_vote(
         | Vote::Genesis(_) => return false,
     }
     let vote_slot = vote.vote.slot();
-    if vote_slot.saturating_add(NUM_SLOTS_FOR_REWARD) < root_slot {
+    if vote_slot.saturating_add(NUM_SLOTS_FOR_REWARD) <= root_slot {
         return false;
     }
     let my_pubkey = cluster_info.id();
@@ -161,13 +161,13 @@ impl ConsensusRewards {
     }
 
     /// Adds received [`VoteMessage`] from other validators.
-    fn add_vote(&mut self, bank: &Bank, vote: &VoteMessage) {
+    fn add_vote(&mut self, root_bank: &Bank, vote: &VoteMessage) {
         let slot = vote.vote.slot();
-        let Some(max_validators) = get_max_validators(bank, slot) else {
+        let Some(max_validators) = get_max_validators(root_bank, slot) else {
             warn!("failed to look up max_validators for slot {slot}");
             return;
         };
-        let root_slot = bank.slot();
+        let root_slot = root_bank.slot();
         // drop state that is too old based on how the root slot has progressed
         self.votes = self.votes.split_off(
             &(root_slot
--- a/votor/src/consensus_rewards/entry.rs
+++ b/votor/src/consensus_rewards/entry.rs
@@ -80,15 +80,11 @@ impl Entry {
                 warn!("Build skip reward cert failed with {e}");
                 None
             }
-            Ok((signature, bitmap)) => {
-                match SkipRewardCertificate::try_new(slot, signature, bitmap) {
-                    Ok(c) => Some(c),
-                    Err(e) => {
-                        warn!("Build skip reward cert failed with {e}");
-                        None
-                    }
-                }
-            }
+            Ok((signature, bitmap)) => SkipRewardCertificate::try_new(slot, signature, bitmap)
+                .inspect_err(|e| {
+                    warn!("Build skip reward cert failed with {e}");
+                })
+                .ok(),
         };
         BuildRewardCertsResponse { skip, notar }
     }
--- a/votor/src/consensus_rewards/entry/notar_entry.rs
+++ b/votor/src/consensus_rewards/entry/notar_entry.rs
@@ -60,36 +60,24 @@ impl NotarEntry {
     /// Builds a [`NotarRewardCertificate`] from the collected votes.
     pub(super) fn build_cert(self, slot: Slot) -> Option<NotarRewardCertificate> {
         // we can only submit one notar rewards certificate but different validators may vote for different blocks and we cannot combine notar votes for different blocks together in one cert.
-        // pick the block_id with most votes.
-        let mut max_entry = None;
-        for (block_id, partial) in self.partials {
-            match &mut max_entry {
-                None => max_entry = Some((block_id, partial)),
-                Some((_, max_partial)) => {
-                    if partial.votes_seen() > max_partial.votes_seen() {
-                        max_entry = Some((block_id, partial));
-                    }
-                }
-            }
-        }
-        match max_entry {
-            None => None,
-            Some((block_id, partial)) => match partial.build_sig_bitmap() {
-                Err(e) => {
-                    warn!("Build notar reward cert failed with {e}");
-                    None
-                }
-                Ok((signature, bitmap)) => {
-                    match NotarRewardCertificate::try_new(slot, block_id, signature, bitmap) {
-                        Ok(c) => Some(c),
-                        Err(e) => {
-                            warn!("Build notar reward cert failed with {e}");
-                            None
-                        }
-                    }
-                }
-            },
-        }
+        // ideally we should pick the block_id with the most stake to maximum leader rewards.
+        // we expect this to be rare enough that picking the block_id with the most votes should be fine in most cases.
+
+        let (block_id, partial) = self
+            .partials
+            .into_iter()
+            .max_by_key(|(_block_id, partial)| partial.votes_seen())?;
+        let (signature, bitmap) = partial
+            .build_sig_bitmap()
+            .inspect_err(|e| {
+                warn!("Build notar reward cert failed with {e}");
+            })
+            .ok()?;
+        NotarRewardCertificate::try_new(slot, block_id, signature, bitmap)
+            .inspect_err(|e| {
+                warn!("Build notar reward cert failed with {e}");
+            })
+            .ok()
     }
 }
 


votor/src/consensus_rewards/entry/partial_cert.rs
```diff
@@ -0,0 +1,216 @@
...
+                None => false,
+                Some(ind) => !*ind,
+            },
+        }
+    }
+
+    /// Adds the given [`VoteMessage`] to the aggregate.
+    pub(super) fn add_vote(&mut self, vote: &VoteMessage) -> Result<(), AddVoteError> {
+        match self {
+            Self::Done(_) => Err(AddVoteError::Duplicate),
```
AshwinSekar: I think it could be cleaner and provide better type safety if the `PartialCert` enum didn't exist:
- `new(), wants_vote()` and `add_vote()` are functions for `InProgressCert`
- `build_sig_bitmap()` consumes an `InProgressCert` and returns a `DoneCert`

Then we can ensure at compile time that we don't try to add a vote to an already finished cert

votor/src/consensus_rewards/entry/partial_cert.rs
```diff
@@ -0,0 +1,216 @@
...
+                            return Err(AddVoteError::Duplicate);
+                        }
+                        state
+                            .signature
+                            .aggregate_with(std::iter::once(&vote.signature))?;
+                        *ind = true;
+                    }
+                }
+                state.cnt = state.cnt.saturating_add(1);
+                if state.cnt == state.bitvec.len() {
```
AshwinSekar: i think it's unintuitive that the `add_vote` could end up building the cert and I don't think this gets us anything. It seems cleaner to wait for the request to actually come in before building.

votor/src/consensus_rewards/entry.rs
```diff
@@ -0,0 +1,201 @@
...
+    max_validators: usize,
+}
+
+impl Entry {
+    /// Creates a new instance of [`Entry`].
+    pub(super) fn new(max_validators: usize) -> Self {
+        Self {
+            skip: PartialCert::new(max_validators),
+            // under normal operations, all validators should vote for a single block id, still allocate space for a few more to hopefully avoid allocations.
+            notar: HashMap::with_capacity(5),
```
AshwinSekar: wondering if we should also verify that we don't have votes from the same validator for multiple block ids.

It would be a minor performance hit but might be worth it to stop an attack where the same validator spams us with imaginary block ids in an attempt to force more allocations.

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    /// Builds reward certificates.
+    fn build_certs(
+        &self,
+        slot: Slot,
+    ) -> (
+        Option<SkipRewardCertificate>,
+        Option<NotarRewardCertificate>,
+    ) {
+        match self.votes.get(&slot) {
```
AshwinSekar: if you like the suggestion from here https://github.com/anza-xyz/alpenglow/pull/639/changes#r2639468328

this could remove the entry, and consume it (which consumes the partial certs) to produce a `BuildRewardsCertResponse` instead

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// The current root slot.
+    pub root_slot: Slot,
+    /// List of [`AddVoteEntry`], one per vote.
+    pub votes: Vec<AddVoteEntry>,
+}
+
+/// Data structure for sending per vote state in the [`AddVoteMessage`].
+pub struct AddVoteEntry {
+    /// Maximum number of validators in the slot that this vote is for.
+    pub max_validators: usize,
```
AshwinSekar: nit: this isn't expected to change until the epoch boundary, would it be better to cache an epoch -> max_validators  in the service rather than having it in every message?

votor/src/consensus_rewards/entry.rs
akhi3030: Agreed!  Just pushed a patch to handle this.

votor/src/consensus_rewards.rs
akhi3030: @AshwinSekar: yes, I think that would be cleaner and make the messages smaller which is good.  

Do you have a suggestion on how to build such a map in the service?  

The only way I can think of would require the service to store `SharableBanks` to look up the `Bank` and from there look up the epoch stakes map.  Kinda similar to what is happening in https://github.com/anza-xyz/alpenglow/blob/7a55fc16ef35405967e164cb6f1a1b9d42617ab9/core/src/bls_sigverify/bls_sigverifier.rs#L49.  And this does not seem very clean to me.

votor/src/consensus_rewards/entry/partial_cert.rs
akhi3030: I thought we wanted as much as possible to not slow down the block creation loop?  That's why I thought building the cert ahead of time would help with that.  Is that not the case?

votor/src/consensus_rewards/entry/partial_cert.rs
akhi3030: Yeah, this is an interesting idea and goes into a bit of a rabbit hole in my head.

> `build_sig_bitmap()` consumes an `InProgressCert` and returns a `DoneCert`

Yes, I agree with this.  However, we would still need some kind of enum somewhere in `Entry` as we would be storing either a `InProgressCert` or a `DoneCert` there?

Further, now I am thinking if we make the assumptions that the block creation loop will ask to build reward certs for a given slot at most once and the slots will only ever be in a strictly increasing order.

If we make those assumptions, then we can clean things up a bit because when `build_sig_bitmap()` is called, it can consume `self` and the relevant state can be dropped because it will not longer be needed.  Does this though process make sense?

votor/src/consensus_rewards.rs
akhi3030: I suppose you are hinting that we can make the assumptions that I alluded to in https://github.com/anza-xyz/alpenglow/pull/639#discussion_r2661385630?  Then yes, I think we could make this change.

votor/src/consensus_rewards/entry/partial_cert.rs
AshwinSekar: > Further, now I am thinking if we make the assumptions that the block creation loop will ask to build reward certs for a given slot at most once and the slots will only ever be in a strictly increasing order.

Yes I think this is something we should expect. I would expect we only store `InProgressCert` and then when BCL needs a cert we consume and remove the entry

votor/src/consensus_rewards/entry/partial_cert.rs
AshwinSekar: That's not the case. When we're building the block, BCL essentially will just be `sleep(400ms)`. During this time we have 4 threads in banking stage that are scheduling and building the block. The rest of the validator is essentially asleep.

So at this point it's fine for this service to construct the aggregates

votor/src/consensus_rewards.rs
AshwinSekar: Instead of passing in the root slot via message, I would just have this service hold onto the shared root bank.

Then when a new vote comes in, we know the root slot for filtering, and can look up the map for the epoch from the root bank.

votor/src/consensus_rewards/entry/partial_cert.rs
akhi3030: cool.  Made the suggested change.

votor/src/consensus_rewards/entry/partial_cert.rs
akhi3030: ah got it.  Makes sense then to not prebuild.  Made the suggested changes.

votor/src/consensus_rewards.rs
akhi3030: made the suggested change.

votor/src/consensus_rewards.rs
akhi3030: introduced `SharableBanks`.

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,265 @@
...
+) -> bool {
+    match vote.vote {
+        Vote::Notarize(_) | Vote::Skip(_) => (),
+        Vote::Finalize(_)
+        | Vote::NotarizeFallback(_)
+        | Vote::SkipFallback(_)
+        | Vote::Genesis(_) => return false,
+    }
+    let vote_slot = vote.vote.slot();
+    if vote_slot.saturating_add(NUM_SLOTS_FOR_REWARD) < root_slot {
```
AshwinSekar:
```suggestion
    if vote_slot.saturating_add(NUM_SLOTS_FOR_REWARD) <= root_slot {
```

votor/src/consensus_rewards.rs
```diff
@@ -0,0 +1,265 @@
...
+            return false;
+        }
+        let Some(entry) = self.votes.get(&vote.vote.slot()) else {
+            return true;
+        };
+        entry.wants_vote(vote)
+    }
+
+    /// Adds received [`VoteMessage`] from other validators.
+    fn add_vote(&mut self, bank: &Bank, vote: &VoteMessage) {
```
AshwinSekar:
```suggestion
    fn add_vote(&mut self, root_bank: &Bank, vote: &VoteMessage) {
```

votor/src/consensus_rewards/entry/notar_entry.rs
```diff
@@ -0,0 +1,187 @@
...
+        if res.is_err() {
+            self.voted.remove(&rank);
+        }
+        res
+    }
+
+    /// Builds a [`NotarRewardCertificate`] from the collected votes.
+    pub(super) fn build_cert(self, slot: Slot) -> Option<NotarRewardCertificate> {
+        // we can only submit one notar rewards certificate but different validators may vote for different blocks and we cannot combine notar votes for different blocks together in one cert.
+        // pick the block_id with most votes.
```
AshwinSekar: I think this is fine as written, but just wanted to note that technically we'd want to maximize our leader reward so stake should also be taken into consideration here.

this is a rare enough issue that it shouldn't matter in practice.

votor/src/consensus_rewards/entry/notar_entry.rs
```diff
@@ -0,0 +1,187 @@
...
+        let mut max_entry = None;
```
AshwinSekar: maybe easier as
```rust 
let max_entry = self.partials
    .iter()
    .max_by_key(|(_block_id, partial)| partial.votes_seen());
```

votor/src/consensus_rewards/entry/notar_entry.rs
```diff
@@ -0,0 +1,187 @@
...
+            match &mut max_entry {
+                None => max_entry = Some((block_id, partial)),
+                Some((_, max_partial)) => {
+                    if partial.votes_seen() > max_partial.votes_seen() {
+                        max_entry = Some((block_id, partial));
+                    }
+                }
+            }
+        }
+        match max_entry {
```
AshwinSekar: to reduce nesting maybe something like this:

```rust
let (block_id, partial) = max_entry?;
let (signature, bitmap) = partial.build_sig_bitmap().map_err(|e| warn!(...)).ok()?;
NotarRewardCertificate::try_new(..).map_err(|e| warn!(..)).ok()
```

votor/src/consensus_rewards/entry/notar_entry.rs
akhi3030: yeah good point.  I'll add a note about it here in case we hit this issue in the future.

votor/src/consensus_rewards/entry/notar_entry.rs
akhi3030: nice!

votor/src/consensus_rewards/entry/notar_entry.rs
akhi3030: much cleaner!

PR #640, opened 2025-12-19
feat: Introduces merkle tree struct

akhi3030:
#### Problem

This is roughly a backport of https://github.com/anza-xyz/agave/pull/9180.

Currently a merkle tree is represented as a simple list of nodes.  We are in the process of introducing double merkle for alpenglow and this creates some problems for that.

In particular, we are going to have different types of merkle trees: chain and double.  It will be good to represent the different types of trees using different types to improve readability of the code and to enforce some checks.  See: https://github.com/anza-xyz/alpenglow/pull/518/files#diff-3c963fba687007333971f9a8049f494d9241e1b23693ee81da9c8f569aa1c376R29 for what we are envisioning.

#### Summary of Changes

This PR introduces a small initial change to represent a merkle tree using a struct instead of a list of nodes.  This has additional benefits of removing some panics when we look up the root of the tree and that the callers do not need to know precisely where in the list the root lives.

### DIFF
--- a/bls-cert-verify/src/cert_verify.rs
+++ b/bls-cert-verify/src/cert_verify.rs
@@ -33,24 +33,33 @@ pub enum CertVerifyError {
     Base3EncodingOnUnexpectedCert(CertificateType),
 }
 
-fn aggregate_keys_from_bitmap<F>(
+fn aggregate_keys_and_stakes_from_bitmap<F>(
     bit_vec: &BitVec<u8, Lsb0>,
-    rank_to_pubkey: &F,
-) -> Option<PubkeyProjective>
+    rank_to_pubkey_and_stake: &F,
+) -> Option<(PubkeyProjective, u64)>
 where
-    F: Fn(usize) -> Option<BlsPubkey> + Sync,
+    F: Fn(usize) -> Option<(BlsPubkey, u64)> + Sync,
 {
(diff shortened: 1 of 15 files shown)


ledger/src/shred/merkle_tree.rs
```diff
@@ -19,59 +19,81 @@ pub(crate) const MERKLE_HASH_PREFIX_NODE: &[u8] = b"\x01SOLANA_MERKLE_SHREDS_NOD
...
+    /// Tries to constructs a `MerkleTree`.
+    ///
+    /// # Errors:
+    ///
+    /// [`Error::EmptyIterator`] if the function is called with an empty iterator.
+    /// [`Error`] if any of the elements in the iterator contains an [`Error`].
+    pub(crate) fn try_new(
+        shreds: impl ExactSizeIterator<Item = Result<Hash, Error>>,
+    ) -> Result<MerkleTree, Error> {
+        if shreds.len() == 0 {
```
ksn6, line 40:
```suggestion
        if shreds.is_empty() {
```

akhi3030: `shreds` is an iterator here and does not have the `is_empty()` method.  As far as I know, `len() == 0` is the cleanest way to do this check.

PR #641, opened 2025-12-19
feat: introduces `NonZeroStake`

akhi3030:
#### Problem

We have a bunch of code that either implicitly (divides by 0) or explicitly (using `unwrap`, etc.) assumes that stake cannot be zero.

#### Summary of Changes

Introduces a new type: `NonZeroStake` that is guaranteed to not be 0.  And uses it in a bunch of places allowing us to remove a bunch of panics and unwraps.


ksn6:
The priority right now is getting feature-complete. Refactors like this are better suited for after we've merged upstream.

PR #642, opened 2025-12-23
fast leader handover: core logic

ksn6:
This PR implements Milestone 1 as articulated here: https://github.com/anza-xyz/alpenglow/issues/293#issuecomment-3145962346.

We implement the "happy case" for fast leader handover; i.e., the case where the optimistic parent matches the confirmed parent obtained via the parent ready channel.

For the "sad case," the leader disseminates an `UpdateParent` message. Leaders and non-leaders both purge the bank and replace it with a new one. The block essentially is a no-op.

Follow-up PRs:
- Milestone 2: Allow new transactions to play on top of this new bank
- Local cluster testing

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -4790,46 +4790,41 @@ impl ReplayStage {
 
                 debug_assert!(!progress.contains_key(&child_slot));
 
-                // Determine actual parent and replay offset from ParentMeta.
-                let (actual_parent_bank, replay_offset) =
-                    if !migration_status.should_allow_block_markers(child_slot) {
-                        (parent_bank.clone(), None)
-                    } else {
-                        let Some(parent_meta) = blockstore
-                            .get_parent_meta(child_slot, BlockLocation::Original)
-                            .expect("Blockstore should not fail")
-                        else {
-                            continue;
-                        };
-
-                        // No lock is held between get_slots_since and get_parent_meta, so
-                        // ParentMeta could have been updated in between. If so, continue and
-                        // the next iteration of generate_new_bank_forks will have consistent values.
-                        if parent_meta.parent_slot != parent_slot {
-                            continue;
-                        }
-
-                        // UpdateParent: use new parent and calculate replay offset
-                        let Some(new_parent) = frozen_banks.get(&parent_meta.parent_slot) else {
-                            // New parent not frozen yet, skip
-                            continue;
-                        };
-                        let num_shreds = u64::from(parent_meta.replay_fec_set_index);
-                        info!(
-                            "slot {child_slot}: UpdateParent detected, using parent {} with \
-                             replay offset {} shreds",
-                            parent_meta.parent_slot, num_shreds
-                        );
-                        (new_parent.clone(), Some(num_shreds))
+                // Determine replay offset from ParentMeta.
+                let replay_offset = if !migration_status.should_allow_block_markers(child_slot) {
+                    None
+                } else {
+                    let Some(parent_meta) = blockstore
+                        .get_parent_meta(child_slot, BlockLocation::Original)
+                        .expect("Blockstore should not fail")
+                    else {
+                        continue;
                     };
 
+                    // No lock is held between get_slots_since and get_parent_meta, so
+                    // ParentMeta could have been updated in between. If so, continue and
+                    // the next iteration of generate_new_bank_forks will have consistent values.
+                    if parent_meta.parent_slot != parent_slot {
+                        continue;
+                    }
+
+                    let num_shreds = u64::from(parent_meta.replay_fec_set_index);
+                    if num_shreds > 0 {
+                        info!(
+                            "slot {child_slot}: replay offset {} shreds from parent {}",
+                            num_shreds, parent_meta.parent_slot
+                        );
+                    }
+                    Some(num_shreds)
+                };
+
                 let leader = leader_schedule_cache
-                    .slot_leader_at(child_slot, Some(&actual_parent_bank))
+                    .slot_leader_at(child_slot, Some(parent_bank))
                     .unwrap();
                 info!(
                     "new fork:{} parent:{} root:{}",
                     child_slot,
-                    actual_parent_bank.slot(),
+                    parent_bank.slot(),
                     forks.root()
                 );
                 // Migration period banks are VoM
@@ -4840,7 +4835,7 @@ impl ReplayStage {
                     info!("Replaying block in slot {child_slot} in VoM");
                 }
                 let child_bank = Self::new_bank_from_parent_with_notify(
-                    actual_parent_bank.clone(),
+                    parent_bank.clone(),
                     child_slot,
                     forks.root(),
                     &leader,
@@ -4853,7 +4848,7 @@ impl ReplayStage {
                 if let Some(num_shreds) = replay_offset {
                     let prev_leader_slot = progress.get_bank_prev_leader_slot(&child_bank);
                     let fork_progress = ForkProgress::new(
-                        actual_parent_bank.last_blockhash(),
+                        parent_bank.last_blockhash(),
                         prev_leader_slot,
                         None,
                         0,
@@ -4868,7 +4863,7 @@ impl ReplayStage {
                     progress,
                     empty,
                     vec![leader],
-                    actual_parent_bank.slot(),
+                    parent_bank.slot(),
                     &forks,
                 );
                 new_banks.insert(child_slot, child_bank);


core/src/block_creation_loop.rs
```diff
@@ -566,15 +732,22 @@ fn start_leader_retry_replay(
...
+                    // TODO(ksn): this isn't right - fix this
+                    ctx.replay_highest_frozen
+                        .freeze_notification
+                        .wait_timeout_while(
+                            highest_frozen_slot,
+                            Duration::from_millis(800),
+                            |hfs| *hfs < parent_slot,
+                        )
+                        .unwrap()
+                };
```
ksn6: @AshwinSekar - thoughts on what to do in the `else` case here?

In particular, in the `else` case, `ParentReady` hasn't happened yet. That's because `timeout` is `None`; i.e., the skip timer doesn't exist yet. How long should we be waiting prior to time out?

core/src/block_creation_loop.rs
```diff
@@ -448,43 +576,81 @@ fn shutdown_and_drain_record_receiver(
...
             break;
         }
-
-        let Ok(record) = record_receiver.recv_timeout(remaining_slot_time) else {
-            continue;
+        let timeout = remaining.unwrap_or(Duration::from_secs(1));
+
+        // Wait for either a record or (if optimistic) parent ready notification
+        let result = if optimistic_parent.is_some() {
+            select_biased! {
```
ksn6: I'm writing the code this way to avoid a busy loop where I keep checking between (1) a parent ready and (2) a new incoming transaction from poh recorder.

I needed to explicitly expose `ctx.record_receiver.receiver` as a result of this, though. Curious as to whether there's a nicer way of doing this.

core/src/block_creation_loop.rs
```diff
@@ -238,40 +247,75 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
+        // Race between parent ready notification and optimistic parent events.
+        // Parent ready is checked first and has priority if both channels are ready.
+        let window_source = select_biased! {
+            recv(ctx.leader_window_info_receiver) -> msg => {
+                msg.ok().map(ParentSource::ParentReady)
+            },
+            recv(optimistic_parent_receiver) -> msg => {
+                msg.ok().map(ParentSource::OptimisticParent)
+            },
+            default(Duration::from_secs(1)) => None,
```
AshwinSekar: think we can just `continue` here
```suggestion
            default(Duration::from_secs(1)) => continue,
```

core/src/block_creation_loop.rs
```diff
@@ -238,40 +247,75 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
+                ParentSource::ParentReady(latest)
+            }
+            Some(ParentSource::OptimisticParent(first)) => {
+                let latest = optimistic_parent_receiver
+                    .try_iter()
+                    .last()
+                    .unwrap_or(first);
+                ParentSource::OptimisticParent(latest)
+            }
+            None => continue,
```
AshwinSekar: if we continue on timeout, then we can return here (and info log that one of the channels was disconnect so we're shutting down)

core/src/block_creation_loop.rs
```diff
@@ -238,40 +247,75 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
-                        .try_iter()
-                        .last()
-                        .or(Some(window))
-                })
-            else {
-                continue;
-            };
-
-            info
+        } = match window_source {
```
AshwinSekar: We should just combine this with the previous match and add the trace logs there

core/src/block_creation_loop.rs
```diff
@@ -238,40 +247,75 @@ fn start_loop(config: BlockCreationLoopConfig) {
...
+            }
         };
 
-        // TODO(ksn): fast leader handover. Once we can stream parent ready events over a channel,
-        // we'll have the two channels race each other to determine what to do.
-        //
-        // For now, just drain the channel and don't do anything with optimistic parents.
-        while let Ok(_optimistic_parent) = optimistic_parent_receiver.try_recv() {}
-
         trace!("Received window notification for {start_slot} to {end_slot} parent: {parent_slot}");
```
AshwinSekar: lets include `fast_leader_handover` in this log message

votor/src/event.rs
```diff
@@ -19,7 +19,7 @@ pub struct LeaderWindowInfo {
     pub start_slot: Slot,
     pub end_slot: Slot,
     pub parent_block: Block,
-    pub skip_timer: Instant,
+    pub skip_timer: Option<Instant>,
```
AshwinSekar: I think this should be kept as `Instant` and `replay_stage.rs` should pass in the current time.

Here are my personal thoughts:
- The goal of FLH is to minimize the time during which no shreds are being produced
- The goal is not to allow the leader to extend their first slot

I do understand the other side - end state could be that leader mods their  block timer to start once the ParentReady is viewed anyway.

I think this abuse already exists due to the buffer in skip timeout. It will have to be condemned via metrics anyway. This current approach will end up looking weird in metrics:
- First block ~500 ms
- Rest of the blocks 400 ms

We should aim to have all blocks as 400ms.

core/src/block_creation_loop.rs
```diff
@@ -419,6 +463,81 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
+    new_parent_block: Block,
+) -> Result<(), PohRecorderError> {
+    let (new_parent_slot, new_parent_block_id) = new_parent_block;
+    let update_parent = UpdateParentV1 {
+        new_parent_slot,
+        new_parent_block_id,
+    };
+    let update_parent = VersionedUpdateParent::V1(update_parent);
+    let update_parent = BlockMarkerV1::new_update_parent(update_parent);
+    let update_parent = VersionedBlockMarker::V1(update_parent);
```
AshwinSekar: nit: doesn't have to be in this PR but maybe we just have some `VersionedBlockMarker::new_X()` helpers to avoid this boilerplate.

Will also make it straightforward when we bump versions - can limit the feature set code to `new_X()`

core/src/block_creation_loop.rs
AshwinSekar: here's what i'm thinking - we just remove the branch here:

```rust
loop {
    ...
    let timeout = ...
    select_biased! {
       recv(ctx.leader_window_info_receiver) -> msg => {
           // if the info is > then our current window we're cooked anyway so we should return
           match msg {
              None => continue,
              Ok(info) if info.start_slot > slot => return;
              Ok(info) => ... // maybe send UpdateParent
           }
       recv(ctx.record_receiver) => ... // record the tx
       default(timeout) => continue
    }
}
```
(haven't tested this compiles)

core/src/block_creation_loop.rs
```diff
@@ -448,43 +583,81 @@ fn shutdown_and_drain_record_receiver(
...
+        match result {
+            RecvResult::Record(record) => {
+                ctx.poh_recorder.write().unwrap().record(
+                    record.slot,
+                    record.mixins,
+                    record.transaction_batches,
+                )?;
+            }
+            RecvResult::ParentReady(leader_window_info) => {
+                // UNWRAP: the select_biased! above only produces ParentReady if optimistic_parent
```
AshwinSekar: i think with my suggestion above we also combine this into the select (remove the intermediate `result`) and then wed on't have to worry about this

core/src/block_creation_loop.rs
AshwinSekar: my thoughts on this are here https://github.com/anza-xyz/alpenglow/pull/642/changes#r2645997858

i think we should always have a skip timer. Also should probably rename this to block timer.

core/src/block_creation_loop.rs
ksn6: yep, will make another PR for this

votor/src/event.rs
ksn6: Could you clarify this thought in more detail? Are you suggesting that the block timer commences as soon as either (a) an optimistic parent arrives or (b) a ParentReady is generated?

If so, I think this is incorrect - in theory, a Byzantine leader N could cause leader N + 1's full window to be skipped.

Seems like we should be fixing the metrics issue instead?

ksn6: Discussed offline - block timer != skip timer, and I'm referring to the skip timer in the above argument.

We'll reset the block timer on the sad path of fast leader handover though.

core/src/block_creation_loop.rs
```diff
@@ -448,43 +570,61 @@ fn shutdown_and_drain_record_receiver(
...
+    mut optimistic_parent: Option<(Slot, Hash)>,
+    block_timer: &mut Instant,
     block_timeout: Duration,
 ) -> Result<(), PohRecorderError> {
     loop {
-        let remaining_slot_time = block_timeout.saturating_sub(block_timer.elapsed());
-        if remaining_slot_time.is_zero() {
+        // Check if we've exceeded the block timeout
+        let timeout = time_left(*block_timer, block_timeout);
+        if timeout.is_zero() {
```
AshwinSekar: I think we only break once we receive the ParentReady
```suggestion
        if timeout.is_zero() && optimistic_parent.is_none() {
```

If the cluster is having network issues leading to a slow parent ready we shouldn't punish the leader

core/src/block_creation_loop.rs
```diff
@@ -419,6 +460,82 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
+    send_update_parent(&ctx.poh_recorder, leader_window_info.parent_block)?;
+    shutdown_and_drain_record_receiver(&ctx.poh_recorder, &mut ctx.record_receiver)?;
+
+    let slot = leader_window_info.start_slot;
+    let (old_parent_slot, _) = optimistic_parent_block;
+    let (new_parent_slot, _) = leader_window_info.parent_block;
+
+    ReplayStage::clear_bank(&ctx.bank_forks, slot);
+
+    let Some(parent_bank) = ctx.bank_forks.read().unwrap().get(new_parent_slot) else {
```
AshwinSekar: why not use `start_leader_retry_replay` here?

core/src/block_creation_loop.rs
```diff
@@ -448,43 +570,61 @@ fn shutdown_and_drain_record_receiver(
...
+    }
 
     // Alpentick and clear bank
-    let mut w_poh_recorder = poh_recorder.write().unwrap();
+    let mut w_poh_recorder = ctx.poh_recorder.write().unwrap();
     let bank = w_poh_recorder
         .bank()
         .expect("Bank cannot have been cleared as BlockCreationLoop is the only modifier");
 
-    trace!(
```
AshwinSekar, line 508: why remove this?

core/src/replay_stage.rs
```diff
@@ -3583,6 +3584,21 @@ impl ReplayStage {
         Ok(())
     }
 
+    pub(crate) fn clear_bank(bank_forks: &RwLock<BankForks>, slot: Slot) {
```
AshwinSekar: think this can just be a fn on `bank_forks`

core/src/replay_stage.rs
```diff
@@ -3622,6 +3638,22 @@ impl ReplayStage {
...
+                    )) => {
+                        let parent_slot = match update_parent {
+                            VersionedUpdateParent::V1(x) => x.new_parent_slot,
+                        };
+
+                        Self::clear_bank(bank_forks, bank_slot);
+
+                        if let Some(parent_bank) = bank_forks.read().unwrap().get(parent_slot) {
+                            if let Some(slot_progress) = progress.get_mut(&bank_slot) {
+                                slot_progress.replay_progress.write().unwrap().last_entry =
```
AshwinSekar: maybe a comment to explain why this is necessary, and how it works if the new parent bank is not yet frozen

core/src/replay_stage.rs
```diff
@@ -3622,6 +3638,22 @@ impl ReplayStage {
...
                 match replay_result {
                     Ok(replay_tx_count) => tx_count += replay_tx_count,
+                    Err(BlockstoreProcessorError::BlockComponentProcessor(
+                        BlockComponentProcessorError::AbandonedBank(update_parent),
+                    )) => {
+                        let parent_slot = match update_parent {
+                            VersionedUpdateParent::V1(x) => x.new_parent_slot,
+                        };
+
+                        Self::clear_bank(bank_forks, bank_slot);
```
AshwinSekar: does this work without clearing the ancestor / descendents map? I guess because it's not used and reset on the next iteration?

ksn6:
> can be in a follow up PR but remember we also have to handle the case when we:
> 
> * don't have the original parent, make sure we start replaying from the update parent on the new block
> * the part of the block before the UpdateParent causes the bank to be marked as dead, however the portion after the UpdateParent is valid, make sure we are still able to replay

Yep, this was intentionally left out for a follow-up PR.

core/src/replay_stage.rs
ksn6: `clear_bank` -> `dump_slots` -> `bank.remove()` handles descendants cleanup

I believe ancestors are re-calculated on the fly in `bank_forks.rs` (see `ancestors()`)

core/src/block_creation_loop.rs
```diff
@@ -446,33 +563,57 @@ fn shutdown_and_drain_record_receiver(
...
+            default(timeout) => {},
+        }
     }
 
-    // Shutdown and clear any inflight records
-    // TODO: do we need to lower the block timeout from 400ms to account for this / insertion of the block footer
-    shutdown_and_drain_record_receiver(poh_recorder, record_receiver)?;
+    shutdown_and_drain_record_receiver(&ctx.poh_recorder, &mut ctx.record_receiver)?;
+
+    if optimistic_parent.is_some() {
```
AshwinSekar: nit: can probably just assert that this is none. Should be a programmer error if we still get some at this point

core/src/block_creation_loop.rs
```diff
@@ -417,6 +460,75 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
+        ctx.my_pubkey,
+        optimistic_parent_block,
+        leader_window_info.parent_block
+    );
+
+    // If the optimistic parent doesn't match the finalized parent (specified in ParentReady), then
+    // this resets the block timer to the new parent's timer.
+    *block_timer = leader_window_info.block_timer;
+
+    send_update_parent(&ctx.poh_recorder, leader_window_info.parent_block)?;
```
AshwinSekar: we need to flip the order of these two (`shutdown_and_drain_record_receiver` before sending the update parent), otherwise we could end up sending some records after the update parent for the old bank which causes a divergence between the leader and the replayer

core/src/block_creation_loop.rs
```diff
@@ -417,6 +460,75 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
+    new_parent_block: Block,
+) -> Result<(), PohRecorderError> {
+    let (new_parent_slot, new_parent_block_id) = new_parent_block;
+    let update_parent = UpdateParentV1 {
+        new_parent_slot,
+        new_parent_block_id,
+    };
+    let update_parent = VersionedUpdateParent::V1(update_parent);
+    let update_parent = BlockMarkerV1::new_update_parent(update_parent);
+    let update_parent = VersionedBlockMarker::V1(update_parent);
```
AshwinSekar: nit: can use `VersionedBlockMarker::new_update_parent` now

core/src/replay_stage.rs
```diff
@@ -3622,6 +3623,33 @@ impl ReplayStage {
...
                 match replay_result {
                     Ok(replay_tx_count) => tx_count += replay_tx_count,
+                    Err(BlockstoreProcessorError::BlockComponentProcessor(
+                        BlockComponentProcessorError::AbandonedBank(update_parent),
+                    )) => {
+                        // Handle UpdateParent marker during fast leader handover.
+                        // The leader built on an optimistic parent that didn't match the
+                        // finalized parent, so we need to clear the bank and restart replay
+                        // from the new parent.
+                        let parent_slot = match update_parent {
```
AshwinSekar: we should make sure that this `parent_slot` is < the current parent

core/src/replay_stage.rs
```diff
@@ -3622,6 +3623,33 @@ impl ReplayStage {
...
+                        // updated by BankForks::remove(), and ancestors are recomputed fresh
+                        // from bank_forks state when needed.
+                        bank_forks.write().unwrap().clear_bank(bank_slot);
+
+                        // Reset the replay progress to continue from the new parent.
+                        // If the new parent bank is not yet available (not frozen), the
+                        // replay loop will wait for it on the next iteration when it
+                        // processes this slot again.
+                        if let Some(parent_bank) = bank_forks.read().unwrap().get(parent_slot) {
+                            if let Some(slot_progress) = progress.get_mut(&bank_slot) {
```
AshwinSekar: for my understanding, this should always exist right?

core/src/replay_stage.rs
```diff
@@ -3622,6 +3623,33 @@ impl ReplayStage {
...
+                                slot_progress.replay_progress.write().unwrap().last_entry =
```
AshwinSekar: what happens if we don't have this line?
and where is it set in the case that the new parent is not yet available?

core/src/block_creation_loop.rs
ksn6: yep, good call

core/src/replay_stage.rs
ksn6: yeah - I'll just hard-unwrap here

core/src/replay_stage.rs
ksn6: > what happens if we don't have this line?

we need this line to get entry verification to pass

> and where is it set in the case that the new parent is not yet available?

yeah, good point - just added logic for this; if the new parent isn't available, added logic that does the following:

- in `replay_progress`, we indicate a "pending new parent"
- in `generate_new_bank_forks`, if we're pending a new parent for a slot and the new parent isn't frozen yet, we skip over it and return to it on the next iteration; once the new parent shows up, we update last_entry, clear the pending new parent in the progress struct and keep going

core/src/replay_stage.rs
```diff
@@ -3632,6 +3633,70 @@ impl ReplayStage {
...
+                        };
+
+                        info!(
+                            "AbandonedBank at slot {bank_slot}: switching parent from {} to \
+                             {new_parent_slot}",
+                            bank.parent_slot()
+                        );
+
+                        // The new parent must be earlier than the current slot.
+                        assert!(
```
AshwinSekar: This shouldn't be an assert here, a malicious operator can send us an UpdateParent that causes this assert to fail and bring down the validator.

I think either here or in block_component_processor we should check the following:
- new_parent < current_slot
- new_parent < previous_parent

and mark the block as dead if these are violated.

core/src/replay_stage.rs
```diff
@@ -3632,6 +3633,70 @@ impl ReplayStage {
...
+                            new_parent_slot < bank_slot,
+                            "UpdateParent new_parent_slot ({new_parent_slot}) must be < bank_slot \
+                             ({bank_slot})"
+                        );
+
+                        // Clear the bank from bank_forks. It will be recreated with the
+                        // correct parent by generate_new_bank_forks on the next iteration.
+                        bank_forks.write().unwrap().clear_bank(bank_slot);
+
+                        // Update last_entry to the new parent's blockhash for entry verification.
```
AshwinSekar: Discussed offline we'll instead just clear the progress here and update correctly when creating the new bank

core/src/replay_stage.rs
ksn6: Ah, good call - I'll do this in `block_component_processor.rs`.

core/src/replay_stage.rs
```diff
@@ -4752,13 +4787,40 @@ impl ReplayStage {
                     trace!("child already active or frozen {child_slot}");
                     continue;
                 }
+
+                // Determine actual parent and replay offset.
+                // If progress entry exists, the slot was already set up - use default parent.
+                // If no progress entry, check ParentMeta for UpdateParent info.
+                let (actual_parent_bank, replay_offset) = if progress.contains_key(&child_slot) {
```
AshwinSekar: this case isn't necessary. Now `progress` should never contain `child_slot` since we remove it. Prefer a
`debug_assert!(!progress.contains_key(&child_slot))`

core/src/replay_stage.rs
```diff
@@ -4752,13 +4787,40 @@ impl ReplayStage {
...
+                    (parent_bank.clone(), None)
+                } else {
+                    match blockstore.get_parent_meta(child_slot, BlockLocation::Original) {
```
AshwinSekar: The `ParentMeta` *must* exist, otherwise we have a bug. I would either `.unwrap().unwrap()` here or at the very least spit out some `error!` logs and ignore this bank - no use replaying this bank with the wrong parent.

ledger/src/blockstore.rs
```diff
@@ -896,8 +896,7 @@ impl Blockstore {
         block_location: BlockLocation,
     ) -> Option<Hash> {
         let slot_meta = self
-            .meta_cf
-            .get(slot)
+            .meta_from_location(slot, block_location)
```
AshwinSekar: think this change is from my PR right? can we revert it here?

runtime/src/block_component_processor.rs
```diff
@@ -279,11 +281,23 @@ impl BlockComponentProcessor {
...
     ) -> Result<(), BlockComponentProcessorError> {
         if self.update_parent.is_some() {
             return Err(BlockComponentProcessorError::MultipleUpdateParents);
         }
 
+        let new_parent_slot = match update_parent {
+            VersionedUpdateParent::V1(x) => x.new_parent_slot,
+        };
+
+        if new_parent_slot >= slot {
```
AshwinSekar: do we also want to check that the `new_parent_slot` < `previous_parent_slot` ?

AshwinSekar: I guess we should also have this check in `on_header` too right?

ledger/src/blockstore.rs
ksn6: yep, my bad - ended up using the same checkout to review your PR

runtime/src/block_component_processor.rs
ksn6: > do we also want to check that the new_parent_slot < previous_parent_slot ?

yep, makes sense - I think we'll actually want to check:

(1) `new_parent_slot < original_parent_slot`, or:
(2) `new_parent_slot == original_parent_slot && block ids are different for the two`

since an `UpdateParent` could technically result in an equivocated block being the new parent

> I guess we should also have this check in on_header too right?

not quite - in `on_header`, if we've seen an `UpdateParent`, we should error stating that we've observed a spurious parent; in particular, block component processor runs during replay - so it runs through shreds in order rather than out-of-order

core/src/replay_stage.rs
ksn6: Discussed offline - this isn't quite right due to the following logic:

- Shred index 5 (or any non-zero shred) arrives for slot N
- `get_slot_meta_entry` creates `SlotMeta` with `parent_slot = Some(M)` from `shred.parent()`
- `handle_chaining_for_slot` adds slot `N` to parent's `next_slots`
- `replay_stage` sees slot `N` in `next_slots`, tries to get `ParentMeta`
- `ParentMeta` doesn't exist because shred 0 hasn't arrived yet

Will add in an assert indicating the block isn't full in the other case.

runtime/src/block_component_processor.rs
ksn6: discussed offline - I think we'll want to revert the changes in block component processor entirely in favor of doing this on shred ingest to more cleanly handle the "haiku" case (i.e., the case where we have garbage shreds)

concretely, we'll cease to insert shreds into blockstore upon receiving a conflicting `UpdateParent` / block header

will handle this in the next PR

core/src/block_creation_loop.rs
```diff
@@ -468,39 +584,63 @@ fn shutdown_and_drain_record_receiver(
...
+    block_timeout: Duration,
 ) -> Result<(), PohRecorderError> {
-    build_reward_certs_sender
+    ctx.build_reward_certs_sender
         .send(BuildRewardCertsRequest { slot })
         .map_err(|_| PohRecorderError::ChannelDisconnected)?;
     loop {
-        let remaining_slot_time = block_timeout.saturating_sub(block_timer.elapsed());
-        if remaining_slot_time.is_zero() {
+        // Don't timeout until we've received ParentReady
```
qkniep, line 597: before we received the ParentReady we shouldn't even start the timer in the first place

ksn6: that's not quite correct

you're referring to the "skip timer," which is done in `handle_parent_ready_event` within `event_handler.rs` (see the invocation to `set_timeouts`); in particular, we start the skip timer only upon receipt of a `ParentReady` update

that's not what we're referring to here - the block timer here determines how long to collect transactions per block (e.g., when to shut down the record receiver + insert the footer + insert the alpentick), how long to retry if replay is behind, etc.

core/src/block_creation_loop.rs
```diff
@@ -468,39 +584,63 @@ fn shutdown_and_drain_record_receiver(
...
+        }
     }
 
     // Shutdown and clear any inflight records
-    // TODO: do we need to lower the block timeout from 400ms to account for this / insertion of the block footer
-    shutdown_and_drain_record_receiver(poh_recorder, record_receiver)?;
+    shutdown_and_drain_record_receiver(&ctx.poh_recorder, &mut ctx.record_receiver)?;
+
+    // By now, we should have received ParentReady and called handle_parent_ready().
+    assert!(
```
AshwinSekar: In the case above `L606` where we are so behind that the window has moved on this assert will fail right?

core/src/replay_stage.rs
```diff
@@ -4752,13 +4787,49 @@ impl ReplayStage {
...
+                    let Some(parent_meta) = blockstore
+                        .get_parent_meta(child_slot, BlockLocation::Original)
+                        .expect("Blockstore should not fail")
+                    else {
+                        // Shred 0 shouldn't be present, since parent meta isn't present.
+                        debug_assert!(!blockstore.is_full(child_slot));
+                        continue;
+                    };
+
+                    if parent_meta.populated_from_update_parent() {
```
AshwinSekar: I think we should remove this guard, we should always check the `ParentMeta` here since we'll need to additionally check that the `parent_block_id` matches the parent's block id (you can check that here and `continue` or I can do it in my PR

core/src/replay_stage.rs
```diff
@@ -4752,13 +4787,49 @@ impl ReplayStage {
...
+                        // UpdateParent: use new parent and calculate replay offset
+                        let Some(new_parent) = frozen_banks.get(&parent_meta.parent_slot) else {
```
AshwinSekar: Also `parent_meta.parent_slot == parent_slot` right? Only case I see is because we don't hold the shred lock, a ParentMeta could have been updated in between line 4768 and this get?

Maybe a comment to explain, or simply continue if `parent_meta.parent_slot != parent_slot` as the next iteration of `generate_new_bank_forks` should have the consistent value.

AshwinSekar (commented):
looking good, almost there, just a couple of comments

core/src/block_creation_loop.rs
ksn6: Yep, good catch - I'll capture whether the window has moved on and assert `window_has_moved_on || optimistic_parent.is_none()` instead.

core/src/replay_stage.rs
ksn6: yep, makes sense - would rather we do the block ID checks in your PR; this one's getting long haha

core/src/replay_stage.rs
ksn6: Added in a continue + a comment explaining this case.

core/src/replay_stage.rs
```diff
@@ -4752,13 +4787,51 @@ impl ReplayStage {
...
+                let (actual_parent_bank, replay_offset) =
+                    if !migration_status.should_allow_block_markers(child_slot) {
+                        (parent_bank.clone(), None)
+                    } else {
+                        let Some(parent_meta) = blockstore
+                            .get_parent_meta(child_slot, BlockLocation::Original)
+                            .expect("Blockstore should not fail")
+                        else {
+                            // Shred 0 shouldn't be present, since parent meta isn't present.
+                            debug_assert!(!blockstore.is_full(child_slot));
```
ksn6: I'm removing this assert, since it's technically possible for a TOCTOU to occur with the check in L4799 (in the very rare case that shred 0 is inserted between L4799 and L4803).

core/src/replay_stage.rs
```diff
@@ -4791,21 +4791,26 @@ impl ReplayStage {
...
+                        // No lock is held between get_slots_since and get_parent_meta, so
+                        // ParentMeta could have been updated in between. If so, continue and
+                        // the next iteration of generate_new_bank_forks will have consistent values.
+                        if parent_meta.parent_slot != parent_slot {
+                            continue;
+                        }
 
-                    if parent_meta.populated_from_update_parent() {
                         // UpdateParent: use new parent and calculate replay offset
                         let Some(new_parent) = frozen_banks.get(&parent_meta.parent_slot) else {
```
AshwinSekar: now that you added the `continue` we don't need this anymore right? since this will just be `parent_bank` always (and it's implicitly frozen by construction). Think all this closure should return is `num_shreds`.

PR #646, opened 2025-12-24
repair: add DMR block_id requests

AshwinSekar:
#### Problem
We removed the CMR based requests but need to add the DMR based ones

#### Summary of Changes
Screenshot 2025-12-05 at 1 24 40 PM

- add `ParentAndFecSetCount` equivalent of `getSliceCount`, rather than returning the final FEC set root it returns the parent. Reponse type is (fec_set_count, ParentMeta, proof that ParentMeta -> DMR)
- add `FecSetRoot` equilvalent of `getSliceHash` -> Response type is (fec_set_root, proof that fec_set_root -> DMR)
- `get_shred` was added in the previous PR

### DIFF
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -372,7 +372,7 @@ type PingCache = ping_pong::PingCache<REPAIR_PING_TOKEN_SIZE>;
 
 /// Window protocol messages
 /// Appending new messages here is safe as long as it is feature gated.
-/// Changing the format of an existing message is posisble but not advised.
+/// Changing the format of an existing message is possible but not advised.
 /// Removing a message is possible by first removing the sender and feature gating the response.
 /// The message can then be removed once the feature gate is active and there are no responders.
 #[cfg_attr(


core/src/repair/serve_repair.rs
```diff
@@ -205,6 +210,94 @@ impl RequestResponse for AncestorHashesRepairType {
...
+    #[allow(dead_code)]
+    ParentAndFecSetCount { slot: Slot, block_id: Hash },
+
+    #[allow(dead_code)]
+    FecSetRoot {
+        slot: Slot,
+        block_id: Hash,
+        fec_set_index: u32,
+    },
+}
```
AshwinSekar, line 225: Similar to ancestor hashes this will run via a different service / socket in next PR.

core/src/repair/serve_repair.rs
```diff
@@ -205,6 +210,94 @@ impl RequestResponse for AncestorHashesRepairType {
...
+    type Response = BlockIdRepairResponse;
+
+    fn num_expected_responses(&self) -> u32 {
+        // The only variable in the response are the merkle proofs.
+        // Even assuming an extreme of u64::MAX fec sets, that corresponds to
+        // 63 * 20 = 1260 bytes for a proof which fits in one packet.
+        1
+    }
+
+    fn verify_response(&self, response: &Self::Response) -> bool {
```
AshwinSekar, line 251: recap of how repair verification works:
- repair requests are nonced
- requester will add to OutstandingRequests
- when received we validate against the nonce
- And lookup the OutstandingRequest to validate against the requests

essentially this code will run before any meaningful action is taken, so we're guaranteed that the merkle aspects of the response check out.

AshwinSekar: similar for the already implemented `getShred`:
https://github.com/anza-xyz/alpenglow/blob/572e25685c2cf928361b1e9e5ca0f8afd9265cca/core/src/repair/serve_repair.rs#L167-L176

however this goes through the "main" repair pathway as it needs to be sigverified

ledger/src/shred/merkle_tree.rs
```diff
@@ -124,13 +124,25 @@ where
         .ok_or(Error::InvalidMerkleProof)
 }
 
+/// Given a flattened merkle `proof` for `node` at `index`,
+/// verify the proof against merkle root `root`
+pub fn verify_merkle_proof(node: Hash, index: usize, proof: &[u8], expected_root: Hash) -> Result<bool, Error> {
```
bw-solana: a little awkward returning `Result` wrapping a `bool`.

Could we just return `Result` and have the false case be an error? Or do we expect callers to handle these cases separately?

If the latter, might be nice to call out the differences in function header here (invalid merkle proof vs. mismatched against expected)

core/src/repair/serve_repair.rs
```diff
@@ -583,18 +749,25 @@ impl ServeRepair {
...
+                            *shred_index,
+                            *block_id,
+                            *nonce,
+                        );
+                        if batch.is_none() {
+                            stats.window_index_for_block_id_misses += 1;
+                        }
+                        (batch, "WindowIndexForBlockIdWithNonce")
+                    } else {
+                        (None, "WindowIndexForBlockIdWithNonce")
```
bw-solana: nit: could consider tiny dedup by doing something like
```rust
let batch = if migration status ...
} else {
    None
};

(batch, "WindowIndexForBlockIdWithNonce")
```

core/src/repair/serve_repair.rs
```diff
@@ -270,7 +367,7 @@ type PingCache = ping_pong::PingCache<REPAIR_PING_TOKEN_SIZE>;
 #[cfg_attr(
     feature = "frozen-abi",
     derive(AbiEnumVisitor, AbiExample),
-    frozen_abi(digest = "H4x6ESrXXSWgh9M5Kw3ryDww2okbEUW36V9BJUVZU17V")
+    frozen_abi(digest = "7XVfjbxcWhiwubU3BkE5QPT6qQLkKBraYGhAHZKWx6UZ")
```
bw-solana, line 381: Is the plan that serdes will "just work" given we're appending only?

core/src/repair/serve_repair.rs
```diff
@@ -205,6 +210,94 @@ impl RequestResponse for AncestorHashesRepairType {
...
+                    fec_set_count,
+                    parent_info: (parent_slot, parent_block_id),
+                    parent_proof,
+                },
+            ) => {
+                if *fec_set_count > MAX_FEC_SETS_PER_SLOT {
+                    return false;
+                }
+
+                let proof_size = merkle_tree::get_proof_size(fec_set_count + 1);
```
bw-solana, line 269: why +1?

core/src/repair/serve_repair.rs
```diff
@@ -205,6 +210,94 @@ impl RequestResponse for AncestorHashesRepairType {
...
+                if parent_proof.len() != proof_size as usize * SIZE_OF_MERKLE_PROOF_ENTRY {
+                    return false;
+                }
+
+                let node = hashv(&[&parent_slot.to_le_bytes(), parent_block_id.as_ref()]);
```
bw-solana: we're using node in quite a few places. I'm going back and forth on if we should be more verbose here to make it clear this is a node in the context of the merkle tree (as opposed to the other 100 definitions of node we have)

core/src/repair/repair_handler.rs
```diff
@@ -172,6 +172,121 @@ pub trait RepairHandler {
...
+            .meta_from_location(slot, location)
+            .unwrap()
+            .unwrap()
+            .is_full());
+
+        let double_merkle_meta = self
+            .blockstore()
+            .get_double_merkle_meta(slot, location)
+            .expect("Unable to fetch double merkle meta")
+            .expect("If location exists, double merkle meta must be populated");
```
bw-solana, line 198: do we insert these together as a transaction into blockstore? Just wondering if there is any possible race here

core/src/repair/repair_handler.rs
```diff
@@ -172,6 +172,121 @@ pub trait RepairHandler {
...
+        )
+        .expect("Packet construction cannot fail");
+
+        Some(PacketBatch::from(
+            PinnedPacketBatch::new_unpinned_with_recycler_data(
+                recycler,
+                "run_parent_fec_set_count",
+                vec![packet],
+            ),
+        ))
```
bw-solana: we duplicate this code a bunch of times. Would be nice to rethink the abstraction a bit at some point

core/src/repair/serve_repair.rs
AshwinSekar: yep, I added a comment to clear this up. We're safe here because the usage of these new messages is feature gated https://github.com/anza-xyz/alpenglow/pull/646/commits/7b9ee584aab98ecf2f8d285775cf563c8e16c507

core/src/repair/serve_repair.rs
AshwinSekar: it's because we have an additional leaf (the parent info). I've added a comment https://github.com/anza-xyz/alpenglow/pull/646/commits/7d5a3ebba32a92625337d3b748df752d5b43aefa

ledger/src/shred/merkle_tree.rs
AshwinSekar: we ignore the exact error anyway and convert to false. Just made it return a Result instead https://github.com/anza-xyz/alpenglow/pull/646/commits/c0e710e99c1101cc58b6c88c4847a1828f2070d6

core/src/repair/serve_repair.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/646/commits/1e339af20dc6d3df536c660e0c3c80b6b020a237

core/src/repair/serve_repair.rs
AshwinSekar: that's a fair point renamed

core/src/repair/repair_handler.rs
AshwinSekar: no race here, the location is entirely based on the double merkle meta column now (no more BlockVersions column). So it's impossible for location to exist without the double merkle meta column being populated.

core/src/repair/repair_handler.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/646/commits/adef7f7f93480091c9beb360650de7864d69fa1c

bw-solana (approved):

PR #647, opened 2025-12-24
block component: new constructors simplify component creation

ksn6:
Introducing a few new constructors to `VersionedBlockMarker` to simplify component creation.

### DIFF
--- a/poh/src/poh_recorder.rs
+++ b/poh/src/poh_recorder.rs
@@ -598,7 +598,7 @@ impl PohRecorder {
         let mut footer = footer.clone();
         footer.bank_hash = working_bank.bank.hash();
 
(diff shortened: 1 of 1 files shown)


poh/src/poh_recorder.rs
```diff
@@ -600,9 +598,7 @@ impl PohRecorder {
         let mut footer = footer.clone();
         footer.bank_hash = working_bank.bank.hash();
 
-        let footer = VersionedBlockFooter::V1(footer.clone());
-        let footer = BlockMarkerV1::new_block_footer(footer);
-        let footer = VersionedBlockMarker::V1(footer);
+        let footer = VersionedBlockMarker::new_block_footer(footer.clone());
```
wen-coding: nit: do you still need the .clone() here?

PR #648, opened 2025-12-25
migration: verify genesis certificate block marker

### LINKED ISSUE #609: Verify GenesisCertificate in blockstore processor
#### Problem
Right now if the genesis certificate is ingested from A2A it's verified in bls_sigverifier.
However if it is ingested via a block marker it is not verified.

#### Proposed Solution
Once the verification functions are moved outside of core, update the replay pathway to verify the certificate.

AshwinSekar:
#### Problem
The GenesisCertificate block marker is not actually verified. This means anyone malicious can send an alpenglow block w/ the marker and force the migration to succeed 😅 

#### Summary of Changes
Actually verify the certificate before accepting the block and enabling alpenglow.

Fixes #609

### DIFF
--- a/runtime/src/block_component_processor.rs
+++ b/runtime/src/block_component_processor.rs
@@ -218,10 +218,21 @@ impl BlockComponentProcessor {
                 .get_pubkey_and_stake(rank)
                 .map(|(_, bls_pubkey, stake)| (*bls_pubkey, *stake))
         })
-        .map_err(|_| BlockComponentProcessorError::GenesisCertificateFailedVerification)?;
+        .map_err(|_| {
+            warn!(
+                "Failed to verify genesis certificate for {slot} in bank slot {}",
+                bank.slot()
+            );
+            BlockComponentProcessorError::GenesisCertificateFailedVerification
+        })?;
 
         let genesis_percent = Fraction::new(genesis_stake, NonZeroU64::new(total_stake).unwrap());
         if genesis_percent < GENESIS_VOTE_THRESHOLD {
+            warn!(
+                "Recieived a genesis certificate for {slot} in bank slot {} with \
+                 {genesis_percent} stake < {GENESIS_VOTE_THRESHOLD}",
+                bank.slot()
+            );
             return Err(BlockComponentProcessorError::GenesisCertificateFailedVerification);
         }
 


runtime/src/block_component_processor.rs
```diff
@@ -194,6 +200,35 @@ impl BlockComponentProcessor {
         Ok(())
     }
 
+    fn verify_genesis_certificate(
```
AshwinSekar, line 203: just a copy paste from the bls sigverifier code:

https://github.com/anza-xyz/alpenglow/blob/572e25685c2cf928361b1e9e5ca0f8afd9265cca/core/src/bls_sigverify/bls_sigverifier.rs#L50-L54

https://github.com/anza-xyz/alpenglow/blob/572e25685c2cf928361b1e9e5ca0f8afd9265cca/core/src/bls_sigverify/bls_sigverifier.rs#L508-L515

runtime/src/block_component_processor.rs
```diff
@@ -194,6 +200,35 @@ impl BlockComponentProcessor {
...
+        let epoch = bank.epoch_schedule().get_epoch(slot);
+        let epoch_stakes = bank
+            .epoch_stakes_map()
+            .get(&epoch)
+            .ok_or(BlockComponentProcessorError::GenesisCertificateFailedVerification)?;
+        let key_to_rank_map = epoch_stakes.bls_pubkey_to_rank_map();
+        let total_stake = epoch_stakes.total_stake();
+
+        let genesis_stake =
+            cert_verify::verify_votor_message_certificate(cert, key_to_rank_map.len(), |rank| {
```
bw-solana: can just import `verify_votor_message_certificate`

runtime/src/block_component_processor.rs
```diff
@@ -194,6 +200,35 @@ impl BlockComponentProcessor {
...
+                key_to_rank_map
+                    .get_pubkey_and_stake(rank)
+                    .map(|(_, bls_pubkey, stake)| (*bls_pubkey, *stake))
+            })
+            .map_err(|_| BlockComponentProcessorError::GenesisCertificateFailedVerification)?;
+
+        let genesis_percent = Fraction::new(genesis_stake, NonZeroU64::new(total_stake).unwrap());
```
bw-solana, line 229: should we just check `total_stake` is non-zero and return error here?

AshwinSekar: I did think about it, but if `total_stake` is `0` we're already screwed. Probably `epoch_stakes.total_stake()` should be returning a `NonZero`.

runtime/src/block_component_processor.rs
```diff
@@ -21,6 +26,8 @@ pub enum BlockComponentProcessorError {
     GenesisCertificateInAlpenglowCluster,
     #[error("GenesisCertificate marker detected on a block which is not a child of genesis")]
     GenesisCertificateOnNonChild,
+    #[error("GenesisCertificate was invalid and failed to verify")]
+    GenesisCertificateFailedVerification,
```
wen-coding, line 30: nit: maybe we should give detailed reason why verification failed?

runtime/src/block_component_processor.rs
wen-coding: Do you think I should merge common code into bls_sigverify module?

runtime/src/block_component_processor.rs
```diff
@@ -194,6 +200,34 @@ impl BlockComponentProcessor {
...
+            .ok_or(BlockComponentProcessorError::GenesisCertificateFailedVerification)?;
+        let key_to_rank_map = epoch_stakes.bls_pubkey_to_rank_map();
+        let total_stake = epoch_stakes.total_stake();
+
+        let genesis_stake = verify_votor_message_certificate(cert, key_to_rank_map.len(), |rank| {
+            key_to_rank_map
+                .get_pubkey_and_stake(rank)
+                .map(|(_, bls_pubkey, stake)| (*bls_pubkey, *stake))
+        })
+        .map_err(|_| BlockComponentProcessorError::GenesisCertificateFailedVerification)?;
```
wen-coding: nit: if we choose to discard the error, we probably should just warn! here.

runtime/src/block_component_processor.rs
```diff
@@ -194,6 +200,34 @@ impl BlockComponentProcessor {
...
+
+        let genesis_percent = Fraction::new(genesis_stake, NonZeroU64::new(total_stake).unwrap());
+        if genesis_percent < GENESIS_VOTE_THRESHOLD {
```
wen-coding, line 230: nit: can we add a warn! here and print the genesis_percent we calculated for debugging?

runtime/src/block_component_processor.rs
AshwinSekar: If it's easy go for it, but not a priority

runtime/src/block_component_processor.rs
AshwinSekar: I kept it as one Error but added warn! logs like you suggested

runtime/src/block_component_processor.rs
wen-coding: That's fine, you can submit as is and we can refactor later.

PR #649, opened 2026-01-05
Populate finalization certs in block footer.

wen-coding:
#### Problem
Per SIMD 402, populate finalization certs in block footer.

#### Summary of Changes
- Rework highest_finalized_with_notarized so it now tracks certs in stead of slot and fast_finalize
- Pass highest_finalized information from consensus pool to block_creation_loop
- Also add a constructor for FinalCertificate in block_component so it converts given finalization certs to the format fit for block footer

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13052,6 +13052,7 @@ dependencies = [
  "futures-core",
  "futures-io",
  "futures-sink",
+ "futures-util",
  "pin-project-lite",
  "tokio",
 ]
--- a/bench-streamer/src/main.rs
+++ b/bench-streamer/src/main.rs
@@ -140,7 +140,7 @@ fn main() -> Result<()> {
                 packet_sender,
                 recycler.clone(),
                 stats.clone(),
-                Some(Duration::from_millis(1)), // coalesce
+                None, // coalesce
                 true,
                 None,
                 false,
--- a/bench-vote/src/main.rs
+++ b/bench-vote/src/main.rs
@@ -46,7 +46,6 @@ static GLOBAL: jemallocator::Jemalloc = jemallocator::Jemalloc;
 const SINK_REPORT_INTERVAL: Duration = Duration::from_secs(5);
 const SINK_RECEIVE_TIMEOUT: Duration = Duration::from_secs(1);
 const SOCKET_RECEIVE_TIMEOUT: Duration = Duration::from_secs(1);
-const COALESCE_TIME: Option<Duration> = Some(Duration::from_millis(1));
 
 fn sink(
     exit: Arc<AtomicBool>,
@@ -300,10 +299,10 @@ fn main() -> Result<()> {
                     s_reader,
                     recycler.clone(),
                     stats.clone(),
-                    COALESCE_TIME, // coalesce
-                    true,          // use_pinned_memory
-                    None,          // in_vote_only_mode
-                    false,         // is_staked_service
+                    None,  // coalesce
+                    true,  // use_pinned_memory
+                    None,  // in_vote_only_mode
+                    false, // is_staked_service
                 ));
             }
         }
--- a/core/src/tpu.rs
+++ b/core/src/tpu.rs
@@ -1,9 +1,7 @@
 //! The `tpu` module implements the Transaction Processing Unit, a
 //! multi-stage transaction processing pipeline in software.
 
-pub use {
-    crate::forwarding_stage::ForwardingClientOption, solana_streamer::quic::DEFAULT_TPU_COALESCE,
-};
+pub use crate::forwarding_stage::ForwardingClientOption;
 use {
     crate::{
         admin_rpc_post_init::{KeyUpdaterType, KeyUpdaters},
@@ -138,7 +136,6 @@ impl Tpu {
         replay_vote_receiver: ReplayVoteReceiver,
         replay_vote_sender: ReplayVoteSender,
         bank_notification_sender: Option<BankNotificationSenderConfig>,
-        tpu_coalesce: Duration,
         duplicate_confirmed_slot_sender: DuplicateConfirmedSlotsSender,
         client: ForwardingClientOption,
         turbine_quic_endpoint_sender: AsyncSender<(SocketAddr, Bytes)>,
@@ -188,7 +185,7 @@ impl Tpu {
             &forwarded_packet_sender,
             forwarded_packet_receiver,
             poh_recorder,
-            Some(tpu_coalesce),
+            None, // coalesce
             Some(bank_forks.read().unwrap().get_vote_only_mode_signal()),
             tpu_enable_udp,
         );
@@ -277,7 +274,6 @@ impl Tpu {
             let adapter = VortexorReceiverAdapter::new(
                 sockets,
                 Duration::from_millis(5),
-                tpu_coalesce,
                 non_vote_sender,
                 enable_block_production_forwarding.then(|| forward_stage_sender.clone()),
                 exit.clone(),
--- a/core/src/validator.rs
+++ b/core/src/validator.rs
@@ -28,7 +28,7 @@ use {
         system_monitor_service::{
             verify_net_stats_access, SystemMonitorService, SystemMonitorStatsReportConfig,
         },
-        tpu::{ForwardingClientOption, Tpu, TpuSockets, DEFAULT_TPU_COALESCE},
+        tpu::{ForwardingClientOption, Tpu, TpuSockets},
         tvu::{Tvu, TvuConfig, TvuSockets},
     },
     anyhow::{anyhow, Context, Result},
@@ -315,7 +315,6 @@ pub struct ValidatorConfig {
     pub warp_slot: Option<Slot>,
     pub accounts_db_skip_shrink: bool,
     pub accounts_db_force_initial_clean: bool,
-    pub tpu_coalesce: Duration,
     pub staked_nodes_overrides: Arc<RwLock<HashMap<Pubkey, u64>>>,
     pub validator_exit: Arc<RwLock<Exit>>,
     pub validator_exit_backpressure: HashMap<String, Arc<AtomicBool>>,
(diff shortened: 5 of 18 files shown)


votor/src/consensus_pool.rs
```diff
@@ -307,22 +327,30 @@ impl ConsensusPool {
...
+                    if self.is_highest_finalized_too_old(slot, false) {
+                        let mut highest_finalized = self.highest_finalized.write().unwrap();
+                        if let Some(finalization_cert) = self
+                            .completed_certificates
+                            .get(&CertificateType::Finalize(slot))
+                        {
+                            if let CertificateType::FinalizeFast(_, _) = finalization_cert.cert_type
+                            {
+                                *highest_finalized = (Some(finalization_cert.clone()), None);
+                            } else {
```
ksn6: This doesn't look right; we're doing `self.completed_certificates.get(&CertificateType::Finalize(slot))` just before this. Accordingly, `finalization_cert.cert_type` cannot be `CertificateType::FinalizeFast` here.

votor/src/consensus_pool.rs
```diff
@@ -307,22 +327,30 @@ impl ConsensusPool {
                     // It's fine to set FastFinalization to false here, because
                     // we will report correctly as long as we have FastFinalization cert.
                     events.push(VotorEvent::Finalized((slot, block_id), false));
-                    if self
-                        .highest_finalized_with_notarize
-                        .is_none_or(|(s, _)| s < slot)
-                    {
-                        self.highest_finalized_with_notarize = Some((slot, false));
+                    if self.is_highest_finalized_too_old(slot, false) {
+                        let mut highest_finalized = self.highest_finalized.write().unwrap();
```
ksn6, line 342: I think there's a race condition here - some external thread could modify `self.highest_finalized` in between:

- L330: `self.is_highest_finalized_too_old`, where a `read()` happens
- L331: the invocation to `self.highest_finalized.write().unwrap()`

To fix this, e.g., what if we grabbed a write lock, passed that into `self.is_highest_finalized_too_old` for its operations, and then used the same write lock in L331?

votor/src/consensus_pool.rs
```diff
@@ -488,12 +514,11 @@ impl ConsensusPool {
         Ok(vec![cert])
     }
 
-    /// Get the notarized block in `slot`
-    pub fn get_notarized_block(&self, slot: Slot) -> Option<Block> {
+    pub fn get_notarization_cert(&self, slot: Slot) -> Option<Arc<Certificate>> {
```
ksn6, line 531: nit: reintroduce the comment

core/src/block_creation_loop.rs
```diff
@@ -409,14 +420,32 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
 }
 
+fn produce_final_certificate(
+    highest_finalized: &RwLock<FinalizationCertPair>,
+) -> Option<FinalCertificate> {
+    let (finalization_cert, notarize_cert) = highest_finalized.read().unwrap().clone();
+    let Some(finalization_cert) = finalization_cert else {
+        if notarize_cert.is_some() {
+            warn!("Finalized certificate is None but notarize certificate is Some");
+        }
```
ksn6:
```suggestion
        if notarize_cert.is_some() {
            warn!(
                "Inconsistent state: finalized cert is None but notarize cert is Some (slot: {:?})",
                notarize_cert.as_ref().map(|c| c.cert_type.slot())
            );
        }
```

core/src/block_creation_loop.rs
```diff
@@ -409,14 +420,32 @@ fn produce_block_footer(bank: Arc<Bank>) -> BlockFooterV1 {
...
         // TODO(akhi): include reward certs
         skip_reward_cert: None,
         notar_reward_cert: None,
     }
 }
 
+fn produce_final_certificate(
+    highest_finalized: &RwLock<FinalizationCertPair>,
+) -> Option<FinalCertificate> {
+    let (finalization_cert, notarize_cert) = highest_finalized.read().unwrap().clone();
```
ksn6: Can probably do something like:

```
let (finalization_cert, notarize_cert) = {
    let (finalization_cert, notarize_cert) = highest_finalized.read().unwrap().clone();
    (finalization_cert.clone(), notarize_cert.clone())
};
```

Cloning `Arc` is cheap, and this allows us to not have to hold the lock for the rest of the function.

votor/src/consensus_pool.rs
wen-coding: In the current setup, consensus_pool_service thread is the only thread allowed to write this variable, should I add a comment there?

core/src/block_creation_loop.rs
wen-coding: I believe in
` let (finalization_cert, notarize_cert) = highest_finalized.read().unwrap().clone();`

The lock is released as soon as .clone() is done. And as you have said, cloning (Option<Arc>, Option<Arc>) is cheap.

core/src/block_creation_loop.rs
wen-coding: Done.

votor/src/consensus_pool.rs
wen-coding: good catch, changed.

votor/src/consensus_pool.rs
wen-coding: Done.

core/src/block_creation_loop.rs
```diff
@@ -98,6 +102,7 @@ struct LeaderContext {
     my_pubkey: Pubkey,
     leader_window_info_receiver: Receiver<LeaderWindowInfo>,
     highest_parent_ready: Arc<RwLock<(Slot, (Slot, Hash))>>,
```
AshwinSekar, line 116: nit: if it's easier I would suggest just combining these into one structure `shared_cert_pool_info` or something, and then we can add convenience methods on the struct to get the info we need

votor-messages/src/consensus_message.rs
```diff
@@ -5,8 +5,12 @@ use {
     solana_bls_signatures::Signature as BLSSignature,
     solana_clock::Slot,
     solana_hash::Hash,
+    std::sync::Arc,
 };
 
+/// Define the finalization cert pair type
+pub type FinalizationCertPair = (Option<Arc<Certificate>>, Option<Arc<Certificate>>);
```
AshwinSekar: I would use a better type here to avoid having to deal with inconsistent state. I think an enum where we have `FastFinalize`  with one cert or `SlowFinalize` with both certs.

wen-coding: Done.

core/src/block_creation_loop.rs
wen-coding: Hmm, I assume you mean combining `highest_parent_ready` and `highest_frozen`, the former is currently updated in event loop and the latter in consensus pool. Or do you want me to wrap all the receivers and notifiers and shared variables into one context for `block_creation_loop` only?

AshwinSekar: yeah I was thinking combining `highest_parent_ready` and `highest_finalized` into one structure. If it's more trouble then it's worht then it's fine as you have it now

entry/src/block_component.rs
```diff
@@ -333,31 +333,33 @@ pub struct FinalCertificate {
...
-                if let CertificateType::Finalize(finalize_slot) = final_cert.cert_type {
-                    if slot != finalize_slot {
-                        return Err(format!(
-                            "notarization slot {slot} does not match finalization slot \
-                             {finalize_slot}",
-                        ));
+    pub fn new_from_certificate(final_certs: FinalizationCerts) -> Result<Option<Self>, String> {
+        let (slot, block_id, final_cert, notar_cert) = match final_certs {
+            FinalizationCerts::Finalize(final_cert, notar_cert) => {
+                if let CertificateType::Notarize(slot, notar_block_id) = notar_cert.cert_type {
```
AshwinSekar, line 342: I think we be stronger here and panic if these are not finalization and notarization certs, as it indicates programmer error.

votor/src/consensus_pool.rs
ksn6: Hmm, still, would likely be cleaner to do something like this, which would obviate this problem even if somehow in the future, another thread were to be allowed to write to this variable.

> To fix this, e.g., what if we grabbed a write lock, passed that into self.is_highest_finalized_too_old for its operations, and then used the same write lock in L331?

votor-messages/src/consensus_message.rs
```diff
@@ -5,8 +5,21 @@ use {
...
     solana_clock::Slot,
     solana_hash::Hash,
+    std::sync::Arc,
 };
 
+/// Define the finalization certs type
+#[derive(Default, PartialEq, Debug, Clone)]
+pub enum FinalizationCerts {
+    /// Finalize certificate and Notarization certificate
+    Finalize(Arc<Certificate>, Arc<Certificate>),
```
ksn6: Would strongly prefer having an inner type here to prevent constructing / parsing these in the wrong order.

core/src/block_creation_loop.rs
wen-coding: It's definitely doable, dunno how much tidier it would be, but let's maybe try it in another PR, this PR is already big enough.

votor-messages/src/consensus_message.rs
wen-coding: Done.

entry/src/block_component.rs
wen-coding: Hmm, actually I already did that, but when I was fixing tests, I realized:
1. there are other cases where creating finalization cert may fail, for example, a bad signature
2. final certificates are best-effort and optional, so I'm not sure I want to kill the validator if the raw data is somehow wrong. Let's say a bit corrupted in your cert storage and you now have a bad cert on starting up, we could check and warn, but I'm not sure it's worth panic for.
Of course, I can do that if you insist on it.

wen-coding: I changed warn! to error! in block_creation_loop to make it more obvious.

AshwinSekar: hmm I'm not sure we should tolerate any sort of failure here. TBC I think it's fine to have this return a `Result` but in BCL we should `expect`. Looking at all of the failures in this fn they all look like programmer errors.

wen-coding: I have some concerns about whether we will introduce a global outage, let's say some signature sneaked in but somehow can't compress due to a bug in new code. But okay....
Done

AshwinSekar: hmm this would indicate a bug in the bls verification code right?

If you're super worried I think we can split it up so that the errors in the first half of this function  (above line 376) are unwrapped as a panic (since it's a programmer error) and we treat the signature errors as error! logs instead?

PR #651, opened 2026-01-05
fixes various problems in `block_creation_loop/stats.rs`

akhi3030:
#### Problems and fixes

- `BlockCreationLoopMetrics` should not derive `Default` as that will set `last_report` to 0 which is invalid.
  - Provide a custom implementation of `Default` that sets `last_report` properly.
- `SlotMetrics::slot` was never actually getting set so we were always reporting 0.
  - `slot` is passed in `report` and not stored in the struct.  This also makes deriving `Default` on the struct more sensible.
- When resetting the two structs, use `Default` instead of duplicating code
- Change the visibility of structs and methods to be more restrictive
- Finally, use struct deconstruction in various places so that when adding new fields, the compiler will help ensure that we update all the relevant methods


core/src/block_creation_loop/stats.rs
```diff
@@ -1,174 +1,182 @@
...
                 ),
             );
 
             // reset metrics
-            self.bank_timeout_completion_count = 0;
-            self.window_production_elapsed = 0;
-            self.skipped_window_behind_parent_ready_count = 0;
-            self.bank_timeout_completion_elapsed_hist.clear();
-            self.last_report = now;
+            *self = Self::default()
```
ksn6: This sets `self.loop_count = 0;` whereas before, this wasn't the case. Is this intentional?

cc: @AshwinSekar

akhi3030: good catch, this was not intentional and a bug.  I introduced a new `reset` method to address this in a hopefully clean way.

core/src/block_creation_loop/stats.rs
```diff
@@ -1,174 +1,187 @@
...
 
-#[derive(Default)]
-pub(crate) struct BlockCreationLoopMetrics {
-    pub(crate) last_report: u64,
-    pub(crate) loop_count: u64,
-    pub(crate) bank_timeout_completion_count: u64,
-    pub(crate) skipped_window_behind_parent_ready_count: u64,
+pub(super) struct BlockCreationLoopMetrics {
+    last_report: u64,
+    pub(super) loop_count: u64,
```
ksn6, line 10: Should probably toss in a comment somewhere indicating that this is special.

core/src/block_creation_loop/stats.rs
```diff
@@ -1,174 +1,187 @@
...
+        } = self;
+        0 == *loop_count
+            + *bank_timeout_completion_count
+            + *window_production_elapsed
+            + *skipped_window_behind_parent_ready_count
+            + bank_timeout_completion_elapsed_hist.entries()
     }
 
-    pub(crate) fn report(&mut self, report_interval_ms: u64) {
+    fn reset(&mut self) {
```
ksn6, line 47: Could we toss in a test ensuring that loop count doesn't change after invoking `reset()`?

core/src/block_creation_loop/stats.rs
akhi3030: 👍

core/src/block_creation_loop/stats.rs
akhi3030: Good idea, added one.

PR #652, opened 2026-01-06
Hooks up the `ConsensusRewards` to the bls sigverifier

akhi3030:
### Problem

We need to hook up the consensus rewards container to the bls sigverifier so that it can actually receive votes which it will use to build the reward certs.

- https://github.com/anza-xyz/alpenglow/pull/637 introduced the reward certs
- https://github.com/anza-xyz/alpenglow/pull/639 introduced the container

### Summary of Changes

- bls sigverifier checks which votes the `ConsensusReward` container wants and then forwards them to it
- tvu; votor; etc. are updated to actually instantiate `ConsensusReward`
- **Note**: the `ConsensusReward` container is not yet wired up to the block creation loop

### DIFF
--- a/core/src/tvu.rs
+++ b/core/src/tvu.rs
@@ -272,8 +272,8 @@ impl Tvu {
         )
         .unwrap();
 
-        // At the moment there are roughly 1K validators and the sigverifier receives votes in batches and sends them to the consensus reward container in batches so hopefully using a channel of 1K slots would never block.
-        let (reward_votes_sender, reward_votes_receiver) = bounded(1000);
+        // At the moment there are roughly 1K validators and the sigverifier receives votes in batches and sends them to the consensus reward container in batches so hopefully using a channel of 2K slots would never block.
+        let (reward_votes_sender, reward_votes_receiver) = bounded(2000);
         let alpenglow_sigverify_service = {
             let sharable_banks = bank_forks.read().unwrap().sharable_banks();
             let verifier = BLSSigVerifier::new(


votor/src/votor.rs
```diff
@@ -240,15 +244,29 @@ impl Votor {
         let event_handler = EventHandler::new(event_handler_context);
         let consensus_pool_service = ConsensusPoolService::new(consensus_pool_context);
 
+        let (_build_reward_certs_sender, build_reward_certs_receiver) = unbounded();
```
AshwinSekar: these will need to be instantiated in validator.rs and passed in so that they can be used in BCL.
Also we should keep these channels bounded and log when the channel is full

core/src/tvu.rs
```diff
@@ -268,14 +268,18 @@ impl Tvu {
             alpenglow_quic_server_config,
         )
         .unwrap();
+        let (reward_votes_sender, reward_votes_receiver) = unbounded();
```
AshwinSekar: FYI we want all production channels to be bounded to avoid allocations.

votor/src/votor.rs
AshwinSekar: i suggest passing in the other side of these channels into BCL (and ignoring them) in this PR to keep the plumbing in once place

core/src/tvu.rs
akhi3030: makes sense... added a bound and an explanation for how it was picked.

votor/src/votor.rs
akhi3030: updated the plumbing.

core/src/tvu.rs
```diff
@@ -268,7 +271,9 @@ impl Tvu {
             alpenglow_quic_server_config,
         )
         .unwrap();
-        let (reward_votes_sender, reward_votes_receiver) = unbounded();
+
+        // At the moment there are roughly 1K validators and the sigverifier receives votes in batches and sends them to the consensus reward container in batches so hopefully using a channel of 1K slots would never block.
+        let (reward_votes_sender, reward_votes_receiver) = bounded(1000);
```
AshwinSekar: think we should make this at least 2000 to accommodate future growth. Under alpenglow the validator set is capped at 2k

akhi3030: sure... updated.

AshwinSekar (approved):

PR #653, opened 2026-01-06
blockstore: avoid parent propagation through incomplete slots

ksn6:
We need to avoid parent propagation unless slots are complete.

This PR addresses the following comments:

https://github.com/anza-xyz/alpenglow/pull/622#discussion_r2641051928
https://github.com/anza-xyz/alpenglow/pull/622#discussion_r2641066934
https://github.com/anza-xyz/alpenglow/pull/622#discussion_r2639396914
https://github.com/anza-xyz/alpenglow/pull/622#discussion_r2639443623

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -5460,12 +5460,14 @@ impl Blockstore {
             // Propagate or clear connectivity based on new parent's state.
             if new_parent_meta.borrow().is_connected() {
                 if !slot_meta.borrow().is_parent_connected() {
-                    slot_meta.borrow_mut().set_parent_connected();
-                    self.propagate_parent_connected_to_children(
-                        &slot_meta,
-                        working_set,
-                        new_chained_slots,
-                    )?;
+                    let became_connected = slot_meta.borrow_mut().set_parent_connected();
+                    if became_connected {
+                        self.propagate_parent_connected_to_children(
+                            &slot_meta,
+                            working_set,
+                            new_chained_slots,
+                        )?;
+                    }
                 }
             } else if slot_meta.borrow_mut().clear_parent_connected() {
                 self.traverse_children_mut(


ledger/src/blockstore.rs
```diff
@@ -5448,15 +5462,12 @@ impl Blockstore {
 
             // Propagate or clear connectivity based on new parent's state.
             if new_parent_meta.borrow().is_connected() {
-                if !slot_meta.borrow().is_parent_connected() {
```
AshwinSekar, line 5451: do we still want this if guard here?
is it possible for the slot meta to already be connected and then we traverse for no reason?

ledger/src/blockstore.rs
```diff
@@ -5384,8 +5384,22 @@ impl Blockstore {
...
+    }
+
+    /// If `slot_meta` is connected, propagate `parent_connected` to its children.
+    fn propagate_parent_connected_to_children(
+        &self,
+        slot_meta: &Rc<RefCell<SlotMeta>>,
+        working_set: &HashMap<(BlockLocation, u64), SlotMetaWorkingSetEntry>,
+        new_chained_slots: &mut HashMap<u64, Rc<RefCell<SlotMeta>>>,
+    ) -> Result<()> {
+        if slot_meta.borrow().is_connected() {
```
AshwinSekar: I think it's cleaner if this if guard is a part of the caller instead. In the usage below (line 5466) we already set parent connected before calling this function.
Could make this a debug_assert and note this in the fucntion doc instead.

ledger/src/blockstore.rs
ksn6: yeah, sounds reasonable; was trying to clean this logic up a bit, though it's fair that we lose a bit of perf with this change - will revert

PR #655, opened 2026-01-07
Switch Alpenglow QUIC streamer to spawn_server_with_cancel

wen-coding:
#### Problem
We need #8836 to enable BLS blacklisting

#### Summary of Changes
Try cherry-pick only QUIC related changes. As of #8025, `spawn_server` is replaced with `spawn_server_with_cancel`, so change our uses accordingly.

### DIFF
--- a/votor/src/voting_service.rs
+++ b/votor/src/voting_service.rs
@@ -293,13 +293,7 @@ mod tests {
             consensus_message::{Certificate, CertificateType, ConsensusMessage, VoteMessage},
             vote::Vote,
         },
-        std::{
-            net::SocketAddr,
-            sync::{
-                atomic::{AtomicBool, Ordering},
-                Arc,
-            },
-        },
+        std::{net::SocketAddr, sync::Arc},
         test_case::test_case,
         tokio_util::sync::CancellationToken,
     };


AshwinSekar (approved):
Comparing the original PRs to this cherry-pick I don't see anything obviously wrong. This seems fine.

Before merging please run this on one of the invalidator clusters for a bit to verify that we haven't broken anything.

core/src/ed25519_sigverifier.rs
```diff
@@ -61,7 +61,11 @@ impl SigVerifier for TransactionSigVerifier {
         if let Some(forward_stage_sender) = &self.forward_stage_sender {
             self.banking_stage_sender
                 .send(banking_packet_batch.clone())?;
-            let _ = forward_stage_sender.try_send((banking_packet_batch, self.reject_non_vote));
+            if let Err(TrySendError::Full(_)) =
+                forward_stage_sender.try_send((banking_packet_batch, self.reject_non_vote))
+            {
+                warn!("forwarding stage channel is full, dropping packets.");
+            }
```
ksn6, line 68: I think this is an Alpenglow-specific change?

nit: the disconnected case is ignored. How about:

```rust
  match forward_stage_sender.try_send((banking_packet_batch, self.reject_non_vote)) {
      Err(TrySendError::Full(_)) => warn!("forwarding stage channel is full, dropping packets."),
      Err(TrySendError::Disconnected(_)) => warn!("forwarding stage channel disconnected."),
      Ok(()) => {}
  }
```

wen-coding: Nope, this is https://github.com/anza-xyz/agave/pull/7986

I'd prefer to only add necessary change to make our API agree with that in Agave in this PR.

PR #660, opened 2026-01-08
Introduces `ValidatedRewardCert`

akhi3030:
#### Problem

Now that we are putting the reward certs in the block footers, when the validators receive the reward certs, they need to be able to ensure their validity.

#### Summary of Changes

This PR introduces `ValidatedRewardCert` which validates the skip and notar reward certs and once validated, returns a list of validators that were included in the reward certs.

To keep the scope of the PR small, the new struct is not used anywhere yet.  In subsequent PRs, we will use it in block component processor when we process the block footer.


runtime/src/validated_reward_certificate.rs
```diff
@@ -0,0 +1,263 @@
...
+            .ok_or(Error::MissingRank)?;
+        validators.push(*pubkey);
+        bls_pubkeys.push(*bls_pubkey);
+    }
+    verify_signature(payload, signature, bls_pubkeys)?;
+    Ok(())
+}
+
+/// Returns the rank map corresponding to the provided slot in the provided bank.
+fn get_rank_map(bank: &Bank, slot: Slot) -> Option<&Arc<BLSPubkeyToRankMap>> {
```
AshwinSekar, line 82: this functionality is present in multiple places, it would be nice to unify into some common area:

https://github.com/anza-xyz/alpenglow/blob/a299954839d20acf96d4d2f30ea6d74a86ea0fea/runtime/src/block_component_processor.rs#L208-L213

https://github.com/anza-xyz/alpenglow/blob/a299954839d20acf96d4d2f30ea6d74a86ea0fea/core/src/bls_sigverify/bls_sigverifier.rs#L52-L58

runtime/src/validated_reward_certificate.rs
```diff
@@ -0,0 +1,263 @@
...
+    let aggregate_pubkey = PubkeyProjective::par_aggregate(pubkeys.par_iter())?;
+    if aggregate_pubkey.verify_signature(signature, payload)? {
+        Ok(())
+    } else {
+        Err(Error::VerifySigFalse)
+    }
+}
+
+/// Extracts a list of validator pubkeys from the rank bitmap and pushes them to the validators Vec.
+fn extract_validators(
```
AshwinSekar, line 57: I see we have this functionality for certificates present here https://github.com/anza-xyz/alpenglow/blob/a299954839d20acf96d4d2f30ea6d74a86ea0fea/bls-cert-verify/src/cert_verify.rs#L69

I'm wondering if we can extract a common validation function which takes in a payload and rank map for use by this file and the certificate verification logic? Ideally it can be generic enough to return the aspects that are important for both use cases - here we want a list of validators that participated and cert verification we want to the total stake.

I'd prefer to avoid duplicating the verification code in case we make changes later. cc @wen-coding

AshwinSekar (commented):
overall looks good, i'm wondering if we can consolidate some of this functionality to enable better code reuse

runtime/src/validated_reward_certificate.rs
akhi3030: I think I'll do this cleanup in a separate PR to keep things tidy.

AshwinSekar: sounds good that works for me

PR #664, opened 2026-01-11
Eliminates duplicate code in `ValidatedRewardCert` and `Certificate` verification

akhi3030:
#### Problem
As discussed in https://github.com/anza-xyz/alpenglow/pull/660/changes#r2676635797, verifying a reward cert and `Certificate` has some duplicate code.  

#### Summary of Changes

This PR refactors some of `bls-cert-verify` to eliminate the code duplication.  In particular:

- The closures are `FnMut` allowing caller to mutably capture the environment.
- Eliminates some errors that should not happen.
- function arguments use byte vec instead of `Certificate` allowing other use cases to call them too
- Separate some error types that were previously joined
- Passes signatures as `AsSignature` allowing combining of compressed and uncompressed signatures
- improves documentation
- `verify_base2` is used to eliminate a bunch of duplicate code in `ValidatedRewardCert`

### DIFF
--- a/bls-cert-verify/src/cert_verify.rs
+++ b/bls-cert-verify/src/cert_verify.rs
@@ -106,7 +106,12 @@ pub fn verify_base2<S: AsSignature>(
         Decoded::Base3(_, _) => return Err(Error::WrongEncoding),
     };
 
-    let pk = get_pubkey(&ranks, checked_rank_map(rank_map, &ranks), max_validators)?;
+    let pk = if cfg!(debug_assertions) {
+        get_pubkey(&ranks, checked_rank_map(rank_map, &ranks), max_validators)?
+    } else {
+        get_pubkey(&ranks, rank_map, max_validators)?
+    };
+
     if pk.verify_signature(signature, payload)? {
         Ok(())
     } else {
@@ -160,7 +165,11 @@ fn verify_base3(
     let ranks = decode(ranks, max_validators).map_err(Error::Decode)?;
     match ranks {
         Decoded::Base2(ranks) => {
-            let pk = get_pubkey(&ranks, rank_map, max_validators)?;
+            let pk = if cfg!(debug_assertions) {
+                get_pubkey(&ranks, checked_rank_map(rank_map, &ranks), max_validators)?
+            } else {
+                get_pubkey(&ranks, rank_map, max_validators)?
+            };
             if pk.verify_signature(signature, payload)? {
                 Ok(())
             } else {
@@ -168,20 +177,27 @@ fn verify_base3(
             }
         }
         Decoded::Base3(ranks, fallback_ranks) => {
-            let pubkeys = [
-                get_pubkey(
-                    &ranks,
-                    checked_rank_map(&mut rank_map, &ranks),
-                    max_validators,
-                )?
-                .into(),
-                get_pubkey(
-                    &fallback_ranks,
-                    checked_rank_map(rank_map, &fallback_ranks),
-                    max_validators,
-                )?
-                .into(),
-            ];
+            let pubkeys = if cfg!(debug_assertions) {
+                [
+                    get_pubkey(
+                        &ranks,
+                        checked_rank_map(&mut rank_map, &ranks),
+                        max_validators,
+                    )?
+                    .into(),
+                    get_pubkey(
+                        &fallback_ranks,
+                        checked_rank_map(rank_map, &fallback_ranks),
+                        max_validators,
+                    )?
+                    .into(),
+                ]
+            } else {
+                [
+                    get_pubkey(&ranks, &mut rank_map, max_validators)?.into(),
+                    get_pubkey(&fallback_ranks, rank_map, max_validators)?.into(),
+                ]
+            };
             let verified = SignatureProjective::par_verify_distinct_aggregated(
                 &pubkeys,
                 signature,


bls-cert-verify/src/cert_verify.rs
```diff
@@ -1,158 +1,167 @@
...
+    payload: &[u8],
+    fallback_payload: &[u8],
+    signature: &BlsSignature,
+    ranks: &[u8],
+    max_validators: usize,
+    mut rank_map: impl FnMut(usize) -> Option<BlsPubkey>,
+) -> Result<(), Error> {
+    let ranks = decode(ranks, max_validators).map_err(Error::Decode)?;
+    match ranks {
+        Decoded::Base2(ranks) => {
```
AshwinSekar, line 167: Should this be `Err(Error::WrongEncoding)`? Or is this the case where we only use the second bitvec if needed, e.g. a `NotarizeFallback` certificate only containing `Notarize` votes?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -1,158 +1,167 @@
...
+/// On success, returns th total stake that signed the certificate.
+pub fn verify_cert_get_total_stake(
+    cert: &Certificate,
+    max_validators: usize,
+    mut rank_map: impl FnMut(usize) -> Option<(u64, BlsPubkey)>,
+) -> Result<u64, Error> {
+    let mut total_stake = 0u64;
+    let rank_map = |ind: usize| {
+        rank_map(ind).map(|(stake, pubkey)| {
+            total_stake = total_stake.saturating_add(stake);
```
AshwinSekar, line 43: clever. guess we're relying on the fact that this `rank_map` will never be accessed for a rank corresponding to a 0 in the bit vec

bls-cert-verify/src/cert_verify.rs
akhi3030: Indeed.  

- https://github.com/anza-xyz/alpenglow/blob/1e6b8bf78a30a175117985873faf89f4badee886/votor/src/consensus_pool/certificate_builder.rs#L242
- https://github.com/anza-xyz/alpenglow/blob/1e6b8bf78a30a175117985873faf89f4badee886/votor/src/consensus_pool/certificate_builder.rs#L229

If we do not have any fallback votes, we use base2 encoding.

bls-cert-verify/src/cert_verify.rs
akhi3030: Yes indeed.  Do you think that might be a problem?

AshwinSekar: It's a bit brittle but I think we can live with it.

Another question I had, what happens if we have a base3 certificate where they've indicated that the same validator sent both a Notarize/Skip and a NotarizeFallback/SkipFalback. 

Is this handled properly to make sure the stake is not double counted? I'm guessing this certificate as a whole should be marked invaild as an honest vaildator won't do that

qkniep: Yes, counting such a cert as invalid is probably the safest option

akhi3030: > It's a bit brittle but I think we can live with it.

I think I'll add some asserts in the functions below where we have decoded the bitmap to ensure that this doesn't happen and add some documentation.  It will be a bit of an improvement.

> Another question I had, what happens if we have a base3 certificate where they've indicated that the same validator sent both a Notarize/Skip and a NotarizeFallback/SkipFalback.

> Is this handled properly to make sure the stake is not double counted? I'm guessing this certificate as a whole should be marked invaild as an honest vaildator won't do that

Ouch!  Good catch.  Didn't think of that as the existing code doesn't seem to be handling that case either.  I'll fix this and add tests for it.

wen-coding: The current certificates use either:
1. base2 encoding where there is only one bitmap
2. base3 encoding where (true, true) is not encodable at all
So it didn't have to check that there is double counting of stake.

AshwinSekar: > base3 encoding where (true, true) is not encodable at all

To confirm this means if we try to create a certificate with a malicious voter's Notarize & NotarizeFallback vote it will fail?
And if we were to somehow receive a certificate that indicated that it would also fail to verify?

wen-coding: Yes, in base3 encodings it's only possible to encode
(false, false)
(false, true)
(true, false)
There is no encoding for (true, true), so the attacker can't express that, which is why it's base3 encoding.

We chose base3 encoding mostly to shrink bitmaps for 4096 validators into 1500-byte packet. The attacker being unable to specify (true, true) is a nice side-effect.

wen-coding: @samkim-crypto to confirm.

akhi3030: I can confirm: https://github.com/anza-xyz/solana-sdk/blob/e99b0f1e11b674c9d0b32a81eee8b7ef27b56f13/signer-store/src/lib.rs#L147.

AshwinSekar: great, so seems like this concern is covered then?

akhi3030: I think so.  The only remaining bit is the following:

> > It's a bit brittle but I think we can live with it.
> 
> I think I'll add some asserts in the functions below where we have decoded the bitmap to ensure that this doesn't happen and add some documentation. It will be a bit of an improvement.

I'll fix that shortly and then ask for a re-review.

akhi3030: added `checked_rank_map`.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -103,20 +105,39 @@ pub fn verify_base2<S: AsSignature>(
...
+    ranks: &BitVec<u8>,
+) -> impl FnMut(usize) -> Option<BlsPubkey> + use<'_, F>
+where
+    F: FnMut(usize) -> Option<BlsPubkey>,
+{
+    move |ind: usize| {
+        let pos = ranks
+            .get(ind)
+            .unwrap_or_else(|| panic!("{ind} is not valid in {ranks:?}"));
+        assert!(pos == true, "{ind} is not set in {ranks:?}");
```
AshwinSekar, line 134: nit: i think this can be a `debug_assert!` or only use the `checked_rank_map` in a non release context

PR #667, opened 2026-01-12
Uses `ValidatedRewardCert` in block component processor and block creation loop

akhi3030:
#### Problem

Now that we are able to parse reward certs, we need to start using them.  Further, on the block creation side, we do not want to re-validate the reward certs as that is time consuming and requires handling unnecessary errors that shouldn't happen.

#### Summary of Changes

- When `ConsensusReward` builds the reward certs, it also includes a list of validators in the certs so that the block creation loop can use the list directly instead of having to parse the certs.
- The above change requires using the rank map in `ConsensusReward` which added some complexity to its tests.
- block_component_processor actually parses the reward certs.
- tests in block_component_processor are updated to use `None` for the reward certs because the above change requires creating valid reward certs.  And as these tests are not designed to test reward certs logic, making them more complex to handle this is unnecessary.  Further, the reward certs logic is already tested in other places there the logic lies.
- **Note**: `update_bank_with_footer` is now passed a list of validators to reward but as of now it discards this list.  This will be handled in subsequent PRs.

### DIFF
--- a/nextest.toml
+++ b/nextest.toml
@@ -7,7 +7,7 @@ build-sbf = { max-threads = 1 }
 [profile.ci]
 failure-output = "immediate-final"
 slow-timeout = { period = "60s", terminate-after = 1 }
-retries = { backoff = "fixed", count = 1, delay = "1s" }
+retries = { backoff = "fixed", count = 3, delay = "1s" }
 
 [profile.ci.junit]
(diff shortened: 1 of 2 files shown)


akhi3030:
:tada: all tests are passing now!

PR #669, opened 2026-01-13
repair: for block id shred request change shred index type to u32

AshwinSekar:
#### Problem
Historical code uses u64 for shred index in some places (e.g. repair/blockstore), even though the value is actually 
a u32 https://github.com/anza-xyz/alpenglow/blob/39fc1655da9ab8caeff913a2959c244a8de50fbc/ledger/src/shred.rs#L255

#### Summary of Changes
Lets stop making this mistake in new code, for the new repair request ShredForBlockId use u32 over the wire

### DIFF
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -117,7 +117,7 @@ pub enum ShredRepairType {
     /// Requesting the missing shred at a particular index for a specific block ID
     ShredForBlockId {
         slot: Slot,
-        index: u64,
+        index: u32,
         fec_set_merkle_root: Hash,
         // Double merkle block id
         block_id: Hash,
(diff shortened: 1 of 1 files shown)


core/src/repair/serve_repair.rs
ksn6, line 113: Can we turn this into a `u32` as well, in addition to the `Shred` variant below?

ksn6, line 160: I think we can remove this function if we make the above couple of changes.

AshwinSekar: this requires a couple releases to do since this is an existing message used by eager repair.
`ShredForBlockId` is the new Alpenglow message (not yet released so changeable).

PR #671, opened 2026-01-13
repair: Add RepairRequestProtocol enum to encapsulate UDP/QUIC

AshwinSekar:
#### Problem
Repair is in a weird state where services support both UDP and QUIC however QUIC support is planning to be dropped.
The new repair service will only use UDP so it doesn't have a quic sender

#### Summary of Changes
Add an enum that makes the sender optional if the protocol is UDP

Fixes #

### DIFF
--- a/core/src/repair/serve_repair.rs
+++ b/core/src/repair/serve_repair.rs
@@ -105,6 +105,13 @@ const SIGNED_REPAIR_TIME_WINDOW: Duration = Duration::from_secs(60 * 10); // 10 
 #[cfg(test)]
 static_assertions::const_assert_eq!(MAX_ANCESTOR_RESPONSES, 30);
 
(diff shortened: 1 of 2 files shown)


core/src/repair/serve_repair.rs
ksn6, line 1449: Should probably get this guy as well. The `initiate_ancestor_hashes_requests_for_duplicate_slot` function in `ancestor_hashes_service.rs` invokes this function, and there's a quic sender there as well that can be folded in to the enum you introduced.

AshwinSekar: this function doesn't take a sender though. Ancestor hashes service supports both QUIC and UDP and so it has a quic sender. 

I'd rather not refactor ancestor hashes further since we're planning on deleting it post alpenglow.

PR #672, opened 2026-01-13
repair_handler: add tests and bug fixes

AshwinSekar:
- `fec_set_index` is in shred space, so we need to divide by 32 before accessing the proof
- `fec_set_index` on the request is unverified, do not panic if it's out of range

Added tests

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -658,6 +658,18 @@ impl Blockstore {
         self.parent_meta_cf.get((slot, location))
     }
 
+    /// Sets the ParentMeta for the specified slot and location.
+    /// Only available for testing.
+    #[cfg(feature = "dev-context-only-utils")]
+    pub fn put_parent_meta(
+        &self,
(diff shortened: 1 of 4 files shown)


ledger/src/blockstore.rs
```diff
@@ -3140,8 +3152,9 @@ impl Blockstore {
         ]
         .into_iter()
         .find(|location| {
-            self.get_double_merkle_root(slot, *location)
-                .is_some_and(|dmr| dmr == block_id)
+            self.get_double_merkle_meta(slot, *location)
```
AshwinSekar: sneaking this in here, this can jump the gun and compute the double merkle meta which is confusing.

i'll be slightly reworking the double merkle meta computation to happen atomically with shred completion in an upcoming PR. This stop gap is to just avoid issues.

AshwinSekar: reverted, moved DMR computation to atomic in #675 instead

PR #665, opened 2026-01-12
blockstore: DMR use correct slot meta column

AshwinSekar:
#### Problem
When trying to compute the DMR out of a different column (not `BlockLocation::Original`) I noticed that the slot meta check would sometimes fail. Turns out we were always checking the original slot meta 🤦 

#### Summary of Changes
Use `meta_from_location` to check the correct slot meta column

PR #675, opened 2026-01-14
blockid: move DMR computation to shred insert

AshwinSekar:
#### Problem
Currently the DoubleMerkleMeta is only computed and inserted once the bank has finished replay (or in the leader case finished broadcast).
This can cause problems in repair as we use this blockstore column to check if we have the block or we otherwise need to repair it.

#### Summary of Changes
Instead we move this computation up to shred insert. Namely once the block is full (all shreds are inserted) we compute the double merkle root & proofs. Perhaps as an optimization we can defer computing the proofs until we actually need them (when responding to repair requests) in the future.

We do this atomically in the write batch. This guarantees the invariant that any alpenglow block that is fully ingested will have this column populated.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -18,10 +18,8 @@ use {
         leader_schedule_cache::LeaderScheduleCache,
         next_slots_iterator::NextSlotsIterator,
         shred::{
-            self,
-            merkle_tree::{get_proof_size, MerkleTree, SIZE_OF_MERKLE_PROOF_ENTRY},
-            ErasureSetId, ProcessShredsStats, ReedSolomonCache, Shred, ShredData, ShredFlags,
-            ShredId, ShredType, Shredder, DATA_SHREDS_PER_FEC_BLOCK,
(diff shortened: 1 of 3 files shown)


ledger/src/blockstore.rs
```diff
@@ -1497,6 +1392,141 @@ impl Blockstore {
...
+        slot: Slot,
+        location: BlockLocation,
+        last_index: u64,
+        parent_metas: &HashMap<(BlockLocation, Slot), ParentMetaWorkingSetEntry>,
+        merkle_root_metas: &HashMap<(BlockLocation, ErasureSetId), WorkingEntry<MerkleRootMeta>>,
+        write_batch: &mut WriteBatch,
+    ) -> Result<()> {
+        let fec_set_count = (last_index / (DATA_SHREDS_PER_FEC_BLOCK as u64) + 1) as usize;
+
+        // Get parent meta from tracker first, then fall back to db
```
AshwinSekar, line 1433: this is just copy pasted from above (and also fixes the bugs from #665

PR #676, opened 2026-01-15
blockstore: avoid inserting bad parent marker shreds

ksn6:
#### Problem

When a block contains both a `BlockHeader` and an `UpdateParent` with conflicting parent information, `should_write_parent_meta` returns an error but the shreds were still being inserted into the blockstore. This resulted in invalid blocks being marked as "full" even though they should be rejected.

#### Summary of Changes

- When `should_write_parent_meta` returns an error, mark the slot as dead
- Reorder `check_insert_data_shred` to call `maybe_update_parent_meta` before `insert_data_shred`, ensuring validation happens before the shred is persisted
- Test the above behavior

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -2342,6 +2342,17 @@ impl Blockstore {
             }
         }
 
+        if shred.index() % DATA_SHREDS_PER_FEC_BLOCK as u32 == 0 || shred.data_complete() {
+            self.maybe_update_parent_meta(
+                &shred,
+                location,
+                slot,
+                just_inserted_shreds,
+                parent_meta_working_set,
+                write_batch,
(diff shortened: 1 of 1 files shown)


ledger/src/blockstore.rs
```diff
@@ -2342,25 +2342,26 @@ impl Blockstore {
...
 
+        let completed_data_sets = self.insert_data_shred(
+            slot_meta,
+            index_meta.data_mut(),
+            &shred,
+            location,
+            write_batch,
+            shred_source,
+        )?;
+
```
ksn6, line 2364: Moving this block after `self.maybe_update_parent_meta` allows us to skip inserting the data shred in the event that `ParentMeta` compatibility checks fail.

PR #677, opened 2026-01-15
fast leader handover: handle haiku problems

ksn6:
During block production, malicious leader A may send the following leader B a unique block that doesn't exist, upon which B optimistically starts constructing via fast leader handover. Upon realizing A's actions, B then disseminates an `UpdateParent` event, associated with a valid parent from a `ParentReady` message. Validators that vote on B's block need to replay B's block - however, replaying from shred 0 is impossible, since B's original starting parent is invalid. Instead, these validators replay from the `UpdateParent` message onwards. If the block was marked as dead when processing starting from shred 0, it's important that we resurrect the block should the block's contents be valid from the `UpdateParent` message onwards.

In anticipation of this, it's entirely possible for B to e.g., place a haiku (or a sonnet, or other forms of garbage) in the shreds prior to the `UpdateParent` message. For now, we allow such forms of garbage. We disallow garbage in FEC set 0, which must contain a valid block header which respects the constraints associated with the `UpdateParent` marker, however.

If a block was marked as dead upon receipt of an `UpdateParent` marker, we ask replay to re-consider replaying messages from the `UpdateParent` marker onwards, thereby ignoring garbage shreds prior.

### DIFF
--- a/ledger/src/blockstore.rs
+++ b/ledger/src/blockstore.rs
@@ -13552,24 +13552,24 @@ pub mod tests {
 
         blockstore
             .insert_shreds(
-                create_block_header_shreds(30, 25, Hash::new_unique()),
+                create_block_header_shreds(32, 25, Hash::new_unique()),
                 None,
                 true,
             )
             .unwrap();
-        verify_next_slots(&blockstore, 25, &[30]);
+        verify_next_slots(&blockstore, 25, &[32]);
 
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(30, 22, Hash::new_unique(), 32, true),
+                create_update_parent_shreds(32, 22, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
             .unwrap();
 
-        assert_eq!(blockstore.meta(30).unwrap().unwrap().parent_slot, Some(22));
+        assert_eq!(blockstore.meta(32).unwrap().unwrap().parent_slot, Some(22));
         verify_next_slots(&blockstore, 25, &[]);
-        verify_next_slots(&blockstore, 22, &[30]);
+        verify_next_slots(&blockstore, 22, &[32]);
     }
 
     #[test]
@@ -13578,22 +13578,22 @@ pub mod tests {
         let blockstore = Blockstore::open(ledger_path.path()).unwrap();
 
         let parent_id = Hash::new_unique();
-        for slot in [41, 40, 42] {
+        for slot in [44, 40, 48] {
             blockstore
                 .insert_shreds(create_block_header_shreds(slot, 35, parent_id), None, true)
                 .unwrap();
         }
-        verify_next_slots(&blockstore, 35, &[40, 41, 42]);
+        verify_next_slots(&blockstore, 35, &[40, 44, 48]);
 
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(41, 32, Hash::new_unique(), 32, true),
+                create_update_parent_shreds(44, 32, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
             .unwrap();
-        verify_next_slots(&blockstore, 35, &[40, 42]);
-        verify_next_slots(&blockstore, 32, &[41]);
+        verify_next_slots(&blockstore, 35, &[40, 48]);
+        verify_next_slots(&blockstore, 32, &[44]);
 
         blockstore
             .insert_shreds(
@@ -13602,7 +13602,7 @@ pub mod tests {
                 true,
             )
             .unwrap();
-        verify_next_slots(&blockstore, 35, &[42]);
+        verify_next_slots(&blockstore, 35, &[48]);
         verify_next_slots(&blockstore, 33, &[40]);
     }
 
@@ -13765,32 +13765,32 @@ pub mod tests {
         blockstore.insert_shreds(shreds, None, true).unwrap();
         assert!(blockstore.meta(5).unwrap().unwrap().is_connected());
 
-        // Slot 10 with BlockHeader pointing to connected parent 5
+        // Slot 8 with BlockHeader pointing to connected parent 5
         blockstore
             .insert_shreds(
-                create_block_header_shreds(10, 5, Hash::new_unique()),
+                create_block_header_shreds(8, 5, Hash::new_unique()),
                 None,
                 true,
             )
             .unwrap();
-        let meta = blockstore.meta(10).unwrap().unwrap();
+        let meta = blockstore.meta(8).unwrap().unwrap();
         assert!(meta.is_parent_connected());
 
-        // Slot 20 chains to slot 10
-        let (shreds, _) = make_slot_entries(20, 10, 5);
+        // Slot 20 chains to slot 8
+        let (shreds, _) = make_slot_entries(20, 8, 5);
         blockstore.insert_shreds(shreds, None, true).unwrap();
 
-        // UpdateParent switches slot 10 to disconnected parent 3 (lower, doesn't exist)
+        // UpdateParent switches slot 8 to disconnected parent 3 (lower, doesn't exist)
         blockstore
             .insert_shreds(
-                create_update_parent_shreds(10, 3, Hash::new_unique(), 32, true),
+                create_update_parent_shreds(8, 3, Hash::new_unique(), 32, true),
                 None,
                 true,
             )
             .unwrap();
 
-        // Slot 10 should have parent_connected cleared
-        let meta = blockstore.meta(10).unwrap().unwrap();
+        // Slot 8 should have parent_connected cleared
+        let meta = blockstore.meta(8).unwrap().unwrap();
         assert_eq!(meta.parent_slot, Some(3));
         assert!(!meta.is_parent_connected());
         // Slot 20 should also have parent_connected cleared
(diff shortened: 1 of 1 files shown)


core/src/replay_stage.rs
```diff
@@ -4749,7 +4752,23 @@ impl ReplayStage {
...
+        // Resurrect dead slots that received an UpdateParent marker
+        while let Ok(signal) = update_parent_receiver.try_recv() {
+            let dominated_by_dead_block = progress.get(&signal.slot).is_some_and(|p| p.is_dead)
+                || blockstore.is_dead(signal.slot);
+            if dominated_by_dead_block {
+                info!(
+                    "resurrecting slot {} via UpdateParent (new parent: {})",
+                    signal.slot, signal.parent_slot
+                );
+                let _ = blockstore.remove_dead_slot(signal.slot);
```
AshwinSekar: we should expect on blockstore operations

core/src/replay_stage.rs
```diff
@@ -4749,7 +4752,23 @@ impl ReplayStage {
         progress: &mut ProgressMap,
         replay_timing: &mut ReplayLoopTiming,
         migration_status: &MigrationStatus,
+        update_parent_receiver: &UpdateParentReceiver,
     ) {
+        // Resurrect dead slots that received an UpdateParent marker
+        while let Ok(signal) = update_parent_receiver.try_recv() {
```
AshwinSekar, line 4815: this should be a top level function in the replay loop, not in generate new bank forks

core/src/replay_stage.rs
```diff
@@ -4749,7 +4752,23 @@ impl ReplayStage {
...
+            let dominated_by_dead_block = progress.get(&signal.slot).is_some_and(|p| p.is_dead)
+                || blockstore.is_dead(signal.slot);
+            if dominated_by_dead_block {
```
AshwinSekar: don't think you need this guard, should just dump the block even if it's not dead yet (no need to waste time replaying the portion before the UpdateParent).

Also avoids the current issue of receiving the signal before the block is dead.

ledger/src/blockstore.rs
```diff
@@ -5839,6 +5877,12 @@ fn send_signals(
             }
         }
     }
+
+    for signal in update_parent_signals {
+        for sender in update_parent_senders {
+            let _ = sender.try_send(signal.clone());
```
AshwinSekar: handle errors correctly, this channel shouldn't be backed up

ledger/src/blockstore.rs
```diff
@@ -528,14 +542,18 @@ impl Blockstore {
         let (ledger_signal_sender, ledger_signal_receiver) = bounded(MAX_REPLAY_WAKE_UP_SIGNALS);
         let (completed_slots_sender, completed_slots_receiver) =
             bounded(MAX_COMPLETED_SLOTS_IN_CHANNEL);
+        let (update_parent_sender, update_parent_receiver) =
+            bounded(MAX_COMPLETED_SLOTS_IN_CHANNEL);
```
AshwinSekar: think 100k is overkill here, only the first slot in each leader window can have the update parent probably can get away with something like 10k

core/src/replay_stage.rs
```diff
@@ -4740,6 +4749,23 @@ impl ReplayStage {
...
+        update_parent_receiver: &UpdateParentReceiver,
+    ) {
+        while let Ok(signal) = update_parent_receiver.try_recv() {
+            info!(
+                "resurrecting slot {} via UpdateParent (new parent: {})",
+                signal.slot, signal.parent_slot
+            );
+            blockstore.remove_dead_slot(signal.slot).expect("Db error");
+            progress.remove(&signal.slot);
+            bank_forks.write().unwrap().clear_bank(signal.slot);
```
AshwinSekar: this will panic if we don't have a bank for `signal.slot` yet right?

I think we should be checking if we actually have a bank, otherwise no action is needed.

Also should've pointed it out in the previous PR but `clear_bank` currently writes the `bank_hash_details` file (for tracking down bank hash divergences). Since we're clearing the bank outside of an equivocation context we should not write this file (but still write it during the `purge_uncofirmed_slot` call chain)

core/src/replay_stage.rs
```diff
@@ -4740,6 +4749,23 @@ impl ReplayStage {
         Ok(())
     }
 
+    fn resurrect_update_parent_slots(
```
AshwinSekar: I think a better name here is, "interrupt_update_parent_slots" since the bank does not necessarily need to be dead

core/src/replay_stage.rs
ksn6: > this will panic if we don't have a bank for signal.slot yet right?

was considering this before, though - if the slot was marked dead in blockstore, doesn't that imply that the bank should be in bank forks?

that being said, can toss in the check just in case

> Also should've pointed it out in the previous PR but clear_bank currently writes the bank_hash_details file (for tracking down bank hash divergences). Since we're clearing the bank outside of an equivocation context we should not write this file (but still write it during the purge_uncofirmed_slot call chain)

sounds good - done

AshwinSekar: slots not necessarily dead right? we just changed that haha

And yeah I think we should just make `dump_slots` lenient. In the past it was only used in TowerBFT situations where we had already filtered to make sure the bank exists. We should adjust it to not be as restrictive. 

also now that rooting happens in a different thread, the bank could technically get rugged so best not to enforce it exists in `dump_slots`

core/src/replay_stage.rs
```diff
@@ -3656,7 +3667,7 @@ impl ReplayStage {
 
                         // Clear the bank from bank_forks. It will be recreated with the
                         // correct parent by generate_new_bank_forks on the next iteration.
-                        bank_forks.write().unwrap().clear_bank(bank_slot);
+                        bank_forks.write().unwrap().clear_bank(bank_slot, false);
```
AshwinSekar: need a guard here as well, if the bank was cleaned up due to votor rooting something between line 3640 and here, this will panic

core/src/replay_stage.rs
```diff
@@ -4740,6 +4751,53 @@ impl ReplayStage {
...
+        progress: &mut ProgressMap,
+        update_parent_receiver: &UpdateParentReceiver,
+    ) {
+        while let Ok(signal) = update_parent_receiver.try_recv() {
+            // Skip if we're the leader
+            let is_leader = bank_forks
+                .read()
+                .unwrap()
+                .get(signal.slot)
+                .is_some_and(|bank| bank.collector_id() == my_pubkey);
```
AshwinSekar: if we haven't started replaying the bank yet (it's not present in bank forks) then we should skip.
Maybe something like:
 
```suggestion
                .is_none_or(|bank| bank.collector_id() == my_pubkey);
```

ledger/src/blockstore.rs
```diff
@@ -5839,6 +5877,14 @@ fn send_signals(
             }
         }
     }
+
+    for signal in update_parent_signals {
+        for sender in update_parent_senders {
+            sender
+                .try_send(signal.clone())
+                .expect("update_parent channel full");
```
AshwinSekar: This feels dangerous. I know we have a 100k bound on this channel, but I'm thinking if there's an attack pattern whereby a group of malicious validators sends UpdateParent shreds for their leader slots in the future in order to overwhelm this channel.

Just to be safe, I would prefer to do a `datapoint_error` and `error!` log if this channel gets full rather than expect

ksn6: Wouldn't setting the max length of this channel to 108k solve this problem entirely?

Signals are sent at most once per slot with slot number 0 mod 4, of which there are 432k / 4 = 108k per epoch.

Just in case though, tossed in error-ing logic (and updated the max length of the channel to 108k).

ledger/src/blockstore.rs
```diff
@@ -2413,6 +2449,10 @@ impl Blockstore {
             },
         };
 
+        if slot % NUM_CONSECUTIVE_LEADER_SLOTS != 0 {
+            return Err(BlockstoreError::UpdateParentOnInvalidSlot(slot));
+        }
+
```
ksn6, line 2504: Added this in to prevent `UpdateParent` message shreds from going into invalid slots.

ledger/src/blockstore.rs
```diff
@@ -1551,6 +1571,15 @@ impl Blockstore {
                 continue;
             }
 
+            if parent_meta.as_ref().populated_from_update_parent() {
```
AshwinSekar: Sorry I realized that we only want to do this for `BlockLocation::Original` at the moment, we want to make sure an UpdateParent from a different block in the slot doesn't trigger an interrupt for the replayed block

AshwinSekar (commented):
almost there, just one more comment

ledger/src/blockstore.rs
AshwinSekar: we actually allow for 2 * SLOTS_PER_EPOCH from our root to be ingested haha. It's probably overkill to make the channel 200k

ksn6: Agreed it's overkill, though 216k should be fine I think?

AshwinSekar (approved):

PR #681, opened 2026-01-15
Bug: use the correct slot for building reward certs

akhi3030:
#### Problem

Previously we were building the reward certs for the same slot as the bank slot.  And this showed up as a problem in #667.

#### Summary of Changes

Passes in the current slot to build the reward certs.


AshwinSekar (approved):
nice catch!

PR #682, opened 2026-01-15
Calculates voting reward without actually paying them

akhi3030:
#### Problem
Now that we have parsed the list of validators that need to be paid the voting reward, we need to calculate their rewards.

#### Summary of Changes

Calculate the voting reward without actually paying them yet.  Some more details:

- In order to calculate the reward, we need to look up the stakes for which we need the slot for which the reward certs were generated.  `BuildRewardCertsResponse` is updated to return the reward slot as well as the list of validators.
- `ValidatedRewardCertificate` is updated to track and return the reward slot and the list of validators.
- `block_component_processor/voting_reward.rs` module is introduced which currently just calculates the rewards.  Subsequent PRs will actually pay them.
-

### DIFF
--- a/runtime/src/block_component_processor/voting_reward.rs
+++ b/runtime/src/block_component_processor/voting_reward.rs
@@ -104,11 +104,8 @@ fn calculate_voting_reward(
     debug_assert!(epoch_inflation < 1.0);
     debug_assert!(epoch_inflation > 0.0);
 
-    let per_slot_inflation =
+    let per_slot_inflation_lamports =
         total_supply_lamports as f64 * epoch_inflation / slots_per_epoch as f64;
-    debug_assert!(per_slot_inflation < 1.0);
-    debug_assert!(per_slot_inflation > 0.0);
-
-    let fractional_stake = validator_stake_lamports as f64 / total_stake_lamports as f64;
-    round(fractional_stake * per_slot_inflation)
+    let fractional_stake_lamports = validator_stake_lamports as f64 / total_stake_lamports as f64;
+    round(fractional_stake_lamports * per_slot_inflation_lamports)
 }


runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,104 @@
...
+    let epoch_stakes = bank
+        .epoch_stakes_from_slot(reward_slot)
+        .ok_or(Error::MissingEpochStakes)?;
+    let vote_accounts = epoch_stakes.stakes().vote_accounts().as_ref();
+    let mut total_leader_reward_lamports = 0u64;
+    for validator in validators {
+        let (validator_stake_lamports, _) = vote_accounts
+            .get(&validator)
+            .ok_or(Error::MissingValidator)?;
+        let reward_lamports =
```
akhi3030, line 47: Question: if this fails, are we able to make the assumption that the failure is deterministic, pay the validators 0 rewards and accept the block?   I am thinking that this failure could lead to liveness issues where we do not accept any new blocks.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+    let total_stake_lamports = epoch_stakes.total_stake();
+    let yearly_inflation = bank.inflation().total(bank.slot_in_year_for_inflation());
+    let slots_per_epoch = bank.epoch_schedule().slots_per_epoch;
+    let slots_per_year = bank.slots_per_year();
+
+    debug_assert!(validator_stake_lamports <= total_stake_lamports);
+    debug_assert!(total_stake_lamports <= total_supply_lamports);
+    debug_assert!(slots_per_epoch as f64 <= slots_per_year);
+
+    let epoch_inflation = (1f64 + yearly_inflation).powf(slots_per_year) - 1f64;
```
ksn6: This doesn't look right - given that `0.015 <= bank.inflation().total(...)`, this expression here evaluates to `\infty`.

We probably want something like:

```rust
  let per_epoch_rate = (1f64 + yearly_inflation).powf(slots_per_epoch as f64 / slots_per_year) - 1f64;
  let per_slot_inflation = total_supply_lamports as f64 * per_epoch_rate / slots_per_epoch as f64;
```

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+        total_leader_reward_lamports =
+            total_leader_reward_lamports.saturating_add(leader_reward_lamports);
+    }
+    Ok(())
+}
+
+/// Different types of errors possible when converting a [`f64`] to a [`u64`].
+#[derive(Debug, PartialEq, Eq, Error)]
+pub enum FloatError {
+    #[error("input {0} is inifinite")]
```
ksn6, line 66: s/inifinite/infinite

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+    Ok(input.round() as u64)
+}
+
+/// Computes the voting reward in Lamports.
+fn calculate_voting_reward(
+    bank: &Bank,
+    epoch_stakes: &VersionedEpochStakes,
+    validator_stake_lamports: u64,
+) -> Result<u64, FloatError> {
```
ksn6, line 91: Let's add tests for this to ensure that the math is right.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+
+    debug_assert!(validator_stake_lamports <= total_stake_lamports);
+    debug_assert!(total_stake_lamports <= total_supply_lamports);
+    debug_assert!(slots_per_epoch as f64 <= slots_per_year);
+
+    let epoch_inflation = (1f64 + yearly_inflation).powf(slots_per_year) - 1f64;
+    let per_slot_inflation =
+        total_supply_lamports as f64 * epoch_inflation / slots_per_epoch as f64;
+    let fractional_stake = validator_stake_lamports as f64 / total_stake_lamports as f64;
+    round(fractional_stake * per_slot_inflation)
```
ksn6: We lose precision by casting `u64` -> `f64`. For correctness, we should cast the `u64`'s to `f128`'s, do the math in `f128` space, and then return `round(...)` as `u64`.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+                |e| Error::RewardCalculation {
+                    validator,
+                    validator_stake_lamports: *validator_stake_lamports,
+                    error: e,
+                },
+            )?;
+        let _validator_reward_lamports = reward_lamports / 2;
+        let leader_reward_lamports = reward_lamports / 2;
+        total_leader_reward_lamports =
+            total_leader_reward_lamports.saturating_add(leader_reward_lamports);
```
ksn6: FYI, we currently lose a lamport here, since when x == 1 mod 2, x/2 + x/2 != x.

We probably want `let leader_reward_lamports = reward_lamports - reward_lamports / 2`.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
+use {
+    crate::{bank::Bank, epoch_stakes::VersionedEpochStakes},
+    solana_clock::Slot,
+    solana_pubkey::Pubkey,
+    thiserror::Error,
+};
+
+/// Different types of errors that can happen when calculating and paying voting reward.
+#[derive(Debug, PartialEq, Eq, Error)]
+pub enum Error {
```
ksn6: Rename to `VotingRewardError` or something else for clarity.

runtime/src/block_component_processor/voting_reward.rs
ksn6: `calculate_voting_reward` should never fail, or else something pretty horrible has happened. Looking through, the only ways for it to fail are:

- epoch_inflation is infinite (impossible, since the exponent will look like 2/365 << 1, and an exponent < 1 leads to contraction)
- slots_per_epoch = 0 (impossible)
- total_stake = 0 (impossible unless all the validators pack up and go home and leave 0 dust)

In any of these cases, a live network somehow seems worse than a non-live network 😅

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,107 @@
...
+        let leader_reward_lamports = reward_lamports / 2;
+        total_leader_reward_lamports =
+            total_leader_reward_lamports.saturating_add(leader_reward_lamports);
+    }
+    Ok(())
+}
+
+/// Different types of errors possible when converting a [`f64`] to a [`u64`].
+#[derive(Debug, PartialEq, Eq, Error)]
+pub enum FloatError {
```
ksn6, line 65: I think we can remove this error and make `calculate_voting_reward` return `u64` rather than an error, and hard-unwrap inside.

runtime/src/block_component_processor/voting_reward.rs
akhi3030: renamed.

runtime/src/block_component_processor/voting_reward.rs
akhi3030: good point.  Updated.

runtime/src/block_component_processor/voting_reward.rs
akhi3030: good catch.  Made improvements here.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,114 @@
...
+    debug_assert!(total_stake_lamports <= total_supply_lamports);
+    debug_assert!(slots_per_epoch as f64 <= slots_per_year);
+
+    let epoch_inflation = (1f64 + yearly_inflation).powf(1.0 / epochs_per_year) - 1f64;
+    debug_assert!(epoch_inflation < 1.0);
+    debug_assert!(epoch_inflation > 0.0);
+
+    let per_slot_inflation =
+        total_supply_lamports as f64 * epoch_inflation / slots_per_epoch as f64;
+    debug_assert!(per_slot_inflation < 1.0);
```
qkniep: this assert is wrong, since `per_slot_inflation` is in lamports, right?

akhi3030: yes, this assert is wrong.  Now removed.  Thanks.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    debug_assert!(validator_stake_lamports <= total_stake_lamports);
+    debug_assert!(total_stake_lamports <= total_supply_lamports);
+    debug_assert!(slots_per_epoch as f64 <= slots_per_year);
+
+    let epoch_inflation = (1f64 + yearly_inflation).powf(1.0 / epochs_per_year) - 1f64;
+    debug_assert!(epoch_inflation < 1.0);
+    debug_assert!(epoch_inflation > 0.0);
+
+    let per_slot_inflation_lamports =
```
AshwinSekar, line 107: It would be nice to reuse existing functionality wherever possible to ensure we haven't made a mistake. Checkout https://github.com/anza-xyz/alpenglow/blob/97b4d63a9d62832efdf148cca98bcb02a63ef585/runtime/src/bank.rs#L2394 
This will calculate the inflation rewards for this epoch, which we can divide by `slots_per_epoch` to get the `per_slot_inflation_lamports`.

A difference here is that you are using the current bank's capitalization as the starting total supply when I believe you should be using a bank from the end of the previous epoch.

akhi3030: reusing code is good, I'll make the suggested change shortly.  

For my knowledge, could you explain why we should use the previous epoch's capitalization instead of the current epoch's please?

AshwinSekar: Because of the large amount of stake accounts, we pay rewards over the first ~2k slots of the epoch. This causes the capitalization to increase during this time (would be good to read the PER code to confirm, since you'll need to modify that next to pay the reward).

If we use the current bank's capitalization in order to calculate rewards this would cause slots in the early parts of the epoch to receive less rewards in comparison to later slots. This doesn't seem ideal. 

I suggested the capitalization at the end of the previous epoch so that the rewards are standard across slots, however maybe a better solution is to additionally add the inflationary rewards produced by the last epoch. Or use the current bank's capitalization and add the yet to be paid PER.

runtime/src/block_component_processor/voting_reward.rs
```diff
@@ -0,0 +1,111 @@
...
+
+/// Computes the voting reward in Lamports.
+fn calculate_voting_reward(
+    bank: &Bank,
+    epoch_stakes: &VersionedEpochStakes,
+    validator_stake_lamports: u64,
+) -> Result<u64, FloatError> {
+    let total_supply_lamports = bank.capitalization();
+    let total_stake_lamports = epoch_stakes.total_stake();
+    let yearly_inflation = bank.inflation().total(bank.slot_in_year_for_inflation());
```
ksn6, line 94: Based on offline discussions, I don't believe we can use this, since `total()` is calculated as:

```rust
    /// inflation rate at year
    pub fn total(&self, year: f64) -> f64 {
        assert!(year >= 0.0);
        let tapered = self.initial * ((1.0 - self.taper).powf(year));

        if tapered > self.terminal {
            tapered
        } else {
            self.terminal
        }
    }
```

which may be non-deterministic based on platform.

Issue #486, opened 2025-09-27
Update blockhash, blockhash_queue, status cache, downstream etc. using block_id


AshwinSekar:
txs are currently signed with the blockhash - the last PoH tick.
Status cache uses blockhash for double spend verification.

In Alpenglow we don't have poh, replace blockhash with block id for alpenglow banks.

Update relevant blockhash and double spend code , ensure that RPCs and downstream tx signing doesn't break

axaysagathiya:
This looks like a great issue to get started with! I’d love to work on it — could you please assign it to me?

AshwinSekar:
@axaysagathiya are you still planning on tackling this?

AshwinSekar:
During replay we use this call chain to figure out if the transactions have already been processed on this fork
https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/runtime/src/bank/check_transactions.rs#L280-L284

The status cache stores this info https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/runtime/src/status_cache.rs#L23

`Bank` also maintains a recent blockhash queue which is stored on chain in the sysvar account https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/runtime/src/bank.rs#L2406

This is written to when we freeze a bank on this fork https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/runtime/src/bank.rs#L2823-L2825
We'll want to update this to use block id for alpenglow banks

This is also exposed via the RPC https://github.com/anza-xyz/alpenglow/blob/2a55cd12f72b1568aefd1453acfde945194e341e/rpc/src/rpc.rs#L2366


Off the top of my head these are the things that need to be updated, ideally it's as simple as writing block id to the block hash queue / status cache and things "just work" but there might be unknowns.

To verify that this works we can run the invalidator cluster with bench tps and ensure that no duplicate transactions are landing.

akhi3030:
starting to look at this issue and wrapping my head around what needs to be done.

My understanding is that we would have to update `is_transaction_already_processed` to something like the following:

```
fn is_transaction_already_processed(...) {
  if alpenglow_is_enabled {
    // Get block_id from the sanitized_tx
    // get status from the status cache using the block_id
  } else {
    // existing code
  }
}
```

[...]

PR #686, opened 2026-01-16
wip: register new block ids

akhi3030:
This is a proposal to address #486, on how we could use block ids in various places instead of block hashes with Alpenglow.

### DIFF
--- a/votor/src/root_utils.rs
+++ b/votor/src/root_utils.rs
@@ -36,6 +36,7 @@ pub(crate) struct RootContext {
 pub(crate) fn set_root(
     my_pubkey: &Pubkey,
     new_root: Slot,
+    block_id: &Hash,
     ctx: &SharedContext,
     vctx: &mut VotingContext,
     rctx: &RootContext,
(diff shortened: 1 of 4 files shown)


akhi3030:
@AshwinSekar: did you have something like this in mind for addressing #486?

votor/src/root_utils.rs
```diff
@@ -128,6 +131,7 @@ where
         .unwrap()
         .get(new_root)
         .expect("Root bank doesn't exist");
+    root_bank.register_recent_blockhash(block_id, &BankWithScheduler::no_scheduler_available());
```
AshwinSekar, line 134: this is not the correct place to do it, we need to be registering the blockhash when we replay the end of the block.

This is currently done via `register_tick`. Alpenglow blocks still have one fake tick at the end so I think this is a good place to do it:
- If it's a TowerBFT block, use the tick's hash as the blockhash
- If it's an Alpenglow block, we need `register_tick` to instead use the `block_id`.

PR #687, opened 2026-01-16
Verify the bank hash from the footer

### LINKED ISSUE #683: Verify the bank hash from the footer
#### Problem
The bank hash value from the footer is not yet verified

#### Proposed Solution
When processing the footer, store the value in `Bank` (something like `expected_bank_hash: Option<Hash>` but do not perform the verification.

Later when freezing the bank, verify that the computed bank hash matches the expected bank hash. If so, take no action. If it does not match, mark the block as dead and produce an error log, bank hash details file, and a metric.

axaysagathiya:
#### Problem
The bank hash value from the footer was not verified

#### Summary of Changes
introduces a validation mechanism that verifies the computed bank hash against the expected hash provided in the block footer. updates the Bank struct to store this footer-derived hash and modifies the `block_component_processor` to populate it during processing. When the bank is frozen, the system now compares the two hashes; if a mismatch occurs, it logs an error, saves debug details, and emits an observability metric to alert developers

Fixes #683

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -88,7 +88,6 @@ use {
     solana_vote::vote_transaction::VoteTransaction,
     solana_votor::{
         consensus_metrics::{ConsensusMetricsEventReceiver, ConsensusMetricsEventSender},
-        consensus_rewards::{AddVoteMessage, BuildRewardCertsRequest, BuildRewardCertsResponse},
         event::{
             CompletedBlock, LeaderWindowInfo, VotorEvent, VotorEventReceiver, VotorEventSender,
         },
@@ -102,6 +101,7 @@ use {
     solana_votor_messages::{
         consensus_message::ConsensusMessage,
         migration::{MigrationStatus, GENESIS_VOTE_REFRESH},
+        reward_certificate::{AddVoteMessage, BuildRewardCertsRequest, BuildRewardCertsResponse},
         vote::Vote,
     },
     std::{
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12000,6 +12000,7 @@ dependencies = [
  "solana-frozen-abi-macro",
  "solana-hash",
  "solana-program",
+ "solana-signer-store",
  "solana-vote-interface",
  "solana-votor-messages",
  "spl-pod",
--- a/ci/_
+++ b/ci/_
@@ -5,7 +5,7 @@
 # |source| me
 #
 
-base_dir=$(realpath --strip "$(dirname "$0")/..")
+base_dir=$(realpath "$(dirname "$0")/..")
 
 _() {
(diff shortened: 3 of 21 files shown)


AshwinSekar (commented):
This looks really good, there's only one thing that we need to do now:

When a block is dead, we want to ensure that we don't replay descendants of the block. We find the descendants to replay here https://github.com/anza-xyz/alpenglow/blob/05e39fee2edf78018dd161e38d91a181819dd029/core/src/replay_stage.rs#L4760

Usually the block is marked dead before being frozen so we'd be fine. but this is a new order of events (bank being frozen -> dead). 

So we'll want to adjust this filter to account for this scenario where a bank is frozen but dead. We can use the `progress.get(slot).unwrap().is_dead` to filter these out

core/src/replay_stage.rs
```diff
@@ -3772,13 +3772,58 @@ impl ReplayStage {
                     let _ = cluster_slots_update_sender.send(vec![bank_slot]);
                 }
 
-                bank.freeze();
+                let verify_result = if migration_status.is_alpenglow_enabled() {
```
AshwinSekar:
```suggestion
                let verify_result = if migration_status.should_allow_block_markers(bank.slot()) {
```

axaysagathiya: Interesting!

AshwinSekar:
> I've fixed the linting issues.
> 
> Just a heads up: while running `ci/test-checks.sh` locally, I encountered an error because `realpath --strip` isn't supported on macOS. I've raised a separate PR #690 to fix that compatibility issue.

ah yeah our ci runners are really meant for ubuntu. you can usually just run the commands directly from the script if that's a problem.

Also we failed on some of the local-cluster tests. I realize this is because we need to additionally filter the verification for non leader banks. If we are the one that produced the block, then obviously we won't have access the expected bank hash (this footer value is only populated correctly after the freeze).

Do you mind adding this patch to fix it:
```diff
diff --git a/core/src/replay_stage.rs b/core/src/replay_stage.rs
index 2bf046bf5d..14ba2c205b 100644
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -3772,7 +3772,12 @@ impl ReplayStage {
                     let _ = cluster_slots_update_sender.send(vec![bank_slot]);
                 }

-                let verify_result = if migration_status.should_allow_block_markers(bank.slot()) {
+                // For alpenglow blocks, verify that the bank hash matches the expected bank hash
+                // sent by the leader in the block footer. If we are the leader, skip verification as
+                // the bank hash in the footer is not yet populated.
+                let verify_result = if migration_status.should_allow_block_markers(bank.slot())
+                    && bank.collector_id() != my_pubkey
+                {
                     bank.freeze_and_verify_bank_hash()
                 } else {
                     bank.freeze();
```

PR #688, opened 2026-01-16
Improvements to calculating epoch inflation rewards

akhi3030:
#### Problem and summary

- `calculate_rewards_for_partitioning()` assumes that it will be called at the start of the epoch before PER has started.  In particular, it is using `self.capitalization()` to perform the inflation calculations which can change once PER has started.  This makes this function a bit fragile.
  - The PR changes this function to also accept the capitalization to break this assumption.

- `epoch_duration_in_years(prev_epoch)`: `prev_epoch` here is poorly named.  This function doesn't care which epoch it is called with and it will calculate the duration for that epoch.
  - renames `prev_epoch` to `epoch`

- Similarly `calculate_previous_epoch_inflation_rewards()` is poorly named.  It doesn't care which epoch / capitalization it is called for and it will perform the calculations.
  - renames `calculate_previous_epoch_inflation_rewards` to `calculate_epoch_inflation_rewards` and the return struct to `EpochInflationRewards` and also some fields in the in the struct as well.

### DIFF
--- a/runtime/src/bank.rs
+++ b/runtime/src/bank.rs
@@ -959,10 +959,15 @@ impl Default for BankTestConfig {
 }
 
 #[derive(Debug)]
-struct PrevEpochInflationRewards {
-    validator_rewards: u64,
-    prev_epoch_duration_in_years: f64,
+/// Data returned from [`calculate_epoch_inflation_rewards`].
+struct EpochInflationRewards {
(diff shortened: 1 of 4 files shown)


akhi3030:
These changes are mainly to help me understand and onboard onto a bunch of new code that I'm trying to wrap my head around.

ksn6:
To minimize divergence with agave with a refactor PR like this, let's please make this change in agave and then backport here.

PR #691, opened 2026-01-20
Some pre work to get ready to do voting reward calculations

akhi3030:
#### Problem

This work is carved out of https://github.com/anza-xyz/alpenglow/pull/682 as that is going to get too big.

- As seen in #682, The reward slot is needed to do calculations.  As such, the consensus rewards container is updated to return the reward slot along with the list of validators.
- To keep the types a bit saner, `BuildRewardCertsResponse` is updated to return errors instead of `None` on failures.
- `PohRecorderError` is updated to return errors when building reward certs fails.  This required moving some types around.
- Other minor related changes

### DIFF
--- a/votor-messages/src/reward_certificate.rs
+++ b/votor-messages/src/reward_certificate.rs
@@ -149,7 +149,7 @@ pub struct BuildRewardCertsRespSucc {
     /// Notar reward certificate.  None if no notar votes were registered.
     pub notar: Option<NotarRewardCertificate>,
     /// If at least one of the certs above is present, then this contains the slot for which the reward certs were built and the list of validators in the certs.
-    pub reward_slot_and_validators: Option<(Slot, Vec<Pubkey>)>,
+    pub validators: Vec<Pubkey>,
 }
 
 /// Error returned when build reward certs fails.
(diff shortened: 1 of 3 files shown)


votor-messages/src/reward_certificate.rs
```diff
@@ -125,3 +128,41 @@ impl NotarRewardCertificate {
...
+
+/// Response when the reward certs are built successfully.
+#[derive(Default)]
+pub struct BuildRewardCertsRespSucc {
+    /// Skip reward certificate.  None if no skip votes were registered.
+    pub skip: Option<SkipRewardCertificate>,
+    /// Notar reward certificate.  None if no notar votes were registered.
+    pub notar: Option<NotarRewardCertificate>,
+    /// If at least one of the certs above is present, then this contains the slot for which the reward certs were built and the list of validators in the certs.
+    pub reward_slot_and_validators: Option<(Slot, Vec<Pubkey>)>,
```
AshwinSekar: nit: maybe it would be better to put these with the `skip` and `notar`:
`skip: Option<(Slot, SkipRewardCertificate)>`
or even directly in the *RewardCertificate type?

That way you have type enforcement of "cert present, slot present"

akhi3030: actually came up with a even cleaner approach.

AshwinSekar (approved):

PR #692, opened 2026-01-20
bcl: populate finalization certificate in block footer

AshwinSekar:
#### Problem
Populate the latest finalization certificate (FastFinalize or Slow + Notarize) in block footer

#### Summary of Changes
* Add `FinalCertificates` enum to be shared by consensus_pool and bcl
* remove existing `highest_finalized_with_notarize` to track the finalization certs directly instead
* When a new slot is finalized update the RwLock for use in bcl
* BCL reads from the lock to populate the footer

Taken over from #649

### DIFF
--- a/votor/src/consensus_pool.rs
+++ b/votor/src/consensus_pool.rs
@@ -496,10 +496,9 @@ impl ConsensusPool {
     }
 
     /// Get the Finalize certificate for a slot
-    fn get_finalize_cert(&self, slot: Slot) -> Option<Arc<Certificate>> {
+    fn get_finalize_cert(&self, slot: Slot) -> Option<&Arc<Certificate>> {
         self.completed_certificates
             .get(&CertificateType::Finalize(slot))
-            .cloned()
     }
 
     /// Get the highest finalized slot (slow or fast)


votor-messages/src/consensus_message.rs
```diff
@@ -160,6 +171,24 @@ pub struct Certificate {
...
+///
+/// This enum captures the two ways a slot can be finalized:
+/// - `Finalize`: Standard finalization with both a Finalize certificate and a Notarize certificate
+/// - `FastFinalize`: Fast finalization with a single FinalizeFast certificate that combines both
+#[derive(Clone, Debug)]
+pub enum FinalizationCerts {
+    /// Standard finalization: requires both a Finalize cert and a Notarize cert for the same slot
+    Finalize {
+        /// The finalize certificate
+        finalize_cert: Arc<Certificate>,
```
akhi3030, line 184: Why is `Arc` used here?  `Certificate` does not seem to be a very big struct.  Could we not keep a copy of it here?

Using just `Certificate` might not be super clean in terms of type checking.  I could put any cert in here and even I use the right type of cert, I can still mess up the slot numbers.  We could use a struct like below maybe?
```    
Finalize {
        finalize_sigature: BLSSignature,
        finalize_bitmap: Vec<u8>,
        notar_signature: BLSSignature,
        notar_bitmap: Vec<u8>,
        slot: Slot,
        block_id: Hash,
    },

```

votor-messages/src/consensus_message.rs
```diff
@@ -160,6 +171,24 @@ pub struct Certificate {
     pub bitmap: Vec<u8>,
 }
 
+/// Represents the highest finalized slot's certificates for block footer inclusion.
+///
+/// This enum captures the two ways a slot can be finalized:
+/// - `Finalize`: Standard finalization with both a Finalize certificate and a Notarize certificate
+/// - `FastFinalize`: Fast finalization with a single FinalizeFast certificate that combines both
+#[derive(Clone, Debug)]
+pub enum FinalizationCerts {
```
akhi3030: I wonder if we can use a different name for this.  We already have `Finalize` certificate.  Maybe `HighestFinalizedSlotCert`?  It's a mouthful but more descriptive.

votor/src/consensus_pool.rs
akhi3030: I find the logic here a bit confusing.  `is_finalized()` is true if we have a fast final cert.  If we have one, then we would've already reported `VotorEvent::Finalized` before.  I am wondering if the logic should be similar to how we handle the finalize case below:

```
if let Some(finalize_cert) = self.get_finalize_cert(slot) {
  events.push(...);
  if self.highest_finalization_slot().is_none_or(|s| s < slot) {
    self.highest_finalization_certs = ...;
  }
}
```

votor/src/consensus_pool.rs
```diff
@@ -488,6 +496,26 @@ impl ConsensusPool {
...
+        self.completed_certificates
+            .iter()
+            .find_map(|(cert_type, cert)| match cert_type {
+                CertificateType::Notarize(s, _) if slot == *s => Some(cert.clone()),
+                _ => None,
+            })
+    }
+
+    /// Get the slot of the highest finalization certificates
+    fn highest_finalization_slot(&self) -> Option<Slot> {
```
akhi3030: I wonder if it will be cleaner if this is a method on `FinalizationCerts`?  And where we are using this function, we do something like: `self.highest_finalization_certs.map(|c| c.slot())`

The name of the function is also confusing because we also have a field `highest_finalized_slot` in the struct.  I am also not sure why we need the field now though.

votor/src/consensus_pool.rs
```diff
@@ -488,6 +496,26 @@ impl ConsensusPool {
             })
     }
 
+    /// Get the Notarize certificate for a slot
+    fn get_notarize_cert(&self, slot: Slot) -> Option<Arc<Certificate>> {
+        self.completed_certificates
```
akhi3030, line 490: Are we worried about the linear search here?  In a separate PR probably but we could optimise this.

votor/src/consensus_pool.rs
```diff
@@ -304,22 +306,29 @@ impl ConsensusPool {
...
+                            .completed_certificates
+                            .get(&CertificateType::Finalize(slot))
+                        {
+                            self.highest_finalization_certs = Some(FinalizationCerts::Finalize {
+                                finalize_cert: finalize_cert.clone(),
+                                notarize_cert: cert.clone(),
+                            });
+                        }
                     }
                 }
```
akhi3030, line 315: It seems like the finalize and the finalizeFast case update `self.highest_finalization_slot` which we are not doing here.  That seems odd.

akhi3030: looked at this again, this makes sense still the logic is very subtle.

AshwinSekar: agreed, i deleted the field to simplify the logic

votor/src/consensus_pool.rs
AshwinSekar: a valid concern, I think we can profile and address separately

votor/src/consensus_pool.rs
AshwinSekar: agreed that's nicer, i've removed `highest_finalized_slot` 
[76c1acc ](https://github.com/anza-xyz/alpenglow/pull/692/commits/76c1acc786e9735bda74e2340b17f75cffec48a7)

votor/src/consensus_pool.rs
AshwinSekar: agreed that's nicer, we still will end up pushing the event twice in case of slow and finalize but I don't think it should be too big of an issue https://github.com/anza-xyz/alpenglow/pull/692/commits/2da03bd648280abc3f483b53178b8d4007eee48e

votor-messages/src/consensus_message.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/692/commits/a0e4de42a9d6ba9431bb6095e21c8df70c0d1938

votor-messages/src/consensus_message.rs
AshwinSekar: Yeah i was kind of torn on this, the way that we have things setup there's not really an easy way to incorporate this into the type system.
Even with the change you suggested I could put in a different signature or bitmap. Instead I left it up to some debug_asserts to check that the cert types and slots are correct.

this will be heavily written to but rarely read from, I left the work to convert Certificate -> footer compatible structure up to BCL. Originally I had something similar to what you suggested (just keeping the BlockMarker type in consensus pool) but it felt wasteful to do the conversion.

akhi3030: sounds fine... we can live with it for now.

votor/src/consensus_pool.rs
akhi3030: looks much cleaner, thanks!

votor/src/consensus_pool.rs
```diff
@@ -488,6 +485,28 @@ impl ConsensusPool {
...
+                _ => None,
+            })
+    }
+
+    /// Get the Finalize certificate for a slot
+    fn get_finalize_cert(&self, slot: Slot) -> Option<Arc<Certificate>> {
+        self.completed_certificates
+            .get(&CertificateType::Finalize(slot))
+            .cloned()
+    }
```
akhi3030:
```suggestion
    fn get_finalize_cert(&self, slot: Slot) -> Option<&Arc<Certificate>> {
        self.completed_certificates
            .get(&CertificateType::Finalize(slot))
    }
```

nit: returning a reference seems to work fine.

PR #693, opened 2026-01-22
block component: verify and ingest finalization certs from footer

AshwinSekar:
#### Problem
We need to verify the slow / fast finalization certificates from the block footer and ingest it into our consensus pool.
Ingestion here is key for unstaked nodes.

#### Summary of Changes

- rename `HighestFinalizedSlotCert` -> `BlockFinalization` since this is used both in the consensus pool to track highest finalized block and for extraction of the footer
- removed the `Arc` from `BlockFinalization`, copying should be cheap enough ~ once every 400ms
- refactor some functions to allow for code reuse in verification w/ the genesis certificate 
- reorganize the limit map so it's accessible by block component processor
- send the verified certs to consensus pool

### DIFF
--- a/ci/do-audit.sh
+++ b/ci/do-audit.sh
@@ -59,6 +59,15 @@ cargo_audit_ignores=(
   # URL:       https://rustsec.org/advisories/RUSTSEC-2024-0376
   # Solution:  Upgrade to >=0.12.3
   --ignore RUSTSEC-2024-0376
+
+  # Crate:     bytes
+  # Version:   1.10.1
+  # Title:     Integer overflow in `BytesMut::reserve`
+  # Date:      2026-02-03
+  # ID:        RUSTSEC-2026-0007
+  # URL:       https://github.com/advisories/GHSA-434x-w66g-qw3r
+  # Solution:  Upgrade to >=1.11.1
+  --ignore RUSTSEC-2026-0007
 )
 scripts/cargo-for-all-lock-files.sh audit "${cargo_audit_ignores[@]}" | $dep_tree_filter
 # we want the `cargo audit` exit code, not `$dep_tree_filter`'s


runtime/src/block_component_processor.rs
```diff
@@ -119,6 +122,7 @@ impl BlockComponentProcessor {
         bank: Arc<Bank>,
         parent_bank: Arc<Bank>,
         marker: VersionedBlockMarker,
+        finalization_cert_sender: Option<&Sender<ConsensusMessage>>,
```
AshwinSekar, line 125: this is optional (similar to `replay_vote_sender` in TowerBFT) because we don't need it for startup blocks

votor-messages/src/consensus_message.rs
```diff
@@ -158,6 +161,51 @@ impl CertificateType {
             _ => None,
         }
     }
+
+    /// Returns the stake fraction required for certificate completion and the
+    /// `VoteType`s that contribute to this certificate.
+    ///
+    /// Must be in sync with `Vote::to_cert_types`
+    pub const fn limits_and_vote_types(&self) -> (Fraction, &'static [VoteType]) {
```
AshwinSekar, line 171: just moved these over from `votor::common` to here

votor-messages/src/consensus_message.rs
```diff
@@ -171,29 +219,30 @@ pub struct Certificate {
...
-/// This enum captures the two ways a slot can be finalized:
-/// - `Finalize`: Standard finalization with both a Finalize certificate and a Notarize certificate
+/// This enum captures the two ways a block can be finalized:
+/// - `Finalize`: Slow finalization with both a Finalize certificate and a Notarize certificate
 /// - `FastFinalize`: Fast finalization with a single FinalizeFast certificate that combines both
 #[derive(Clone, Debug)]
-pub enum HighestFinalizedSlotCert {
-    /// Standard finalization: requires both a Finalize cert and a Notarize cert for the same slot
+#[allow(clippy::large_enum_variant)]
+pub enum BlockFinalization {
```
AshwinSekar: open to naming suggestions

entry/src/block_component.rs
```diff
@@ -387,6 +388,36 @@ impl FinalCertificate {
             notar_aggregate: None,
         }
     }
+
+    /// Converts the block marker representation of a finalization certificate into verification friendly
+    /// representation of Slow or Fast finalization certificates
+    pub fn to_certificates(self) -> Result<BlockFinalization, BlsError> {
```
akhi3030: if this fails, we will get a generic bls error that will not explain which verification failed.  Might make things hard to debug from logs.

runtime/src/block_component_processor.rs
```diff
@@ -243,11 +265,80 @@ impl BlockComponentProcessor {
...
 
+    fn verify_final_cert(
+        bank: &Bank,
+        final_cert: FinalCertificate,
+    ) -> Result<BlockFinalization, BlockComponentProcessorError> {
+        let (slot, block_id) = (final_cert.slot, final_cert.block_id);
+
+        let final_certs = final_cert
+            .to_certificates()
+            .map_err(|_| BlockComponentProcessorError::InvalidFinalizationCertificate)?;
```
akhi3030: if we return a more informed error here then this might have to be updated.  I think just printing out the error in the `warn` would suffice.

entry/src/block_component.rs
akhi3030: nit: since we are consuming `self`, the function should be called `into_...`.

runtime/src/block_component_processor.rs
```diff
@@ -243,11 +265,80 @@ impl BlockComponentProcessor {
...
+
+        match &final_certs {
```
akhi3030: nit: I wonder if this verification can be moved into `to_certificates()` so that we cannot even construct `BlockFinalization` unless all the checks pass and then we do not have to worry about whether or not such checks were performed when we are using a object of type `BlockFinalization`.  Bonus points for using a new type e.g. `ValidatedBlockFinalization` 😄

runtime/src/block_component_processor.rs
```diff
@@ -270,7 +361,26 @@ impl BlockComponentProcessor {
...
+        // Verify finalization certificate and send to consensus pool
+        if let Some(final_cert) = footer.final_cert {
+            let verified_certs = Self::verify_final_cert(&bank, final_cert)?;
+            if let Some(sender) = finalization_cert_sender {
+                match verified_certs {
+                    BlockFinalization::Finalize {
+                        finalize_cert,
+                        notarize_cert,
+                    } => {
+                        let _ = sender.send(ConsensusMessage::from(notarize_cert));
```
akhi3030: should we print a `warn` if this fails to help with debugging?  If someone changes the `unbounded` or the receiver is disconnected, we would at least capture some logs.

votor-messages/src/consensus_message.rs
akhi3030: I might be tempted to to add a `Cert` suffix otherwise this is definitely an improvement.

votor-messages/src/consensus_message.rs
```diff
@@ -171,29 +219,30 @@ pub struct Certificate {
...
+    FastFinalize(Certificate),
 }
 
-impl HighestFinalizedSlotCert {
+impl BlockFinalization {
     /// The slot that is finalized
     pub fn slot(&self) -> Slot {
         match self {
-            HighestFinalizedSlotCert::Finalize {
+            BlockFinalization::Finalize {
```
akhi3030:
```suggestion
            Self::Finalize {
```

So we do not have to update this in case of renaming.

runtime/src/validated_block_finalization.rs
```diff
@@ -0,0 +1,315 @@
...
+        cert: CertificateType,
+        got: Fraction,
+        required: Fraction,
+    },
+}
+
+/// A validated proof that a block has been finalized.
+///
+/// This type can only be constructed through validation methods, providing
+/// compile-time guarantees that the contained certificates are valid.
```
akhi3030: Unfortunately, since this is an `enum`, this can be constructed directly.  E.g. the compiler will let us just do `let foo = ValidatedBlockFinalizationCert::FastFinalize(...);`.

The way to get this guarantee is to turn this into a struct.  Maybe we can define it as:

```
struct ValidatedBlockFinalizationCert {
  finalize_cert: Certificate,
  notar_cert: Option<Certificate>,
}
```

runtime/src/validated_block_finalization.rs
```diff
@@ -0,0 +1,315 @@
...
+                    slot,
+                    block_id,
+                    final_aggregate: VotesAggregate::from_certificate(cert),
+                    notar_aggregate: None,
+                }
+            }
+        }
+    }
+
+    fn verify_certificate(
```
akhi3030, line 313: could you add some doc comments to explain what is being returned in the `u64`s please?

runtime/src/validated_block_finalization.rs
```diff
@@ -0,0 +1,315 @@
...
+    ///
+    /// This method performs both BLS signature conversion and stake threshold verification
+    /// against the provided bank's epoch stakes.
+    ///
+    /// # Errors
+    /// Returns an error if:
+    /// - BLS signature conversion fails
+    /// - Certificate signature verification fails
+    /// - The certificates don't meet the required stake thresholds
+    pub fn try_from_footer(
```
akhi3030, line 77: It is fine to defer it for now but we should add some unit tests to ensure that this function does fail for invalid certs.

runtime/src/validated_block_finalization.rs
```diff
@@ -0,0 +1,315 @@
...
+                    Fraction::new(finalize_stake, NonZeroU64::new(total_stake).unwrap());
+                let notarize_threshold = notarize_cert.cert_type.limits_and_vote_types().0;
+                let finalize_threshold = finalize_cert.cert_type.limits_and_vote_types().0;
+
+                if notarize_percent < notarize_threshold {
+                    warn!(
+                        "Received a slow finalization in the footer for slot {slot} block_id \
+                         {block_id:?} in bank slot {} with notarize {notarize_percent} stake \
+                         expecting at least {notarize_threshold}",
+                        bank.slot()
```
akhi3030, line 122: should we take the slot from the footer instead of the bank here?  Given the call tree, they should match up but it isn't clear in this function that that will always be the case.

core/src/block_creation_loop.rs
```diff
@@ -742,7 +743,7 @@ fn record_and_complete_block(
     );
 
     BlockComponentProcessor::update_bank_with_footer(
-        working_bank.bank.clone_without_scheduler(),
+        &working_bank.bank.clone_without_scheduler(),
```
akhi3030: nit: we are taking a reference to a cloned object.  Could this just be `&working_bank.bank`?

akhi3030 (commented):
thanks, looks much better.  Just a couple of small comments now.

core/src/block_creation_loop.rs
AshwinSekar: good catch https://github.com/anza-xyz/alpenglow/pull/693/changes/1c4c67ddac8a6d609a02ddd5a59f732a3a99916b

runtime/src/validated_block_finalization.rs
AshwinSekar: ah I did have some basic unit tests but forgot to move them over, done here https://github.com/anza-xyz/alpenglow/pull/693/commits/7acac02abb6f017a251f21a949ca99d0559474ce

runtime/src/validated_block_finalization.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/693/commits/7e0b41181928cd67ed05139244bc43b4c27dc1f0

runtime/src/validated_block_finalization.rs
AshwinSekar: this is a really good point, I do like keeping the enum instead of having the option. I hid it in an inner type here https://github.com/anza-xyz/alpenglow/pull/693/commits/c903a112f547bc64351f199507ef881cc22ca339 seems like an idiomatic approach

runtime/src/validated_block_finalization.rs
AshwinSekar: the footer itself doesn't have a slot.
Here we log both the finalization cert slot and the bank slot.

ci/do-audit.sh
```diff
@@ -59,6 +59,15 @@ cargo_audit_ignores=(
...
   --ignore RUSTSEC-2024-0376
+
+  # Crate:     bytes
+  # Version:   1.10.1
+  # Title:     Integer overflow in `BytesMut::reserve`
+  # Date:      2026-02-03
+  # ID:        RUSTSEC-2026-0007
+  # URL:       https://github.com/advisories/GHSA-434x-w66g-qw3r
+  # Solution:  Upgrade to >=1.11.1
+  --ignore RUSTSEC-2026-0007
```
AshwinSekar, line 70: ignoring this new audit warning for now - will backport the fix from agave later

PR #694, opened 2026-01-22
Calculates voting reward using an off curve account

akhi3030:
## Problem

Based on the reward certs, we need to calculate and pay the voting rewards.

## Summary of changes

This PR tackles the calculation part of the problem.

- It uses an off curve account to store the metadata needed for calculating vote rewards.
- the off curve account is updated each time a new epoch starts.
- the data from the account is used to calculate validator and the leader rewards
- currently, the calculated rewards are discarded.  Subsequent PRs will work on paying them.


runtime/src/bank.rs
```diff
@@ -1683,6 +1678,22 @@ impl Bank {
                 &mut rewards_metrics,
             ));
 
+        let per_validator_rewards = self
```
akhi3030: from reading the code, it seems to me that the PER calculations have finished at this point in time.  Maybe someone else can double check.

runtime/src/genesis_utils.rs
```diff
@@ -125,13 +127,20 @@ pub fn create_genesis_config_with_alpenglow_vote_accounts(
...
-    create_genesis_config_with_vote_accounts_and_cluster_type(
+    let mut genesis_config_info = create_genesis_config_with_vote_accounts_and_cluster_type(
         mint_lamports,
         voting_keypairs,
         stakes,
         ClusterType::Development,
         true,
-    )
+    );
+    VoteRewardAccountState::genesis_insert_account(
```
akhi3030: I made this change to fix some test failures.

Without this change, after the epoch changes for the first time, the capitalization increases and some tests were comparing capitalization and failing because of that.

I am not sure if this is the correct way to handle this problem.

runtime/src/bank/partitioned_epoch_rewards/calculation.rs
```diff
@@ -107,7 +109,7 @@ impl Bank {
         parent_slot: Slot,
         parent_block_height: u64,
         rewards_metrics: &mut RewardsMetrics,
-    ) {
+    ) -> u64 {
```
akhi3030, line 112: also started a discussion on discord.  I do not understand how I am able to return this value here already.

runtime/src/genesis_utils.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "dev-context-only-utils")]
```
akhi3030: changes to this file were necessary to ensure that the account is setup during tests and local cluster tests.  Otherwise, when the epoch changes, the capitalization changes or we simply fail to find the account and various tests fail.  Not sure if this is the right way to address these test failures.  Happy to get feedback.

runtime/src/bank/partitioned_epoch_rewards/calculation.rs
akhi3030: figured this out now.

runtime/src/bank/tests.rs
```diff
@@ -5364,14 +5364,14 @@ fn test_bank_hash_consistency() {
             assert_eq!(bank.epoch(), 1);
             assert_eq!(
                 bank.hash().to_string(),
-                "6h1KzSuTW6MwkgjtEbrv6AyUZ2NHtSxCQi8epjHDFYh8"
+                "BURo5b1ZCkgvjE4z9KviRgz6iRoQRad9JKWE82wjXPSc"
```
akhi3030: I am not sure if this is the right way to fix this test.  Suggestions are welcome.

akhi3030: Hmm... this test is succeeding for me locally but failing on CI.  I am not sure how to fix it yet.  Investigating... if someone has hints, suggestions welcome.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,195 @@
...
+fn calculate_voting_reward(
+    slots_per_epoch: u64,
+    epoch_validator_rewards_lamports: u64,
+    total_stake_lamports: u64,
+    validator_stake_lamports: u64,
+) -> Result<u64, FloatError> {
+    let per_slot_inflation_lamports =
+        epoch_validator_rewards_lamports as f64 / slots_per_epoch as f64;
+    let fractional_stake_lamports = validator_stake_lamports as f64 / total_stake_lamports as f64;
+    round(fractional_stake_lamports * per_slot_inflation_lamports)
```
ksn6: As discussed in https://github.com/anza-xyz/alpenglow/pull/682, please avoid using `f64`'s in rewards calculations. `f64` calculations are not deterministic across all CPU architectures, which may lead to non-determinism in rewards calculations, which could in turn lead to bank hash mismatches.

Instead:

```rust
let num = (epoch_validator_rewards_lamports as u128) * (validator_stake_lamports as u128);
let denom = (slots_per_epoch as u128) * (total_stake_lamports as u128);

num / denom
```

We should hard unwrap `num / denom` to a `u64`, since it's mathematically impossible for `num / denom` to not be able to fit into a `u64` (probably a good idea to test this).

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,195 @@
...
+        return Err(FloatError::Infinite(input.to_string()));
+    }
+    if input < 0.0 {
+        return Err(FloatError::Overflow(input.to_string()));
+    }
+    if input > u64::MAX as f64 {
+        return Err(FloatError::Overflow(input.to_string()));
+    }
+    Ok(input.round() as u64)
+}
```
ksn6: After implementing https://github.com/anza-xyz/alpenglow/pull/694/changes#r2724940357 , we can remove this function entirely. uint128 math is nice, and we'll avoid this entire class of potential errors.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,195 @@
...
+        prev_epoch: Epoch,
+        prev_epoch_capitalization: u64,
+        additional_validator_rewards: u64,
+    ) {
+        let PrevEpochInflationRewards {
+            validator_rewards,
+            prev_epoch_duration_in_years: _,
+            validator_rate: _,
+            foundation_rate: _,
+        } = bank.calculate_previous_epoch_inflation_rewards(prev_epoch_capitalization, prev_epoch);
```
ksn6: This implicitly relies on `f64` math, though it looks like `agave` does this as well?

https://github.com/anza-xyz/agave/blob/master/runtime/src/bank.rs#L2396-L2420

Given that we're changing rewards for Alpenglow anyways, I think we should just get this correct on the first go / use ints all the way through. Otherwise, we'll have to re-launch / in the meantime, feels like this is ripe for a bank hash mismatch.

That being said, can do this in a follow-up PR for readability; getting the pow() math to work with ints is going to be pretty gnarly.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,195 @@
...
+        bank.store_account_and_update_capitalization(&VOTE_REWARD_ACCOUNT_ADDR, &account);
+    }
+
+    /// Calculates and serializes a new version of [`Self`] into the accounts in the [`Bank`] when a new epoch starts.
+    ///
+    /// At the start of a new epoch, over several slots we pay the inflation rewards from the previous epoch.
+    /// This is called Partitioned Epoch Rewards (PER).
+    /// As such, the capitalization keeps increasing in the first slots of the epoch.
+    /// Vote rewards are calculated as a function of the capitalization and we do not want voting in the initial slots to earn less rewards than voting in the later rewards.
+    /// As such this function is called with [`additional_validator_rewards`] which should be the total rewards that will be paid by PER and we use the capitalization from the previous epoch plus this value to compute the vote rewards.
```
ksn6: Long comments

runtime/src/bank/tests.rs
ksn6: Do we know what precise change triggers this bank hash change?

runtime/src/genesis_utils.rs
AshwinSekar: could you expand on the capitalization change issue?

runtime/src/genesis_utils.rs
```diff
@@ -125,13 +127,15 @@ pub fn create_genesis_config_with_alpenglow_vote_accounts(
...
-    create_genesis_config_with_vote_accounts_and_cluster_type(
+    let mut genesis_config_info = create_genesis_config_with_vote_accounts_and_cluster_type(
         mint_lamports,
         voting_keypairs,
         stakes,
         ClusterType::Development,
         true,
-    )
+    );
+    VoteRewardAccountState::genesis_insert_account(&mut genesis_config_info.genesis_config, 0);
```
AshwinSekar: is this necessary if we have the code in `activate_all_features_alpengow`? 

also wondering if we can just account for the first epoch in the rewards code when reading the account rather than having to insert the account at genesis. the insertion of the `GENESIS_CERTIFICATE_ACCOUNT` is a bit of a hack that will be cleaned up once alpenglow is the default, i'd prefer if we didn't have to insert more special accounts during genesis

runtime/src/block_component_processor/vote_reward.rs
akhi3030: I like the suggestion and made the change.  

> since it's mathematically impossible for num / denom to not be able to fit into a u64 (probably a good idea to test this).

Not sure I understand this point though.  I would imagine that the result should fit in `u64` because we expect the inflation to fit in u64.  Or to put it another way, could you provide some details on how I can do this test please?

akhi3030: 176df52c5405753a9a0f0e686ba90376ab7809d0

runtime/src/block_component_processor/vote_reward.rs
akhi3030: Agreed.  I made a note in https://docs.google.com/document/d/1xMuv4BALFuyuzNaabZ31NktARZvxUpYd3nsKSPXWY0k/edit?tab=t.0.

runtime/src/block_component_processor/vote_reward.rs
ksn6: > since it's mathematically impossible for num / denom to not be able to fit into a u64 

We're saying the same thing - the result must fit into a `u64`.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+    // per_slot_inflation = epoch_validator_rewards_lamports / slots_per_epoch
+    // fractional_stake = validator_stake / total_stake_lamports
+    // rewards = fractional_stake * per_slot_inflation
+    //
+    // The code below is equivalent but changes the order of operations to maintain precision
+
+    let numerator = epoch_validator_rewards_lamports as u128 * validator_stake_lamports as u128;
+    let denominator = slots_per_epoch as u128 * total_stake_lamports as u128;
+
+    // SAFETY: the result should fit in u64 because we do not expect the inflation in a single epoch to exceed u64::MAX.
```
ksn6: Long comment here + in a number of places in this PR - may be worth adding a gutter in your IDE.

runtime/src/bank/tests.rs
```diff
@@ -12543,8 +12543,11 @@ fn test_bank_burn_vat(enable_alpenglow: bool, enable_vat: bool) {
     assert!(bank_epoch_1.epoch_stakes(3).is_none());
     let capitalization_epoch_1 = bank_epoch_1.capitalization();
     if enable_alpenglow && enable_vat {
-        // VAT should be burned
-        assert!(capitalization_epoch_1 < capitalization_epoch_0);
+        // We expect the VAT to be burned so epoch 1's capitalization should be lower however
```
akhi3030: I suppose a more elegant way to fix this test would be to have it run over 3 epochs.  In the first epoch the vote reward account doesn't exist, in the second epoch it is created, and in the third epoch we enable AG and burn the VAT.  Should I try to make that change or is this way of handling the issue good enough?

runtime/src/genesis_utils.rs
akhi3030: good idea, made the change.

runtime/src/genesis_utils.rs
akhi3030: this issue has been addressed with the recent commit.

runtime/src/bank/tests.rs
akhi3030: I am a bit stumped with this test.  It is failing on CI and locally.  The hash my local machine computes is `BURo5b1ZCkgvjE4z9KviRgz6iRoQRad9JKWE82wjXPSc` and CI computes `GAcpLy2beH4eyygZaprWWzn4geSCd3xvLvnn2tudvhy1`.  

I think the hash is changing because we are adding a new account to the bank.  But why are the computed hashes different locally and on CI?

CC: @AshwinSekar

runtime/src/bank/tests.rs
AshwinSekar: If it's an easy change, I would go for it as it improves our testing coverage

runtime/src/bank/tests.rs
AshwinSekar: were you able to track this down? for debugging bank hash mismatches you can use the utilities from this file https://github.com/anza-xyz/agave/blob/master/runtime/src/bank/bank_hash_details.rs

It will print all sorts of data about the bank and contained accounts and you should be able to diff between your machine and CI

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+#[derive(Deserialize, Serialize)]
+pub(crate) struct VoteRewardAccountState {
+    /// The rewards (in lamports) that would be paid to a validator whose stake is equal to the capitalization and it voted in every slot in the epoch.
+    /// This is also the epoch inflation.
+    epoch_validator_rewards_lamports: u64,
+}
+
+impl VoteRewardAccountState {
+    /// Returns the deserialized [`Self`] from the accounts in the [`Bank`].
+    fn get_state(bank: &Bank) -> Self {
```
ksn6: nit: name is a bit misleading, since this sounds like a "getter"; it's a bit confusing, since we also invoke `set_state` here.

should probably toss this into `Bank` and call it, e.g., `alpenglow_reward_state()`; alternatively, we could call this function `construct_from_bank` or something

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+        let account_size = bincode::serialized_size(&Self {
+            epoch_validator_rewards_lamports: 0,
+        })
+        .unwrap();
+        Rent::default().minimum_balance(account_size as usize)
+    }
+
+    /// Serializes and updates [`Self`] into the accounts in the [`Bank`].
+    fn set_state(&self, bank: &Bank) {
+        let account_size = bincode::serialized_size(&self).unwrap();
```
ksn6, line 55: add a TODO here to replace this with wincode

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+        let (validator_stake_lamports, _) = vote_accounts
+            .get(&validator)
+            .ok_or(PayVoteRewardError::MissingValidator)?;
+        let reward_lamports = calculate_voting_reward(
+            bank.epoch_schedule().slots_per_epoch,
+            epoch_validator_rewards_lamports,
+            total_stake_lamports,
+            *validator_stake_lamports,
+        );
+        let validator_reward_lamports = reward_lamports / 2;
```
ksn6, line 153: Likely worth adding an explicit comment stating that validators and leaders split the rewards equally per the Alpenglow SIMD.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+            epoch_validator_rewards_lamports: 0,
+        })
+        .unwrap();
+        Rent::default().minimum_balance(account_size as usize)
+    }
+
+    /// Serializes and updates [`Self`] into the accounts in the [`Bank`].
+    fn set_state(&self, bank: &Bank) {
+        let account_size = bincode::serialized_size(&self).unwrap();
+        let lamports = Rent::default().minimum_balance(account_size as usize);
```
ksn6: Rather than using `Rent::default()` here, should we be using `bank.rent_collector().rent`?

cc: @AshwinSekar

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+        additional_validator_rewards: u64,
+    ) {
+        let EpochInflationRewards {
+            validator_rewards_lamports,
+            epoch_duration_in_years: _,
+            validator_rate: _,
+            foundation_rate: _,
+        } = bank.calculate_epoch_inflation_rewards(prev_epoch_capitalization, prev_epoch);
+        let epoch_validator_rewards_lamports =
+            validator_rewards_lamports + additional_validator_rewards;
```
ksn6: The math here is incorrect.

The rewards calculation is based on:

- capitalization from the previous epoch
- PER payouts, which haven't fully been distributed yet

Currently, the code computes:

`validator_rewards_lamports + additional_validator_rewards`

where:

- `validator_rewards_lamports = validator_rate * capitalization * epoch_duration_in_years`
- `additional_validator_rewards` refers to the PER payouts.

Instead, we should really be paying out:

`validator_rate * (capitalization + additional_validator_rewards) * epoch_duration_in_years`

which expands to:

`validator_rewards_lamports + validator_rate * additional_validator_rewards * epoch_duration_in_years`

Rough numbers, assuming:

- validator_rate of ~5%
- total supply of 500M SOL
- epoch fraction of 1/182

Per epoch, that's 137k of SOL paid out => validators make ~6.8k in SOL from inflation. The current formulation overpays validators by nearly this entire amount. The correction factor to add onto `validator_rewards_lamports` works out to be:

`0.05 * 6.8k * 1/182 ~ 2 SOL`

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -0,0 +1,169 @@
...
+        let validator_reward_lamports = reward_lamports / 2;
+        let leader_reward_lamports = reward_lamports - validator_reward_lamports;
+        total_leader_reward_lamports =
+            total_leader_reward_lamports.saturating_add(leader_reward_lamports);
+    }
+    Ok(())
+}
+
+/// Computes the voting reward in Lamports.
+fn calculate_voting_reward(
```
ksn6, line 162: Could we add tests for this + the other functions in this file? Would be useful to cover cases where the products below would overflow u64, etc.

runtime/src/bank/tests.rs
akhi3030: Created https://github.com/anza-xyz/alpenglow/pull/713.  Once that lands, I'll update this PR.

runtime/src/block_component_processor/vote_reward.rs
AshwinSekar: yeah good catch, they should be the same but if we have access to the bank one lets use that

runtime/src/bank/tests.rs
akhi3030: OK, finally managed to figure out what data is different between my machine and CI.
https://github.com/anza-xyz/alpenglow/pull/694/changes#diff-c5a8f8c9ab946fc3c6aa9e58090bb7fa6e7cd732c71e829e221d7b88c89a2b0aR18.  The pubkey for the account that is being computed is different on CI and on my machine.  I imagine there is some difference between how the rng is being seeded?  I am surprised to see that.  But I'll keep investigating.

akhi3030: OK problem identified and fixed.  The problem was that we have two different keys for the AG feature and I was using one locally and the CI was using the other.

runtime/src/block_component_processor/vote_reward.rs
akhi3030: good catch.  I think a3fa30420981bbb2eeef4df02adce5c51faafde0 should fix this issue.

runtime/src/block_component_processor/vote_reward.rs
akhi3030: added a bunch of tests.  The remaining function not currently tested can we tested once we add support for actually paying the rewards.

AshwinSekar (commented):
LGTM but will defer to @ksn6  to make sure his comments about the math were addressed

PR #695, opened 2026-01-22
fast leader handover: sad leader handover test

ksn6:
We introduce a local cluster test that verifies that fast leader handover works; in particular, the sad case path.

Note that the other Alpenglow tests implicitly check the happy case - in a future PR, we'll explicitly test for this.

### DIFF
--- a/votor-messages/Cargo.toml
+++ b/votor-messages/Cargo.toml
@@ -16,10 +16,7 @@ frozen-abi = [
     "dep:solana-frozen-abi-macro",
     "solana-hash/frozen-abi",
 ]
-dev-context-only-utils = [
-    "dep:bincode",
-    "dep:solana-keypair",
-]
+dev-context-only-utils = ["dep:bincode", "dep:solana-keypair"]
 
 [dependencies]
 agave-feature-set = { workspace = true }


local-cluster/tests/local_cluster.rs
```diff
@@ -8002,3 +8002,285 @@ fn test_alpenglow_add_missing_parent_ready() {
...
+    let mut validator_config = ValidatorConfig::default_for_test();
+    validator_config.fixed_leader_schedule = Some(leader_schedule);
+    validator_config.voting_service_test_override = Some(VotingServiceOverride {
+        additional_listeners: vec![vote_listener_socket.local_addr().unwrap()],
+        alpenglow_port_override: AlpenglowPortOverride::default(),
+    });
+
+    let mut validator_configs = make_identical_validator_configs(&validator_config, NUM_NODES);
+    for (i, config) in validator_configs.iter_mut().enumerate() {
+        config.sad_leader_handover_counter = Some(flh_counters[i].clone());
```
AshwinSekar: what is the purpose of this counter? it seems unused and I don't think we need to expose this info in the first place

ksn6: Not ready for review yet; code's still a bit messy as well.

local-cluster/tests/local_cluster.rs
```diff
@@ -8002,3 +8010,350 @@ fn test_alpenglow_add_missing_parent_ready() {
...
+            }
+            false
+        }
+    }
+
+    let vote_listener_thread = std::thread::spawn({
+        let connection_cache = cluster.connection_cache.clone();
+
+        move || {
+            let mut state = SadPathState::new();
```
ksn6, line 8285: To understand control flow more easily, start reading from this line moving forwards. Double-click into the relevant parts of `SadPathState` as you continue reading.

votor/src/consensus_metrics.rs
```diff
@@ -298,7 +298,6 @@ impl ConsensusMetrics {
 
     /// This function can be called if there is a new [`Epoch`] and it will carry out end of epoch reporting.
     fn maybe_new_epoch(&mut self, epoch: Epoch) {
-        assert!(epoch >= self.current_epoch);
```
ksn6, line 301: This invariant is incorrect, since replay can play blocks out of order; near an epoch boundary, this assert triggers.

We're rewriting `consensus_metrics.rs` here to fix this: https://github.com/anza-xyz/alpenglow/pull/698. For now, removing this assert allows tests to pass.

local-cluster/tests/local_cluster.rs
```diff
@@ -6536,6 +6536,14 @@ fn broadcast_vote(
     }
 }
 
+fn sign_and_construct_vote(vote: Vote, keypair: &Keypair, rank: u16) -> ConsensusMessage {
```
AshwinSekar: can we make this a DCOU fn in votor-messages instead?

local-cluster/tests/local_cluster.rs
```diff
@@ -8002,3 +8010,350 @@ fn test_alpenglow_add_missing_parent_ready() {
...
+/// - At the last slot of each leader's rotation (slot % 4 == 3), test injects delayed
+///   skip votes from nodes 2 and 3
+/// - The delay ensures the next leader initially starts with OptimisticParent pointing
+///   to the current slot's block
+/// - When skip votes arrive, slot S gets skipped, causing the next leader at slot S+1
+///   to (hopefully) enter the "sad path" - their optimistic parent doesn't match
+///   finalized parent
+/// - Leader must wait for finalization before producing blocks
+///
+/// ## Phase 3: Observe Liveness (slots 40+)
```
AshwinSekar, line 8035: just a suggestion to reduce LoC

you could remove the explicit phase 3, and just start the the spy and then `cluster_tests::check_for_new_roots(64, ...` (passing just the alive nodes `ContactInfo`). 

if you share the `finalized_sad_path_slots` from the spy thread then after `check_for_new_roots` has completed you can assert that there's at least enough finalized sad path slots.

PR #698, opened 2026-01-24
feat: rewrite consensus metrics to not assume epoch monotonicity

ksn6:
`consensus_metrics.rs` is currently written incorrectly.

We can't assume the invariant that `maybe_new_epoch` is invoked with a monotonically non-decreasing `epoch`, given that blocks from different forks may play out of order during replay.

To rectify, this PR introduces a map from epoch to metrics + updates the appropriate structure upon receiving a block.

When it comes to metrics emission, suppose that slot `S` is the highest possible slot in some epoch called `S.epoch`. We emit "end of epoch" metrics for `S.epoch` at the earlier of:
  - We have a finalized slot in the following epoch, `S.epoch + 1`
  - If `S` finalizes (if `S` doesn't finalize, then default to (1)).

### DIFF
--- a/votor/src/consensus_metrics.rs
+++ b/votor/src/consensus_metrics.rs
@@ -7,7 +7,7 @@ use {
     solana_pubkey::Pubkey,
     solana_votor_messages::vote::Vote,
     std::{
-        collections::{BTreeMap, HashSet},
+        collections::{BTreeMap, BTreeSet},
         sync::{
             atomic::{AtomicBool, Ordering},
             Arc,
@@ -151,7 +151,7 @@ pub struct ConsensusMetrics {
     epoch_metrics: BTreeMap<Epoch, EpochMetrics>,
 
     /// Epochs that have already been emitted (to prevent duplicate emissions).
-    emitted_epochs: HashSet<Epoch>,
+    emitted_epochs: BTreeSet<Epoch>,
 
     /// The highest finalized slot we've seen.
     highest_finalized_slot: Option<Slot>,
@@ -167,7 +167,7 @@ impl ConsensusMetrics {
     fn new(epoch_schedule: EpochSchedule, receiver: ConsensusMetricsEventReceiver) -> Self {
         Self {
             epoch_metrics: BTreeMap::default(),
-            emitted_epochs: HashSet::default(),
+            emitted_epochs: BTreeSet::default(),
             highest_finalized_slot: None,
             epoch_schedule,
             receiver,
@@ -290,25 +290,6 @@ impl ConsensusMetrics {
         self.maybe_emit_completed_epochs();
     }
 
-    /// Returns true if metrics for `epoch` should be emitted.
-    fn should_emit_epoch(
-        &self,
-        epoch: Epoch,
-        finalized_epoch: Epoch,
-        highest_finalized: Slot,
-    ) -> bool {
-        if self.emitted_epochs.contains(&epoch) {
-            return false;
-        }
-        // Emit if finalized slot is in a later epoch
-        if finalized_epoch > epoch {
-            return true;
-        }
-        // Emit if the last slot in epoch is finalized
-        let last_slot = self.epoch_schedule.get_last_slot_in_epoch(epoch);
-        highest_finalized >= last_slot
-    }
-
     /// Checks if any epochs are ready to be emitted and emits them.
     fn maybe_emit_completed_epochs(&mut self) {
         let Some(highest_finalized) = self.highest_finalized_slot else {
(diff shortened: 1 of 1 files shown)


votor/src/consensus_metrics.rs
```diff
@@ -236,15 +266,71 @@ impl ConsensusMetrics {
...
+        if self.emitted_epochs.contains(&epoch) {
+            return false;
+        }
+        // Emit if finalized slot is in a later epoch
+        if finalized_epoch > epoch {
+            return true;
+        }
+        // Emit if the last slot in epoch is finalized
+        let last_slot = self.epoch_schedule.get_last_slot_in_epoch(epoch);
+        highest_finalized >= last_slot
```
AshwinSekar: Why do we need this condition here? It seems just having the condition above `finalized_epoch > epoch` should be fine for reporting?

And I guess a broader question, why did we change reporting to happen on finalization rather than when the `Block` event is emitted for a slot in the new epoch?

ksn6: > Why do we need this condition here? It seems just having the condition above finalized_epoch > epoch should be fine for reporting?

Perhaps overkill, though this often gets us end-of-epoch metrics 1 slot faster - if we detect that the last slot of an epoch has been finalized, we can immediately emit epoch statistics.

> And I guess a broader question, why did we change reporting to happen on finalization rather than when the `Block` event is emitted for a slot in the new epoch?

Ideally, we emit epoch statistics precisely once per epoch. My thinking here is that as soon as we observe a slot in epoch E that has been finalized, we have enough information to disseminate epoch statistics for epoch E - 1 (alternatively, the last slot in E - 1 has been finalized). Simply observing a block event in epoch E isn't sufficient to decide that we may emit epoch statistics, given that blocks can play out of order.

votor/src/consensus_metrics.rs
```diff
@@ -285,23 +372,155 @@ impl ConsensusMetrics {
...
                 i64
             ),
         );
+    }
+
+    /// Cleans up old epoch data to prevent unbounded memory growth.
+    fn cleanup_old_epochs(&mut self, finalized_epoch: Epoch) {
+        let cutoff_epoch = finalized_epoch.saturating_sub(EPOCHS_TO_RETAIN);
+        self.epoch_metrics.retain(|&epoch, _| epoch >= cutoff_epoch);
+        self.emitted_epochs.retain(|&epoch| epoch >= cutoff_epoch);
```
akhi3030: nit: we should only ever have 2 epochs here so doesn't really matter.  Still `split_off` would be more efficient here.  But to use `split_off` in `emitted_epoch`, that will have to be changed to a BTreeSet.

votor/src/consensus_metrics.rs
```diff
@@ -236,15 +266,71 @@ impl ConsensusMetrics {
...
+            .epoch_metrics
+            .keys()
+            .filter(|&&epoch| self.should_emit_epoch(epoch, finalized_epoch, highest_finalized))
+            .copied()
+            .collect();
+
+        for epoch in epochs_to_emit {
+            self.emit_epoch_metrics(epoch);
+            self.emitted_epochs.insert(epoch);
+        }
```
akhi3030:
```suggestion
        for &epoch in self.epoch_metrics.keys() {
            if self.should_emit_epoch(epoch, finalized_epoch, highest_finalized) {
                self.emit_epoch_metrics(epoch);
                self.emitted_epochs.insert(epoch);
            }
        }
```

I think this does the same thing.  It is fewer lines of code and it does not allocate new Vecs.

votor/src/consensus_metrics.rs
```diff
@@ -236,15 +266,71 @@ impl ConsensusMetrics {
...
-        for (addr, metrics) in &self.node_metrics {
+    /// Handles a slot finalization event.
+    fn handle_slot_finalized(&mut self, finalized_slot: Slot) {
+        let current = self.highest_finalized_slot.unwrap_or(0);
+        self.highest_finalized_slot = Some(current.max(finalized_slot));
+        self.maybe_emit_completed_epochs();
+    }
+
+    /// Returns true if metrics for `epoch` should be emitted.
+    fn should_emit_epoch(
```
akhi3030: wouldn't it be cleaner if this function computed finalized_epoch and highest_finalized itself instead of the caller passing it in.  It seems like the info can be computed from `self`.

votor/src/consensus_metrics.rs
akhi3030: These are epoch level metrics and reporting them one slot earlier probably doesn't buy us very much.  I would rather keep the logic simpler and delay the reporting.

But I absolutely agree that we should report when blocks are finalized rather when the `Block` events happen.

votor/src/consensus_metrics.rs
```diff
@@ -236,15 +266,71 @@ impl ConsensusMetrics {
...
+        }
+
+        self.cleanup_old_epochs(finalized_epoch);
+    }
+
+    /// Emits metrics for the given epoch.
+    fn emit_epoch_metrics(&self, epoch: Epoch) {
+        let Some(epoch_metrics) = self.epoch_metrics.get(&epoch) else {
+            return;
+        };
```
akhi3030:
```suggestion
    fn emit_epoch_metrics(&self, epoch: Epoch, epoch_metrics:&EpochMetrics) {
```

nit: the caller is already performing a lookup in the map (currently using `keys()`).  The caller could be updated to look up the metrics as well and pass that in.  This way, we won't need to do two lookups in the map and we save some lines of code.

votor/src/votor.rs
```diff
@@ -235,7 +235,7 @@ impl Votor {
             root_context,
         };
 
-        let root_epoch = sharable_banks.root().epoch();
+        let epoch_schedule = sharable_banks.root().epoch_schedule().clone();
```
akhi3030, line 245: I don't remember seeing any code so far where we were able to copy the epoch schedule out of the bank.  I guess I assumed that it is possible that the epoch schedule can change in future banks (and that is why it is stored in banks).  In other words, just want to make sure that this is indeed safe and would the metrics container need access to sharable_banks instead?

akhi3030 (commented):
Thanks for catching this problem and for the fix.  Generally looks good a few small comments though.

votor/src/consensus_metrics.rs
ksn6: Cool - let's go with simplifying the logic.

votor/src/consensus_metrics.rs
ksn6: we can entirely get rid of this function, it turns out

votor/src/votor.rs
ksn6: This should be safe - `bank.rs` sets its epoch schedule by cloning from `GenesisConfig` while children build their epoch schedules by cloning from their respective parents.

PR #699, opened 2026-01-26
consensus metrics: avoid `Histogram` in favor of Welford stats

ksn6:
We aim to capture running first and second order statistics of a number of values in consensus metrics. `consensus_metrics.rs` uses `Histogram`, which:

- Unnecessarily approximates mean and standard deviation calculations
- Introduces an extra dependency

We replace `Histogram` with `welford_stats.rs`, which calculates online means and standard deviations within machine precision (i.e., no approximation), paying attention to numerical stability.


votor/src/welford_stats.rs
```diff
@@ -0,0 +1,135 @@
+use num_traits::NumCast;
+
+/// Welford's online algorithm for computing running mean, variance, and standard deviation.
+#[derive(Debug, Clone, Default)]
+pub(crate) struct WelfordStats {
```
akhi3030, line 5: I think we should name structs after the "what" they are providing and not how they are providing it.  So that if we end up changing the underlying mechanism of how we are providing it, it does not make the name confusing.  Maybe call this `Stats`?

ksn6: Would rather stick with `WelfordStats`; it's well-known + a classic online algorithm, and readers immediately know how the algorithm works / can look it up if they haven't come across it. `Stats` is vague, since there are all kinds of statistics out there, and it's not immediately obvious as to what the time / space complexities involved are.

Happy to rename if we ever need to track values derived from sufficient statistics outside of n-th order moments.

akhi3030: fair enough

PR #700, opened 2026-01-26
Improve macOS compatibility for CI script `limit-threads.sh`

axaysagathiya:
#### Problem
The `ci/common/limit-threads.sh` script fails on macOS with `nproc: command not found`. This occurs because `nproc` is a Linux-specific utility (part of coreutils) and is not available by default on macOS systems.

#### Summary of Changes
* Use `sysctl -n hw.ncpu` to determine the processor count on macOS.
* Retain `nproc` usage for Linux environments


axaysagathiya:
Hi @AshwinSekar,
Whenever you have time, could you please take a look? It’s been pending review for a few weeks. Thanks!

PR #701, opened 2026-01-29
fast leader handover: reschedule pre-`UpdateParent` transactions

ksn6:
Fast Leader Handover (FLH) allows a new leader to start building its block before the parent is finalized, using an "optimistic parent" (the highest notarized block).

However, when the finalized parent differs from the optimistic parent (the "sad path"):
1. The leader has been building on the wrong parent
2. All executed transactions must be discarded (bank is cleared)
3. The leader sends an `UpdateParent` and starts constructing on the new bank

Previously, transactions executed before the `UpdateParent` were simply dropped. This isn't great, since high-value transactions may have been scheduled prior to the `UpdateParent`.

This commit resends all of these transactions to banking for potential rescheduling.

Note: in the interest of keeping this PR simple, we'll have local cluster testing for this behavior in the next PR.

### DIFF
--- a/core/src/banking_trace.rs
+++ b/core/src/banking_trace.rs
@@ -422,6 +422,7 @@ impl BankingTracer {
     }
 }
 
+#[derive(Clone)]
 pub struct TracedSender {
     label: ChannelLabel,
     sender: Sender<BankingPacketBatch>,
@@ -464,10 +465,6 @@ impl TracedSender {
 
     pub fn is_empty(&self) -> bool {
         self.len() == 0
-    }
-
-    pub fn sender(&self) -> &Sender<BankingPacketBatch> {
-        &self.sender
     }
 }
 
--- a/core/src/block_creation_loop.rs
+++ b/core/src/block_creation_loop.rs
@@ -6,10 +6,9 @@
 //! Once alpenglow is active, this is the only thread that will touch the [`PohRecorder`].
 use {
     crate::{
-        banking_trace::BankingTracer,
+        banking_trace::{BankingPacketSender, BankingTracer},
         replay_stage::{Finalizer, ReplayStage},
     },
-    agave_banking_stage_ingress_types::BankingPacketBatch,
     crossbeam_channel::{select_biased, Receiver, Sender},
     solana_clock::Slot,
     solana_entry::block_component::{
@@ -116,7 +115,7 @@ pub struct BlockCreationLoopConfig {
     pub reward_certs_receiver: Receiver<BuildRewardCertsResponse>,
 
     /// Sender for packets to banking stage (used to re-inject transactions after sad leader handover).
-    pub non_vote_sender: Sender<BankingPacketBatch>,
+    pub banking_stage_sender: BankingPacketSender,
 }
 
 struct LeaderContext {
@@ -137,7 +136,7 @@ struct LeaderContext {
     build_reward_certs_sender: Sender<BuildRewardCertsRequest>,
     reward_certs_receiver: Receiver<BuildRewardCertsResponse>,
     highest_finalized: Arc<RwLock<Option<HighestFinalizedSlotCert>>>,
-    non_vote_sender: Sender<BankingPacketBatch>,
+    banking_stage_sender: BankingPacketSender,
 
     // Metrics
     metrics: BlockCreationLoopMetrics,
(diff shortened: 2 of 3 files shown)


core/src/block_creation_loop.rs
```diff
@@ -555,6 +565,28 @@ fn handle_parent_ready(
 
     ctx.bank_forks.write().unwrap().clear_bank(slot, false);
 
+    // Re-inject accumulated transactions back to banking stage for rescheduling
+    let packets: Vec<BytesPacket> = accumulated_txs
+        .into_iter()
+        .filter_map(|tx| {
+            let serialized = bincode::serialize(tx).ok()?;
```
OliverNChalk: This should never fail right @apfitzge ? We're taking Packet -> VersionedTransacion -> Record ->  Packet.

core/src/banking_trace.rs
```diff
@@ -465,6 +465,10 @@ impl TracedSender {
     pub fn is_empty(&self) -> bool {
         self.len() == 0
     }
+
+    pub fn sender(&self) -> &Sender<BankingPacketBatch> {
+        &self.sender
+    }
```
OliverNChalk: @apfitzge what is the idea behind the tracing senders? should BCL use tracing or just the inner sender?

OliverNChalk (commented):
Don't see any bugs, just some questions around codebase conventions for andrew

core/src/banking_trace.rs
apfitzge: The traced senders will populate our trace files which is how we can look at ingressed txs with timestamps after leader slots

apfitzge: It seems like it should use the traced sendes, otherwise we don't have a complete picture of the transactions that the scheduler saw at that time? i.e. trace files no longer represent reality in case of a bad leader handover

core/src/block_creation_loop.rs
apfitzge: can we use wincode instead?

apfitzge: it should never fail if we use *wincode*. bincode will no longer be supported soon (txv1 breaks it)

apfitzge: that said since this is non-essential behavior for the re-injection...probably better to just continue with the filter_map instead of an unwrap here or something.
Would be good to log an error if this ever fails though...though again it should not unless there's a bug in the serialization code.

core/src/block_creation_loop.rs
```diff
@@ -629,6 +663,11 @@ fn record_and_complete_block(
                 if let Ok(record) = msg {
                     ctx.record_receiver
                         .on_received_record(record.transaction_batches.len() as u64);
+
+                    record.transaction_batches.iter().for_each(|batch| {
```
AshwinSekar: we should gate this behind `optimistic_parent.is_some()`, no need to populate this buffer unless we are actually producing the first leader block optimistically

core/src/block_creation_loop.rs
```diff
@@ -111,6 +114,9 @@ pub struct BlockCreationLoopConfig {
     pub build_reward_certs_sender: Sender<BuildRewardCertsRequest>,
     /// Channel to receive the built reward certs.
     pub reward_certs_receiver: Receiver<BuildRewardCertsResponse>,
+
+    /// Sender for packets to banking stage (used to re-inject transactions after sad leader handover).
+    pub non_vote_sender: Sender<BankingPacketBatch>,
```
AshwinSekar:
```suggestion
    pub banking_stage_sender: Sender<BankingPacketBatch>,
```

at least in the context of bcl, the distinction of "non_vote" doesn't make sense

core/src/block_creation_loop.rs
ksn6: Added a TODO for when we upstream; there's some versioning stuff which prevents us from wincod-ing "advanced structures" in Alpenglow.

ksn6: My thinking is that it should be impossible for the tx to fail serialization conditioned on in having being scheduled - added in an error log just in case.

core/src/banking_trace.rs
ksn6: I'll plumb through the full traced sender in that case.

core/src/block_creation_loop.rs
ksn6: > it should never fail if we use wincode. bincode will no longer be supported soon (txv1 breaks it)

yep - see https://github.com/anza-xyz/alpenglow/pull/701#discussion_r2744537719

apfitzge: Gotcha, as a note this will break with txv1 so we can't forget to do it

PR #654, opened 2026-01-07
wip: bls_sigverify: holistic refactor for simplicity in upstreaming

AshwinSekar:
#### Problem
code is too complex to upstream

#### Summary of Changes
see if we can refactor and remove legacy ed25119 artifacts

PR #703, opened 2026-01-30
[core] Refactor BLS sigverify logic

samkim-crypto:
#### Summary of Changes

General refactoring of the BLS sigverify logic to make the code more modular and readable.

The first two commits [ba2d09a](https://github.com/anza-xyz/alpenglow/pull/703/commits/ba2d09a986f6180ce76d902a39b6a33a2a6d5ca6), [376c053](https://github.com/anza-xyz/alpenglow/pull/703/commits/376c053086750e3c21512a174bc81a2ad9d202fc) straightforwardly extract logic from `verify_and_send_batches` and `preprocess_batches` into separate helper functions.

[6099cb1](https://github.com/anza-xyz/alpenglow/pull/703/commits/6099cb1a9c33557df1541710711cbeab7a018da2) is the most technical refactor:
- Extracted vote and certificate signature verification into separate modules.
- Previously, the logic rebuilt vectors of votes and certs for every packet batch. To minimize heap allocations (which are small compared to crypto operations, but unnecessary nonetheless), I added persistent buffers for votes, certs, and metrics.

[a243975](https://github.com/anza-xyz/alpenglow/pull/703/commits/a243975e8b09055edbc67a5b23c6d68e55373594): Moved errors into a separate module.

[2ec6d80](https://github.com/anza-xyz/alpenglow/pull/703/commits/2ec6d80fc5584f3202d949e69a14109efca93d9a), [0f7ab07](https://github.com/anza-xyz/alpenglow/pull/703/commits/0f7ab07f01a0fd3f4187eb826f2a532866bc5cdd): Refactored the BLS SigVerify service.
- Previously, the `BLSSigVerifier` had to be instantiated externally and passed to the service constructor. I updated the service to instantiate the verifier internally.
- Cleaned up the service loop logic to be more modular, following https://github.com/anza-xyz/alpenglow/pull/654.

I considered merging `BLSSigVerifyService` and `BLSSigVerifier` as in https://github.com/anza-xyz/alpenglow/pull/654, but decided against it to keep the file sizes manageable. I think it makes sense to maintain `BLSSigVerifyService` as the runner/orchestrator and `BLSSigVerifier` as the worker performing the signature verification.

While upstream uses `spawn` for `BLSSigVerifyService`, I chose `new` to maintain consistency with other services. I am open to renaming it to `spawn` though if preferred.

Follow-up Todos:
- Decouple the Send logic from the signature verify logic in `BLSSigVerifier`. This will improve modularity with minimal performance cost.
- Clean up and document `bls_vote_sigverify.rs` and `bls_cert_sigverify.rs` (which handle the core cryptographic checks). This PR leaves them untouched.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -113,7 +113,7 @@ impl BLSSigVerifier {
         self.cert_buffer.clear();
         self.metric_buffer.clear();
 
-        self.extract_and_filter_messages(batches, &mut last_voted_slots, &root_bank);
+        self.extract_and_filter_messages(batches, &root_bank);
 
         preprocess_time.stop();
         self.stats.preprocess_count.fetch_add(1, Ordering::Relaxed);
@@ -121,27 +121,41 @@ impl BLSSigVerifier {
             .preprocess_elapsed_us
             .fetch_add(preprocess_time.as_us(), Ordering::Relaxed);
 
+        // Destructure self to borrow fields independently for the closure
+        let vote_buffer = &self.vote_buffer;
+        let vote_payload_cache = &self.vote_payload_cache;
+        let stats = &self.stats;
+        let cluster_info = &self.cluster_info;
+        let leader_schedule = &self.leader_schedule;
+        let message_sender = &self.message_sender;
+        let votes_for_repair_sender = &self.votes_for_repair_sender;
+        let metric_buffer = &mut self.metric_buffer;
+        let cert_buffer = &mut self.cert_buffer;
+        let verified_certs = &self.verified_certs;
+
         // Perform signature verification
         let (votes_result, certs_result) = rayon::join(
             || {
                 verify_and_send_votes(
-                    &self.vote_buffer,
+                    vote_buffer,
                     &root_bank,
-                    &self.vote_payload_cache,
-                    &self.stats,
-                    &self.cluster_info,
-                    &self.leader_schedule,
-                    &self.message_sender,
-                    &self.votes_for_repair_sender,
+                    vote_payload_cache,
+                    stats,
+                    cluster_info,
+                    leader_schedule,
+                    message_sender,
+                    votes_for_repair_sender,
+                    &mut last_voted_slots,
+                    metric_buffer,
                 )
             },
             || {
                 verify_and_send_certificates(
-                    &mut self.cert_buffer,
+                    cert_buffer,
                     &root_bank,
-                    &self.verified_certs,
-                    &self.stats,
-                    &self.message_sender,
+                    verified_certs,
+                    stats,
+                    message_sender,
                 )
             },
         );
@@ -173,12 +187,7 @@ impl BLSSigVerifier {
             .retain(|vote, _| vote.slot() > root_bank.slot());
     }
 
-    fn extract_and_filter_messages(
-        &mut self,
-        batches: Vec<PacketBatch>,
-        last_voted_slots: &mut HashMap<Pubkey, Slot>,
-        root_bank: &Bank,
-    ) {
+    fn extract_and_filter_messages(&mut self, batches: Vec<PacketBatch>, root_bank: &Bank) {
         for packet in batches.iter().flatten() {
             self.stats.received.fetch_add(1, Ordering::Relaxed);
 
@@ -202,13 +211,6 @@ impl BLSSigVerifier {
                     else {
                         continue;
                     };
-
-                    Self::record_vote_metrics(
-                        pubkey,
-                        vote_message.vote,
-                        last_voted_slots,
-                        &mut self.metric_buffer,
-                    );
 
                     if self.check_stale_vote(&vote_message, root_bank) {
                         continue;
@@ -341,20 +343,6 @@ impl BLSSigVerifier {
             })?;
 
         Some((*pubkey, *bls_pubkey))
-    }
-
-    fn record_vote_metrics(
-        pubkey: Pubkey,
-        vote: Vote,
-        last_voted_slots: &mut HashMap<Pubkey, Slot>,
-        consensus_metrics: &mut Vec<ConsensusMetricsEvent>,
-    ) {
-        if vote.is_notarization_or_finalization() {
-            let existing = last_voted_slots.entry(pubkey).or_insert(vote.slot());
-            *existing = (*existing).max(vote.slot());
-        }
-
-        consensus_metrics.push(ConsensusMetricsEvent::Vote { id: pubkey, vote });
     }
 
     fn check_stale_vote(&self, vote_message: &VoteMessage, root_bank: &Bank) -> bool {
--- a/core/src/bls_sigverify/bls_vote_sigverify.rs
+++ b/core/src/bls_sigverify/bls_vote_sigverify.rs
@@ -17,7 +17,7 @@ use {
     solana_measure::measure::Measure,
     solana_pubkey::Pubkey,
     solana_runtime::bank::Bank,
-    solana_votor::consensus_rewards,
+    solana_votor::{consensus_metrics::ConsensusMetricsEvent, consensus_rewards},
     solana_votor_messages::{
         consensus_message::{ConsensusMessage, VoteMessage},
         reward_certificate::AddVoteMessage,
@@ -38,6 +38,7 @@ pub(crate) struct VoteToVerify {
 
 /// Verifies votes and sends verified votes to the consensus pool.
 /// Also returns a copy of the verified votes that the rewards container is interested is so that the caller can send them to it.
+#[allow(clippy::too_many_arguments)]
 pub(crate) fn verify_and_send_votes(
     votes_to_verify: &[VoteToVerify],
     root_bank: &Bank,
@@ -47,6 +48,8 @@ pub(crate) fn verify_and_send_votes(
     leader_schedule: &LeaderScheduleCache,
     message_sender: &Sender<ConsensusMessage>,
     votes_for_repair_sender: &VerifiedVoteSender,
+    last_voted_slots: &mut HashMap<Pubkey, Slot>,
+    consensus_metrics: &mut Vec<ConsensusMetricsEvent>,
 ) -> Result<AddVoteMessage, BLSSigVerifyError> {
     let verified_votes = verify_votes(votes_to_verify, vote_payload_cache, stats);
 
@@ -67,6 +70,18 @@ pub(crate) fn verify_and_send_votes(
     let mut verified_votes_by_pubkey: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
     for vote in verified_votes {
         stats.received_votes.fetch_add(1, Ordering::Relaxed);
+
+        if vote.vote_message.vote.is_notarization_or_finalization() {
+            let existing = last_voted_slots
+                .entry(vote.pubkey)
+                .or_insert(vote.vote_message.vote.slot());
+            *existing = (*existing).max(vote.vote_message.vote.slot());
+        }
+        consensus_metrics.push(ConsensusMetricsEvent::Vote {
+            id: vote.pubkey,
+            vote: vote.vote_message.vote,
+        });
+
         if vote.vote_message.vote.is_notarization_or_finalization()
             || vote.vote_message.vote.is_notarize_fallback()
         {


core/src/bls_sigverify/bls_sigverify_service.rs
```diff
@@ -1,168 +1,219 @@
...
+        let mut last_print = Instant::now();
+        const MAX_DEDUPER_AGE: Duration = Duration::from_secs(2);
+        const DEDUPER_FALSE_POSITIVE_RATE: f64 = 0.001;
+        const DEDUPER_NUM_BITS: u64 = 63_999_979;
 
-pub struct BLSSigverifyService {
-    thread_hdl: JoinHandle<()>,
-}
+        let mut rng = rand::thread_rng();
+        let mut deduper = Deduper::<2, [u8]>::new(&mut rng, DEDUPER_NUM_BITS);
```
AshwinSekar: When upstreaming the simple bls verifier it was noted that we don't need the deduper for this service as quic flow control will naturally help us limit spam

core/src/bls_sigverify/bls_sigverify_service.rs
```diff
@@ -1,168 +1,219 @@
...
+            // Packet Receive Step
+            let (mut batches, num_packets, recv_duration) =
+                match Self::receive_packets(&packet_receiver) {
+                    ReceiveAction::Process(b, n, d) => (b, n, d),
+                    ReceiveAction::Continue => continue,
+                    ReceiveAction::Break => break,
+                };
+
+            let batches_len = batches.len();
+            debug!(
```
AshwinSekar: lets remove these old debug logs, they're kind of weird as they pass timestamp as well.
If needed we can add trace logs

core/src/bls_sigverify/bls_sigverify_service.rs
```diff
@@ -1,168 +1,219 @@
...
+            if deduper.maybe_reset(&mut rng, DEDUPER_FALSE_POSITIVE_RATE, MAX_DEDUPER_AGE) {
+                stats.num_deduper_saturations += 1;
+            }
+            let (discard_or_dedup_fail, dedup_time_us) =
+                Self::dedup_packets(&deduper, &mut batches);
+
+            // BLS Signature Verification (and Send) Step
+            //
+            // TODO(sam): Currently, verification result is sent inside the verification function.
+            //            Consider refactoring out the send step out of the signature verification
```
AshwinSekar, line 80: I was actually thinking of the other way where we just call `verifier.process_batches()` and update stats in this loop to keep it simple.

Something similar to what we have upstream:
https://github.com/anza-xyz/agave/blob/bc7dccde1f7c7b206736fd8d4a37a7cdb3c59964/core/src/bls_sigverifier.rs#L63 + updating stats about the batch recieve.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -80,22 +49,115 @@ pub struct BLSSigVerifier {
...
+        // Send verified votes to the rewards container
+        self.send_reward_votes(add_vote_msg);
+
+        self.stats.maybe_report_stats();
+        Ok(())
+    }
+
+    fn prune_caches(&mut self, root_bank: &Bank) {
+        // Prune entries <= the new root slot
         if self.last_checked_root_slot < root_bank.slot() {
```
AshwinSekar: I think it's cleaner to have this if conditional in the main loop before calling `prune_caches`

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -107,80 +169,73 @@ impl BLSSigVerifier {
                 .unwrap()
                 .retain(|vote, _| vote.slot() > root_bank.slot());
         }
+    }
+
+    fn extract_and_filter_messages(
+        &mut self,
+        batches: &mut [PacketBatch],
```
AshwinSekar: wdyt about just taking ownership of the batches here?

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -107,80 +169,73 @@ impl BLSSigVerifier {
...
-                Ok(msg) => msg,
-                Err(_) => {
-                    self.stats
-                        .received_malformed
-                        .fetch_add(1, Ordering::Relaxed);
-                    packet.meta_mut().set_discard(true);
-                    continue;
-                }
+            let Ok(message) = packet.deserialize_slice::<ConsensusMessage, _>(..) else {
+                packet.meta_mut().set_discard(true);
```
AshwinSekar: I don't think we need to be setting discard anymore right? this packet batch is not used further

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -244,368 +291,69 @@ impl BLSSigVerifier {
...
-            .votes_batch_parallel_verify_count
-            .fetch_add(1, Ordering::Relaxed);
-        self.stats
-            .votes_batch_parallel_verify_elapsed_us
-            .fetch_add(votes_batch_parallel_verify_time.as_us(), Ordering::Relaxed);
-        verified_votes
-    }
+    ) -> Option<(Pubkey, BlsPubkey)> {
+        // Missing epoch stakes
+        let (key_to_rank_map, _) = get_key_to_rank_map(root_bank, vote_message.vote.slot())
```
AshwinSekar, line 319: nit: maybe we could cache the key_to_rank_map for current epoch, current epoch + 1 so we don't have to go through the multi step process for every vote?

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+            Err(e @ TrySendError::Disconnected(_)) => {
+                return Err(e.into());
+            }
+        }
+    }
+
+    // Send votes
+    for (pubkey, slots) in verified_votes_by_pubkey {
+        match verified_votes_sender.try_send((pubkey, slots)) {
```
AshwinSekar: just noting that this is for use in repair, perhaps clearer to replace the verified_votes prefix with repair_votes or votes_for_repair

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+    let (distinct_payloads, distinct_pubkeys): (Vec<_>, Vec<_>) =
+        grouped_pubkeys.into_iter().unzip();
+    let aggregate_pubkeys_result: Result<Vec<PubkeyProjective>, _> = distinct_pubkeys
+        .into_iter()
+        .map(|pks| PubkeyProjective::par_aggregate(pks.into_par_iter()))
+        .collect();
+
+    let verified_optimistically = if let Ok(aggregate_pubkeys) = aggregate_pubkeys_result {
+        let signatures = votes_to_verify
+            .par_iter()
```
AshwinSekar, line 166: curious if you've noticed speedup with the par_iters in this function or if our batches would be small enough that it adds overhead

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -107,80 +169,73 @@ impl BLSSigVerifier {
...
 
-                    // Invalid rank
-                    let Some((solana_pubkey, bls_pubkey, _stake)) =
-                        key_to_rank_map.get_pubkey_and_stake(vote_message.rank.into())
-                    else {
-                        self.stats.received_bad_rank.fetch_add(1, Ordering::Relaxed);
-                        packet.meta_mut().set_discard(true);
-                        continue;
-                    };
+                    Self::record_vote_metrics(
```
AshwinSekar: it seems wrong that this happens here (i know the old code had it here).
It seems we'd want to notify the consesus metrics *after* the vote has been verified no? cc @akhi3030

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+    stats: &BLSSigVerifierStats,
+) -> Vec<bool> {
+    if certs_to_verify.is_empty() {
+        return vec![];
+    }
+    stats.certs_batch_count.fetch_add(1, Ordering::Relaxed);
+    let mut certs_batch_verify_time = Measure::start("certs_batch_verify");
+
+    let verified_results: Vec<bool> = certs_to_verify
+        .into_par_iter()
```
AshwinSekar, line 90: similar question here, my intuition is that we'd end up getting a burst of certificates of the same type in the batch (e.g. block gets finalized we get a finalization certificate from everyone).

It seems that could lead to a waste verifying these certificates in parallel, if the first one is successful we don't have to verify the rest

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Yeah on my dev server, when I run the [benches](https://github.com/anza-xyz/solana-sdk/blob/master/bls-signatures/benches/bls_signatures.rs), I get the following result:

```
... Seq faster than Par until...
Batch size 128: 1.25ms Seq vs 1.29ms Par
Batch size 256: 1.48ms Seq vs 1.33ms Par
```

For these numbers, it does seem like sticking to non-parallel verify might be the way to go. But let me tackle the `bls_vote_sigverify` and `bls_cert_sigverify` in a follow-up and just focus on refactoring the bls-sigverifier here.

core/src/bls_sigverify/bls_sigverifier.rs
samkim-crypto: Yeah it seems like you are right. Let me wait for @akhi3030 for another confirm and then I can update the code.

core/src/bls_sigverify/bls_cert_sigverify.rs
samkim-crypto: This is a good point. We have some [filtering](https://github.com/anza-xyz/alpenglow/pull/703/files#diff-44998bea37a652beea78de28073041a995c20d87f97e42c57fd7bd85466cbb39R245) already, but we don't filter within the same batch, so maybe we should add that logic.

We shouldn't be verifying the certs in parallel, but rather they should be aggregated first and then verified, so even if there are lots of duplicates in the same batch (same cert message, but different sigs), it shouldn't blow-up cost linearly.

But looking at the current cert verify logic, it seems like it is quite removed from the original code that I added. Let me defer this to a follow-up PR where I will go through the cert verify logic carefully and do lots of clean-ups and add docs.

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Okay, I updated the prefix within the `bls_sigverify` module ([6894d2a](https://github.com/anza-xyz/alpenglow/pull/703/commits/6894d2aea9fb68418927e2cdfa5af040cffbfa22)). I did not update the names in TVU or other parts of core, but I can go ahead and update it in other places if you think it will be clearer there as well.

core/src/bls_sigverify/bls_sigverifier.rs
samkim-crypto: Good point! Added in [9292539](https://github.com/anza-xyz/alpenglow/pull/703/commits/9292539453328864255b25c975a5c709bbea49f3). I had to do some refactoring of the cert filtering logic to satisfy the borrow checker.

core/src/bls_sigverify/bls_sigverifier.rs
samkim-crypto: Yep, done in [0b72c6e](https://github.com/anza-xyz/alpenglow/pull/703/commits/0b72c6ef3ffe619b63ea7e85411451c7cf4157cc).

core/src/bls_sigverify/bls_sigverifier.rs
samkim-crypto: I was considering this as well. I was thinking about whether we can use a zero-copy deserialization as noted [here](https://github.com/anza-xyz/alpenglow/pull/703/files#diff-44998bea37a652beea78de28073041a995c20d87f97e42c57fd7bd85466cbb39R224). If we do, then we would need the `extract_and_filter_messages` function to hold a reference to the batches because the `VotesToVerify` will hold references to bytes in the packet batches and if packet batches are freed, then the references will be dangling.

We don't use zero-copy at the moment, so I think it does make sense to give ownership of the batches. Done in [53336f9](https://github.com/anza-xyz/alpenglow/pull/703/commits/53336f94928203553cea6b9d1a903981798d8dac).

core/src/bls_sigverify/bls_sigverifier.rs
samkim-crypto: Agreed. Done in [36a4d9c](https://github.com/anza-xyz/alpenglow/pull/703/commits/36a4d9c6cbc3163c439d6f03c64a041ed5705d8e).

core/src/bls_sigverify/bls_sigverify_service.rs
samkim-crypto: Done in [016982c](https://github.com/anza-xyz/alpenglow/pull/703/commits/016982c10a0458a77afaabf5d09b9efa53a08006).

core/src/bls_sigverify/bls_sigverify_service.rs
samkim-crypto: Oh I thought quic deduper applied only for within same connection and not if there are duplicate packets across different connections. But I do agree that dedup logic doesn't really belong within sigverify where we should focus primarily on cryptographic checks. Removed in [3a5784d](https://github.com/anza-xyz/alpenglow/pull/703/commits/3a5784d156d02952b0485f1d2763e6015bbf1f26).

core/src/bls_sigverify/bls_sigverify_service.rs
samkim-crypto: Yeah I think I misread the upstream. I saw that the signature verification and the send results are [split](https://github.com/anza-xyz/agave/blob/bc7dccde1f7c7b206736fd8d4a37a7cdb3c59964/core/src/bls_sigverifier.rs#L94-L97), so I thought we wanted to ultimately split these logic for readability.

I actually prefer that we merge the verify and send logic into a single function and we should get slightly better performance this way.

Now that we removed dedup, I think there is no real reason to have separate `bls_sigverifier.rs` and `bls_sigverify_service.rs`. I can go ahead and merge these two (i.e. merge `BLSSigVerifyService::run` with `BLSSigVerifier::verify_and_send_batches`) though I'd prefer to do this (and remove the TODO above) in a follow-up if I can 🙏

core/src/bls_sigverify/bls_sigverify_service.rs
AshwinSekar: Yeah we only rate limit per connection but we have an overall limit of only staked connections that can be made so we get an upper bound.

I also see we have some logic here to dedup already verified things, so I feel like this should be enough without the deduper. we'll test some dos vectors in invalidator to make sure

core/src/bls_sigverify/bls_vote_sigverify.rs
AshwinSekar: just here is fine, thanks!

core/src/bls_sigverify/bls_cert_sigverify.rs
AshwinSekar: Makes sense, fine to do as a follow up

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+pub(crate) fn verify_and_send_certificates(
+    certs_buffer: &mut Vec<Certificate>,
+    bank: &Bank,
+    verified_certs: &RwLock<HashSet<CertificateType>>,
+    stats: &BLSSigVerifierStats,
+    message_sender: &Sender<ConsensusMessage>,
+) -> Result<(), BLSSigVerifyError> {
+    let results = verify_certificates(certs_buffer, bank, verified_certs, stats);
+
+    let valid_count = results.iter().filter(|&&valid| valid).count();
```
akhi3030, line 53: we are iterating over `results` once here and once again below.  I wonder if it might be cleaner to increment the counter below and then update the stats.  We may have to handle the disconnected error then.  However, I imagine that if the channel is disconnected then we are probably shutting down and missing some stats near the end is OK.

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    #[error("Cert Verification Error {0:?}")]
+    CertVerifyFailed(#[from] BlsCertVerifyError),
+
+    #[error("Not enough stake {0}: {1} < {2}")]
+    NotEnoughStake(u64, Fraction, Fraction),
+}
+
+pub(crate) fn verify_and_send_certificates(
+    certs_buffer: &mut Vec<Certificate>,
```
akhi3030, line 45: In line 58 below, we always drain the whole buffer.  Would it be cleaner for this argument to be just `Vec<Certificate>` then?

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+    stats
+        .total_valid_packets
+        .fetch_add(valid_count as u64, Ordering::Relaxed);
+
+    for (cert, is_valid) in certs_buffer.drain(..).zip(results) {
+        // Send the BLS certificate message to certificate pool.
+        if is_valid {
+            match message_sender.try_send(ConsensusMessage::Certificate(cert)) {
+                Ok(()) => {
+                    stats.sent.fetch_add(1, Ordering::Relaxed);
```
akhi3030, line 63: atomic operations have a big impact on the CPU performance.  Could we keep a local counter here and then do the atomic operation once after the for loop?

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+    Ok(())
+}
+
+fn verify_certificates(
+    certs_to_verify: &[Certificate],
+    bank: &Bank,
+    verified_certs: &RwLock<HashSet<CertificateType>>,
+    stats: &BLSSigVerifierStats,
+) -> Vec<bool> {
```
akhi3030, line 82: nit: this function is allocating and returning a vec that the caller is using to decide if we should send the msg or not.  If we wanted to optimise and not have the allocation, we could try to return an iterator here.  But then we get into nasty borrow checking issues.  The alternative is to combine the two functions into one.  Can also be done as a follow up of course.

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+        stats.received_verified.fetch_add(1, Ordering::Relaxed);
+        return Ok(());
+    }
+
+    let slot = cert_to_verify.cert_type.slot();
+    let Some((key_to_rank_map, total_stake)) = get_key_to_rank_map(bank, slot) else {
+        return Err(CertVerifyError::KeyToRankMapNotFound(slot));
+    };
+
+    let (required_stake_fraction, _) = certificate_limits_and_vote_types(&cert_to_verify.cert_type);
```
akhi3030, line 141: this logic has a lot of overlap with https://github.com/anza-xyz/alpenglow/pull/693/changes#diff-ed47b4a0198313377e091bb3957bbbc63d937805426d1b2b6de39d0a50d32a0cR5117.  That code has not landed yet but maybe we can add a note to try to remove duplicate code after both this and that PR land.

core/src/bls_sigverify/bls_cert_sigverify.rs
```diff
@@ -0,0 +1,163 @@
...
+    bank: &Bank,
+    verified_certs: &RwLock<HashSet<CertificateType>>,
+    stats: &BLSSigVerifierStats,
+) -> Result<(), CertVerifyError> {
+    if verified_certs
+        .read()
+        .unwrap()
+        .contains(&cert_to_verify.cert_type)
+    {
+        stats.received_verified.fetch_add(1, Ordering::Relaxed);
```
akhi3030, line 132: we will grab a lock for cert and we will do atomic ops for each cert.  A lot of atomic operations and barrier instructions.  I wonder if we can grab the lock once in `verify_certificates` and filter out the cached certs to minimise this.

core/src/bls_sigverify/bls_sigverify_service.rs
```diff
@@ -1,168 +1,165 @@
...
+                    ReceiveAction::Continue
+                }
+            },
+        }
     }
 
-    fn verifier<const K: usize>(
-        deduper: &Deduper<K, [u8]>,
-        recvr: &Receiver<PacketBatch>,
+    fn verify_packets(
```
akhi3030, line 124: I wonder if we should measure time even if this function returns an error.

We also have this handy macro https://github.com/anza-xyz/alpenglow/blob/68e6794eaee2552a85e1454bd1c7dd4fd2208365/measure/src/macros.rs#L88 that's used quite widely.

core/src/bls_sigverify/stats.rs
akhi3030, line 27: not part of the PR but this seems wrong.  It should be something like bls-packet?

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct VoteToVerify {
+    pub vote_message: VoteMessage,
+    pub bls_pubkey: BlsPubkey,
+    pub pubkey: Pubkey,
+}
+
+/// Verifies votes and sends verified votes to the consensus pool.
+/// Also returns a copy of the verified votes that the rewards container is interested is so that the caller can send them to it.
```
akhi3030, line 40:
```suggestion
/// Also returns a `AddVoteMessage` that the caller should use to send the verified votes to the rewards container.
```

outdated comment.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+                .then_some(vote)
+        })
+        .collect();
+    let add_vote_msg = AddVoteMessage { votes };
+
+    stats
+        .total_valid_packets
+        .fetch_add(verified_votes.len() as u64, Ordering::Relaxed);
+
+    let mut verified_votes_by_pubkey: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
```
akhi3030, line 70:
```suggestion
    let mut verified_votes_by_pubkey = HashMap::new();
```

type inference seems to work here.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+    for vote in verified_votes {
+        stats.received_votes.fetch_add(1, Ordering::Relaxed);
```
akhi3030, line 72: we could pull the atomic operation out of the loop.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+        if vote.vote_message.vote.is_notarization_or_finalization()
+            || vote.vote_message.vote.is_notarize_fallback()
+        {
+            let slot = vote.vote_message.vote.slot();
+            let cur_slots: &mut Vec<Slot> =
+                verified_votes_by_pubkey.entry(vote.pubkey).or_default();
+            if !cur_slots.contains(&slot) {
+                cur_slots.push(slot);
+            }
+        }
```
akhi3030, line 94: can we rename `verified_votes_by_pubkey` to something like `votes_for_repair` please?  Otherwise this logic doesn't immediately make sense.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+                    .votes_for_repair_sent_failed
+                    .fetch_add(1, Ordering::Relaxed);
+            }
+        }
+    }
+
+    Ok(add_vote_msg)
+}
+
+fn verify_votes(
```
akhi3030, line 128: I think there is scope for some simplification in this function.  I believe that this is copied over from the previous version.  I am happy to offer a follow up PR after this lands.

core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -80,159 +49,242 @@ pub struct BLSSigVerifier {
...
+        Ok(())
+    }
+
+    fn prune_caches(&mut self, root_bank: &Bank) {
+        self.last_checked_root_slot = root_bank.slot();
+        self.verified_certs
+            .write()
+            .unwrap()
+            .retain(|cert| cert.slot() > root_bank.slot());
+        self.vote_payload_cache
```
akhi3030, line 184: for reward purposes we process older votes too.  So this will drop entries from cache too quickly that are still useful.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -0,0 +1,242 @@
...
+    stats
+        .votes_batch_parallel_verify_count
+        .fetch_add(1, Ordering::Relaxed);
+    stats
+        .votes_batch_parallel_verify_elapsed_us
+        .fetch_add(votes_batch_parallel_verify_time.as_us(), Ordering::Relaxed);
+    verified_votes
+}
+
+fn get_vote_payload(
```
akhi3030, line 238: I wondered about this code in the previous iteration too.  We are serializing a very small struct here.  I am surprised that it is expensive enough that we need to have a cache for it.  We could also use wincode to speed things up.  Maybe we start without a cache and see where the bottleneck is?

akhi3030 (commented):
I am very happy for this to land as is and for the comments below to be addressed in separate follow up PRs.  I am also happy to help make the improvements here.  This is a big module and keeping this PR open till it is "perfect" will keep the PR open for too long.

I'll let @AshwinSekar decide though.

core/src/bls_sigverify/bls_sigverifier.rs
akhi3030: I think it makes sense to capture the vote after verification.  But it would also make sense to capture metrics in the consensus metrics container about how about many votes / certs we are receiving that fail verification.  So I propose that we move this recording to after verification and then as a follow up after the PR lands, I do another PR to capture metrics about failed verification votes and certs.

samkim-crypto: Okay great, I updated the code in [3da5b3f](https://github.com/anza-xyz/alpenglow/pull/703/commits/3da5b3f20c41a6ebf57ee027965e0336ca35e2c9).

AshwinSekar (approved):
From the discussion it looks like we'll tackle the outstanding items in a follow up PR.  Approving this refactor.

PR #706, opened 2026-02-02
[bls-cert-verify] Optimize bls-cert-verify and improve code quality

samkim-crypto:
#### Summary of Changes

Cleaning up the `solana-bls-cert-verify` crate.

[766cde8](https://github.com/anza-xyz/alpenglow/pull/706/commits/766cde8455a253d6cb870ee600756dd3d876865f): Benches for the certificate verification logic exists in `solana-core`, but I thought it made sense to add benches here too to identify any specific bottle necks related to certificate verification.

[735274f](https://github.com/anza-xyz/alpenglow/pull/706/commits/735274fbd22d77938741e4c57924031d4214ba6a): Updated the public key aggregation logic to take advantage of the mixed point additions.

The `solana-bls-signatures` v3.0 crate added support for mixed point additions (https://github.com/anza-xyz/solana-sdk/pull/506). Previously, we had to convert all points to projective before summing these together to aggregate public keys, but we can now add affine or even unserialized points directly to project points using a simpler group add formula.

Using mixed point additions, the bench results seem to improve quite drastically:

```
BLS Cert Verify/Base2_Notarize/1000
                        time:   [2.8434 ms 2.8479 ms 2.8525 ms]
                        change: [-92.215% -92.192% -92.171%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
BLS Cert Verify/Base3_NotarizeFallback/1000
                        time:   [3.5388 ms 3.5573 ms 3.5776 ms]
                        change: [-90.838% -90.788% -90.732%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  8 (8.00%) high mild
  2 (2.00%) high severe
```

I am glad that this gives quite a significant performance boost, but maybe we should have used mixed-point addition from the start.. 🙏 

[01c9687](https://github.com/anza-xyz/alpenglow/pull/706/commits/01c96873a2f38b67c80663615806f30643a91922): For the base3 certificate verification, we have to aggregate two sets of public keys (pertaining to primary votes and fallback votes). Currently, these two sets are aggregated sequentially. If we aggregate these two sets in parallel, then we get decent performance improvements, so I applied this change.

```
BLS Cert Verify/Base3_NotarizeFallback/1000
                        time:   [3.1656 ms 3.1710 ms 3.1765 ms]
                        change: [-11.390% -10.858% -10.366%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
```

The rest of the commits [db2b01e](https://github.com/anza-xyz/alpenglow/pull/706/commits/db2b01eda574fed311571e8dcc494fb5f37254c1), [315a77c](https://github.com/anza-xyz/alpenglow/pull/706/commits/315a77c919c72688c7805807c241171e4587ff6f), [8b3615f](https://github.com/anza-xyz/alpenglow/pull/706/commits/8b3615fb28f967073f99bd5291be658b911926f4), [f99187d](https://github.com/anza-xyz/alpenglow/pull/706/commits/f99187d39e4a53f609d81c3310ad37b6f5ee4307) are general refactoring and code quality improvements. In particular, I renamed the function `verify_cert_get_total_stake` to just `verify_certificate` because that seemed cleaner and follows a more standard rust naming convention, but I can revert this if others are opposed.

Finally, I added more precise bench logic in [6721329](https://github.com/anza-xyz/alpenglow/pull/706/commits/6721329a99411da5a86536bf5c189b8c2c5fff22) to include:
- A single BLS signature verification. This serves as a good reference since we can't possibly hope to beat a single sig verification
- Collect public key
- Aggregate public key

Here are the results:
```
| Validators | Collect Keys | Aggregate Keys | Verify Sig (Constant) | Total (Actual) | Overhead |
| **500** | 0.003 ms     | 0.85 ms        | 1.26 ms               | **2.26 ms** | ~0.15 ms |
| **1000** | 0.007 ms     | 1.46 ms        | 1.26 ms               | **2.88 ms** | ~0.15 ms |
| **1500** | 0.010 ms     | 1.92 ms        | 1.26 ms               | **3.39 ms** | ~0.20 ms |
| **2000** | 0.013 ms     | 2.51 ms        | 1.26 ms               | **3.97 ms** | ~0.19 ms |
```

Distribution:

```
| Validators | Aggregate % | Verify % | Collect % |
| **500**   | 38%         | 56%     | < 0.2%    |
| **1000** | 51%         | 44%      | < 0.3%    |
| **1500** | 57%         | 37%      | < 0.3%    |
| **2000** | 63%         | 32%      | < 0.3%    |
```

Roughly for 1000~1500 validators, the public key aggregation logic takes a little over 50% of the cost of verifying a certificate.

I created in issue https://github.com/anza-xyz/alpenglow/issues/708 to cut down public key aggregation even further. If we take approach 2, then I think public key aggregation can almost be an order of magnitude faster. I'll tackle this issue on a follow-up.

### DIFF
--- a/bls-cert-verify/benches/cert_verify.rs
+++ b/bls-cert-verify/benches/cert_verify.rs
@@ -111,7 +111,7 @@ fn bench_verify_cert(c: &mut Criterion) {
         let pubkeys_ref = &pubkeys;
 
         // Base2 Setup
-        // Assume 2/3rds of validataors sign
+        // Assume 2/3rds of validators sign
         let num_signers_base2 = (size * 2) / 3;
         let cert_base2 = create_base2_cert(&keypairs, num_signers_base2);
 


bls-cert-verify/benches/cert_verify.rs
```diff
@@ -0,0 +1,185 @@
...
+    let mut builder = CertificateBuilder::new(cert_type);
+    builder.aggregate(&vote_messages).unwrap();
+    builder.build().unwrap()
+}
+
+#[allow(clippy::arithmetic_side_effects)]
+fn bench_verify_cert(c: &mut Criterion) {
+    let mut group = c.benchmark_group("BLS Cert Verify");
+    let validator_sizes = [500, 1000, 1500, 2000];
+    const TEST_STAKE: u64 = 30; // assume each validator has stake 30 (arbitrary number)
```
akhi3030, line 91:
```suggestion
    // assume each validator has stake 30 (arbitrary number)
    const TEST_STAKE: u64 = 30; 
```

nit: this seems like the more rusty way to do comments.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -27,72 +27,105 @@ pub enum Error {
...
-        ),
+            accumulating_rank_map,
+        )
     }?;
+
     Ok(total_stake)
 }
 
-/// Verifies the [`signature`] of the [`payload`] which is signed by at most [`max_validators`] validators in the base2 encoded [`ranks`] using the [`rank_map`] to lookup the [`BlsPubkey`].
+fn get_vote_payloads(cert_type: &CertificateType) -> (Vote, Option<Vote>) {
```
akhi3030, line 101:
```suggestion
fn get_vote_payloads(cert_type: CertificateType) -> (Vote, Option<Vote>) {
```

nit: if a function takes something as a reference and then clones or copy it, it is better to expose that to the caller.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -27,72 +27,105 @@ pub enum Error {
...
+/// # Arguments
+///
+/// * `cert` - The certificate to verify.
+/// * `max_validators` - The maximum number of validators (used for decoding the bitmap).
+/// * `rank_map` - A closure that maps a validator's rank (index) to their stake and public key.
+///
+/// # Returns
+///
+/// On success, returns the total stake.
+pub fn verify_certificate(
```
akhi3030, line 60: I had originally picked the name `verify_cert_get_total_stake` and I am happy with this renaming.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -27,72 +27,105 @@ pub enum Error {
...
+            accumulating_rank_map,
+        )
     }?;
+
     Ok(total_stake)
 }
 
-/// Verifies the [`signature`] of the [`payload`] which is signed by at most [`max_validators`] validators in the base2 encoded [`ranks`] using the [`rank_map`] to lookup the [`BlsPubkey`].
+fn get_vote_payloads(cert_type: &CertificateType) -> (Vote, Option<Vote>) {
+    match *cert_type {
```
akhi3030, line 102: Factoring out this logic from the function does make it easier to read that function.  Just a note that now we are having to create an additional `Option` and we have two branches instead of one.  I doubt this is causing any perf hits but something to keep in mind if we feel like we need to squeeze more perf.

bls-cert-verify/src/cert_verify.rs
```diff
@@ -27,72 +27,105 @@ pub enum Error {
...
+            Some(Vote::new_skip_fallback_vote(slot)),
+        ),
+    }
+}
+
+fn serialize_vote(vote: &Vote) -> Vec<u8> {
+    // `expect` is safe because the `Vote` struct is composed entirely of primitive
+    // types (u64, Hash, enums) that are inherently serializable and it is constructed
+    // locally within this module.
+    bincode::serialize(vote).expect("Vote serialization should never fail for valid Vote structs")
```
akhi3030, line 123: I plan on introducing wincode serialization in bls sigverifier.  I think I will refactor this at the same time when I do that.

akhi3030 (dismissed):
Thanks, looks great!

bls-cert-verify/src/cert_verify.rs
samkim-crypto: Good call. Fixed!

bls-cert-verify/src/cert_verify.rs
samkim-crypto: I think the original name was fine as well! Just thought to keep it simple.

bls-cert-verify/src/cert_verify.rs
AshwinSekar: might want to bump wincode to the latest release as well (or do this when we upstream), think what we have available here is a little outdated, but maybe it has enough features for votes and bls

bls-cert-verify/benches/cert_verify.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    for &size in &validator_sizes {
+        let keypairs = create_bls_keypairs(size);
+
+        // Pre-calculate public keys to simulate efficient Bank lookup
+        let pubkeys: Vec<BlsPubkey> = keypairs.iter().map(|kp| kp.public.into()).collect();
+        let pubkeys_ref = &pubkeys;
+
+        // Base2 Setup
+        // Assume 2/3rds of validataors sign
```
ksn6: s/validataors/validators/

bls-cert-verify/src/cert_verify.rs
```diff
@@ -107,53 +140,28 @@ pub fn verify_base2<S: AsSignatureAffine>(
...
 
-/// Verifies the [`signature`] of [`payload`] and [`fallback_payload`] which is signed by the validators in the base3 encoded [`ranks`]  using the [`rank_map`] to lookup the [`BlsPubkey`].
+/// Verifies a signature for a split-vote scenario involving two distinct payloads.
 ///
-/// [`rank_map`] is [`FnMut`] allowing caller to perform computation based on which validators signed the payload.
-fn verify_base3(
+/// This function handles the "Base3" case where participating validators are split
+/// between signing a primary `payload` and a `fallback_payload`. This path is used
+/// only in rare edge cases where a fallback vote is required.
+pub fn verify_base3(
```
ksn6, line 164: This is now pub - is that intentional?

bls-cert-verify/src/cert_verify.rs
```diff
@@ -107,53 +140,28 @@ pub fn verify_base2<S: AsSignatureAffine>(
...
-where
-    F: FnMut(usize) -> Option<BlsPubkey>,
-{
-    move |ind: usize| {
-        let pos = ranks
-            .get(ind)
-            .unwrap_or_else(|| panic!("{ind} is not valid in {ranks:?}"));
-        assert!(pos == true, "{ind} is not set in {ranks:?}");
-        rank_map(ind)
-    }
```
ksn6, line 135: These checks seemed nice to have, since they'd panic only during tests - I guess because we're doing `iter_ones()` in `collect_pubkeys`, we no longer need these; is that right?

bls-cert-verify/src/cert_verify.rs
samkim-crypto: Yeah I just made the functions public to make it easier to bench them directly.

I guess we can use nightly to benchmark private functions, but since this crate only really makes sense in the agave setting, I thought I would just make it public. If you are opposed to this, then I can definitely limit the visibility and figure out another way to run the benches.

bls-cert-verify/src/cert_verify.rs
samkim-crypto: That's right. We have `iter_ones()`, so the extra check becomes redundant here.

PR #710, opened 2026-02-03
replay: switch incompatible banks on ParentReady

AshwinSekar:
#### Problem
We need to switch out banks for duplicate block scenarios.
From talking to Kobi the final rules we decided on are:
> Whenever I emit a ParentReady event for some parent-block b, I can interrupt execution/delete Banks on hand that prevent executing b, and commence executing the branch leading down to b.

> About finalization fixing: check the highest slot for which we have finalized anything, then don’t switch for lower/equal slots

Full doc https://docs.google.com/document/d/1s0ToP6JvkDjdY5SGffyxThEWq4_92-6tpyiAZsh_cv0/edit?tab=t.0

#### Summary of Changes
Implement switching logic on ParentReady filtered by highest finalization.

First commit is impl, second and third are for setting up a basic equivocation test

### DIFF
--- a/core/src/replay_stage.rs
+++ b/core/src/replay_stage.rs
@@ -718,7 +718,7 @@ impl ReplayStage {
                 unfrozen_gossip_verified_vote_hashes,
                 epoch_slots_frozen_slots,
             };
-            let mut pending_switches = HashMap::new();
+            let mut pending_switch = None;
 
             let (working_bank, in_vote_only_mode) = {
                 let r_bank_forks = bank_forks.read().unwrap();
@@ -886,7 +886,7 @@ impl ReplayStage {
                     Self::process_switch_bank_events(
                         &my_pubkey,
                         &switch_bank_receiver,
-                        &mut pending_switches,
+                        &mut pending_switch,
                         &blockstore,
                         &bank_forks,
                         &mut progress,
@@ -2134,7 +2134,7 @@ impl ReplayStage {
             .expect("must exist based on earlier check");
     }
 
-    /// Process switch block events from votor or replay.
+    /// Process switch block events from votor
     ///
     /// When receiving a switch request for block b we attempt to switch out the bank in slot(b)
     /// for b if the bank in slot(b) does not match the block id for b.
@@ -2145,42 +2145,41 @@ impl ReplayStage {
     ///
     /// Then to perform the switch for b and all of its ancestors identified above we:
     /// - Clear the existing bank (if one exists) in the slot from bank forks and progress
-    /// - Use `blockstore::switch_block_from_alternate` to atommically switch the shreds fetched
+    /// - Use `blockstore::switch_block_from_alternate` to atomically switch the shreds fetched
     ///   by informed repair into the original column
     ///
     /// At this point generate_new_bank_forks can replay the fork up to b
     ///
-    /// If there are multiple switch requests for a slot, we use the newest one.
+    /// We only care about the latest switch event. When deferring while waiting for repair we store
+    /// this in `pending_switch`
     fn process_switch_bank_events(
         my_pubkey: &Pubkey,
         switch_bank_receiver: &SwitchBankEventReceiver,
-        pending_switches: &mut HashMap<Slot, Hash>,
+        pending_switch: &mut Option<(Slot, Hash)>,
         blockstore: &Blockstore,
         bank_forks: &RwLock<BankForks>,
         progress: &mut ProgressMap,
     ) {
         let root = bank_forks.read().unwrap().root();
 
-        switch_bank_receiver
-            .try_iter()
-            .for_each(|switch_bank_event| {
-                let (slot, block_id) = switch_bank_event.block();
-                if slot <= root {
-                    return;
-                }
+        if let Some(switch_bank_event) = switch_bank_receiver.try_iter().last() {
+            let (slot, block_id) = switch_bank_event.block();
+            if slot <= root {
+                return;
+            }
 
-                // Overwrite any pending switches, later switches take precedence
-                if let Some(prev_block_id) = pending_switches.insert(slot, block_id) {
-                    info!(
-                        "{my_pubkey}: Overwriting previous switch request in {slot} for \
-                         {prev_block_id:?} to {block_id}"
-                    );
-                } else {
-                    info!("{my_pubkey}: Adding switch request in {slot} to {block_id}");
-                }
-            });
+            // Overwrite the pending switch, later switches take precedence
+            if let Some(prev_switch_request) = pending_switch.replace((slot, block_id)) {
+                info!(
+                    "{my_pubkey}: Overwriting previous switch request {prev_switch_request:?}
+                         with (slot, {block_id})"
+                );
+            } else {
+                info!("{my_pubkey}: Received switch request in {slot} to {block_id}");
+            }
+        };
 
-        pending_switches.retain(|slot, block_id| {
+        *pending_switch = pending_switch.filter(|(slot, block_id)| {
             if bank_forks.read().unwrap().block_id(*slot) == Some(*block_id) {
                 // Nothing to switch
                 return false;
--- a/local-cluster/tests/local_cluster.rs
+++ b/local-cluster/tests/local_cluster.rs
@@ -6261,6 +6261,7 @@ fn test_alpenglow_basic_equivocation() {
     a_validator_config.repair_handler_type = RepairHandlerType::Malicious(MaliciousRepairConfig {
         bad_shred_slot_frequency: Some(2),
         bad_shred_index_frequency: Some(32), // Only equivocate for indices where index % 32 == 0
+        slot_range: Some((0, 50)),           // Only for the first 50 slots
     });
 
     // Cluster config
@@ -6286,6 +6287,8 @@ fn test_alpenglow_basic_equivocation() {
     let cluster = LocalCluster::new_alpenglow(&mut cluster_config, SocketAddrSpace::Unspecified);
 
     // Ensure all nodes are rooting
+    // Although the low staked node might be behind while the leader is equivocating,
+    // once the leader stops equivocating it will be able to catch up
     cluster.check_for_new_roots(
         32,
         "test_alpenglow_basic_equivocation",
--- a/core/src/repair/malicious_repair_handler.rs
+++ b/core/src/repair/malicious_repair_handler.rs
@@ -21,6 +21,8 @@ pub struct MaliciousRepairConfig {
     pub bad_shred_slot_frequency: Option<Slot>,
     /// If set, respond maliciously for shred indices where `index % frequency == 0`
     pub bad_shred_index_frequency: Option<u64>,
+    /// If set, only respond maliciously for slots within this range (inclusive)
+    pub slot_range: Option<(Slot, Slot)>,
 }
 
 pub struct MaliciousRepairHandler {
@@ -49,6 +51,12 @@ impl MaliciousRepairHandler {
 
     /// Check if we should respond maliciously for this slot and shred index
     fn should_respond_maliciously(&self, slot: Slot, shred_index: u64) -> bool {
+        if let Some((start, end)) = self.config.slot_range {
+            if slot < start || slot > end {
+                return false;
+            }
+        }
+
         let slot_matches = self
             .config
             .bad_shred_slot_frequency


core/src/replay_stage.rs
```diff
@@ -857,28 +883,21 @@ impl ReplayStage {
...
                         &mut progress,
                     );
-                }
 
-                let forks_root = bank_forks.read().unwrap().root();
-                if !migration_status.is_alpenglow_enabled() {
+                    // Banks might have been switched above, these maps are no longer accurate
+                    drop(ancestors);
+                    drop(descendants);
+                } else if !migration_status.is_alpenglow_enabled() {
```
AshwinSekar: reordered so i could `else if` here as ancestors / descendants are no longer accurate so they need to be dropped

core/src/replay_stage.rs
```diff
@@ -4849,6 +4992,15 @@ impl ReplayStage {
                         continue;
                     }
 
+                    if Some(parent_meta.parent_block_id) != parent_bank.block_id()
```
AshwinSekar, line 4995: this is the AG equivalent of https://github.com/solana-foundation/solana-improvement-documents/pull/340

we'll still do 340 to help out firedancer but here in AG we can just validate the DMR matches what's in the header / update parent

ledger/src/blockstore.rs
```diff
@@ -1949,6 +1976,84 @@ impl Blockstore {
...
+            shreds,
+            None, // leader_schedule
+            true, // is_trusted
+            None, // should_recover_shreds
+            &mut BlockstoreInsertionMetrics::default(),
+        )
+        .expect("Blockstore insertion must succeed");
+    }
+
+    /// Switch the block in `slot` from an alternate location to the original location.
```
AshwinSekar, line 2006: I just did the simplest thing to switch banks in this hopefully never used scenario - copy the shreds.

We can of course replay from the alternate shred column directly but it requires a little more complexity:
- Modifying slot meta chaining to be aware of alt columns
- Changing generate_new_bank_forks / bank to be column aware
- Update blockstore_processor to replay entries from alt column
- RPC & ledger tool to serve data from alt columns.

I think for V1 we just do the dumbest thing and in the future revisit

local-cluster/tests/local_cluster.rs
```diff
@@ -6216,6 +6219,82 @@ fn test_alpenglow_imbalanced_stakes_catchup() {
     );
 }
 
+/// We start 2 nodes, where the first node A holds 90% of the stake.
+/// B has turbine disabled, and receives duplicate blocks through eager repair.
+/// However, through informed repair it is able to repair the correct blocks and keep up with A.
+#[test]
+#[serial]
+fn test_alpenglow_basic_equivocation() {
```
AshwinSekar, line 6227: super simple test that serves duplicate shreds via eager repair.

I did have a more complex test https://github.com/AshwinSekar/alpenglow/commit/7d8122a90610f3426aa4c93a2462ea9041b210d9#diff-5e9f940ea065adfd3066ef0d8ef0cfb5029b5ba478d96fb484b36954e464505cR6307

But unlike the simple test this requires multiple nodes to vote for consensus to progress. Test ended up being flaky because eager repair is too slow sometimes for the skip timeout.

I am working on a malicious broadcast impl instead which should work but I think it's probably best to just defer to internal test cluster for the complex tests - dunno if i should merge a *working* malicious broadcaster

core/src/validator.rs
```diff
@@ -266,6 +266,46 @@ pub struct GeneratorConfig {
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[repr(u8)]
+pub enum TurbineMode {
+    /// Normal operation - turbine and repair both enabled
+    #[default]
+    Enabled = 0,
+    /// Turbine disabled but repair still works (for equivocation tests)
+    TurbineDisabled = 1,
+    /// Both turbine and repair disabled (for partition tests)
+    TurbineAndRepairDisabled = 2,
```
AshwinSekar, line 279: this is the default behavior of the current atomic bool, however for the equivocation test we need repair to work so split this out.

existing tests all use TurbineAndRepairDisabled.

core/src/replay_stage.rs
```diff
@@ -822,7 +828,26 @@ impl ReplayStage {
                     migration_status.as_ref(),
                     &votor_event_sender,
                 );
+                replay_active_banks_time.stop();
```
bw-solana, line 831: bit of a re-definition on what replaying active banks means, but I'm fine with it

core/src/replay_stage.rs
```diff
@@ -822,7 +828,26 @@ impl ReplayStage {
...
                 let did_complete_bank = !new_frozen_slots.is_empty();
+
+                // Check if we've completed the migration conditions
+                if migration_status.is_ready_to_enable() {
```
bw-solana, line 835: is this tied to the switching banks changes or was this just a latent bug to not check migration status before handling alpenglow things?

core/src/replay_stage.rs
bw-solana: can this not just be an else?

core/src/replay_stage.rs
```diff
@@ -857,28 +882,21 @@ impl ReplayStage {
...
                         progress.handle_new_root(&bank_forks_r);
                     }
-                }
-                replay_active_banks_time.stop();
 
-                // Check if we've completed the migration conditions
-                if migration_status.is_ready_to_enable() {
-                    Self::enable_alpenglow(
-                        &exit,
+                    Self::process_switch_bank_events(
```
bw-solana, line 886: maybe we should have this return a bool to indicate if banks were switched. Could then tuck the drops in an if conditional on this return value

core/src/replay_stage.rs
```diff
@@ -4781,6 +4923,7 @@ impl ReplayStage {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
```
bw-solana: delete?

core/src/replay_stage.rs
```diff
@@ -4849,6 +4992,15 @@ impl ReplayStage {
                         continue;
                     }
 
+                    if Some(parent_meta.parent_block_id) != parent_bank.block_id()
+                        // Genesis doesn't have a block id
+                        && parent_slot != 0
+                    {
+                        // There were duplicate blocks in this slot and we have the wrong one replayed
+                        // Do not continue down this fork
```
bw-solana, line 5000: expectation is that votor/event handler will observe votes and kick off any necessary repair to get block IDs matching, right? Might be nice to indicate that in comment here

core/src/tvu.rs
```diff
@@ -403,6 +403,9 @@ impl Tvu {
             completed_slots_receiver,
         };
 
+        // Create switch block event channel for ReplayStage
+        let (switch_bank_sender, switch_bank_receiver) = bounded(100);
```
bw-solana, line 413: I'm assuming 100 is small enough while being way overprovisioned. Might be nice to add a comment here or make constant and comment near the constant

core/src/tvu.rs
```diff
@@ -475,7 +478,7 @@ impl Tvu {
             rpc_subscriptions: rpc_subscriptions.clone(),
             consensus_metrics_sender,
             migration_status: migration_status.clone(),
-            highest_finalized,
+            highest_finalized: highest_finalized.clone(),
```
bw-solana: need clone?

ledger/src/blockstore/blockstore_purge.rs
```diff
@@ -143,12 +143,28 @@ impl Blockstore {
...
+        self.purge_slot_cleanup_chaining_impl(slot, /* keep_alt_columns */ false)
+    }
+
+    /// Like `purge_slot_cleanup_chaining` but preserves alternate block columns.
+    /// Used when switching from an alternate block to allow repair data to be retained.
+    pub(crate) fn purge_slot_cleanup_chaining_keep_alt(&self, slot: Slot) -> Result<bool> {
+        self.purge_slot_cleanup_chaining_impl(slot, /* keep_alt_columns */ true)
+    }
+
+    fn purge_slot_cleanup_chaining_impl(&self, slot: Slot, keep_alt_columns: bool) -> Result<bool> {
```
bw-solana: I think this is more human readable if we flip the polarity:
`keep_alt_columns` --> `purge_alt_columns`

ledger/src/blockstore.rs
bw-solana: I think this should be okay... as you mentioned, should be extremely rare, so a little performance hiccup is fine-ish.

I'm slightly worried about holding the lock for a long time. Worst case, we're copying 32k shreds or 32MB of data x2 (once for orig, once for alt). The shred insertion pipe is stalled during this time, but shouldn't take more than milliseconds to do the fully copies

ledger/src/blockstore.rs
```diff
@@ -1949,6 +1976,84 @@ impl Blockstore {
...
+        assert!(
+            !matches!(location, BlockLocation::Original),
+            "Cannot switch from Original location"
+        );
+
+        let lock = self.insert_shreds_lock.lock().unwrap();
+
+        // 1. Backup the original block if needed
+        if let Some(dmr) = self.get_double_merkle_root(slot, BlockLocation::Original) {
+            let backup_location = BlockLocation::Alternate { block_id: dmr };
```
bw-solana, line 2026: we effectively have infinite `Alternate` locations, right? Because we index by slot+block_id. Just double checking there's no real space constraint here

ledger/src/blockstore.rs
```diff
@@ -1949,6 +1976,84 @@ impl Blockstore {
...
+            if self.get_double_merkle_root(slot, backup_location).is_none() {
```
bw-solana, line 2027: this could only be some if we're flipping back and forth or something, right?

ledger/src/blockstore.rs
```diff
@@ -1949,6 +1976,84 @@ impl Blockstore {
...
+        // 3. Copy shreds from alternate location to original
+        let alt_meta = self
+            .meta_from_location(slot, location)
+            .expect("Database read failed")
+            .expect("Alternate slot must have SlotMeta");
+        assert!(alt_meta.is_full(), "Alternate slot must be full");
+
+        self.copy_shreds_locked(&lock, slot, location, BlockLocation::Original);
+
+        // Verify the switch was successful
```
bw-solana: nit: `// 4.`

ledger/src/blockstore.rs
```diff
@@ -1949,6 +1976,84 @@ impl Blockstore {
...
+    }
+
+    /// Switch the block in `slot` from an alternate location to the original location.
+    /// This atomically:
+    /// 1. Back up the original column data if it's a valid block that we don't have
+    /// 2. Purges the original column data while preserving alternate columns
+    /// 3. Copies shreds from the alternate location to the original location
+    ///
+    /// Holds `insert_shreds_lock` for the entire operation.
+    pub fn switch_block_from_alternate(&self, slot: Slot, location: BlockLocation) {
```
bw-solana: nit: s/location/from_location/

ledger/src/blockstore.rs
```diff
@@ -3226,6 +3338,40 @@ impl Blockstore {
...
+    /// Gets all data shreds for a slot from the specified location.
+    /// Returns shreds in index order starting from `start_index`.
+    pub fn get_data_shreds_for_slot_from_location(
+        &self,
+        slot: Slot,
+        start_index: u64,
+        location: BlockLocation,
+    ) -> Result<Vec<Shred>> {
+        let Some(index) = self.get_index_from_location(slot, location)? else {
+            return Ok(Vec::new());
```
bw-solana, line 3345: Is this a valid scenario? Do we expect this API to be called before having all (or any?) shreds?

ledger/src/blockstore.rs
```diff
@@ -3226,6 +3338,40 @@ impl Blockstore {
...
+        };
+
+        index
```
bw-solana: should be able to pre-allocate a vec here. I believe the index has shred count.

of course, hopefully this is rare enough that we don't need it to be performant 😆

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
             .expect("must exist based on earlier check");
     }
 
+    /// Process switch block events from votor or replay.
+    ///
+    /// When receiving a switch request for block b we attempt to switch out the bank in slot(b)
+    /// for b if it the bank in slot(b) does not match the block id for b.
```
bw-solana: nit: delete 'it'

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
...
+    ///
+    /// When we need to switch a bank b, we first defer until we've repaired the ancestory of b:
+    /// - We must have block b and all of its ancestors up to any ancestor we've already replayed
+    /// - If no ancestor is replayed and it links back <= root, we can ignore this request
+    ///
+    /// Then to perform the switch for b and all of its ancestors identified above:
+    /// 1. If the existing turbine block in the slot is full, backup in Alternate column
+    /// 2. Purge any existing bank in the slot and clear the shreds from the turbine column
+    /// 3. Copy shreds from the Alternate column of the requested block to the Original column
```
bw-solana: redundant with the information in the blockstore comment. Can probably point to that function or leave out the details here

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
...
+    ) {
+        let root = bank_forks.read().unwrap().root();
+
+        switch_bank_receiver
+            .try_iter()
+            .for_each(|switch_bank_event| {
+                // Overwrite any pending switches, later switches take precedence
+                let (slot, block_id) = switch_bank_event.block();
+                if slot <= root {
+                    return;
```
bw-solana: `continue`?

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
...
+        blockstore: &Blockstore,
+        bank_forks: &RwLock<BankForks>,
+        progress: &mut ProgressMap,
+    ) {
+        let root = bank_forks.read().unwrap().root();
+
+        switch_bank_receiver
+            .try_iter()
+            .for_each(|switch_bank_event| {
+                // Overwrite any pending switches, later switches take precedence
```
bw-solana: this comment would probably be better down below (immediately above the insert)

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
...
+                        "{my_pubkey}: Overwriting previous switch request in {slot} for \
+                         {prev_block_id:?} to {block_id}"
+                    );
+                } else {
+                    info!("{my_pubkey}: Adding switch request in {slot} to {block_id}");
+                }
+            });
+
+        pending_switches.retain(|slot, block_id| {
+            if bank_forks.read().unwrap().block_id(*slot) == Some(*block_id) {
```
bw-solana, line 2183: I guess we should only have a couple entries here at a given time, but the rapid read lock acquisition here is... not my favorite

core/src/replay_stage.rs
```diff
@@ -2116,6 +2134,130 @@ impl ReplayStage {
...
+                return false;
+            }
+
+            // Check if we have received the block and all of its ancestors and collect the ones we
+            // need to switch out
+            let mut ancestor_slot = *slot;
+            let mut ancestor_block_id = *block_id;
+            let mut blocks_to_switch = vec![];
+            loop {
+                if ancestor_slot <= root {
```
bw-solana, line 2194: < makes sense here.

But `if `ancestor_slot` == root` and `blocks_to_switch` is non-empty, don't we want to just `break` here?

AshwinSekar: the loop has a little bit of a confusing flow (suggestions appreciated):
- On the first iteration if `slot == ancestor_slot == root` then we should ignore this request as it is an outdated attempt to switch out the root
- If we are actually traversing (`blocks_to_switch` is non-empty), notice that we only traverse if the condition on line 2221 is false. This means that `ancestor_slot == root` is not currently replayed in bank forks
  - In this case we are attempting to switch a fork that would end up switching out the root, so we should also ignore this request

The loop kind of has a structure of a more traditional `do while` loop where we only traverse until we have an ancestor replayed. When we get to the condition `ancestor_slot <= root` we know that block we want to switch in `ancestor_slot` is not replayed. So if it equals the root then we should ignore this request.

core/src/replay_stage.rs
AshwinSekar: yeah agreed. if you think it's worthwhile I can set up a cache similar to https://github.com/anza-xyz/agave/blob/master/runtime/src/bank_hash_cache.rs to avoid grabbing this read lock as frequently.

We should be grabbing the write lock roughly every 400ms, and *hopefully* only checking this read lock once every 1.6 seconds (one parent ready per leader window) so I figured it's fine to not cache, however i don't mind creating the simple cache.

core/src/replay_stage.rs
AshwinSekar: it's actually a return since we're inside a `for_each`

ledger/src/blockstore.rs
AshwinSekar: yep the more likely case is:
- we received garbage mixed shreds during turbine / eager repair
- this will set the blockstore.is_dead to true and *NOT* insert the shreds and fill out the DMR 
- thus we'll skip this backup step

In the rare case we received a coherent block and need to switch it out, this optimization is to avoid having to repair again if we need to switch back and forth later.

If you think it's overkill i'm fine omitting it

ledger/src/blockstore.rs
AshwinSekar: yep there's no explicit restriction on the # of `Alternate` locations in the code. 

If you check the doc from the PR description, Kobi was kind enough to write a proof on the upper bound of blocks we'd need to repair and thus a bound on the `Alternate` locations, so don't have to worry about any attack vectors trying to fill up this column

AshwinSekar: Adding https://github.com/anza-xyz/alpenglow/issues/722 so we can be notified if there's any bugs that cause us to repair more than 11 blocks in a slot

core/src/replay_stage.rs
AshwinSekar: The drops here are legacy mimicking the existing drops in replay stage after maybe_start_leader / dump_and_repair

They're a warning so that we don't accidentally add code past this. The idea is that switch_bank_events should be the last logic thing in the replay loop, I don't think we need to tuck in the drops but perhaps if we have some crazy use case in the future we can rework.

core/src/replay_stage.rs
AshwinSekar: no bug, just reordered it because i think it's cleaner:
(1) Check if we're ready to enable and potentially enable alpenglow
(2) if alpenglow -> ...
(3) if !alpenglow -> ...

Previously we had (1) and (2) switched which is kind of weird

core/src/replay_stage.rs
AshwinSekar: yeah just putting this back to how it was pre Alpenglow, replay_active_banks should just measure actual replay time.

I messed it up when initially adding the alpenglow stuff.

ledger/src/blockstore.rs
AshwinSekar: Yeah i'm hoping that copy is pretty fast. If leader submits duplicate blocks and is this successful such that we require a lot of switches, consensus is pretty stalled anyway for a bit.

Either way I'll add some metrics so we can get a gauge on things during testing https://github.com/anza-xyz/alpenglow/issues/723

core/src/replay_stage.rs
AshwinSekar: yeah d'oh 😅 
https://github.com/anza-xyz/alpenglow/pull/710/commits/a70c4da28f4611a5592a1bfb7b093ecf4774fe8d

core/src/replay_stage.rs
AshwinSekar: yep good catch https://github.com/anza-xyz/alpenglow/pull/710/commits/4f4f62a387f10a885cb5e4471597f2e6b9bccd69

core/src/replay_stage.rs
AshwinSekar: Yep that's correct, clairfied here
https://github.com/anza-xyz/alpenglow/pull/710/commits/b5fd8da63fb4b5d098ba1d42c5b577c5a3994a78

core/src/tvu.rs
AshwinSekar: Yep good call I added a comment here https://github.com/anza-xyz/alpenglow/pull/710/commits/5bc9e9205ebea3fc4e48c72da113ae56e6870c19

core/src/tvu.rs
AshwinSekar: 🤦 yeah we don't. Been using my clanker to do the boilerplate plumbing of things I need to the actual logic i implement. For whatever reason it's very clone happy haha. 

https://github.com/anza-xyz/alpenglow/pull/710/commits/0c99b4b393e307beef713431e105e87d1c2cab84

ledger/src/blockstore/blockstore_purge.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/710/commits/b3327d4b6a78030d62792b10ce7729da3fc91762

ledger/src/blockstore.rs
AshwinSekar: https://github.com/anza-xyz/alpenglow/pull/710/commits/638c5a4bd63be14472e9e548e6e27b2ef85b03e8

ledger/src/blockstore.rs
AshwinSekar: yeah good call, I also realized I should be just using the rocksdb iterator as it's better for performance https://github.com/anza-xyz/alpenglow/pull/710/commits/ee3dbdea7b37aef2c9778118fb0cf1374a773ae4

ledger/src/blockstore.rs
AshwinSekar: I think this is generic enough that it could be used elsewhere in the future. I'm not opposed to returning SlotUnavailable if you prefer

for the scope of this PR we check that the slot is full before we call it

core/src/replay_stage.rs
AshwinSekar: yeah good call https://github.com/anza-xyz/alpenglow/pull/710/commits/b2c15db29145a897aed2bb2bbd67a8da4ac5e1db

core/src/replay_stage.rs
AshwinSekar: https://github.com/anza-xyz/alpenglow/pull/710/commits/b2c15db29145a897aed2bb2bbd67a8da4ac5e1db

core/src/replay_stage.rs
AshwinSekar: done https://github.com/anza-xyz/alpenglow/pull/710/commits/b2c15db29145a897aed2bb2bbd67a8da4ac5e1db

ledger/src/blockstore.rs
bw-solana: I think it's fine 👍

core/src/replay_stage.rs
bw-solana: we've got enough to do, can punt on it

core/src/replay_stage.rs
bw-solana: I see.. the case I'm describing is impossible because we will hit the break down below, right?

bw-solana: Tucking your comment inside this if conditional would probably help:
```rust
if ancestor_slot <= root {
    // This is either (1) an outdated attempt to switch out the
    // root or (2) attempt to switch to a fork that would end up
    // switching out the root, so we should ignore
    return false;
}
```

AshwinSekar: > I see.. the case I'm describing is impossible because we will hit the break down below, right?

yep exactly, if it actually links to the version of `root` that we have replayed, we'll hit the break and perform the switch. this return false is only if it links to an alternate block in slot `root`

bw-solana (dismissed):

Issue #708, opened 2026-02-03
Speed up BLS cert verification (BLS public key aggregation)


samkim-crypto:
#### Problem

About half of the time spent for verifying alpenglow certs is public key aggregation.

Currently, every time we verify a certificate, we loop over all the relevant BLS public keys and perform:
1. Read: Read the 96 bytes
2. Deserialize & Check: Convert bytes to `G1Affine`. This is the expensive part. It validates that the point is on the curve and in the correct subgroup.
3. Math: Add `G1Affine` to the `PubkeyProjective` accumulator.

Since BLS public keys stored in vote accounts have already been verified via Proof of Possession, we can safely skip these expensive point checks during certificate verification.

#### Proposed Changes

There are two approaches to removing the redundant point checks:

Approach 1: Add `par_aggregate_unchecked` to the `solana-bls-signatures` crate.

- Pros: Keeps the optimization contained mostly within the crypto/cert libraries. No major changes to runtime data structures.
- Cons: Still paying the cost of deserializing bytes to curve points, just without the validation overhead.

Approach 2: Cache PubkeyAffine Points (The "Zero-Cost" Option)

We change how the Bank/Runtime stores these keys in memory. Instead of storing the Pubkey (bytes) in the in-memory representation of the Vote Account, we store the deserialized PubkeyAffine struct.

- Pros: Deserialization happens exactly once (at account load/creation). Aggregation becomes pure math (just adding points). This effectively eliminates the aggregation overhead entirely.
- Cons: Touches runtime structures and might increase memory usage slightly (Affine points are larger than compressed bytes).

PR #712, opened 2026-02-04
[core] Optimize bls-vote-sigverify and improve code quality

samkim-crypto:
#### Summary of Changes

The BLS vote verify logic is in a pretty rough shape, so I refactored the code. I also tried a bunch of optimization to see if there are ways to improve performance, so this PR took a while...

[ab31175](https://github.com/anza-xyz/alpenglow/pull/712/commits/ab3117512bbf75c6f4bd1ac36db4d586e8f20141): I first refactored the `verify_votes` function to sub-functions to be able to bench optimistic and fallback sigverify as well as signature and public key aggregations.

[501aecf](https://github.com/anza-xyz/alpenglow/pull/712/commits/501aecfcd9f2a0bed11b1adc0cb430530c0718df): The bench code that we had wasn't so useful, so I re-wrote the bench code focused on the vote verification logic. The vote verification performance depends on the batch size as well as the number of distinct messages, so I made the bench code run with different combinations of batch sizes and number of distinct messages. I am not sure yet how large these values are going to be in practice for the number of validator sets that we have, but we should be able to later update these test numbers easily.

[b5a0679](https://github.com/anza-xyz/alpenglow/pull/712/commits/b5a06796b94b9c1e93b6c4372e61fd698d340bc2): Ignore this PR. This was already done in https://github.com/anza-xyz/alpenglow/pull/717.

[fdc708b](https://github.com/anza-xyz/alpenglow/pull/712/commits/fdc708bead1d381b0f5d70153a81f23caf3dc9a8): Since we are now serializing/deserializing on the fly, I thought it made sense to use `Vote` as the key for the hashmap on public key aggregation. This allows us to also serialize deduplicated votes.

[6f72f85](https://github.com/anza-xyz/alpenglow/pull/712/commits/6f72f85e5fe7faf74ecff65cb2bad4f84e1d0ef9), [18a33a4](https://github.com/anza-xyz/alpenglow/pull/712/commits/18a33a467ba35e0d59f35f73d8cf5713019050ca), [3c711d9](https://github.com/anza-xyz/alpenglow/pull/712/commits/3c711d9b1379b919fb69a10e58e6f007d65c773f): These are just general refactoring to make the code more readable. One notable change though is that I moved the reward send logic to inside `bls_vote_verify.rs` since that seemed a bit cleaner.

[dcb8acb](https://github.com/anza-xyz/alpenglow/pull/712/commits/dcb8acb70b457421bb9633886f80bd6863d0f139): Parallelized public key and signature aggregation. This seemed to give about ~7-10% improvement in optimistic sigverify.

[eb7d3bd](https://github.com/anza-xyz/alpenglow/pull/712/commits/eb7d3bd92601f92c7f87da93f9d787a89c565725): I wrote some todo's to improve public key and signature aggregation. This is along the lines of https://github.com/anza-xyz/alpenglow/issues/708. We should be able to cut these aggregation costs by quite a bit, but will tackle this in a follow-up (will also need to update the `solana-bls-signatures` crate).

Other than optimizing aggregation, I have a couple more ideas I want to try:
- For BLS signature verification we need to hash the message to the curve point in an expensive way. Assuming that a lot of the votes are going to be on the same message, we should be able to cache these hashes.
- Once aggregation optimization and caching hashes are in, we should measure precisely how the fallback sigverify compares to the optimistic sigverify. For small batch sizes, the fallback performs better, so maybe we can have logic to proceed directly to fallback verification.

This PR leaves the cert verify logic and doesn't yet address https://github.com/anza-xyz/alpenglow/pull/703#pullrequestreview-3734323265. I'll address these in a separate follow-up.

### DIFF
--- a/core/src/bls_sigverify/stats.rs
+++ b/core/src/bls_sigverify/stats.rs
@@ -129,13 +129,6 @@ pub(crate) struct BLSSigVerifierStats {
 
 impl Default for BLSSigVerifierStats {
     fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl BLSSigVerifierStats {
-    #[cfg_attr(feature = "dev-context-only-utils", qualifiers(pub))]
-    pub(crate) fn new() -> Self {
         Self {
             total_valid_packets: AtomicU64::new(0),
 
@@ -169,7 +162,9 @@ impl BLSSigVerifierStats {
             consensus_reward_send_failed: AtomicU64::new(0),
         }
     }
+}
 
+impl BLSSigVerifierStats {
     /// If sufficient time has passed since last report, report stats.
     pub(crate) fn maybe_report_stats(&mut self) {
         let now = Instant::now();
@@ -305,6 +300,6 @@ impl BLSSigVerifierStats {
                 i64
             ),
         );
-        *self = BLSSigVerifierStats::new();
+        *self = BLSSigVerifierStats::default();
     }
 }
--- a/core/benches/bls_vote_sigverify.rs
+++ b/core/benches/bls_vote_sigverify.rs
@@ -105,7 +105,7 @@ fn bench_verify_votes_optimistic(c: &mut Criterion) {
 
     for (batch_size, num_distinct) in get_matrix_params() {
         let votes = generate_test_data(num_distinct, batch_size);
-        let stats = BLSSigVerifierStats::new();
+        let stats = BLSSigVerifierStats::default();
         let label = format!("msgs_{num_distinct}/batch_{batch_size}");
 
         group.bench_function(&label, |b| {
@@ -122,7 +122,7 @@ fn bench_aggregate_pubkeys(c: &mut Criterion) {
 
     for (batch_size, num_distinct) in get_matrix_params() {
         let votes = generate_test_data(num_distinct, batch_size);
-        let stats = BLSSigVerifierStats::new();
+        let stats = BLSSigVerifierStats::default();
         let label = format!("msgs_{num_distinct}/batch_{batch_size}");
 
         group.bench_function(&label, |b| {
@@ -158,7 +158,7 @@ fn bench_verify_individual_votes(c: &mut Criterion) {
     for &batch_size in BATCH_SIZES {
         // Distinctness doesn't affect the cost of N individual verifications.
         let votes = generate_test_data(1, batch_size);
-        let stats = BLSSigVerifierStats::new();
+        let stats = BLSSigVerifierStats::default();
         let label = format!("batch_{batch_size}");
 
         group.bench_function(&label, |b| {
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -75,7 +75,7 @@ impl BLSSigVerifier {
             votes_for_repair_sender,
             reward_votes_sender,
             message_sender,
-            stats: BLSSigVerifierStats::new(),
+            stats: BLSSigVerifierStats::default(),
             verified_certs: RwLock::new(HashSet::new()),
             consensus_metrics_sender,
             last_checked_root_slot: 0,


samkim-crypto:
The CI is red due to a rust security advisory (unrelated to the PR changes), but this PR should be ready for review 🙏

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -21,19 +22,36 @@ use {
...
+        };
+        self.bls_pubkey
+            .verify_signature(&self.vote_message.signature, &payload)
+            .is_ok()
+    }
+}
+
 /// Verifies votes and sends verified votes to the consensus pool.
-/// Also returns a copy of the verified votes that the rewards container is interested is so that the caller can send them to it.
+/// Also returns a `AddVoteMessage` that the caller should use to send the
```
akhi3030: should this be updated as the function is not returning anything anymore.  I guess we could say that the verified votes are sent on the `reward_votes_sender` channel.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -56,54 +109,86 @@ pub(crate) fn verify_and_send_votes(
...
+        }
+        Err(TrySendError::Disconnected(_)) => {
+            warn!("could not send votes to reward container, receive side of channel is closed");
+        }
+    }
+}
 
-    let mut verified_votes_by_pubkey: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
+#[allow(clippy::too_many_arguments)]
+fn process_and_send_votes_to_consensus(
```
akhi3030, line 132: nit: can we get some documentation on this function please.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -56,54 +109,86 @@ pub(crate) fn verify_and_send_votes(
...
-        }
+        let vote_msg = vote.vote_message;
+        let slot = vote_msg.vote.slot();
+
         consensus_metrics.push(ConsensusMetricsEvent::Vote {
             id: vote.pubkey,
             vote: vote.vote_message.vote,
         });
 
+        if vote.vote_message.vote.is_notarization_or_finalization() {
```
akhi3030, line 151: Follow up: I think it will be cleaner if we had some kind of `Repair::wants_vote()` function here instead to encapsulate the logic of repair in one place.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -56,54 +109,86 @@ pub(crate) fn verify_and_send_votes(
...
+        }
+
         if vote.vote_message.vote.is_notarization_or_finalization()
             || vote.vote_message.vote.is_notarize_fallback()
         {
-            let slot = vote.vote_message.vote.slot();
-            let cur_slots: &mut Vec<Slot> =
-                verified_votes_by_pubkey.entry(vote.pubkey).or_default();
+            let cur_slots: &mut Vec<Slot> = votes_for_repair.entry(vote.pubkey).or_default();
             if !cur_slots.contains(&slot) {
```
akhi3030, line 162: Follow up: Hmm... would be great if we used a `HashSet` instead but it seems like that will require changing the type of the channel, etc.  So we could attempt to do that in a follow up.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -56,54 +109,86 @@ pub(crate) fn verify_and_send_votes(
...
-    let mut verified_votes_by_pubkey: HashMap<Pubkey, Vec<Slot>> = HashMap::new();
+#[allow(clippy::too_many_arguments)]
+fn process_and_send_votes_to_consensus(
+    verified_votes: &[VoteToVerify],
+    message_sender: &Sender<ConsensusMessage>,
+    last_voted_slots: &mut HashMap<Pubkey, Slot>,
+    consensus_metrics: &mut Vec<ConsensusMetricsEvent>,
+    stats: &BLSSigVerifierStats,
+) -> Result<HashMap<Pubkey, Vec<Slot>>, BLSSigVerifyError> {
+    let mut votes_for_repair = HashMap::new();
```
akhi3030, line 139: Follow up: not super happy with the name of the variable as it seems to be a map from `Pubkey` to `Slot`s.  However, the current name is fine given the name of the channel, etc.  So I think we could consider renaming if we improve the channel name, etc. in a follow up.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -21,19 +22,36 @@ use {
...
+    }
+}
+
 /// Verifies votes and sends verified votes to the consensus pool.
-/// Also returns a copy of the verified votes that the rewards container is interested is so that the caller can send them to it.
+/// Also returns a `AddVoteMessage` that the caller should use to send the
+/// verified votes to the rewards container.
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn verify_and_send_votes(
     votes_to_verify: &[VoteToVerify],
```
akhi3030, line 55: Follow up: it should be possible to use `Vec<VoteToVerify>` to reduce allocations happening in this function.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -127,96 +210,142 @@ fn verify_votes(
...
-
     stats.votes_batch_count.fetch_add(1, Ordering::Relaxed);
+
+    // Try optimistic verification - fast to verify, but cannot identify invalid votes
+    if verify_votes_optimistic(votes_to_verify, stats) {
+        return votes_to_verify.to_vec();
+    }
+
+    // Fallback to individual verification
+    verify_votes_fallback(votes_to_verify, stats)
```
akhi3030: nit: the name of the function sounds like we will be validating votes like notar fallback, etc.  How about `verify_votes_individually` instead?

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -226,3 +355,7 @@ fn verify_votes(
         .fetch_add(votes_batch_parallel_verify_time.as_us(), Ordering::Relaxed);
     verified_votes
 }
+
+fn get_vote_payload(vote: &Vote) -> Arc<Vec<u8>> {
```
akhi3030: I think this snuck in after being removed.  We can keep the function but we don't need to return the `Arc`

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -127,96 +210,142 @@ fn verify_votes(
...
 
-    let (distinct_payloads, distinct_pubkeys): (Vec<_>, Vec<_>) =
-        grouped_pubkeys.into_iter().unzip();
-    let aggregate_pubkeys_result: Result<Vec<PubkeyProjective>, _> = distinct_pubkeys
-        .into_iter()
-        .map(|pks| PubkeyProjective::par_aggregate(pks.into_par_iter()))
+    let (distinct_vote_structs, distinct_pubkeys_groups): (Vec<_>, Vec<_>) =
+        grouped_votes.into_iter().unzip();
+
+    let distinct_payloads: Vec<Arc<Vec<u8>>> = distinct_vote_structs
```
akhi3030: This wouldn't have an `Arc` on it if we update `get_vote_payload` and then we also do not need to create a `Vec` here.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -21,19 +22,36 @@ use {
...
+            .is_ok()
+    }
+}
+
 /// Verifies votes and sends verified votes to the consensus pool.
-/// Also returns a copy of the verified votes that the rewards container is interested is so that the caller can send them to it.
+/// Also returns a `AddVoteMessage` that the caller should use to send the
+/// verified votes to the rewards container.
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn verify_and_send_votes(
```
akhi3030, line 54: follow up: this function should return its own stats struct instead of taking the general one.  Then we could eliminate all the atomic operations to keep the stats updated.

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -43,11 +61,46 @@ pub(crate) fn verify_and_send_votes(
     leader_schedule: &LeaderScheduleCache,
     message_sender: &Sender<ConsensusMessage>,
     votes_for_repair_sender: &VerifiedVoteSender,
+    reward_votes_sender: &Sender<AddVoteMessage>,
     last_voted_slots: &mut HashMap<Pubkey, Slot>,
     consensus_metrics: &mut Vec<ConsensusMetricsEvent>,
-) -> Result<AddVoteMessage, BLSSigVerifyError> {
+) -> Result<(), BLSSigVerifyError> {
```
akhi3030, line 65: follow up: I think we should return a more specific error here instead of the general one.  It will make tracking down issues when debugging a bit easier.  It seems like there aren't that many different types of error this function can return other than channel disconnected.

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Ah I missed this stale docs. Updated!

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Updated!

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Yeah I was mainly working with the assumption that the channel is fixed, but I can definitely address it in a follow-up.

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Good point! I renamed it to `verify_individual_votes` to make it slightly more inline with the rest of the function names.

core/src/bls_sigverify/bls_vote_sigverify.rs
samkim-crypto: Yeah great point. I removed the `Arc` and also removed the intermediate `Vec`.

samkim-crypto:
Thanks for all the useful comments @akhi3030! As you suggested, I'll address the "follow-up" suggestions on a follow-up PR 🙏

core/src/bls_sigverify/bls_vote_sigverify.rs
```diff
@@ -127,96 +215,140 @@ fn verify_votes(
...
+    // Try optimistic verification - fast to verify, but cannot identify invalid votes
+    if verify_votes_optimistic(votes_to_verify, stats) {
+        return votes_to_verify.to_vec();
+    }
+
+    // Fallback to individual verification
+    verify_individual_votes(votes_to_verify, stats)
+}
+
+#[cfg_attr(feature = "dev-context-only-utils", qualifiers(pub))]
```
akhi3030, line 229: TIL.  This is quite useful to know!

core/src/bls_sigverify/stats.rs
```diff
@@ -121,10 +124,17 @@ pub(crate) struct BLSSigVerifierStats {
     pub(crate) received_votes: AtomicU64,
     pub(crate) last_stats_logged: Instant,
 
-    pub(super) consensus_reward_send_failed: u64,
+    pub(super) consensus_reward_send_failed: AtomicU64,
+}
+
+impl Default for BLSSigVerifierStats {
+    fn default() -> Self {
```
akhi3030, line 131: Should we get rid of the `new()` and inline the code here?  We don't need both.

akhi3030 (dismissed):
LGTM!  Looks great.  I am happy for this to land.  Found one minor nit about a `Default` impl.  But that can be addressed here or later.

core/src/bls_sigverify/stats.rs
samkim-crypto: Oh yeah good catch. I just updated this in this PR.

akhi3030 (approved):

PR #714, opened 2026-02-04
consensus_pool: handle intra-window safe to notar

AshwinSekar:
#### Problem
We currently don't handle the intrawindow safe to notar case specified here:
image

Now that we have the repair service working, we can handle it.

#### Summary of Changes
For intrawindow blocks:
- Do not immediately emit the event, add to pending safe to notar set
- Send repair request
- Wait until block has been received
- Check if the parent is notarize-fallback (or stronger)
- Emit the event

### DIFF
--- a/votor/src/consensus_pool_service.rs
+++ b/votor/src/consensus_pool_service.rs
@@ -306,6 +306,14 @@ impl ConsensusPoolService {
                 return Self::handle_channel_disconnected(&mut ctx, "Votor event receiver");
             }
 
+            let wait_timeout = if pending_safe_to_notar.is_empty() {
+                Duration::from_secs(1)
+            } else {
+                // If there are pending blocks that are waiting for repair in order to emit
+                // SafeToNotar events, use a shorter timeout
+                Duration::from_millis(20)
+            };
+
             let messages: Vec<ConsensusMessage> = select! {
                 recv(ctx.consensus_message_receiver) -> msg => {
                     let Ok(first) = msg else {
@@ -313,7 +321,7 @@ impl ConsensusPoolService {
                     };
                     std::iter::once(first).chain(ctx.consensus_message_receiver.try_iter()).collect()
                 },
-                default(Duration::from_secs(1)) => continue
+                default(wait_timeout) => continue
             };
 
             for message in messages {


votor/src/consensus_pool_service.rs
```diff
@@ -436,6 +451,72 @@ impl ConsensusPoolService {
...
+            // If repair hasn't been sent, send it now
+            if !*repair_sent {
+                let repair_event = RepairEvent::FetchBlock { slot, block_id };
+                if ctx.repair_event_sender.try_send(repair_event).is_ok() {
+                    *repair_sent = true;
+                    stats.pending_safe_to_notar_repair_sent += 1;
+                }
+            }
+
+            // Check if we've received the full block in blockstore
```
AshwinSekar, line 505: this could be optimized in the future:
- If we receive the block through turbine we need the full block (to check block_id matches) before we can check the parent
- If we're receiving the block through informed repair, technically all we need is the results of the first request ParentAndFecSetCount to be able to check the parent, don't need to wait for the full block to be ingested

local-cluster/tests/local_cluster.rs
```diff
@@ -7398,7 +7398,7 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
 /// - Protocol determines it's "SafeToNotar" under condition 2 and issues NotarizeFallback
 ///
 /// ## Phase 3: Recovery and Liveness Verification
-/// After observing 5 NotarizeFallback votes from Node C:
```
AshwinSekar: because turbine & repair are disabled for C in this test, it now can only notarize-fallback the first block, as the intrawindow blocks require us to repair to check parent

Once https://github.com/anza-xyz/alpenglow/pull/715 is merged I'll revisit this test to create a setup that allows repair to work while keeping turbine disabled

local-cluster/tests/local_cluster.rs
```diff
@@ -7410,6 +7410,7 @@ fn test_alpenglow_ensure_liveness_after_intertwined_notar_and_skip_fallbacks() {
 /// - Recovery is seamless once partition is resolved
 #[test]
 #[serial]
+#[ignore]
```
AshwinSekar: because turbine & repair are disabled for C in this test, it now can only notarize-fallback the first leader block, as the intrawindow blocks require us to repair to check parent (previously we just voted notarize-fallback w/o having the block).

This doesn't work for this test, because we can't fully control when turbine/repair are disabled (could end up taking effect intrawindow), and the test requires at least 5 slots of notarize-fallbacks - we can only do 1 if we're lucky.

Once https://github.com/anza-xyz/alpenglow/pull/715 is merged I'll revisit this test to create a setup that allows repair to work while keeping turbine disabled.

votor/src/consensus_pool/slot_stake_counters.rs
```diff
@@ -70,13 +78,22 @@ impl SlotStakeCounters {
...
-                events.push(VotorEvent::SafeToNotar((slot, *block_id)));
-                stats.event_safe_to_notarize = stats.event_safe_to_notarize.saturating_add(1);
                 self.safe_to_notar_sent.push(*block_id);
+
+                if is_first_in_leader_window {
+                    // First block in leader window - emit event immediately
+                    events.push(VotorEvent::SafeToNotar((slot, *block_id)));
+                    stats.event_safe_to_notarize = stats.event_safe_to_notarize.saturating_add(1);
+                } else {
+                    // Intrawindow block - add to pending for later processing
```
akhi3030, line 94: should we add something in stats for this case?

local-cluster/tests/local_cluster.rs
AshwinSekar: #715  was merged and this was fixed in https://github.com/anza-xyz/alpenglow/pull/714/commits/66dd9623f1f3c3ec0a0bd605d2058d20c912ff66

akhi3030 (approved):

PR #715, opened 2026-02-04
testing: add turbine disabled but repair enabled mode

AshwinSekar:
Split from #710 

#### Problem
Now that we have repair working, it would be nice to write some tests that have turbine disabled but repair enabled

#### Summary of Changes
Extend the turbine disabled atomic to support this mode. All existing tests use the "both disabled" mode


core/src/shred_fetch_stage.rs
```diff
@@ -69,7 +72,7 @@ impl ShredFetchStage {
         name: &'static str,
         flags: PacketFlags,
         repair_context: Option<&RepairContext>,
-        turbine_disabled: Arc<AtomicBool>,
+        turbine_mode: Arc<AtomicU8>,
```
akhi3030: nit: it seems that we can pass a `&AtomicU8` here to save some atomic operations.  We could also consider passing in a `&TurbineMode` to get some more type safety.

core/src/validator.rs
```diff
@@ -266,6 +266,46 @@ pub struct GeneratorConfig {
     pub starting_keypairs: Arc<Vec<Keypair>>,
 }
 
+/// Controls turbine and repair behavior for testing network partitions.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+#[repr(u8)]
+pub enum TurbineMode {
```
akhi3030: It would be nice if this struct hid the `AtomicU8` from the users.  Something like:

```
enum TurbineModeKind {
  // the same enum as now.
}

struct TurbineMode {
  mode: Arc<AtomicU8>
}

impl TurbineMode {
  fn get_mode(&self)...

  fn set_mode(&self)...
}
```

core/src/shred_fetch_stage.rs
AshwinSekar: hid it as TurbineMode for type safety. We still end up having to check the underlying atomic in the outer loop so not sure if we're saving any atomic operations

akhi3030: I meant we would save some increment and decrements by not having to clone the `Arc`.

local-cluster/tests/local_cluster.rs
```diff
@@ -6944,12 +6944,8 @@ fn test_alpenglow_ensure_liveness_after_double_notar_fallback() {
             );
         }
 
-        fn handle_node_c_vote(
-            &mut self,
-            vote: &Vote,
-            node_c_turbine_disabled: &Arc<AtomicBool>,
-        ) -> bool {
-            let turbine_disabled = node_c_turbine_disabled.load(Ordering::Acquire);
```
akhi3030, line 6952: hmm... I wonder if there is a reason this was `Acquire`?  I cannot see a reason for it yet but I might be missing something.

AshwinSekar: hmm i'm not actually sure. I do see that this test was previously flaky and apparently I i `#[ignore]`-ed it https://github.com/anza-xyz/alpenglow/pull/260

Perhaps @ksn6 can weigh in. I'll merge this for now and we can look to reenable this test.

PR #716, opened 2026-02-05
Makes ` get_key_to_rank_map` a method on the `Bank`

akhi3030:
### Problem

In the bls sigverifier and in consensus rewards we need to look up the rank map from the bank.  Both modules had a bunch of duplicate code to handle functionality

### Summary of change

- Introduces `Bank::get_rank_map` that the two modules are using
- Removes the caching of the rank maps in the sigverifier.  It doesn't seem to be saving much in terms of compute and only adds complexity.  We also do not have any data to suggest that this caching is helping with a bottleneck
- Misc. cleanup in consensus rewards where some modules do not need to rely on the rank map being wrapped inside an `Arc`

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -303,10 +303,15 @@ impl BLSSigVerifier {
         vote_message: &VoteMessage,
         root_bank: &Bank,
     ) -> Option<(Pubkey, BlsPubkey)> {
-        let key_to_rank_map = root_bank.get_rank_map(vote_message.vote.slot())?;
+        let Some(rank_map) = root_bank.get_rank_map(vote_message.vote.slot()) else {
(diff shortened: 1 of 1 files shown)


core/src/bls_sigverify/bls_sigverifier.rs
```diff
@@ -36,6 +36,11 @@ use {
     },
 };
 
+fn get_key_to_rank_map(bank: &Bank, slot: Slot) -> Option<(&Arc<BLSPubkeyToRankMap>, u64)> {
```
AshwinSekar: wdyt about making this a fn on `bank` instead? Seemes generic enough that it doesn't have to be tied to the sigverifier

akhi3030: good idea... this actually leads to quite a bit of cleanup.

PR #717, opened 2026-02-05
Removes caching of serialized data

akhi3030:
#### Problem

bls sigverifier is currently caching serialized `Vote`.  This cache is also shared between multiple threads.  The struct being serialized is quite trivial so it should be relative fast and sharing the cache between threads is probably adding some additional overhead.  Finally we do not have any concrete profiling data to suggest that this is a bottleneck.

#### Summary of Changes

Removes the caching.  This simplifies the code and also removes some allocations.

### DIFF
--- a/core/src/bls_sigverify/bls_sigverifier.rs
+++ b/core/src/bls_sigverify/bls_sigverifier.rs
@@ -27,7 +27,6 @@ use {
     solana_votor_messages::{
         consensus_message::{Certificate, CertificateType, ConsensusMessage, VoteMessage},
         reward_certificate::AddVoteMessage,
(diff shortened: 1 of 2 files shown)


samkim-crypto (approved):
Yes, I removed it in https://github.com/anza-xyz/alpenglow/pull/712 ([b9ff168](https://github.com/anza-xyz/alpenglow/pull/712/commits/b9ff1689ce18ac94fe696ac4852c7dd2cf710d4c)) as well. I ran the benches and it doesn't seem to impact performance that much.

You can merge this and then I can rebase in https://github.com/anza-xyz/alpenglow/pull/712.

PR #725, opened 2026-02-09
Pays the vote accounts the calculated vote rewards

akhi3030:
#### Problem

In https://github.com/anza-xyz/alpenglow/pull/694, we calculate the vote rewards.  We now need to pay the rewards.

#### Summary of Changes

- Sometimes the reward slot might be in an older epoch than the current slot, so we need to store information about 2 epochs in the `VoteRewardAccountState`
- Renames `VoteRewardAccountState` to `EpochInflationAccountState`
- Moves `EpochInflationAccountState` into its own module
- Improves testing for `EpochInflationAccountState`
- Improves the errors returned from calculating and paying rewards
- Looks up the accounts from the bank and pays the rewards into them.
- Temporarily explicitly refers to `VoteStateV4` which will be cleaned up when we upstream.

### DIFF
--- a/runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
+++ b/runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
@@ -87,11 +87,7 @@ impl EpochInflationAccountState {
     pub(super) fn set_state(&self, bank: &Bank) {
         // TODO: use wincode instead.
         let account_size = bincode::serialized_size(&self).unwrap();
-        let lamports = bank
-            .rent_collector()
-            .rent
-            .minimum_balance(account_size.try_into().unwrap())
-            .max(1);
+        let lamports = bank.get_minimum_balance_for_rent_exemption(account_size as usize);
         let account = AccountSharedData::new_data(lamports, &self, &system_program::ID).unwrap();
         bank.store_account_and_update_capitalization(&VOTE_REWARD_ACCOUNT_ADDR, &account);
     }
@@ -174,10 +170,7 @@ impl EpochInflationAccountState {
             prev: None,
         };
         let account_size = bincode::serialized_size(&state).unwrap();
-        bank.rent_collector()
-            .rent
-            .minimum_balance(account_size as usize)
-            .max(1)
+        bank.get_minimum_balance_for_rent_exemption(account_size as usize)
     }
 }
 


runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,77 +156,129 @@ impl VoteRewardAccountState {
...
+
+fn pay_reward(bank: &Bank, pubkey: Pubkey, account: VoteAccount, reward: u64) {
+    let data = account.account().data_clone();
+    // XXX: need to use V4 here.
+    let mut vote_state: VoteStateV3 = bincode::deserialize(&data).unwrap();
+    vote_state.increment_credits(bank.epoch(), reward);
+    let shared_account =
+        AccountSharedData::new_data(account.lamports(), &vote_state, account.owner()).unwrap();
+    // XXX: is the stake getting set properly?
+    bank.store_account(&pubkey, &shared_account);
```
AshwinSekar: we should use `store_accounts` for better performance, so would change the flow here to first load all the vote accounts, modify, and then store

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,77 +156,129 @@ impl VoteRewardAccountState {
...
+    // As per the Alpenglow SIMD, the rewards are split equally between the validators and the leader.
+    let validator_reward_lamports = reward_lamports / 2;
+    let leader_reward_lamports = reward_lamports - validator_reward_lamports;
+    (validator_stake_lamports, leader_reward_lamports)
+}
+
+fn pay_reward(bank: &Bank, pubkey: Pubkey, account: VoteAccount, reward: u64) {
+    let data = account.account().data_clone();
+    // XXX: need to use V4 here.
+    let mut vote_state: VoteStateV3 = bincode::deserialize(&data).unwrap();
```
AshwinSekar: yes we'll need to use V4 here, VoteStateV4 is in the process of being activated (just was activated on testnet).

You'll want to use something like this https://github.com/anza-xyz/agave/blob/527a52e794af3b8d8eba7d63a8cb751034a12747/programs/vote/src/vote_state/mod.rs#L75 which will convert V3 / V4 into a handler type that you can use to increment credits. 

At the moment you can ignore (don't pay rewards) for anything that fails this deserialization + conversion pathway. Once fully activated there will be a deprecation effort for VoteStateV3.

runtime/src/block_component_processor/vote_reward.rs
akhi3030: Moved to using `store_accounts` now.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,82 +156,201 @@ impl VoteRewardAccountState {
...
+    let denominator = epoch_state.slots_per_epoch as u128 * total_stake_lamports as u128;
 
     // SAFETY: the result should fit in u64 because we do not expect the inflation in a single
     // epoch to exceed u64::MAX.
-    (numerator / denominator).try_into().unwrap()
+    let reward_lamports: u64 = (numerator / denominator).try_into().unwrap();
+    // As per the Alpenglow SIMD, the rewards are split equally between the validators and the leader.
+    let validator_reward_lamports = reward_lamports / 2;
+    let leader_reward_lamports = reward_lamports - validator_reward_lamports;
+    (validator_stake_lamports, leader_reward_lamports)
```
ksn6: Shouldn't this be `(validator_reward_lamports, leader_reward_lamports)`?

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,82 +156,201 @@ impl VoteRewardAccountState {
...
+                current_epoch,
+                validator_to_reward,
+                current_slot_account,
+                validator_reward,
+                &mut accounts_to_store,
+            );
+        }
     }
+    let Some((_, leader_account)) = current_slot_accounts.get(&current_slot_leader) else {
+        // XXX: should I return an error here or panic?
```
ksn6: Haven't thought too closely about this - are there weird epoch boundary things that could trigger this? E.g. what happens if a leader closes their vote account at/around an epoch boundary and they're not around to get paid?

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -35,8 +52,15 @@ impl VoteRewardAccountState {
...
                 // will become active in an epoch before the epoch in which Alpenglow is activated.
+                let epoch_state = EpochState {
+                    epoch_validator_rewards: 0,
+                    slots_per_epoch: bank.epoch_schedule.slots_per_epoch,
+                };
                 let state = Self {
-                    epoch_validator_rewards_lamports: 0,
+                    current_epoch: bank.epoch(),
+                    current_epoch_state: epoch_state.clone(),
+                    prev_epoch: bank.epoch() - 1,
```
ksn6: saturating_sub

runtime/src/block_component_processor/vote_reward.rs
akhi3030: discussed with Ashwin offline.  There are indeed some edge cases where a leader has closed their account but they haven't been removed from the schedule yet.  So We need to actually "ignore" the issue here and continue onward.

runtime/src/block_component_processor/vote_reward.rs
akhi3030: changed to `checked_sub.unwrap()` as this issue we would hit most often in testing and it is good to catch it earlier than to see some stranger failure.

runtime/src/block_component_processor/vote_reward.rs
akhi3030: good catch, thank you!

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -18,13 +21,27 @@ static VOTE_REWARD_ACCOUNT_ADDR: LazyLock<Pubkey> = LazyLock::new(|| {
     pubkey
 });
 
+/// The per epoch info stored in the off curve account.
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+struct EpochState {
+    /// The rewards (in lamports) that would be paid to a validator whose stake is equal to the
+    /// capitalization and it voted in every slot in the epoch.  This is also the
+    /// epoch inflation.
+    epoch_validator_rewards: u64,
```
AshwinSekar: i feel like this name is a little weird maybe something like 
```suggestion
    epoch_inflation_lamports: u64,
```

or even something like `max_possible_validator_reward`

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -18,13 +21,27 @@ static VOTE_REWARD_ACCOUNT_ADDR: LazyLock<Pubkey> = LazyLock::new(|| {
...
 /// The state stored in the off curve account used to store metadata for calculating and paying
 /// voting rewards.
+///
+/// Info for the previous and the current epoch is stored.
 #[derive(Debug, PartialEq, Eq, Deserialize, Serialize)]
 pub(crate) struct VoteRewardAccountState {
-    /// The rewards (in lamports) that would be paid to a validator whose stake is equal to the
-    /// capitalization and it voted in every slot in the epoch.  This is also the epoch inflation.
-    epoch_validator_rewards_lamports: u64,
+    current_epoch: Epoch,
```
AshwinSekar: I would consider putting the Epoch inside EpochState, just so it's less likely that we make an error like updating `current_epoch` but forgetting to update `current_epoch_state`

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -18,13 +21,27 @@ static VOTE_REWARD_ACCOUNT_ADDR: LazyLock<Pubkey> = LazyLock::new(|| {
...
+    /// Number of slots in the epoch.
+    slots_per_epoch: u64,
+}
+
 /// The state stored in the off curve account used to store metadata for calculating and paying
 /// voting rewards.
+///
+/// Info for the previous and the current epoch is stored.
 #[derive(Debug, PartialEq, Eq, Deserialize, Serialize)]
 pub(crate) struct VoteRewardAccountState {
```
AshwinSekar: also maybe a better name for the account and state would be something like `EpochInflation...`

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,82 +156,208 @@ impl VoteRewardAccountState {
...
+    let reward_epoch = bank.epoch_schedule.get_epoch(reward_slot);
+    let epoch_state = VoteRewardAccountState::new_from_bank(bank)
+        .get_epoch_state(reward_epoch)
+        .ok_or(PayVoteRewardError::NoEpochValidatorStake {
+            reward_epoch,
+            current_slot,
+        })?;
+
+    let current_slot_accounts = bank.vote_accounts();
+    let mut accounts_to_store = Vec::with_capacity(validators_to_reward.len());
```
AshwinSekar: nit: should we use + 1 here? to account for the leader not being part of the aggregates

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,82 +156,208 @@ impl VoteRewardAccountState {
...
+    // info about older slots as well.
+    let reward_epoch = bank.epoch_schedule.get_epoch(reward_slot);
+    let epoch_state = VoteRewardAccountState::new_from_bank(bank)
+        .get_epoch_state(reward_epoch)
+        .ok_or(PayVoteRewardError::NoEpochValidatorStake {
+            reward_epoch,
+            current_slot,
+        })?;
+
+    let current_slot_accounts = bank.vote_accounts();
```
AshwinSekar: nit: 
```suggestion
    let current_vote_accounts = bank.vote_accounts();
```

runtime/src/block_component_processor/vote_reward.rs
AshwinSekar:
```suggestion
    let mut paid_vote_accounts = Vec::with_capacity(validators_to_reward.len());
```

runtime/src/block_component_processor/vote_reward.rs
akhi3030: picked `max_possible_validator_reward`

runtime/src/block_component_processor/vote_reward.rs
akhi3030: In normal situations, the leader would've also voted and be part of the aggregates.  Doesn't hurt to add one in the off chance that it wasn't around I suppose.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -107,82 +155,210 @@ impl VoteRewardAccountState {
...
+    let data = account.account().data_clone();
+    // TODO (akhi): this is a stop gap till we upstream.
+    let Ok(mut vote_state) = bincode::deserialize(&data) else {
+        return;
+    };
+    increment_credits(&mut vote_state, epoch, reward);
+    accounts_to_store.push((
+        pubkey,
+        AccountSharedData::new_data(account.lamports(), &vote_state, account.owner()).unwrap(),
+    ));
```
ksn6: Where do we actually end up increasing the lamport count for the validator? Is that supposed to be here or elsewhere?

akhi3030: the lamport count is increased in the `increment_credits` function call above.

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
```diff
@@ -0,0 +1,250 @@
...
+        Self { current, prev }
+    }
+
+    /// Returns the deserialized [`Self`] from the accounts in the [`Bank`].
+    pub(super) fn new_from_bank(bank: &Bank) -> Self {
+        match bank.get_account(&VOTE_REWARD_ACCOUNT_ADDR) {
+            None => {
+                // this can happen in the first epoch when the account has not been created yet.
+                // we create a dummy state to handle this case with the assumption that this code
+                // will become active in an epoch before the epoch in which Alpenglow is activated.
```
AshwinSekar: So based on our previous discussion we'll want this to happen the same epoch that alpenglow is plugged in, we should update the comment.

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
```diff
@@ -0,0 +1,250 @@
...
+    ) {
+        let prev_state = Self::new_from_bank(bank);
+        let current = EpochInflationState::new_from_bank(
+            bank,
+            prev_epoch,
+            prev_epoch_capitalization,
+            additional_validator_rewards,
+        );
+        let state = Self {
+            prev: prev_state.current,
```
AshwinSekar: how does this work out for the first epoch? wouldn't `prev_state.current` also be pointing to the current epoch?

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
akhi3030: Updated.

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
akhi3030: Yeah this was a bug.  I reworked some of the logic here and now store `prev` inside an `Option` to support the first epoch when AG is enabled.  I think this should be better.

akhi3030:
Seems to be failing due to a rustsec which should be addressed in https://github.com/anza-xyz/alpenglow/pull/765.

runtime/src/block_component_processor/vote_reward.rs
```diff
@@ -1,188 +1,311 @@
...
+            None => {
+                warn!("EpochInflationAccountState does not exist, ensure this is a test");
+                EpochInflationState {
+                    max_possible_validator_reward: 0,
+                    slots_per_epoch: bank.epoch_schedule.slots_per_epoch,
+                    epoch: bank.epoch(),
+                }
+            }
+        }
+    };
```
akhi3030, line 98: @AshwinSekar: I'm not happy about this bit of code.  I had to add it because of how the local cluster tests are started.  It'll be good to discuss maybe in our next 1-1 if this is the best way or if there might be another solution.

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl EpochInflationAccountState {
+    /// Returns the deserialized [`Self`] from the accounts in the [`Bank`].
+    ///
+    /// Returns `None` if the `bank` does not contain the account.
+    pub(super) fn new_from_bank(bank: &Bank) -> Option<Self> {
+        bank.get_account(&VOTE_REWARD_ACCOUNT_ADDR).map(|acct| {
+            // unwrap should be safe as the data being deserialized was serialized by us in
+            // [`Self::set_state`].
+            acct.deserialize_data().unwrap()
```
AshwinSekar: nit: would be better to `expect` here and capture the comment you had in the msg

runtime/src/block_component_processor/vote_reward.rs
AshwinSekar: I assume this is because the local-cluster tests start Alpenglow at genesis?
I think the cleanest way would be to populate the account for these dev clusters similar to how we populate the migration genesis certificate for these test clusters
https://github.com/anza-xyz/agave/blob/95d34a61120c9e606cc6f0ceb77960b16e3a46c6/runtime/src/genesis_utils.rs#L344

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
```diff
@@ -90,7 +90,8 @@ impl EpochInflationAccountState {
         let lamports = bank
```
AshwinSekar: can we just use `bank.get_minimum_balance_for_rent_exemption`?

runtime/src/block_component_processor/vote_reward/epoch_inflation_account_state.rs
```diff
@@ -170,6 +177,7 @@ impl EpochInflationAccountState {
         bank.rent_collector()
```
AshwinSekar: same here

akhi3030:
@ksn6: I would like to land this change so I will dismiss your request changes review.  Feel free to review when you get a chance and I will do follow up PRs.

